PR #2980, opened 2016-06-15
Master

8v060htwyc:


yishaigalatzer:
Thanks for the PR, looks like it is not rebased correctly. Can you please resubmit as such (or if ok with you, I can just apply your changes manually)

PR #8777, opened 2019-10-31
NuGet Restore Manager spec images

oQuantumDecayo:


zivkan:
@oQuantumDecayo  Is this image referenced from somewhere?

Issue #912, opened 2015-07-09
Support pre-release packages with floating versions


pranavkm:
Consider a project.json specification such as

```
{
   "A: "1.0.*"
}
```

If the only packages in the feed are prerelease packages such as `A.1.0.0-alpha`, `A.1.0.1-beta`, the restore fails stating it was unable to resolve package `A`.

------------
[anangaur] Adding the requirement as evolved during the discussions:

As Noel, who uses NuGet packages in PackageReference based projects,

- I would like to use floating versions that only resolves to stable versions of the package (Current behavior)
- I would like to use floating version that can resolve to the latest version even if the latest happens to be a pre-release version.
- I should be able to use both the braces/regex formats i.e. 1.* or [1.0.0, 2.0.0) that can include pre-release versions
- I would like to additionally define whether I want to resolve to rc, beta, alpha or all pre-releases.

------------
**Link to spec** - https://github.com/NuGet/Home/wiki/Support-pre-release-packages-with-floating-versions
**Status** - reviewed

yishaigalatzer:
We can discuss this for 3.2

yishaigalatzer:
I looked at how NPM does it, and it looks like it will not allow pre-releases with such a specifier.

We need to work on how we extend our syntax to support ranges (do we go with something similar to npm? Do we also allow `1.*-*`? Do we allow for version ranges ala packages.config/nuspec?

pranavkm:
> Do we also allow 1._-_

AFAIK it stops when it sees the `*`. @emgarten would know this better.

yishaigalatzer:
Yes it does, my comment was a question about what we should do in the future :)

CZEMacLeod:
Wondering if there is any progress on this as for CI purposes it would really help us to support `7.0.*-*` or something like

```
{
      "version": "7.0.*",
      "prerelease": true
}
```

yishaigalatzer:
Not yet, a good indication of progress would be if the item is assigned to someone for work.

CZEMacLeod:
Unfortunately that's what I thought... Just been open for over a year and bumped again and again... :(

spotlesscoder:
Please implement this

ghost:
So this ticket is still in Priority 1 and has been bumped for a couple of years.

We use CI and end up spitting out a lot of packages, the maintenance of keeping up with them is only just worth the value of keeping everything separate. Being able to specificy "Latest prerelease" in our dev branches would really be very useful.

[...]

PR #9104, opened 2020-01-27
Floating stable and prerelease versions in the same version range

nkolev92:
The implementation of the design is completed in https://github.com/NuGet/NuGet.Client/tree/dev-nkolev92-w. 

Design for #912

Note that the only thing out of date is the link to the tool to test implementation. I will comment when that's done. 

fyi @joelverhagen


designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+
+| Floating Version | Available Versions | Best matching version |
+| -----------------|--------------------|-----------------------|
+| 5.* | 5.0.0 <br> 5.1.0 <br> 5.2.0 <br> 5.3.0 <br> 5.4.0 <br> 5.5.0-preview.1 <br> 5.5.0-preview.2 | 5.4.0 |
+
+There is no way to float to both stable and prerelease versions at the same time.
+
+## Who are the customers
+
+PackageReference customers. A popular customer with > 80 comments & 38+ up-votes.
```
joelverhagen: customer _issue_?

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Requirements
+
+From the issue:
+
+As Noel, who uses NuGet packages in PackageReference based projects,
+
+I would like to use floating versions that only resolves to stable versions of the package (Current behavior)
```
joelverhagen: only **resolve**

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+I would like to use floating version that can resolve to the latest version even if the latest happens to be a pre-release version.
```
joelverhagen: floating **versions**

joelverhagen: or maybe **a** floating version

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+There are a few reasons why we decided against that approach:
+
+1. The PackageReference opt in into prerelease version for ranges is dependant on the version requested. PackageReference allows prereleases as potential versions for dependencies due to (link issue)
+1. You can not specify which prerelease versions you want included. For  example, you can not express the following version from the proposed approach `5.1.*-rc.*`. 
+1. It adds an additional element which could realistically be added anywhere. PackageReference is integrated within MSBuild, you could easily add 
+1. It does not apply in certain situations, creating redundancies.
+
+* How will old clients behave
+
+Old clients fail to parse this version. This is arguably the desired behavior, as it allows the customer to recognize very early that things will not work the way they expect them to.
```
joelverhagen: In the past we hid such packages by default and exposed them via the `semVerLevel=2.0.0`/new registration hive protocol enhancements.

Do you want to consider this approach? It an expensive server side changes but it is possible if we expect a lot of friction here.

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+| 1.*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 | 1.1.0 | Only stable versions starting 1.* are matched | 
+| 1.2.0-*          | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.2.0-rc.1 <br> 1.2.0-rc.2 | 1.2.0-rc.2 | The highest matching prerelease is selected |
+| *-*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 3.0.0-beta.1 | 3.0.0 is the highest version |
+| 1.*-*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 | 1.2.0-rc.2 | A prerelease version is the highest matching version |
+| *-rc.*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 | 2.0.0 | A stable version is the highest matching version |
+
+For convenience, the implementation of this proposal has been uploaded to a helper service where you can try out the version parsing and the version selection. https://nugettoolsdev.azurewebsites.net/5.3.0
```
joelverhagen: Send me those bits, man!

designs/FloatingStableAndPrerelease.md
joelverhagen: Oh wait, people can't put these in .nuspecs, so it's a decision only package consumers make, right?

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. The PackageReference opt in into prerelease version for ranges is dependant on the version requested. PackageReference allows prereleases as potential versions for dependencies due to (link issue)
+1. You can not specify which prerelease versions you want included. For  example, you can not express the following version from the proposed approach `5.1.*-rc.*`. 
+1. It adds an additional element which could realistically be added anywhere. PackageReference is integrated within MSBuild, you could easily add 
+1. It does not apply in certain situations, creating redundancies.
+
+* How will old clients behave
+
+Old clients fail to parse this version. This is arguably the desired behavior, as it allows the customer to recognize very early that things will not work the way they expect them to. 
+
```
joelverhagen, line 136: Could you provide an example error message that an old client will dump out? In general it would be nice if someone searches for that error message they should know that they must update client to use the feature via docs.

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+For example:
+
+| Floating Version | Available Versions | Versions that match |
+| -----------------|--------------------|-----------------------|
+| 5.*-rc.* | 4.9.0 <br> 5.1.0 <br> 5.2.0-beta.1 <br> 5.2.0-rc.3 <br> 6.0.0-rc.1 | 5.1.0 <br> 5.2.0-rc.3 |
+
+Note that 5.2.0-beta.1 and 6.0.0-rc.1 do not match.
+
+The behavior is best illusrated with examples
```
joelverhagen: **illustrated**

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,95 @@
...
+| -----------------|--------------------|-------------------|--------------|-------|
+| *                | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 2.0.0 | 2.0.0 | Only stable versions are matched | 
+| 1.*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 | 1.1.0 | Only stable versions starting 1.* are matched | 
+| 1.2.0-*          | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.2.0-rc.1 <br> 1.2.0-rc.2 | 1.2.0-rc.2 | The highest matching prerelease is selected |
+| *-*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 3.0.0-beta.1 | 3.0.0 is the highest version |
+| 1.*-*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 | 1.2.0-rc.2 | A prerelease version is the highest matching version |
+| *-rc.*              | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 <br> 3.0.0-beta.1 | 1.1.0 <br> 1.2.0-rc.1 <br> 1.2.0-rc.2 <br> 2.0.0 | 2.0.0 | A stable version is the highest matching version |
```
joelverhagen: please include a sample where the star is not immediately after a dot.

`1.0.0-rc*`
and
`1.0.0-rc.1*`

designs/FloatingStableAndPrerelease.md
nkolev92: As discussed, we do not expect this to be an issue as floating version ranges are only relevant in the client scenarios.

designs/FloatingStableAndPrerelease.md
nkolev92: Unfortunately that's difficult. 
The reason for that is the old clients can't differentiate between legit failures and ones that might work in the future.

joelverhagen: That's fine, even if it's a generic message it's good to understand what people will be seeing

designs/FloatingStableAndPrerelease.md
nkolev92: fixed.

nkolev92:
@joelverhagen addressed all your feedback.

designs/FloatingStableAndPrerelease.md
nkolev92: Cool, I'm adding that to the spec.

joelverhagen (approved):
Consider making each example link to the tool as proof

joelverhagen:
Are there any other possible syntaxes? We are choosing what seems like the most logical... But once choose it is is chosen forever.

anangaur:
Is there a short write up of what has been implemented. Some examples of how to use?

anangaur:
Specifically how are we handling `alpha`, `beta`, `rc` labels in the pre-releases

nkolev92:
> Is there a short write up of what has been implemented. Some examples of how to use?

The spec has examples. 

> Specifically how are we handling alpha, beta, rc labels in the pre-releases

Prerelease floating is already a thing, you just couldn't combine it with stable, now you can. 

I'll update the examples to link to the tool.

zkat (commented):
This looks good, although my gut feeling is concerned that big, backwards-incompatible changes like these can have adverse effects on the ecosystem. Have you considered, for example, the possibility of preventing **publication** of any packages that use these new specifiers, until our analytics show that legacy clients have fallen out of use and it's safe to use these publicly? There's potential for ecosystem breakage here, and that concerns me.

nkolev92:
> Have you considered, for example, the possibility of preventing publication of any packages that use these new specifiers, until our analytics show that legacy clients have fallen out of use and it's safe to use these publicly? There's potential for ecosystem breakage here, and that concerns me.

Floating versions are only allowed on the consumer side. 
There's no risk of breaking existing customers. They are not allowed in a nuspec.

At pack time floating versions get simplified to an equivalent range with the selected version as the minimum version instead.

zkat:
Great! We're all good, then :)

zkat (approved):

nkolev92:
@NuGet/nuget-client Anyone interested in a spec discussion or do you find this doc satisfactory?

Specifically, @anangaur, @aortiz-msft, @rrelyea

nkolev92:
A design discussion for this is pending.

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,131 @@
...
+I would like to use floating versions that can resolve to the latest version even if the latest happens to be a pre-release version.
+I would like to additionally define whether I want to resolve to rc, beta, alpha or all pre-releases.
+
+## Goals
+
+* Define a versioning strategy that allows floating to both stable and prerelease versions simultaneously. Floating only prerelease and only stable versions is possible today with `1.0.0-*` and `1.*` respectively.
+
+## Non-Goals
+
+* Support for more than 1 prerelease label in floating versions. This is out of scope, because there is no way today with the prerelease only floating versions. As such it can be treated as a separate feature.
```
zivkan: I know that this spec only mentions `PackageReference`, but I think it would be good either in the goals or non-goals section to explicitly say that it's only about refernecing packages, and it's invalid for a package to use this syntax for its dependencies.

nkolev92: Makes sense, will add that.

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,135 @@
...
+<ItemGroup>
+    <PackageReference Include="NuGet.Packaging" Version="5.*" />
+</ItemGroup>
+```
+
+Right now this will only include stable package versions.
+
+| Floating Version | Available Versions | Best matching version |
+| -----------------|--------------------|-----------------------|
+| 5.* | 5.0.0 <br> 5.1.0 <br> 5.2.0 <br> 5.3.0 <br> 5.4.0 <br> 5.5.0-preview.1 <br> 5.5.0-preview.2 | 5.4.0 |
```
terrajobst: This renders funny. I think you'll want 5.4 to align vertically between `Available versions` and `Best matching versions`

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,135 @@
...
+* I would like to use floating versions that can resolve to the latest version even if the latest happens to be a pre-release version.
+* I would like to additionally define whether I want to resolve to rc, beta, alpha or all pre-releases.
+
+## Goals
+
+* Define a versioning strategy that allows floating to both stable and prerelease versions simultaneously. Floating only prerelease and only stable versions is possible today with `1.0.0-*` and `1.*` respectively.
+
+## Non-Goals
+
+* Support for more than 1 prerelease label in floating versions. This is out of scope, because there is no way today with the prerelease only floating versions. As such it can be treated as a separate feature.
```
terrajobst: I don't understand this statement. My understanding is that pre-release labels are sorted, i.e. `beta` would be considered newer than `alpha`. In that sense it supports multiple labels? Or are you saying that consumers won't be able to say "resolve latest in the `beta` band", as opposed to "latest version across all labels"?

designs/FloatingStableAndPrerelease.md
```diff
@@ -0,0 +1,135 @@
...
+* Floating versions are only supported for top level packages in the project, not in the nuspec. This is not going to change.
```
terrajobst: I understand that you won't allow them in .nuspec files. But does top-level mean that you can only have them in a single project in your solution? That would be weird. I assume you can specify them in any project in the graph, but it means that the "root" of the graphs wildcard will take precedent of lower versions in the dependency graph, like how fixed version numbers work today. True?

terrajobst (commented):
Left some comments

designs/FloatingStableAndPrerelease.md
nkolev92: >  But does top-level mean that you can only have them in a single project in your solution? That would be weird.

No, you can have them in project. 

>  I assume you can specify them in any project in the graph, but it means that the "root" of the graphs wildcard will take precedent of lower versions in the dependency graph, like how fixed version numbers work today. True?

That's correct. The "nearest wins" rule applies to floating version ranges the same way it applies to fixed version ranges.

designs/FloatingStableAndPrerelease.md
nkolev92: When you say `1.1.0-alpha.*` it means only the prereleases with an alpha label will be preferred. 
See [example](https://nugettoolsdev.azurewebsites.net/5.5.0-floating.7611/find-best-version-match?versionRange=1.1.0-alpha.*&versions=1.1.0-alpha.3%0D%0A1.1.0-beta.1%0D%0A1.1.0). 

If you specify `1.1.0-*` then all `1.0.0-` prerelease versions are in the preferred range. 

See similar [example](https://nugettoolsdev.azurewebsites.net/5.5.0-floating.7611/find-best-version-match?versionRange=1.1.0-*&versions=1.1.0-alpha.3%0D%0A1.1.0-beta.1%0D%0A1.1.0) how the priority changes. 

So customers can say: 

* Resolve latest beta
* Resolve latest version across all labels

but can't say

* Resolve latest alpha or beta. 

Does that make sense? 
I think it's worth adding an explanation but I want to confirm it's understandable.

karann-msft (approved):
Please publish to Home/Wiki using this template - https://github.com/NuGet/Home/wiki/Flag-vulnerable-packages

dogmasega:
i wounder why mark @$$

PR #9109, opened 2020-01-28
Extensibility services guidelines

### LINKED ISSUE #9062: Guidelines for asynchronous NuGet extensibility services

nkolev92:
Fixes: https://github.com/NuGet/Home/issues/9062
A separate PR for the already reviewed extensibility services doc. 

@NuGet/nuget-client can someone quickly review this so I can merge it?


designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+As such a lot of our customers as a pattern invoke the NuGet APIs from the UI thread.
+
+This leads to a lot of performance and reliability issues with NuGet as the culprit.
+
+Despite being synchronous IVs* some of our APIs can be invoked off the UI thread, but that's not consistently documented.
+Given that most of our code is already asynchronous, there's no need why we cannot expose services that are asynchronous.
+
+## Who are the customers
+
+All Visual Studio users wil reap the benefits.
```
dominoFire:
```suggestion
All Visual Studio users will reap the benefits.
```

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace NuGet.VisualStudio.Contracts
+{
+    public interface INuGetPackageInstaller
+    {
+        Task InstallPackageAsync(
```
dominoFire, line 79: I would like to use a Task that returns a value with information specific of the Task result

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Deprecation of old services
+
+Most of the initial extensibility service work will involve migrating existing services.
+The old services should be marked as obsolete when the replacement ones are created.
+
+## Considerations
+
+* Should we use ValueTask over Task in all of our interfaces?
```
dominoFire: We should use types that is available on most platforms

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+On the client side one would do:
+
+```cs
+    IBrokeredServiceContainer brokeredServiceContainer = await this.GetServiceAsync<SVsBrokeredServiceContainer, IBrokeredServiceContainer>();
+    Assumes.Present(brokeredServiceContainer);
+    IServiceBroker sb = brokeredServiceContainer.GetFullAccessServiceBroker();
+    INuGetPackageInstaller client = await sb.GetProxyAsync<INuGetPackageInstaller>(NuGetPackageInstallerServiceDescriptor, this.DisposalToken);
+```
+
+### Free threadedness
```
dominoFire:
```suggestion
### Free threading considerations
```

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+
+Performance is always one of the primary considerations.
```
dominoFire, line 116: Do we have a baseline to compare performance ?

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Requirements
+
+* Asynchronous and performant NuGet extensibility services
+* Ability to invoke these through RPC. These days there are components that are running out of devenv.exe
+
+## Goals
+
+* Define a guideline for designing asynchronous NuGet services
+* Introduce a new NuGet.VisualStudio.Contracts
```
dominoFire:
```suggestion
* Introduce a new assembly called `NuGet.VisualStudio.Contracts`
```

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+* Expose them through an extensibility mechanism to allow out of process invocation
+
+## Non-Goals
+
+* Define the APIs for all the services that will be migrated.
```
dominoFire, line 40: Will we have one-on-one equivalent with current NuGet IVs* interfaces?

dominoFire (dismissed):
Looks good, just address the typos.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: That's fine, you can define a service that does that. 
This is just an example.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: ValueTask is available on all platforms through https://www.nuget.org/packages/System.Threading.Tasks.Extensions/

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: This is not really about a specific implementation, it's more about how we implementation. 
It's really a call to "build performant services"

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: This just covers the guidelines, not an implementation of a specific service. 
As such the goal of this architecture doc is not to define the services. 

All of the services that need migration will be evaluated at their own merits.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
zivkan: No we should not use `ValueTask` everywhere. From [this blog post](https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/):

> Should every new asynchronous API return ValueTask / ValueTask<TResult>?
> In short, no: the default choice is still Task / Task<TResult>.

nkolev92:
@dominoFire Addressed feedback where appropriate/answered questions where appropriate. 

Please take another look.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: I guess I can change it to, should we use ValueTask in the extensibility interfaces. 

It has it's uses, but I'm trying to address the question for the extensbiility services specifically.

nkolev92: @zivkan fixed.

zivkan: It's been a while since I read the blog post, but I expect it's been covered there. Basically, `ValueTask` is only beneficial when the code returns synchronously the majority of the time (I think on a ASP.NET Community Standup video, someone said something like 90%+), as it's harmful when it's actually async. Since the requirements section of this document says it should work in RPC environments (where we know that 100% of calls will not be synchronuos), it means that `ValueTask` is not all at appropriate.

I can't find any documented problems, but I do remember in another ASP.NET Community Standup someone said that they were helping customers debug issues by "incorrect" usage of `ValueTask`. I think if multiple codepaths await the same ValueTask that completes async, then bugs happen, so ValueTask can only be used when the task will only be awaited once. But I'm not 100% on this. My point is that ValueTask is less safe than `Task`, and should only be used when we know it's safe, meaning a case-by case decision.

nkolev92: Yeah, you are correct. 
My description links the blog post you are referring to :) 

In general, because this is RPC, the answer is no!

dtivel (approved):

nkolev92:
@dominoFire 

Can you review again?
I'd like to merge this.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+### Deprecation of old services
+
+Most of the initial extensibility service work will involve migrating existing services.
+The old services should be marked as obsolete when the replacement ones are created.
+
+## Considerations
+
+* Should we use ValueTask over Task in our new extensibility interfaces?
```
zivkan: Keeping this as a question makes it sound like it's still open for discussion/decision. Even the text below makes it sound undecided. I suggest making the bullet point a statement that we are not using it, and the text below to say that although it was considered because it might have this advantage, it won't be used because of these problems.

zivkan (approved):
As I wrote in the comment, the question under considerations makes it sound like we haven't made a decision yet. I suggest that be changed.

designs/NuGetExtensibility/NuGetExtensibilityServices.md
nkolev92: Makes sense, I'll fix it.

nkolev92: fixed

nkolev92:
@dominoFire 
🔔

PR #9182, opened 2020-02-15
Add spec for CPVM updates in VS

zivkan:
Fixes: https://github.com/NuGet/Home/pull/9182

### DIFF
--- a/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
+++ b/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
@@ -309,7 +309,9 @@ A new service should be created. This service will have multiple responsibilitie
 * Find which `Directory.Packages.props` file needs to be changed given a package install for a specific project.
 * Orchestrate efficient updating and nomination of projects. In particular, when project files and `Directory.Packages.props` files need to change due to a single customer gesture, avoid multiple project nominations and multiple restores due to multiple file changes.
 
-Proposed API.
+## Proposal
+
+This is the proposed API, to be used as a starting point, but it is subject to change as the Project System team own implementing it and may adapt it to their own coding standards.
 
 ```cs
 enum PackageReferenceAction
@@ -383,12 +385,6 @@ interface IPackageReferenceUpdateService
   Task ApplyChangesAsync(IReadOnlyList<PackageReferenceChangeRequest> actions, CancellationToken token);
 }
 ```
-
-open questions:
-
-* The convention in .NET is to add `CancellationToken` to async APIs. However, what is the expected behaviour if cancellation is requested in the middle of processing a single change request?
-* Which project types will this service support? C# and VB only?  f#? c++? non-Microsoft project systems?
-
 
 ## References
 
--- a/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
+++ b/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
@@ -314,75 +314,78 @@ A new service should be created. This service will have multiple responsibilitie
 This is the proposed API, to be used as a starting point, but it is subject to change as the Project System team own implementing it and may adapt it to their own coding standards.
 
 ```cs
-enum PackageReferenceAction
+public abstract class PackageReferenceChangeRequest
 {
-  /// <summary>
-  /// To catch bugs where values are left uninitialized, but otherwise usused and invalid.
-  /// </summary>
-  Unknown = 0,
+    /// <summary>
+    /// GUIDs for the projects in the solution to which a package reference change should be applied.
+    /// </summary>
+    public IReadOnlyList<Guid> ProjectGuids { get; }
 
-  /// <summary>
-  /// The change should add or update PackageReference and or PackageVersion items.
-  /// </summary>
-  AddOrUpdate = 1,
+    /// <summary>
+    /// The package identity.
+    /// </summary>
+    public string PackageIdentity { get; }
 
-  /// <summary>
-  /// The change should remove matching PackageReference items only, do not remove PackageVersion items.
-  /// </summary>
-  RemovePackageReference = 2,
-
-  /// <summary>
-  /// This change should remove both PackageReference and PackageVersion items.
-  /// </summary>
-  /// <remarks>
-  /// Out of scope for the initial version. It has been defined to avoid needing an API change to add this in the future.
-  /// </remarks>
-  Remove = 3
+    /// <summary>
+    /// Initializes a new <see cref="PackageReferenceChangeRequest" /> instance.
+    /// </summary>
+    public PackageReferenceChangeRequest(IReadOnlyList<Guid> projectGuids, string packageIdentity)
+    {
+        this.ProjectGuids = projectGuids;
+        this.PackageIdentity = packageIdentity;
+    }
 }
 
-class PackageReferenceChangeRequest
+public class AddOrUpdatePackageReferenceChangeRequest : PackageReferenceChangeRequest
 {
-  /// <summary>
-  /// The action to apply
-  /// </summary>
-  public PackageReferenceAction Action { get; set;}
+    /// <summary>
+    /// The version of the package to add.
+    /// </summary>
+    public string Version { get; }
 
-  /// <summary>
-  /// Which projects in the solution should apply this PackageReference action
-  /// </summary>
-  public IReadOnlyList<string> Projects { get; set; }
+    /// <summary>
+    /// Additional metadata to set on the PackageReference item for the given package identity.
+    /// </summary>
+    /// <remarks>
+    /// The package must create or set metadata defined by the keys of this dictionary, with the value defined by the dictionary item's value.
+    /// It must also remove any metadata already defined in the PackageReference if the value for that metadata key is null or empty.
+    /// </remarks>
+    public IReadOnlyDictionary<string, string> AdditionalMetadata { get; }
 
-  /// <summary>
-  /// The package ID 
-  /// </summary>
-  public string PackageId { get; set; }
-
-  /// <summary>
-  /// The package version
-  /// </summary>
-  /// <remarks>
-  /// * AddOrUpdate must set the Version attribute in the Directory.Packages.props fie to this version. If the Directory.Packages.props file does not already contain an item for this PackageId, then it should be added.
-  /// * Remove and RemovePackageReference will ignore any value in Version
-  /// </remarks>
-  public string Version { get; set; }
-
-  /// <summary>
-  /// Additional metadata for this PackageId to set on the PackageReference item.
-  /// </summary>
-  /// <remarks>
-  /// * AddOrUpdate must create or set metadata defined by the keys of this dictionary, with the value defined by the dictionary item's value.
-  /// * AddOrUpdate must remove any metadata already defined in the PackageReference, if there is no key for that metadata in the dictionary.
-  /// * RemovePackage will ignore any value in AdditionalMetadata.
-  /// </remarks>
-  public IReadOnlyDictionary<string, string> AdditionalMetadata { get; set; }
+    /// <summary>
+    /// Initializes a new <see cref="AddOrUpdatePackageReferenceChangeRequest" /> instance.
+    /// </summary>
+    public AddOrUpdatePackageReferenceChangeRequest(IReadOnlyList<Guid> projectGuids, string packageIdentity, string version, IReadOnlyDictionary<string, string> additionalMetadata)
+        : base(projectGuids, packageIdentity)
+    {
+        this.Version = version;
+        this.AdditionalMetadata = additionalMetadata;
+    }
 }
 
-interface IPackageReferenceUpdateService
+public class RemovePackageReferenceChangeRequest : PackageReferenceChangeRequest
 {
-  /// <summary>
-  /// Apply changes to the project(s) provided.
-  /// </summary>
-  Task ApplyChangesAsync(IReadOnlyList<PackageReferenceChangeRequest> actions, CancellationToken token);
+    /// <summary>
+    /// Whether PackageVersion items should be removed (in addition to PackageReference items).
+    /// </summary>
+    public bool RemovePackageVersionItems { get; }
+
+    /// <summary>
+    /// Initializes a new <see cref="PackageReferenceChangeRequest" /> instance.
+    /// </summary>
+    public RemovePackageReferenceChangeRequest(IReadOnlyList<Guid> projectGuids, string packageIdentity, bool removePackageVersionItems = false)
+        : base(projectGuids, packageIdentity)
+    {
+        this.RemovePackageVersionItems = removePackageVersionItems;
+    }
+}
+
+public interface IPackageReferenceManager
+{
+    /// <summary>
+    /// Apply package reference changes to the projects provided.
+    /// </summary>
+    Task ApplyChangesAsync(IReadOnlyList<PackageReferenceChangeRequest> actions, CancellationToken token);
 }
 ```
 
--- a/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
+++ b/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
@@ -315,7 +315,7 @@ Proposed API.
 enum PackageReferenceAction
 {
   /// <summary>
-  /// To catch uninitialized values, but otherwise usused and invalid.
+  /// To catch bugs where values are left uninitialized, but otherwise usused and invalid.
   /// </summary>
   Unknown = 0,
 
@@ -348,7 +348,7 @@ class PackageReferenceChangeRequest
   /// <summary>
   /// Which projects in the solution should apply this PackageReference action
   /// </summary>
-  public List<string> Projects { get; set; }
+  public IReadOnlyList<string> Projects { get; set; }
 
   /// <summary>
   /// The package ID 
@@ -359,12 +359,8 @@ class PackageReferenceChangeRequest
   /// The package version
   /// </summary>
   /// <remarks>
-  /// * If the project does not use CPVM
-  ///   * AddOrUpdate must set the PackageReference Version metadata to this value
-  ///   * RemovePackage will ignore any value in Version
-  /// * If the project uses CPVM
-  ///   * AddOrUpdate must set the Version attribute in the Directory.Packages.props fie to this version. If the Directory.Packages.props file does not already contain an item for this PackageId, then it should be added.
-  ///   * RemovePackage will ignore any value in Version
+  /// * AddOrUpdate must set the Version attribute in the Directory.Packages.props fie to this version. If the Directory.Packages.props file does not already contain an item for this PackageId, then it should be added.
+  /// * Remove and RemovePackageReference will ignore any value in Version
   /// </remarks>
   public string Version { get; set; }
 
@@ -376,7 +372,7 @@ class PackageReferenceChangeRequest
   /// * AddOrUpdate must remove any metadata already defined in the PackageReference, if there is no key for that metadata in the dictionary.
   /// * RemovePackage will ignore any value in AdditionalMetadata.
   /// </remarks>
-  public IDictionary<string, string> AdditionalMetadata { get; set; }
+  public IReadOnlyDictionary<string, string> AdditionalMetadata { get; set; }
 }
 
 interface IPackageReferenceUpdateService
@@ -384,7 +380,7 @@ interface IPackageReferenceUpdateService
   /// <summary>
   /// Apply changes to the project(s) provided.
   /// </summary>
-  Task<bool> ApplyChangesAsync(List<PackageReferenceChangeRequest> actions, CancellationToken token);
+  Task ApplyChangesAsync(IReadOnlyList<PackageReferenceChangeRequest> actions, CancellationToken token);
 }
 ```
 
--- a/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
+++ b/designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
@@ -1,4 +1,4 @@
-# Centrally managing NuGet package versions - Managing packages in Visual Studio Spec
+# Centrally managing NuGet package versions - Visual Studio NuGet/Project System APIs
 
 * Status: **In Review**
 * Author(s): [Andy Zivkovic](https://github.com/zivkan)
@@ -10,7 +10,7 @@
 
 ## Problem Background
 
-Within Visual Studio, only project systems read and manipulate project files, and provide APIs to other components, such as NuGet. This allows project systems to efficiently handle changes. Although CPVM's `Directory.Packages.props` file is not the project file itself (`csproj`, `vbproj`, and so on), it is imported by the project files and therefore contributes to project evaluation.
+Within Visual Studio, only project systems read and manipulate project files, and provide APIs to other components, such as NuGet. This allows project systems to efficiently handle changes. Although CPVM's `Directory.Packages.props` file is not the project file itself (`csproj`, `vbproj`, and so on), it is an MSBuild file and is imported by the project files and therefore contributes to project evaluation.
 
 To enable NuGet's package management functionality within Visual Studio (Package Manager UI, Package Manager Console, or NuGet's APIs to install packages), NuGet needs agreement with project systems on how APIs should behave with the new feature.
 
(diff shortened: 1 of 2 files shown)


designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+
+* It's valid for some, not all, projects in the solution to use centrally managed package versions.
+* It's valid for different projects using central package versions to use different `Directory.Packages.props` files.
+* As is valid now, it's valid for some projects to use `packages.config`, while others use `PackageReference`. The `packages.config` projects cannot use centrally managed package versions, and individual `PackageReference` projects may opt in or out of central package version management.
+
+## Solution
+
+### Installing a package in a project
+
+Currently NuGet calls project system APIs to add a PackageReference to the project for the given package id and version. However, when the project uses centrally managed versions, the project file must not add the version on the `PackageReference` element.
```
cristinamanum: Would be possible to add more information about the Project System APIs used to add/update PackageReferences?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+
+* API aware of centrally managed package versions
+
+Presumably this would mean there's a separate API to add a `PackageVersion` to the `Directory.Packages.props` file. But since this file is shared across multiple projects, would the API be independent from the projects, or should NuGet pick one project and call the API on that project?
+
+Should new APIs be added to the project systems considering the version attribute must not be written? Or NuGet could call the API with a null version. Or NuGet passes the version and the project system must know if the project uses centrally managed package versions or not.
+
+### Removing a package from a project
+
+Currently NuGet calls APIs on the project system to remove the `PackageReference` for that package given a package id. The current API is sufficient and the same behavior should continue. The `Directory.Packages.props` file should not be modified to remove the `PackageVersion`, as other projects may still be using it.
```
cristinamanum: Is there any difference in behavior between older PR type projects and SDK PR projects?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Solution
+
+### Installing a package in a project
+
+Currently NuGet calls project system APIs to add a PackageReference to the project for the given package id and version. However, when the project uses centrally managed versions, the project file must not add the version on the `PackageReference` element.
+
+There are two categories of solutions.
+
+* API ignorant of centrally managed package versions
```
cristinamanum: API == NuGet API or Project System API? Can the current ProjectSystem Apis ignore the version and add only the PackageReference item without a Version metadata? This question is especially important for not SDK Package Ref style projects.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+
+Project system APIs would not be changed, but the implementation would need to be aware of how to detect if the feature is enabled, and know when to add a version to the `PackageReference` and when not to, and when to modify the `Directory.Packages.props` and when not not.
+
+There would also be a potential issue when a single user gesture causes a package to be installed into multiple projects, particularly when the `Directory.Packages.props` file does not already contain a `PackageVersion` for the package. If NuGet calls APIs to modify all the projects in parallel, there might be a race condition with trying to update `Directory.Packages.props`.
```
cristinamanum: In addition to Directory.Packages.props updates -we would need to consider that the project file and Directory.Packages.props needs to be updated inside of a transaction like model.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+### Change the version of a package in `Directory.Packages.props`
+
+This will affect, at minimum, all the loaded projects in the solution that have a `PackageReference` to the package whose version is changed. For SDK-style projects, it appears this should work already, by watching the file for changes and then sending NuGet a project nomination with the new information. Currently non-SDK style projects need to be unloaded and reloaded to take into account changes to imported MSBuild files.
+
+Given this single user gesture will often affect multiple projects, should the API to modify the `Directory.Packages.props` file be on a project system interface, or a new interface that's decoupled from individual projects?
+
+Once the file is modified, should NuGet notify projects that it's changed, or should they detect it themselves?
+
+In particular, given that SDK style projects currently use file watchers to detect changes to imported files and automatically reload, but non-SDK style projects do not and continue to use outdated information until the project is reloaded, how can we avoid a bad user experience when a customer uses Package Manager UI/Console to change package versions?
```
cristinamanum: Currently if a project file is updated outside visual Studio with a new package Version , Visual Studio detects the change and asks for reload.
We can go on the same route or only present the buttons "Reload All" / "Ignore All"  .

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: In case you haven't yet read to the end of the document, I provided a little detail in annex 1, but I'll add more details.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: Well, the two project systems do use different APIs, but otherwise not really. They both remove PackageReferences based on package ID alone. The package version isn't used in the remove API call.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: Good question. I didn't test, and I didn't look for API docs either. Depending on if I run out of other things to work on before we meet with project system teams, I'll investigate further. Otherwise I hope they can answer this type of question when we meet.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: It would be ideal if it was transation-like, but honestly I don't think it's important. The `Directory.Packages.props` needs to be modified first to add a version for the package, and once that's done the project file needs to be updated. If the project file fails to update, I don't think there's any real harm in the `Directory.Packages.props` file having a package version for a package that isn't referenced by any project.  Once we get agreement on which team will be writing to the props file, if they find it easy to make it transactional, great.

NuGet does a preview restore before "committing" (calling the project system APIs to modify the project). The preview restore determines which packages will be installed (finding all dependencies). By the time NuGet calls project system APIs, we already know that the restore graph is valid. At least in PMUI, I'm not so sure about PMC.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: The non-SDK project system prompts to reload the project only if the project file is updated. If a file that is imported by the project file is modified, the project system ignores the change until the project is reloaded. This is a problem for the `Directory.Packages.props` file.

If project system wants to start watching imported files for changes, great. Forcing customers to reload the project every time any project in the solution updates the props file will give a bad experience, but at least it won't restore the wrong packages. If NuGet is responsible for monitoring the files for changes, and then forcing the project system to reload, well, if we have to, then it's probably better than restoring the wrong version of packages.  But this is something we need to talk to project system teams about. Does anyone have any ideas how to give a better customer experience? If changes to APIs are needed, will anyone commit to it?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,95 @@
...
+
+There would also be a potential issue when a single user gesture causes a package to be installed into multiple projects, particularly when the `Directory.Packages.props` file does not already contain a `PackageVersion` for the package. If NuGet calls APIs to modify all the projects in parallel, there might be a race condition with trying to update `Directory.Packages.props`.
+
+* API aware of centrally managed package versions
+
+Presumably this would mean there's a separate API to add a `PackageVersion` to the `Directory.Packages.props` file. But since this file is shared across multiple projects, would the API be independent from the projects, or should NuGet pick one project and call the API on that project?
+
+Should new APIs be added to the project systems considering the version attribute must not be written? Or NuGet could call the API with a null version. Or NuGet passes the version and the project system must know if the project uses centrally managed package versions or not.
+
+### Removing a package from a project
```
cristinamanum: Another section I think it would be regarding the View of versions in PM UI . Is there anything needed from the SDK and / or previous SDK style projects to display the correct CPVM versions?

zivkan: If NuGet has enough information to restore, then it has enough information to display in PMUI. What that really means is that for SDK style projects, we get the PackageVersion items from nomination, and PMUI needs to get it from our own project cache (using our existing project type abstractions). For non-SDK style projects, NuGet will call the new APIs being added for restore, and will do so every time it needs to populate the PMUI list.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: I've added more details to annex 1 now. Let me know if it doesn't contain the information that you're interested in.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: I added more details into annex 1 about project system API usage. You can see the details about what APIs are calls for the two project systems.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,108 @@
...
+* Author(s): [Andy Zivkovic](https://github.com/zivkan)
+
+## Related Specs
+
+* [Centrally managing NuGet package versions](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions)
+* [Centrally managing NuGet package versions - Restore Design Spec](https://github.com/NuGet/Home/blob/dev/designs/CentrallyManagingNuGetPackageVersions-Restore.md)
+
+## Problem Background
+
+To date, NuGet does not modify any MSBuild files (projects files, or files imported by the projects) either directly or though MSBuild APIs. Instead, NuGet uses APIs provided by the various project systems when projects need to be modified as a result of installing, updating or removing packages. In practise, NuGet has only indirectly caused project files to be modified.
```
nkolev92: All of this is a VS design principle afaik. If you want to modify the project file you go through the project system.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Who are the customers
+
+All customers managing packages on projects that are using central package version management with projects in Visual Studio IDE.
+
+## Goals
+
+1. When installing a package in a centrally managed project, if the package already exists in the `Directory.Packages.props` file, only the project file is modified. The `PackageReference` element added must not have a version applied.
+2. When installing a package in a centrally managed project, if the package does not already exist in the `Directory.Packages.props` file, then the package version is added to `Directory.Packages.props`, in addition to goal 1 above.
+3. When updating a centrally managed package version, the `Directory.Packages.props` file is updated, and all projects that use the package are restored with the new package version.
```
nkolev92: I think we should call it out explicitly, but 100% of the projects affected by the Directory.Packages.Props files should be restored because CPVM affects transitive packages as well.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
nkolev92:
```suggestion
* Centrally managed package versions agnostic API
```

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,108 @@
...
+
+* API ignorant of centrally managed package versions
+
+Project system APIs would not be changed, but the implementation would need to be aware of how to detect if the feature is enabled, and know when to add a version to the `PackageReference` and when not to, and when to modify the `Directory.Packages.props` and when not not.
+
+There would also be a potential issue when a single user gesture causes a package to be installed into multiple projects, particularly when the `Directory.Packages.props` file does not already contain a `PackageVersion` for the package. If NuGet calls APIs to modify all the projects in parallel, there might be a race condition with trying to update `Directory.Packages.props`.
+
+* API aware of centrally managed package versions
+
+Presumably this would mean there's a separate API to add a `PackageVersion` to the `Directory.Packages.props` file. But since this file is shared across multiple projects, would the API be independent from the projects, or should NuGet pick one project and call the API on that project?
```
nkolev92: I am having a bit trouble following the spec. 
Not sure if it's scenario or project system API based, right now it seems to be lumping adding a new packageversion and a new packagereference together and I think that's what's making it tricky to follow.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
nkolev92: When we ask for APIs to modify the PackageVersion, we should ensure a delete call exists. 
While we are not currently sure whether we'll ever use it, I don't think it's a high cost to add and we wouldn't need to get project-system changes to implement a new feature.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
nkolev92: We'd still need to account for the scenario where we might kick off a new restore before all nominations come in. That could lead to a restore failure and then a successful restore later when all the nominations arrive. 

This becomes more complex with the csproj.dll based projects.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,108 @@
...
+Once the file is modified, should NuGet notify projects that it's changed, or should they detect it themselves?
+
+In particular, given that SDK style projects currently use file watchers to detect changes to imported files and automatically reload, but non-SDK style projects do not and continue to use outdated information until the project is reloaded, how can we avoid a bad user experience when a customer uses Package Manager UI/Console to change package versions?
+
+## References
+
+* https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions
+* https://github.com/microsoft/MSBuildSdks/tree/master/src/CentralPackageVersions
+
+## Appendix 1 : Current project system API usage by NuGet
```
nkolev92: You might already be aware, but I tried to capture a similar summary in https://github.com/NuGet/Home/blob/d5c73c912fcd44b0df2de457da78909ff2ca6b78/designs/NuGetExtensibility/PackageInstaller.md#csproj-packagereference-scenarios.

nkolev92:
Meant to add a final comment. 
I feel like we could do a better job at covering the open questions. 

This spec talks about the challenges and potential solution, but I think we can summarize the asks a bit better.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+## Non - Goals
+
+6. This spec is scoped to Visual Studio API interaction between components. Package Manager UI changes are out of scope. Command Line Interface is out of scope.
+7. `Directory.Packages.props` files that import other MSBuild files. If the customer uses this, the behavior is undefined, specifically around which `Directory.Packages.props` file gets updated.
+
+## Scenarios
+
+The following scenarios will be enabled when the proposed design changes will be completed.
+
+### Install a package in a project that already exists in `Directory.Packages.props`
```
cristinamanum: I think that it is worth to mention that these scenarios applies for projects that are opt-in CPVM. For the rest the behavior will not change.

cristinamanum: From a definition stand point : A project is a CPVM opt-in project if and only if the Directory.Packages.props exists and the project has the EnabledCentralPackageVersions set to true.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+  <ItemGroup>
+    <PackageReference Include="PackageA" />
+  </ItemGroup>
+
+</Project>
+```
+
+### Install a package in a project that does not exist in `Directory.Packages.props`
+
+In this scenario, there is a `Directory.Packages.props` file, but it does not have `PackageVersion` for `PackageA`. There is a project `Project1`, which also does not currently have a reference to `PackageA`. When the customer installs `PackageA`, version `1.2.3`, into `Project1`, `Directory.Packages.props` is modified to include `<PackageVersion Include="PackageA" Version="1.2.3" />`, and the project's file is modified to include a `<PackageReference Include="PackageA" />`.
```
cristinamanum: It will be the closest Directory.Packages.props in the folder structure that will be modified.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+  <ItemGroup>
+    <PackageReference Include="PackageB" />
+  </ItemGroup>
+
+</Project>
+```
+
+### Update the version
+
+In this scenario, there is a `Directory.Packages.props` file which defines a version for `PackageA` and `PackageB`. There are two projects, `Project1` has a `PackageReference` to `PackageA`, and `Project2` has a `PackageReference` to `PackageB`. The customer updates `PackageA` to a different version. The `Directory.Packages.props` must be modified to contain the new version. All projects (two in this scenario) must be restored. `Project1` must be restored as it directly references `PackageA`. `Project2` must be restored, because `PackageB` might have a dependency on `PackageA`, and therefore the `PackageA`'s version needs to be pinned.
```
cristinamanum: If the ProjectA is defined in a imported Directory.Packages.props we need to decide this action what result will have. 

Is the parent that will have the version updated. 
Or a different simpler option will be to act as an addition and add the new version to the the closest Directory.Packages.props.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+  /// Out of scope of the first version (minimal viable product). In scope later. Adding it now
+  // avoids changing the interface in shipped versions.
+  /// </remarks>
+  Task<bool> DeletePackageAsync(List<string> projects, string packageId);
+}
+```
+
+Open questions:
+
+* How to handle when a PackageReference needs `IncludeAssets`, `PrivateAssets` and/or `ExcludeAssets`?
```
cristinamanum: This experience should not change. 
Is the ability to IncludeAssets, etc. is supported in Visual Studio only for SDK style projects?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+
+Note that:
+
+* It's valid for some, not all, projects in the solution to use centrally managed package versions.
+* It's valid for different projects using central package versions to use different `Directory.Packages.props` files.
+* As is valid now, it's valid for some projects to use `packages.config`, while others use `PackageReference`. The `packages.config` projects cannot use centrally managed package versions, and individual `PackageReference` projects may opt in or out of central package version management.
+
+## Solution
+
+A new service should be created. This service will have multiple responsibilities:
```
cristinamanum: In addition to using ProjectSystem for adding/updating package refs NuGet uses Microsoft.Build to do updates for dotnet scenarios. 
ProjectSystem seems to use Microsoft.Build as well. 
It would be great if this new service can be used across both scenarios - Visual Studio and dotnet ,

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+
+Proposed API. Very subject to change as we start implementing and discover better ideas:
+
+```cs
+interface IDirectoryPackagesPropsService
+{
+  /// <summary>
+  /// Add a PackageReference to the projects provided. If the packageId does
+  /// </summary>
+  Task<bool> AddPackageAsync(List<string> projects, string packageId, string packageVersion);
```
cristinamanum: projects is like a list of projects' full paths?

cristinamanum: What should be the behavior on error? All or none of the projects updated?

cristinamanum: Can these operations fail because of user settings/behavior? If yes how should these types of errors bubbled up? 

Can these methods throw? If yes in what conditions?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: This spec is about CPVM. Isn't it a safe assumption that projects are impacted by the spec's scenarios have opt-in?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: This spec is about the API contract between NuGet and the project systems, so we should define it. I should extend appendix 1 with more details, but basically the way it currently works is we ask project system for an object that represents the PackageReference item in MSBuild, then we call additional APIs to add the extra information. If NuGet is going to call a new VS service to make modifications, the API for the service needs to allow this scenario.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+* Orchestrate efficient updating and nomination of projects. In particular, when project files and `Directory.Packages.props` files need to change due to a single customer gesture, avoid multiple project nominations and multiple restores due to multiple file changes.
+
+This new service will by used by NuGet, but only for projects that have centrally managed versions enabled. Projects that do not use CPVM will continue to work as they do today.
+
+Proposal: When a project is using `CPVM`, NuGet will only use this new service and will no longer call the existing project system APIs
+
+Proposed API. Very subject to change as we start implementing and discover better ideas:
+
+```cs
+interface IDirectoryPackagesPropsService
```
dominoFire: Why don't we extend or update `IVsPackageInstaller` ? What is the reason to have a separate interface?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+
+The following scenarios will be enabled when the proposed design changes will be completed.
+
+### Install a package in a project that already exists in `Directory.Packages.props`
+
+In this scenario, there is a `Directory.Packages.props` file which specifies a `PackageVersion` for `PackageA`. There is a project `Project1`, which does not currently have a reference to `PackageA`. When the customer installs `PackageA` into `Project1`, the project's file is modified to include a `<PackageReference Include="PackageA" />`, which does not specify a version. `Directory.Packages.props` remains unchanged.
+
+#### Before
+
+`$(repo_root)\Directory.Packages.props`
```
dominoFire: Do this new API will handle CPVM in MSBuild?

https://github.com/microsoft/MSBuildSdks/tree/master/src/CentralPackageVersions

dominoFire (commented):
Just questions on the spec.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+## Requirements
+
+The requirements are defined in the [feature spec](https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions).
+
+Note that:
+
+* It's valid for some, not all, projects in the solution to use centrally managed package versions.
+* It's valid for different projects using central package versions to use different `Directory.Packages.props` files.
```
nkolev92: I'd like to understand how the preview restore works in this situation. Think it's the one hole I see with this design.

nkolev92: I remember David suggested not doing the preview...comes with its own issues ofc :)

designs/CentrallyManagingNuGetPackageVersions-VS-APIs.md
```diff
@@ -0,0 +1,419 @@
...
+  /// </remarks>
+  public IDictionary<string, string> AdditionalMetadata { get; set; }
+}
+
+interface IPackageReferenceUpdateService
+{
+  /// <summary>
+  /// Apply changes to the project(s) provided.
+  /// </summary>
+  Task<bool> ApplyChangesAsync(List<PackageReferenceChangeRequest> actions, CancellationToken token);
```
cristinamanum: If we like to define a list of action to be sent to the service, we would need to define:
1. If actions will be applied in order 
2. What are cases of bad requests (if any) and what the expected behavior will be? Cases like request1(project1 package1 version1 addorupdate), request2(project1 package1 version1 removePackageReference) is it bad request or the result is removePackageReference ?

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: Because the NuGet team is not implementing the service.

designs/CentrallyManagingNuGetPackageVersions-VSManagePackages.md
zivkan: No, the API is for NuGet's CPVM feature.

zivkan:
Closing as project system team took over responsibility to implement this (as per this spec's recommendation)

PR #9312, opened 2020-03-17
Add a design spec template that the NuGet team uses

nkolev92:
Fixes https://github.com/NuGet/Client.Engineering/issues/154

This is basically a summary of the spec we have been using. 

Only change I've made is making all the `Open Questions`, `Considerations` etc. a heading 2. 

Once merged I will update any links in the wiki and one notes I can find talking about a spec template :) 

@skofman1 @xavierdecoster @joelverhagen @agr @shishirx34 @ryuyu @dannyjdev @loic-sharma @zhhyu  
Thoughts/comments welcome. Figured you have your own template, maybe we can unify to using one.

### DIFF
--- a/designs/SPEC_TEMPLATE.md
+++ b/designs/SPEC_TEMPLATE.md
@@ -4,7 +4,6 @@
 * Status: In Review / Reviewed / Implementing / Implemented
 * Author(s): [NuGet](https://github.com/NuGet)
 * Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
-* Type: Feature
 
 ## Problem Background
 
--- a/designs/SPEC_TEMPLATE.md
+++ b/designs/SPEC_TEMPLATE.md
@@ -2,21 +2,17 @@
 # Spec Name
 
 * Status: In Review / Reviewed / Implementing / Implemented
-* Author(s): [NuGet](https://github.com/NuGet)
+* Author(s): [Author Name](https://github.com/NuGet)
 * Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
 
 ## Problem Background
 
 Present the background and the motivation for this design. 
-Present evidence why this is an important problem to tackle
+Present evidence why this is an important problem to tackle.
 
 ## Who are the customers
 
 Customers affected by this change.
-
-## Requirements
-
-Clarify the requirements in a few bullet points
 
 ## Goals
 
--- a/designs/SPEC_TEMPLATE.md
+++ b/designs/SPEC_TEMPLATE.md
@@ -1,0 +1,49 @@
+
+# Spec Name
+
+* Status: In Review / Reviewed / Implementing / Implemented
+* Author(s): [NuGet](https://github.com/NuGet)
+* Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
+* Type: Feature
+
+## Problem Background
+
+Present the background and the motivation for this design. 
+Present evidence why this is an important problem to tackle
+
+## Who are the customers
+
+Customers affected by this change.
+
+## Requirements
+
+Clarify the requirements in a few bullet points
+
+## Goals
+
+Specifically call out the goals for this proposed design.
+
+## Non-Goals
+
+Call out things that are not in consideration for this design. 
+
+## Solution
+
+Describe the proposed solution in as much detail as you deem necessary.
+
+## Future Work
+
+Declare any future work if applicable.
+
+## Open Questions
+
+Any open questions not specifically called out in the design.
+
+## Considerations
+
+Any other solutions or approaches considered? 
+The rejected designs are often just as important as the proposed ones!
+
+### References
+
+References to related issues, features


zkat (approved):

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,49 @@
+
+# Spec Name
+
+* Status: In Review / Reviewed / Implementing / Implemented
+* Author(s): [NuGet](https://github.com/NuGet)
+* Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
+* Type: Feature
```
zivkan: what other types might there be?  Bug fixes shouldn't need a design document. FOr something big like CPVM, where there's an overarching design doc, but then smaller more specific design docs, do we use different values here?  Or maybe just remove this field?

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,49 @@
...
+
+Customers affected by this change.
+
+## Requirements
+
+Clarify the requirements in a few bullet points
+
+## Goals
+
+Specifically call out the goals for this proposed design.
```
zivkan: Aren't goals and requirements just synonyms for the same thing?  How are the two sections different?

zivkan (approved):
maybe we can simplify this a little, but it does seem to be a template for most of the design docs I remember seeing in the team.

designs/SPEC_TEMPLATE.md
nkolev92: Great point. I'll remove it.
Think I had something else in mind when i did this :D

Sometimes, even big bugs lead to a refactoring that might require a spec, but it's really a DCR then. 

It really provides no value, so removing is definitely the best.

designs/SPEC_TEMPLATE.md
nkolev92: You don't have to "satisfy" all requirements.

zivkan: Isn't it the other way around? Using your favourite search engine and asking for "define requirement", I get:

> a thing that is compulsory; a necessary condition

edit: and the definition of goal is

> the object of a person's ambition or effort; an aim or desired result.

nkolev92: Yeah, the other way around. 
Goal is a broader statement. 
Requirements is very specific.

zivkan: Anyway, is there a better way we can represent this? For example, just one goals section, split into mandatory and optional lists?

nkolev92: I've seen people use both so frequently, which is why I've left them in. 

I'd appreciate input from others on this.

nkolev92: @nuget/nuget-client @karann-msft, @chgill-MSFT I'd appreciate input from other people on this topic.

nkolev92: My thought right now is to just remove the "requirements" section. 

Goals is broader and includes more. The problem background can also include the reqs as necessary depending on the author's preference.

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
+
+# Spec Name
+
+* Status: In Review / Reviewed / Implementing / Implemented
+* Author(s): [NuGet](https://github.com/NuGet)
```
donnie-msft: I believe we normally tag ourselves, so is this a static value or should it be the actual author of the spec?

designs/SPEC_TEMPLATE.md
donnie-msft: I'm with Andy, I think "Requirements" and "Goals" sound very similar, and it's hard to differentiate what goes where.

Requirements to me sounds like it should be "Constraints". What is unique to this work that you are having to avoid? 
Example: We don't want to break XYZ until the next Major Version RTM.
This is not a goal... but it is constraining your work in some way.

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
...
+
+Call out things that are not in consideration for this design. 
+
+## Solution
+
+Describe the proposed solution in as much detail as you deem necessary.
+
+## Future Work
+
+Declare any future work if applicable.
```
donnie-msft, line 31: If we use my Constraints suggestion for a section, this is where you'd identify any follow-ups for when those constraints are lifted.
Example: For the next Major Version RTM, we will do XYZ as a follow-up to this work.

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
...
+Any open questions not specifically called out in the design.
+
+## Considerations
+
+Any other solutions or approaches considered? 
+The rejected designs are often just as important as the proposed ones!
+
+### References
+
+References to related issues, features
```
donnie-msft: Just a note that a simple list of Issues isn't always sufficient. A brief note with context might be nice to set as an expectation here.

I find we have a lot of 3+ years-old issues, which may be "Related", but how they are related may be complex and only resolved partially. A good spec should summarize these, at a minimum, so that you don't force the reader to read potentially 50 comments from 2017 or earlier.

Or, you can say that this Section is just a simple Citation list, and all context should be referenced in the other sections 😋

designs/SPEC_TEMPLATE.md
```diff
@@ -0,0 +1,48 @@
...
+
+Specifically call out the goals for this proposed design.
+
+## Non-Goals
+
+Call out things that are not in consideration for this design. 
+
+## Solution
+
+Describe the proposed solution in as much detail as you deem necessary.
```
donnie-msft, line 27: Do you specify the expected Release here? Or are specs completely independent from implementation schedule?
I know PMs often want to know when to announce things, so not sure if estimation comes into play for a spec or not.

donnie-msft (changes requested):
I'm glad we are formalizing templates! 👏

Assume that if something doesn't apply, leave it and write "N/A".

What about Documentation? Do we want to have a section to state the plan for creating New docs / modifying existing docs? This also might help explain what the Spec is doing if you can succinctly list the Documentation it affects, and gets the discussion going earlier rather than wayyyy out at release-time 👍

designs/SPEC_TEMPLATE.md
nkolev92: It's not a static value, I just didn't want a dead link. 
I've done the same for the issue.
I think it's fairly obvious.

designs/SPEC_TEMPLATE.md
nkolev92: > Requirements to me sounds like it should be "Constraints". What is unique to this work that you are having to avoid?

I don't really want this template to be "shocking" to people. 
It's a combination of the 4-5 specs I've seen used by team members and I've not seen constraints used. 

I'll just use Goals as I mentioned. 

It's a barebones guide, it doesn't have to capture everything.

designs/SPEC_TEMPLATE.md
nkolev92: I guess I misunderstood your constraints proposal.

Future Work is a generic enough name for any follow ups, which is why I've left it there.

designs/SPEC_TEMPLATE.md
nkolev92: That's what the considerations sections is for. 
This is exactly what the name suggests, references. 
You can link to a full issue, an issue comment, a PR or an article anywhere else, it's up to the author.

designs/SPEC_TEMPLATE.md
nkolev92: Up to the spec author. 
Some specs will discuss this, others won't. 
It really depends on the specific feature, the spec maturity etc.

nkolev92:
@donnie-msft 
Replied to all comments, remove the requirements section. 
PTAL.

designs/SPEC_TEMPLATE.md
donnie-msft: It doesn't have to capture everything, but why does it have to capture the same thing twice?
I'd still like more explicit difference here if these 2 sections are necessary.

designs/SPEC_TEMPLATE.md
donnie-msft: What part did you misunderstand? Should we discuss offline?

designs/SPEC_TEMPLATE.md
nkolev92: > I'd still like more explicit difference here if these 2 sections are necessary.

Feedback received. I removed the `Requirements` section before I re-requested review.

See: https://github.com/NuGet/Home/pull/9312#issuecomment-604036363

designs/SPEC_TEMPLATE.md
nkolev92: I removed `Requirements` section, everything can be captured goals as it's broader. 

Are you proposing anything for future work here or?

donnie-msft (approved):
"Goals" makes sense, and I'm approving since I believe I could add a Constraints section to a spec if I want, or at least just list them under Goals.

nkolev92:
> "Goals" makes sense, and I'm approving since I believe I could add a Constraints section to a spec if I want, or at least just list them under Goals.

I think adding Constraints/Requirements under goals is fine. 

These are just the bare minimum ones needed.

nkolev92:
Thanks all for your help on this. 

I will update onenote/wiki references to this template

Issue #4989, opened 2017-04-06
extern alias support for NuGet package references


davkean:
* [Design document](https://github.com/NuGet/Home/blob/dev/designs/PackageReference-Extern-Alias.md)

_From @fubar-coder on April 5, 2017 15:59_

Currently, when a NuGet package reference is added, there is no way to set the alias from the project system for the new style csproj projects.

This feature is needed, because NuGet package references don't result in direct assembly references any more and only those can have an alias.

My proposal is to add the alias(es) to all assemblies referenced for the NuGet package, but not the indirectly referenced NuGet packages.

_Copied from original issue: dotnet/project-system#1930_

notes

https://github.com/dotnet/sdk/issues/10947 The build tasks on (.NET Core SDK side)
https://github.com/dotnet/NuGet.BuildTasks/issues/70 The build tasks for the non-SDK based PackageReference
https://github.com/dotnet/project-system/issues/6011 Nomination updates on project-system side.

davkean:
Great suggestion. I'll move this over to the NuGet repo, they own the syntax for PackageReference.

andymac4182:
This would be great since we are running into this issue due to https://www.nuget.org/packages/StackExchange.Redis/ and https://www.nuget.org/packages/StackExchange.Redis.StrongName/

brianpos:
We have this issue to if we want to add both of these
https://www.nuget.org/packages/Hl7.Fhir.DSTU2
https://www.nuget.org/packages/Hl7.Fhir.STU3

wburgers:
@brianpos exactly the same problem here.
We would like to run both fhir versions next to each other.

brianpos:
@wburgers i have a project using the extern alias trick https://github.com/brianpos/FhirPathTester but do have to set manually after inclusion. So can't wait for this to be fixed.

JoseFMP:
+1, can't wait for this. I think this is important because right now different packages implementing same namespaces.

@brianpos could you share your trick please?

fubar-coder:
IIRC you can configure NuGet to store the NuGet packages in an packages folder instead of using the cache in the users home folder using the NuGet.config. This allows you to manually specify project-relative assembly references (via `Reference Update`) which can include the alias. The problem with this approach is, that you have to update the reference whenever the package version changes. However, you **can** restore the packages into folders without the package version of you don't have conflicting package versions.

JoseFMP:
[...]

Issue #2365, opened 2016-03-20
Enable EmbedInteropTypes for PackageReference (matching Packages.Config capability)


AArnott:
In the packages.config world, when references are actually added to the user's project file, the project system will add `<EmbedInteropTypes>true</EmbedInteropTypes>` metadata to the reference when its sniff-test tells it that the assembly should be linked rather than referenced. This is an important distinction for several reasons.

But in the project.json world, this is lost -- the project system has no say in the matter, and the build system seems to reference everything even when it should be linked. This will break scenarios when Visual Studio SDK extensions use NuGet packages for such references as `Microsoft.VisualStudio.Imaging.Interop.14.0.DesignTime` as the types won't be embedded, and although the extension might work on that developer's machine (who has the VS SDK installed), the extension will fail at runtime on the end user's machine because that assembly isn't present. It was supposed to be embedded into the extension author's assembly!

I discussed this in a design meeting with @davidfowl and @jasonmalinowski in mid-September, 2015. But I haven't heard that anything came of it afterward. The design IIRC was that the NuPkg should support a new `link` folder next to `lib` and `ref`. The assemblies in the `link` folder get passed to the compiler with `/link:` instead of `/ref:`. 

So a package with this file layout:

```
MyPackage.nuspec
lib\
   net45\
      MyRegularAssembly.dll
link\
   net45\
      MyContracts.EmbedME.dll
```

When consumed by a C# project would produce a command line like this:

```
csc.exe /ref:MyPackage\1.0.0\lib\net45\MyRegularAssembly.dll /link:MyPackage\1.0.0\link\net45\MyContracts.EmbedME.dll
```

Design Spec - https://github.com/NuGet/Home/wiki/Embed-Interop-Types-with-PackageReference

yishaigalatzer:
I think as a workaround, the vs packages can use msbuild targets for now. I'm moving this out.

AArnott:
Yes, as a workaround I've published http://nuget.org/packages/microsoft.visualstudio.sdk.embedinteroptypes

jaredpar:
@AArnott how would a consumer of a interop package differentiate if they wanted to link or not?  Today I can `/reference:EnvDTE` or `/link:EnvDTE`.  I would assume `EnvDTE` comes in a single package (maybe that's wrong).  As a consumer though how would I control the behavior?

AArnott:
[...]

PR #9319, opened 2020-03-18
Extern alias support for PackageReference 

nkolev92:
Design for https://github.com/NuGet/Home/issues/4989.

### DIFF
--- a/designs/PackageReference-Extern-Alias.md
+++ b/designs/PackageReference-Extern-Alias.md
@@ -1,4 +1,4 @@
-# Extern alias with PackageReference
+# Extern alias support with PackageReference
 
 * Status: In Review
 * Author(s): [Nikolche Kolev](https://github.com/nkolev92)
@@ -8,8 +8,8 @@
 ## Problem Background
 
 In the PackageReference world there is no way to add assembly specific metadata.
-Sometimes multiple packages might carry the same type, and in those cases the solution is use the alias feature, but there's no native way to do that in PackageReference.
-In contrast to packages.config, one can add the extra metadata to the assembly include.
+Sometimes multiple packages might carry the same type, and in those cases the solution is to use the alias feature. However, there's no native way to do that in PackageReference.
+In comparison, in packages.config, one can add the extra metadata to the assembly included because all the assembly references are written in the project file. 
 
 ## Who are the customers
 
--- a/designs/PackageReference-Extern-Alias.md
+++ b/designs/PackageReference-Extern-Alias.md
@@ -52,6 +52,10 @@ However the intuitiveness of this approach as it's analogous to [ProjectReferenc
 TODO NK - the specification. 
 TODO NK - implementation.
 
+NuGet implementation: 
+SDK based projects: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
+Csproj based projects: https://github.com/dotnet/NuGet.BuildTasks/blob/master/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs
+
 ## Future Work
 
 * Package metadata for transitive package references will be considered in the future. 
--- a/designs/PackageReference-Extern-Alias.md
+++ b/designs/PackageReference-Extern-Alias.md
@@ -49,20 +49,66 @@ However the intuitiveness of this approach as it's analogous to [ProjectReferenc
 
 ### Solution - Technical details
 
-TODO NK - the specification. 
-TODO NK - implementation.
+In PackageReference, everything about packages is in the assets file, as such the changes are the following: 
 
-NuGet implementation: 
-SDK based projects: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs
-Csproj based projects: https://github.com/dotnet/NuGet.BuildTasks/blob/master/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs
+In the csproj the specification is: 
+
+```xml
+<ItemGroup>
+    <PackageReference Include="NuGet.Contoso.Library.Signed"  Version="5.6.0" Aliases="signed" />
+</ItemGroup>
+```
+
+In the targets section of the assets file, where the compile items are listed out: 
+```json
+  "targets": {
+    ".NETCore,Version=v.3.0": {
+      "NuGet.Contoso.Library.Signed/5.6.0": {
+        "type": "package",
+        "compile": {
+          "lib/netstandard2.0/NuGet.Contoso.Library.Signed.dll": {
+            "Aliases" : "signed"
+          }
+        },
+        "runtime": {
+          "lib/netstandard2.0/NuGet.Contoso.Library.Signed.dll": {}
+        }
+      }
+    },
+```
+
+Note that only the compile items have the additional properties. 
+
+On the library side there are no changes. 
+
+The (LockFileItem)[https://github.com/NuGet/NuGet.Client/blob/4fef99532f4022504feec5f68c8501cbeadd3aed/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileItem.cs] type which represents an element in the compile list already has a collection of properties. 
+
+Specifically:
+
+```cs
+  public string Path { get; }
+
+  public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
+```
+
+There is no need for a change here. 
+
+The implementation of this feature spans multiple components. 
+Specifically the work items as follows: 
+
+* NuGet.Client makes the changes for processing the new metadata and adding to the project.assets.json
+* The build tasks on (.NET Core SDK side)[https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs] consume these changes
+* The build tasks on the (non-SDK based PackageReference)[https://github.com/dotnet/NuGet.BuildTasks/blob/master/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs] consume these changes.
+* Nomination updates on (project-system)[https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/PackageReference.xaml] side.
 
 ## Future Work
 
-* Package metadata for transitive package references will be considered in the future. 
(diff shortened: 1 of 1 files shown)


nkolev92:
cc @dsplaisted @wli3 @rainersigwald @davkean 
Would appreciate if you can take a look.

dsplaisted (approved):
Looks good.

I don't think you can specify the same alias for multiple different assemblies, so I don't know how this would work if we wanted to support transitive dependencies as is mentioned in the doc.

nkolev92:
> I don't think you can specify the same alias for multiple different assemblies, so I don't know how this would work if we wanted to support transitive dependencies as is mentioned in the doc.

I tested that and I can think it goes both. 

You can specify the same alias more than 1 assembly. 
You can also specify multiple aliases for the same assembly. 

Looks like all the aliasing is handled independently. 
[extern.zip](https://github.com/NuGet/Home/files/4351227/extern.zip)

//cc @jaredpar  Any issues with this proposal?

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,171 @@
...
+    <PackageReference Include="NuGet.Contoso.Library.Signed"  Version="1.26.0" Aliases="signed" />
+</ItemGroup>
+```
+
+* Applies to all the compile time assemblies.
+* Does not apply to transitive dependencies.
+* More than one alias is allowed in an equivalent fashion to [ProjectReference.Aliases](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.projectreference.aliases?). Specifically to define more than alias, specify `first,second`.
+
+The limitation of this approach is that it does not allow per assembly granularity. 
+However the intuitiveness of this approach as it's analogous to [ProjectReference.Aliases](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.projectreference.aliases?), and the fact that a large majority of public packages have only 1 assembly per package (per framework), were the deciding factor to adopt this approach.
```
zkat, line 50: Is there any escape hatch if folks are dealing with multi-assembly-per-fw packages? I assume [using this workaround?](https://github.com/NuGet/Home/pull/9319/files#diff-5e2137566c23ceec0613da9ed3b30dccR119)

I assume you already talked to potential customers about this being an acceptable sacrifice?

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,171 @@
...
+
+```cs
+  public string Path { get; }
+
+  public IDictionary<string, string> Properties { get; } = new Dictionary<string, string>();
+```
+
+There is no need for a change here. 
+
+The value provided in the Aliases attribute will be pass through as far as NuGet is concerned. NuGet will not validate that the value provided is a valid alias, that will be done at build time.
```
zkat: tiny nit:

```suggestion
The value provided in the Aliases attribute will be passed through as far as NuGet is concerned. NuGet will not validate that the value provided is a valid alias, that will be done at build time. 

```

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Support `Aliases` as a PackageReference property & provide a way to address transitively dependency out of the box
+
+A future evolution of the current approach could be the addition of metadata for transitive packages. 
+
+```xml
+<ItemGroup>
+    <PackageReference Include="StackExchange.Redis.StrongName.Signed" Version="1.26.0" Aliases="signed" />
+</ItemGroup>
+```
```
zkat: what's the difference between this one and the example above?? I think I'm missing something.

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,171 @@
...
+
+This approach drastically increases the scope of this work, but given that it's additive. 
+Given that we are not confident that this is how we want to address the transitive package metadata problems this is just listed here for completeness.
+
+### References
+
+* https://github.com/NuGet/Home/issues/4989
+* https://github.com/NuGet/Home/issues/5740
+* https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias
+* https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.projectreference.aliases?view=roslyn-dotnet
```
zkat: Can you add descriptions/titles for these links?

zkat (commented):
Just a first pass through. Sorry if my questions are silly. This is a nice proposal!!

designs/PackageReference-Extern-Alias.md
nkolev92: Yeah, the workaround is one approach. 
You can also define more than one alias per package/assembly, so you can be selective about how you use it. 

I have talked to individuals both internal and external but it's a fairly limited. 
The biggest sell for me is the fact that a very large majority of packages have 1 assembly only.

designs/PackageReference-Extern-Alias.md
nkolev92: Sure, I can do that.

designs/PackageReference-Extern-Alias.md
nkolev92: I think this example needs to be removed and then we just have the transitive metadata proposal.

nkolev92: Fixed it now.

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Solution - Technical details
+
+In PackageReference, everything about packages is in the assets file, as such the changes are the following: 
+
+In the csproj the specification is: 
+
+```xml
+<ItemGroup>
+    <PackageReference Include="NuGet.Contoso.Library.Signed"  Version="5.6.0" Aliases="signed" />
```
davkean, line 60: Call out that this is "comma, separated" so "signed, unsigned" should be allowed.

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,164 @@
...
+Sometimes multiple packages might carry the same type, and in those cases the solution is to use the alias feature. However, there's no native way to do that in PackageReference.
+In comparison, in packages.config, one can add the extra metadata to the assembly included because all the assembly references are written in the project file. 
+
+## Who are the customers
+
+All PackageReference customers
+
+## Requirements
+
+* A feature or guidance to allow the customers to specify an alias for an assembly.
```
davkean, line 20: Was there a embed interop type version of this? I can't remember? If not, can we just bite these both off at the same time.

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,164 @@
...
+## Considerations
+
+In addition to the proposed approach, 2 of other solutions were considered. 
+
+### Do nothing - Recommend the customers use the custom target workaround
+
+The workaround that allows per assembly granularity is the following:
+
+```xml
+  <Target Name="AddCustomAliases" BeforeTargets="FindReferenceAssembliesForReferences;ResolveReferences">
```
dominoFire, line 146: What would happen if NuGet only makes the changes to inject this target in PackageReference world, in a way that users don't need to add this target in their project files, instead of adding new elements to project.assets.json?

As far as I understand, in packages.config world, because all references are added to the project file, it is easier to customize it to add the exte

nkolev92: It's not really something we want users depending on. It's not a firm public extensibility point. 

It's an internal build target. Yes, you can technically play it to make your scenarios work, but we don't make guarantees that these things will always work the way they do now.

designs/PackageReference-Extern-Alias.md
nkolev92: Good point, will add it.

designs/PackageReference-Extern-Alias.md
nkolev92: There is!
That one is an author driven feature.
But hunting through the issues, I don't think it was fully implemented :(
As in NuGet did the work, but the build tasks are not consuming it so not very useful :(

Original issue: 
https://github.com/NuGet/Home/issues/2365

2 tasks that need completed:
https://github.com/dotnet/NuGet.BuildTasks/issues/69
https://github.com/dotnet/sdk/issues/2732

//cc @aortiz-msft We should have someone from NuGet ensure the partner changes are handled.

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,164 @@
...
+
+In the targets section of the assets file, where the compile items are listed out: 
+```json
+  "targets": {
+    ".NETCore,Version=v.3.0": {
+      "NuGet.Contoso.Library.Signed/5.6.0": {
+        "type": "package",
+        "compile": {
+          "lib/netstandard2.0/NuGet.Contoso.Library.Signed.dll": {
+            "Aliases" : "signed"
```
rainersigwald, line 72: What is the JSON representation if multiple aliases are specified?

designs/PackageReference-Extern-Alias.md
```diff
@@ -0,0 +1,164 @@
...
+* NuGet.Client makes the changes for processing the new metadata and adding to the project.assets.json
+* The build tasks on (.NET Core SDK side)[https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs] consume these changes
+* The build tasks on the (non-SDK based PackageReference)[https://github.com/dotnet/NuGet.BuildTasks/blob/master/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs] consume these changes.
+* Nomination updates on (project-system)[https://github.com/dotnet/project-system/blob/master/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/Dependencies/PackageReference.xaml] side.
+
+## Future Work
+
+* Package metadata for transitive package references will be considered in the future. For now the recommendation is to elevate that PackageReference to a direct dependency. 
+
+## Open Questions
```
rainersigwald, line 132: Is there Package Manager UI exposure for this, or project XML only?

rainersigwald (commented):
LGTM overall, couple of nits.

designs/PackageReference-Extern-Alias.md
nkolev92:
```json
"Aliases" : "signed,signed2"
```

I dont want NuGet to validate it, just pass it through. 
The compiler can complain if it's wrong.

designs/PackageReference-Extern-Alias.md
nkolev92: Nope, project XML only. Will call it out.

donnie-msft (approved):

Issue #9240, opened 2020-03-02
net5.0-<TargetPlatformIdentifier><TargetPlatformVersion> support


rrelyea:
A continuation of the work started here: https://github.com/nuget/home/issues/9089
Spec is still under some discussion. (eta for spec draft is 3/6/2020)

PR #9382, opened 2020-04-01
add in notes on plan for net5 platform feature

zkat:
Ref: https://github.com/NuGet/Home/issues/9240

### DIFF
--- a/designs/Net5PlatformAndPlatformVersion.md
+++ b/designs/Net5PlatformAndPlatformVersion.md
@@ -1,0 +1,177 @@
+# net5.0 TFM Platform and PlatformVersion implementation
+
+This document is meant to describe the various parts of the code that need to be updated in order to support platforms in net5.0 target framework identifiers. That is, the optional part after a `-` following `net5.0`. For example, `net5.0-ios14.1`. Note that this ONLY covers the platform part, because [initial support for `net5.0` has already been added](https://github.com/NuGet/NuGet.Client/pull/3265).
+
+For more details on the feature itself, please refer to [the full spec](https://github.com/terrajobst/designs/blob/master/accepted/2020/net5/net5.md).
+
+## class NuGetFramework
+
+Source: src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs
+
+### Fields
+
+Three additional fields will be added:
+
+```csharp
+        /// <summary>
+        /// Framework Platform (net5.0+)
+        /// </summary>
+        public string Platform
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// Framework Platform Version (net5.0+)
+        /// </summary>
+        public Version PlatformVersion
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// True if platform is non-empty (net5.0+)
+        /// </summary>
+        public bool HasPlatform
+        {
+            get { return !string.IsNullOrEmpty(Platform); }
+        }
+
+        /// <summary>
+        /// True if platform version is non-empty (net5.0+)
+        /// </summary>
+        public bool HasPlatformVersion
+        {
+            get { return IsNet5Era && !(PlatformVersion.Major == 0 && PlatformVersion.Minor == 0); }
+        }
+```
+
+### Constructors
+
+An additional constructor should be added that accepts `Version platformVersion` as its 4th argument. The current code in `NuGetFramework/3` should be moved to this new constructor, and `string profile` should be renamed to `string profileOrPlatform`. This field will be EITHER a legacy profile or, if net5.0+, a platform identifier.
+
+```csharp
+        public NuGetFramework(string framework, Version version, string profileOrPlatform)
+            : this(framework, version, profileOrPlatform, FrameworkConstants.EmptyVersion)
+        {
+        }
+
+        private const int Version5 = 5;
+
+        public NuGetFramework(string frameworkIdentifier, Version frameworkVersion, string profileOrPlatform, Version platformVersion)
+        {
+            if (frameworkIdentifier == null)
+            {
+                throw new ArgumentNullException("frameworkIdentifier");
+            }
+
+            if (frameworkVersion == null)
+            {
+                throw new ArgumentNullException("frameworkVersion");
+            }
+
+            _frameworkIdentifier = frameworkIdentifier;
+            _frameworkVersion = NormalizeVersion(frameworkVersion);
+
+            IsNet4Era = (_frameworkVersion.Major >= Version5 && StringComparer.OrdinalIgnoreCase.Equals(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, _frameworkIdentifier));
+
+            _frameworkProfile = IsNet5Era ? profileOrPlatform ?? string.Empty : string.Empty;
+            Platform = IsNet5Era ? profileOrPlatform ?? string.Empty : string.Empty;
+            PlatformVersion = NormalizeVersion(platformVersion ?? FrameworkConstants.EmptyVersion);
+        }
+```
+
+### GetDotNetFrameworkName
+
+This method needs to be updated to append the Platform and PlatformVersion. I believe the format is supposed to be:
+
+```csharp
+string.Format(CultureInfo.InvariantCulture, "Platform={0},PlatformVersion={0}", framework.Platform, framework.PlatformVersion);
+```
+
+### GetShortFolderName
+
+This method needs to be updated such that when `IsNet5Era` is `true` and `Platform` is not `null`, `"-{Platform}[PlatformVersion]` is appended to the `net5.0` short tfm. This should be similar to how profiles are appended.
+
+## class NuGetFrameworkFactory
+
+Source: src/NuGet.Core/NuGet.Frameworks/NuGetFrameworkFactory.cs
+
+### ParseFrameworkName
+
+This seems to be the core of the "long" tfm parsing logic. This needs to be modified to detect `Platform=` and `PlatformVersion=` and extract those accordingly. It should also skip parsing out `Profile` if it detects that the framework it's currently parsing is net5.0+. If an invalid Platform is detected, it should error out.
+
+### ParseFolder
+
+This seems to be the core of the "short" tfm parsing logic. As with `ParseFrameworkName`, this needs to be modified such that, when the identifier is net5.0+, it will check `FrameworkConstants.FrameworkPlatforms` and verify that the given platform is one of the accepted ones (per the net5.0 spec linked at top). If an invalid Platform is detected, it should error out.
+
+### RawParse
+
+This... mostly seems to do what I think it needs to do? It should probably have a big fat comment noting that we're piggybacking off profile parsing (keeping the net5.0+ dependent logic in `ParseFolder`). Maybe renaming variables to `profileOrPlatform` for clarity.
+
+## class FrameworkConstants
+
+Source: src/NuGet.Core/NuGet.Frameworks/FrameworkConstants.cs
+
+### (new!) FrameworkPlatforms
+
+A new `internal static HashSet<string> FrameworkPlatforms` field should be added and populated with all valid platforms (`android`, `ios`, `macos`, `tvos`, `watchos`, `windows`). As of this writing, it's **unclear** what to do about Tizen and Unity, and whether they'll be regular Platforms or a TFI or whatever.
+
+## class CompatibilityProvider
+
+Source: src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
+
+### IsCompatibleWithTargetCore
+
+This method needs to be modified such that if there's a Platform present in the `candidate`, the platform name has to match. Additionally, an `IsVersionCompatible(target.PlatformVersion, candidate.PlatformVersion)` should be made if both have a `PlatformVersion` included. This will need to check for null `PlatformVersion`s, as well.
+
+## class NuGetFrameworkFullComparer
+
+Source: src/NuGet.Core/NuGet.Frameworks/comparers/NuGetFrameworkFullComparer.cs
+
+### Equals
+
+An extra clause should be added here that checks that `Platform` and `PlatformVersion` are compared against.
+
+Additionally, there's maybe some leftover work, as per [this comment in PR #3265](https://github.com/NuGet/NuGet.Client/pull/3265/files#r388004831) as far as what to do with `Unsupported`. This might be getting worked on separately?
+
+## class CompatibilityTests
+
+Source: test/NuGet.Core.Tests/NuGet.Frameworks.Test/CompatibilityTests.cs
+
+`[Theory] Compatibility_FrameworksAreCompatible` data needs to be modified such that Platforms are supported, and all expected platforms are included in the tests.
+
+## class NuGetFrameworkParseTests
+
+Source: test/NuGet.Core.Tests/NuGet.Frameworks.Test/NuGetFrameworkParseTests.cs
+
+`[Theory] NuGetFramework_Basic` needs additional test cases to make sure Platform and PlatformVersion are being parsed correctly from the "long" tfm format.
+
+## class NuGetFrameworkTests
+
+Source: test/NuGet.Core.Tests/NuGet.Frameworks.Test/NuGetFrameworkTests.cs
+
+An additional test should be added to make sure things like `net50-android10` parse to `net5.0-android10.0`.
+
+## Remaining work/questions
+
+1. Ensure Pack target passes TargetPlatformVersion to .nuspec generator. (Use `0.0` as a sentinel)
+1. Fail parsing/creation of `NuGetFramework` when an unknown OS flavor is being used -- slam an Unsupported tfi into it??
+
+## Out of scope for this
+
+1. We need to make sure there's some kind of information/warning available when `net50` is used, since [we're going to warn now if the `.` is missing, starting from `net5.0`](https://github.com/NuGet/Home/issues/9215). I'm not sure what level to do this at.
+1. Design NuGet manifest entries for TargetPlatformVersion and TargetPlatformMinVersion (not sure what "design" is needed -- TPV should be treated as part of the tfm itself. TargetPlatformMinVersion is a separate, new attribute. This is a bit more involved than just passing it through.)
+1. Ensure Pack target passes TargetPlatformMinVersion to .nuspec generator.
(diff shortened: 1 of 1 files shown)


designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// True if platform is non-empty (net5.0+)
+        /// </summary>
+        public bool HasPlatform
+        {
+            get { return !string.IsNullOrEmpty(Platform); }
```
donnie-msft, line 39: If Platform is whitespace, will it fail from other validation?

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        /// <summary>
+        /// True if platform version is non-empty (net5.0+)
+        /// </summary>
+        public bool HasPlatformVersion
+        {
+            get { return IsNet5Era && !(PlatformVersion.Major == 0 && PlatformVersion.Minor == 0); }
```
donnie-msft, line 47: I'm not sure the comment matches the logic, and there may be a better way of identifying "empty" versions. My understanding from the [Version Class](https://docs.microsoft.com/en-us/dotnet/api/system.version?view=netframework-4.8#remarks) is that -1 is the unassigned value, and 0.0 would be a valid version:

> The value of Version properties that have not been explicitly assigned a value is undefined (-1).

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+
+A new `internal static HashSet<string> FrameworkPlatforms` field should be added and populated with all valid platforms (`android`, `ios`, `macos`, `tvos`, `watchos`, `windows`). As of this writing, it's **unclear** what to do about Tizen and Unity, and whether they'll be regular Platforms or a TFI or whatever.
+
+## class CompatibilityProvider
+
+Source: src/NuGet.Core/NuGet.Frameworks/CompatibilityProvider.cs
+
+### IsCompatibleWithTargetCore
+
+This method needs to be modified such that if there's a Platform present in the `candidate`, the platform name has to match. Additionally, an `IsVersionCompatible(target.PlatformVersion, candidate.PlatformVersion)` should be made if both have a `PlatformVersion` included. This will need to check for null `PlatformVersion`s, as well.
```
donnie-msft, line 128: Mention of `null` versions, but my other comment is on logic that refers to 0.0 as a missing Version.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+
+            if (frameworkVersion == null)
+            {
+                throw new ArgumentNullException("frameworkVersion");
+            }
+
+            _frameworkIdentifier = frameworkIdentifier;
+            _frameworkVersion = NormalizeVersion(frameworkVersion);
+
+            IsNet4Era = (_frameworkVersion.Major >= Version5 && StringComparer.OrdinalIgnoreCase.Equals(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, _frameworkIdentifier));
```
donnie-msft, line 78: typo?: IsNet5Era

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Fields
+
+Three additional fields will be added:
+
+```csharp
+        /// <summary>
+        /// Framework Platform (net5.0+)
+        /// </summary>
+        public string Platform
```
cristinamanum, line 19: Suggestion: maybe a different type Platform - to include the name and version?

designs/Net5PlatformAndPlatformVersion.md
cristinamanum, line 78: IsNet4Era or IsNet5Era?

cristinamanum: Will it be a public property?

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+
+### RawParse
+
+This... mostly seems to do what I think it needs to do? It should probably have a big fat comment noting that we're piggybacking off profile parsing (keeping the net5.0+ dependent logic in `ParseFolder`). Maybe renaming variables to `profileOrPlatform` for clarity.
+
+## class FrameworkConstants
+
+Source: src/NuGet.Core/NuGet.Frameworks/FrameworkConstants.cs
+
+### (new!) FrameworkPlatforms
```
cristinamanum, line 118: Could we get these through dotnet api call instead?

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
+# net5.0 TFM Platform and PlatformVersion implementation
+
+This document is meant to describe the various parts of the code that need to be updated in order to support platforms in net5.0 target framework identifiers. That is, the optional part after a `-` following `net5.0`. For example, `net5.0-ios14.1`. Note that this ONLY covers the platform part, because [initial support for `net5.0` has already been added](https://github.com/NuGet/NuGet.Client/pull/3265).
+
+For more details on the feature itself, please refer to [the full spec](https://github.com/terrajobst/designs/blob/master/accepted/2020/net5/net5.md).
```
zivkan, line 5:
```suggestion
For more details on the feature itself, please refer to [the full spec](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md).

```

designs/Net5PlatformAndPlatformVersion.md
zivkan, line 19:
```suggestion
        public string PlatformIdentifier
```
reason: https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#mapping-to-properties

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+        public Version PlatformVersion
+        {
+            get;
+            private set;
+        }
+
+        /// <summary>
+        /// True if platform is non-empty (net5.0+)
+        /// </summary>
+        public bool HasPlatform
```
zivkan, line 37: I know other properties in the class have `HasWhatever` properties, but I don't like the pattern. Where is it used? Why doesn't the caller do `!string.IsNullOrEmpty(PlatformIdentifier)` instead?

designs/Net5PlatformAndPlatformVersion.md
zivkan: Also, what's wrong will null? This is a valid use for null values. It seems wasteful to force the property to be not-null, and then invent our own convention for how to represent "not specified"

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+            }
+
+            _frameworkIdentifier = frameworkIdentifier;
+            _frameworkVersion = NormalizeVersion(frameworkVersion);
+
+            IsNet4Era = (_frameworkVersion.Major >= Version5 && StringComparer.OrdinalIgnoreCase.Equals(FrameworkConstants.FrameworkIdentifiers.NetCoreApp, _frameworkIdentifier));
+
+            _frameworkProfile = IsNet5Era ? profileOrPlatform ?? string.Empty : string.Empty;
+            Platform = IsNet5Era ? profileOrPlatform ?? string.Empty : string.Empty;
+            PlatformVersion = NormalizeVersion(platformVersion ?? FrameworkConstants.EmptyVersion);
```
zivkan, line 82: why empty values to represent values that aren't defined?  It's a perfect use-case for null. I feel like I'm missing information about why `_frameworkProfile` already does this.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+        }
+```
+
+### GetDotNetFrameworkName
+
+This method needs to be updated to append the Platform and PlatformVersion. I believe the format is supposed to be:
+
+```csharp
+string.Format(CultureInfo.InvariantCulture, "Platform={0},PlatformVersion={0}", framework.Platform, framework.PlatformVersion);
```
zivkan, line 91: Maybe we need to check with the .NET team. I'm not sure `Platform=` is right. Might be `PlatformIdentifier=`.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+        /// </summary>
+        public bool HasPlatformVersion
+        {
+            get { return IsNet5Era && !(PlatformVersion.Major == 0 && PlatformVersion.Minor == 0); }
+        }
+```
+
+### Constructors
+
+An additional constructor should be added that accepts `Version platformVersion` as its 4th argument. The current code in `NuGetFramework/3` should be moved to this new constructor, and `string profile` should be renamed to `string profileOrPlatform`. This field will be EITHER a legacy profile or, if net5.0+, a platform identifier.
```
zivkan, line 53: I think platform identifier and platform version should be two new parameters, rather than re-using profile depending on context (framework & version). With the current implementation, it makes it impossible for non-.NET TFMs (like `tizan` or `unity`) to use platform identifier and platform version, if they would ever want to in the future. By having profile and platform as separate parameters, by convention .NET will only use one of the two, depending on the framework version, but it doesn't prohibit other use-cases.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+## class NuGetFrameworkTests
+
+Source: test/NuGet.Core.Tests/NuGet.Frameworks.Test/NuGetFrameworkTests.cs
+
+An additional test should be added to make sure things like `net50-android10` parse to `net5.0-android10.0`.
+
+## Remaining work/questions
+
+1. Ensure Pack target passes TargetPlatformVersion to .nuspec generator. (Use `0.0` as a sentinel)
+1. Fail parsing/creation of `NuGetFramework` when an unknown OS flavor is being used -- slam an Unsupported tfi into it??
```
zivkan, line 161: I'm not sure. I'd like others opinions. In a future .NET, if they want to support new platforms, should they be dependant on NuGet updating our platform whitelist, before they can even progress on other parts of the build system?  I'm happy for NuGetFramework's `IsSupported` to return false if it's not a whitelisted platform, but I just don't know how I feel about returning the unsupported TFI.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Out of scope for this
+
+1. We need to make sure there's some kind of information/warning available when `net50` is used, since [we're going to warn now if the `.` is missing, starting from `net5.0`](https://github.com/NuGet/Home/issues/9215). I'm not sure what level to do this at.
+1. Design NuGet manifest entries for TargetPlatformVersion and TargetPlatformMinVersion (not sure what "design" is needed -- TPV should be treated as part of the tfm itself. TargetPlatformMinVersion is a separate, new attribute. This is a bit more involved than just passing it through.)
```
zivkan, line 166: manifest? Are you referring to the nuspec file, or something else? I guess not given the list item below.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+1. Fail pack when packing equivalent TFMs like net5.0-ios and net5.0-ios13.0 (note: we do want to eventually allow multi-targeting with equivalent TFMs, but that's out of scope for this work, and [seems to be a Big Task](https://github.com/NuGet/Home/issues/5154) -- also, we don't support them right now, so this isn't an issue yet)
+1. Produce an error when a OS flavored .NET 5 TFM is consumed but the package has no TPV in either folder name of the manifest. Is this supposed to be during restore??
+1. Produce a warning when a project consumes a library with a higher `TargetPlatformMinVersion` (related to 5154)
+1. When present, msbuild props like `TargetPlatformIdentifier` and `TargetPlatformVersion` should supersede the corresponding data in `NuGetFramework` objects. (this is related to 5154, too)
+
+## Follow-up
+
+1. Compatibility tables need to be extended to include compatibility for all the Xamarin TFMs to appropriate `net5.0-xxxx1.2` TFMs. I'm not sure we do anything similar right now, so this might actually be new functionality? -- send an email to Immo because our understanding is that we're not actually doing this. cc nikolche/rob
```
zivkan, line 175: My memory from the meetings were that old Xamarin TFMs were going to be incompatible, and target fallback could be used if desired. But it's good to check.

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+1. Do we care about `netcoreapp5.0` and `net5.0` both being used? Are those the "equivalent TFMs" we should error on?
```
zivkan, line 176: I think we should at least warn

designs/Net5PlatformAndPlatformVersion.md
```diff
@@ -0,0 +1,177 @@
...
+1. Produce a warning when a net5.0 project references netcoreapp3.1 (or earlier) with WPF/WinForms (what? How do we even do this????) -- we apparently did this previously, but we decided not to do it in our code and let the SDK take care of it?
```
zivkan, line 177: if NuGet was to do this, we'd have to check the dependency to check if tfi==netcoreapp and tfv=3.1 (or 3.0?) and if there's a framework reference to the WPF/WinForms package. 

edit: the more I think about this, the more I feel like it should be in the SDK, not NuGet.

designs/Net5PlatformAndPlatformVersion.md
nkolev92: The "no null" thing is something I pushed for. 

The reason is consistency. 
The platform version is expected to be declared in a similar way as the target framework version, see: https://nugettoolsdev.azurewebsites.net/5.5.0-preview.2.6382/parse-framework?framework=net. 

The serialization should be predictable too.
You'd get a v0.0 and that's a good enough experience.

If you have a TPI, you need a TPV. 

TPI needs a valid value to create a NuGetFramework, having a not declared TPV is weird. We need a default that makes more sense. 

0.0 is what TFV does, so we should do the same.

designs/Net5PlatformAndPlatformVersion.md
nkolev92: I think they should be new parameters as well.


I think we should just have a 4 val constructor, TFI, TFV, TPI & TPV.

designs/Net5PlatformAndPlatformVersion.md
zkat: it'll fail because it's an invalid TPI. Just as if you'd used `-foo1.2` as a Platform.

designs/Net5PlatformAndPlatformVersion.md
zkat: yeah, this was a long conversation. If you still want to chat about it, we can jump on a call. Let me know, @zivkan and @donnie-msft!

designs/Net5PlatformAndPlatformVersion.md
zkat: oh thank you. I needed to update this after we changed to 0.0 as a sentinel :)

designs/Net5PlatformAndPlatformVersion.md
zkat: hah yes.

zkat: IsNet5Era is an `internal` property.

designs/Net5PlatformAndPlatformVersion.md
zkat: I asked the SDK team about this in our sync meeting, and consensus was that this should probably live in NuGet, but that errors might, actually, be a bit better if we moved it to the SDK. It makes implementation more complicated, potentially, because we'd have to wait on dotnet.

designs/Net5PlatformAndPlatformVersion.md
zkat: I agree, but this is what @nkolev92 pushed for, and I ended up conceding. Happy to jump on a call if you want to bring back the discussion. fwiw, I agree that this is a good use-case for null, in general.

designs/Net5PlatformAndPlatformVersion.md
zkat: `TargetProfileIdentifier` uses `Profile=`, which is why I did it this way.

designs/Net5PlatformAndPlatformVersion.md
nkolev92: For reference :) 
https://github.com/NuGet/Home/pull/9382#discussion_r404417914

designs/Net5PlatformAndPlatformVersion.md
zkat: yeah, nuspec :)

designs/Net5PlatformAndPlatformVersion.md
zkat: Turns out -- they're gonna be compatible! The current spec is correct and up to date.

designs/Net5PlatformAndPlatformVersion.md
zkat: Per the sdk/msbuild sync meeting, we will actually be erroring out on this.

designs/Net5PlatformAndPlatformVersion.md
zkat: This is going to be done by the SDK

designs/Net5PlatformAndPlatformVersion.md
zkat: Doing it this way makes it consistent with how TFI and TFV are handled, though, so I prefer the split version. Also makes those fields more straightforward to access.

designs/Net5PlatformAndPlatformVersion.md
zkat: I think `!string.IsNullOrEmpty()` isn't as descriptive as having a boolean to check against that knows the semantics of the value being missing. That said, I'll remove this for now, because I don't know where we'll use it -- I might bring it back if I find a case where it's actually needed for public API purposes.

designs/Net5PlatformAndPlatformVersion.md
donnie-msft: No need to chat, but maintain that the comment is inaccurate.

JonDouglas:
Hello!

Actions to take:

JonDouglas:
I am closing this PR as it may be abandoned with no reviews since. If it's important to include in the repository, please re-open the PR & assign reviewers.

Issue #9600, opened 2020-05-23
`dotnet list` command supports `--verbosity` option


joelverhagen:
## Details about Problem

NuGet product used: dotnet.exe
dotnet.exe --version: 3.1.202
OS version: Windows 10, version 1909
Worked before? no

## Detailed repro steps so we can see the same problem

1. Create a new project, e.g. `dotnet new console`
1. Add a package, e.g. `dotnet add package NuGet.Packaging`
1. List deprecated packages, e.g. `dotnet list package --deprecated`

No HTTP `GET` and `OK` log lines are visible, unlike `dotnet restore --verbosity normal`, `dotnet nuget push`, and `dotnet nuget delete`. Especially when listing many packages or using `--include-transitive` it can lead to the command hanging for a long time without any visible output.

It would be helpful to have a verbosity switch like `dotnet restore` such as `--verbosity normal` to show the HTTP spew. This would allow understanding why the operation is slow and if one particular source is behaving badly.

You can do this investigation with restore via `dotnet restore --verbosity normal`.

I think it's right that the log lines are not shown by default (there are a lot sometimes) but there should be a way to see them without launching Fiddler or some other MITM proxy.

I think it's hardcoded to not allow this here:
https://github.com/NuGet/NuGet.Client/blob/eb9ad919724e5ef435d6321b775868c9f2c0f9a2/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs#L649

### Verbose Logs

> Please include verbose logs (NuGet.exe <COMMAND> -verbosity detailed | dotnet.exe <COMMAND> --verbosity diag | etc...)

Well, that's the problem... 😂

```
PS C:\Users\joelv\Desktop\testproj> dotnet list package --deprecated --verbosity diag
Unrecognized command or argument '--verbosity'
Unrecognized command or argument 'diag'
Usage: dotnet list <PROJECT | SOLUTION> package [options]

Arguments:
  <PROJECT | SOLUTION>   The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.

Options:
  -h, --help                                Show command line help.
  --outdated                                Lists packages that have newer versions.
  --deprecated                              Lists packages that have been deprecated.
  --framework <FRAMEWORK | FRAMEWORK\RID>   Chooses a framework to show its packages. Use the option multiple times for multiple frameworks.
  --include-transitive                      Lists transitive and top-level packages.
[...]

PR #9630, opened 2020-05-30
Add design for dotnet list package --verbosity

joelverhagen:
Related to https://github.com/NuGet/Home/issues/9600.

See the Markdown view for details and context:
image

### DIFF
--- a/designs/Package-List-Verbosity.md
+++ b/designs/Package-List-Verbosity.md
@@ -158,6 +158,59 @@ Options:
   <b>-v, --verbosity &lt;LEVEL&gt;</b>                   <b>Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].</b>
 </pre>
 
+### Example output
+
+This is output of the invocation of `dotnet list package --outdated --framework netcoreapp2.1` for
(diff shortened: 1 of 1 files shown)


anangaur:
Can you post output for default (minimal?), normal and other verbosity levels?

joelverhagen:
@anangaur - added that to the end of the spec

anangaur:
To me the normal verbosity should be the default and the amount of normal verbosity should be same as what is minimal today :)

However, I see that minimal is default verbosity level for dotnet implying customers see the right verbosity using dotnet.exe. However, if a command is also available with nuget.exe and MSBuild.exe does it mean that customers would see different/more verbose output with these tools? If so, this is odd.

joelverhagen:
> To me the normal verbosity should be the default and the amount of normal verbosity should be same as what is minimal today :)

Yes normal seems like the default word, but the other `dotnet` commands have already made this decision. It just so happens that the default verbosity (minimal) will cause the output of `dotnet list package` to look the same after this change is implemented. So that works in our favor it thing. I believe the right user experience is to hide HTTP messages but default.

I'll note that this does align with nuget.exe list (which has entirely different behavior, but the closest match to dotnet list package). In short nuget.exe list also hides HTTP by default but shows it if you increase verbosity. So that's matching-ish.

> However, if a command is also available with nuget.exe and MSBuild.exe does it mean that customers would see different/more verbose output with these tools?

This command is only available on the `dotnet` CLI so it's a non-issue here.

nkolev92 (approved):
This is a beautifully written and very detailed spec. 

👏 👏 👏 👏 

I am personally not concerned about the extra logs. 
Warnings should be surfaced to the customer.

PR #9657, opened 2020-06-08
Add spec for INuGetProjectServices

zivkan:


designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,99 @@
...
+}
+
+public sealed class GetInstalledPackagesOptions
+{
+    public static GetInstalledPackageOptions Default { get; }
+
+    // in the future may include options such as WaitForNomination, WaitForRestore.
+}
+
+public sealed class NuGetProjectNotReadyException : InvalidOperationException
```
nkolev92: I'm not sure we'd want to extend from InvalidOperationException. This implies a user error. That's not necessarily true. 
Maybe we tell customers they need to couple this with an operation progress call to check if the project is ready.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,99 @@
...
+
+* APIs for:
+  * InstallPackageAsync
+  * UninstallPackageAsync
+
+## Open Questions
+
+* How to handle errors
+  * CPS projects (SDK-style .NET projects) work by: On project load, CPS sends NuGet a "nomination" with project/package information. Therefore, NuGet can't return packages in GetPackagesAsync until CPS nomination complete. With IVsPackageInstallerServices, we throw InvalidOperationException if we haven't got nomination info. Given that the new API needs to work over remoting scenarios, how should it be done?  There are multiple options
+    1. Throw exception (task in faulted state). This might appear as a RemotingException to callers, so maybe hard to distinguish between different types of exceptions?
```
nkolev92: The remote invocation exception has a code. So we can basically just say Code `1` means, this, code `3` means that.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,99 @@
...
+    2. Return additional error information. This might be a tuple or a result class. Should we use a type extending `Exception` (stack trace probably wouldn't transport), or have `string Code` and `string Message`?  Or just a `bool Success`, without details about why it was unsuccessful?
+* GetInstalledPackagesAsync
+  * Should returned package version be the resolved version, or the requested range, or the lower bound of the requested range? Leaning towards requested range. It means
```
nkolev92: Personally, i think we sohuld do the installed version only.
We are not doing that right now for PR, so maybe the first implementation can just do the same thing as before.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,99 @@
...
+    1. Throw exception (task in faulted state). This might appear as a RemotingException to callers, so maybe hard to distinguish between different types of exceptions?
+    2. Return additional error information. This might be a tuple or a result class. Should we use a type extending `Exception` (stack trace probably wouldn't transport), or have `string Code` and `string Message`?  Or just a `bool Success`, without details about why it was unsuccessful?
+* GetInstalledPackagesAsync
+  * Should returned package version be the resolved version, or the requested range, or the lower bound of the requested range? Leaning towards requested range. It means 
+    * Should we return both requested and resolved versions?  Leaning to no: most components & extensions wouldn't care. If wanted, a PackageReference (PR) specific service could be added, which allows it to provide PR specific properties.
+  * Should there be options to wait for nomination, rather than throwing an exception?
+  * If `NuGetPackage.Version` is the resolved version, do we distinguish between project not nominated vs restore not complete?  What about if restore failed?
```
nkolev92: I think this should be handled the same way as project not ready. Maybe we just have a different error code.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,99 @@
...
+* NuGetPackage could have property saying if package is direct or transitive. (future work)
```
nkolev92: I think would be very relevant. I'd be in favor of adding it sooner rather than later if possible.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: My intention was to be compatible with `IVsPackageInstallerServices.GetInstalledPackages`. But since this is a new API not intended to be compatible, then you're right, I don't think it makes sense.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>Package information could not be retrieved because the project is in an invalid state</summary>
+    /// <remarks>If a project has an invalid target framework value, or a package reference has a version value, NuGet may be unable to generate basic project information, such as requested packages.</remarks>
+    ProjectInvalid
+}
+
+/// <summary>Basic information about a package</summary>
+public sealed class InstalledPackage
+{
+    /// <summary>The package id.</summary>
+    public string Id { get; }
```
dominoFire, line 84: Why not [`PackageIdentity`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/Core/PackageIdentity.cs) instead o defining `Id` and `Version` ?

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,122 @@
...
+
+    /// <summary>The package version</summary>
+    /// <remarks>
+    /// If the project uses packages.config, this will be the installed package version.
+    /// If the project uses PackageReference, this would ideally be the resolved package version, but may be the requested package version.
```
dominoFire: Then, you would need a way to know if the project is PackageReference or packages.config style. I suggest adding a field in `GetInstalledPackagesResult` for that

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,122 @@
+# INuGetProjectServices
+
+* Status: In Review
+* Author(s): [Andy Zivkovic](https://github.com/zivkan)
+* Issue: [1](https://github.com/NuGet/Home/issues/9577) IVsPackageInstallerServices should be async
+
+## Problem Background
+
+The existing IVsPackageInstallerServices service is synchronous, and is being called by several components on the UI thread, causing UI delays in Visual Studio. We are creating new async APIs that will allow us to unblock the UI thread, and use the Visual Studio Service Broker to work in Visual Studio CodeSpaces (and maybe Live Share).
```
dominoFire, line 9: Are you going to replace the inner bits of NuGet IVs* with these async interfaces/implementations to avoid code duplication ? Or, Are IVs* and this interface for different purposes?

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,122 @@
...
+
+* This will not be a 1:1 replacement for any of NuGet's existing `IVs` interfaces.
+
+## Solution
+
+```cs
+/// <summary>Service to interact with projects in a solution</summary>
+public interface INuGetProjectServices
+{
+    /// <summary>Gets the list of packages installed in a project.</summmry>
```
dominoFire: In what time? After restore succeeds? at caller time?

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,122 @@
...
+    /// <summary>The package id.</summary>
+    public string Id { get; }
+
+    /// <summary>The package version</summary>
+    /// <remarks>
+    /// If the project uses packages.config, this will be the installed package version.
+    /// If the project uses PackageReference, this would ideally be the resolved package version, but may be the requested package version.
+    /// </remarks>
+    public string Version { get; }
+
```
dominoFire, line 99: Should we expose a data model similar of what represents an installed package in project.assets.json?

dominoFire (commented):
Some comments on the data classes fields

designs/NuGetExtensibility/INuGetProjectServices.md
nkolev92: We never use our own IVs* interfaces. 

They are only used for extensibility purposes.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
+# INuGetProjectServices
+
+* Status: In Review
+* Author(s): [Andy Zivkovic](https://github.com/zivkan)
+* Issue: [1](https://github.com/NuGet/Home/issues/9577) IVsPackageInstallerServices should be async
```
nkolev92: Issue number should be 9577 :D

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+public interface INuGetProjectServices
+{
+    /// <Summary>Gets the list of packages installed in a project.</Summary>
+    /// <param name="projectId">Project ID (GUID).</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>
+    /// The list of packages in the project.
+    /// </returns>
+    /// <exception cref="System.ArgumentException">When projectId is not a guid.</exception>
+    Task<GetInstalledPackagesResult> GetInstalledPackagesAsync(string projectId, CancellationToken cancellationToken);
```
nkolev92: Consider using

`InstalledPacakgesResult` instead.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+}
+
+/// <summary>Result of a call to INuGetProjectServices.GetInstalledPackagesAsync</summary>
+public sealed class GetInstalledPackagesResult
```
nkolev92: Same as above, consider removing the Get out of the name.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+{
+    /// <summary>The status of the result</summary>
+    public GetInstalledPackageResultStatus Status { get; }
+
+    /// <summary>List of packages in the project</summary>
+    /// <remarks>May be null if <see cref="Status"/> was not successful</remarks>
+    public IReadOnlyCollection<NuGetInstalledPackage> Packages { get; }
```
nkolev92, line 50: Given that thius has count, I imagine it doesn't affect performance, but I'm just wondering if we should use IReadOnlyList instead to provide for index access.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+    /// <remarks>This typically happens shortly after the project is loaded, but the project system has not yet informed NuGet about package references</remarks>
+    ProjectNotReady,
+
+    /// <summary>Package information could not be retrieved because the project is in an invalid state</summary>
+    /// <remarks>If a project has an invalid target framework value, or a package reference has a version value, NuGet may be unable to generate basic project information, such as requested packages.</remarks>
+    ProjectInvalid
+}
+
+/// <summary>Basic information about a package</summary>
+public sealed class NuGetInstalledPackage
```
nkolev92, line 81: Looks like we might need the PackagePath but you can verify that with the test team. https://github.com/NuGet/NuGet.Client/blob/f95e1aff6ceb901706ed19cd6cbd63ec05d40b07/src/NuGet.Clients/NuGet.VisualStudio.Implementation/Extensibility/VsPackageInstallerServices.cs#L93

designs/NuGetExtensibility/INuGetProjectServices.md
nkolev92: I say definitely resolved version, use the requested if necessary.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+
+* Other required APIs: (incomplete list)
+  * InstallPackageAsync
+  * UninstallPackageAsync
+
+## Open Questions
+
+* GetInstalledPackagesAsync
+  * Should returned package version be the resolved version, or the requested range, or the lower bound of the requested range? Leaning towards requested range. It means 
+    * Should we return both requested and resolved versions?  Leaning to no: most components & extensions wouldn't care, plus it doesn't make sense in `packages.config` projects. If wanted, a PackageReference (PR) specific service could be added, which allows it to provide PR specific properties.
```
nkolev92: Think resolved makes more sense given the package path.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+  * Should there be options to wait for nomination, rather than returning error?
+  * If `NuGetInstalledPackage.Version` is the resolved version, do we distinguish between project not nominated vs restore not complete?  What about if restore failed?
```
nkolev92: If restore failed and the package is not on disk, i think we just need to provide null in the package path potentially. Or something along those lines to match the current IVs* API.

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+/// <summary>Service to interact with projects in a solution</summary>
+public interface INuGetProjectServices
+{
+    /// <Summary>Gets the list of packages installed in a project.</Summary>
+    /// <param name="projectId">Project ID (GUID).</param>
+    /// <param name="cancellationToken">Cancellation token.</param>
+    /// <returns>
+    /// The list of packages in the project.
+    /// </returns>
+    /// <exception cref="System.ArgumentException">When projectId is not a guid.</exception>
```
nkolev92: Change this to RemoteInvocation exception. 

Ensure AA + other codespaces folks review it.

nkolev92 (commented):
Looks great overall!

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,138 @@
...
+    /// <param name="isDirectDependency"></param>
+    /// <returns></returns>
+    public static NuGetInstalledPackage CreateNuGetInstalledPackage(string id, string version, bool isDirectDependency);
+
+    /// <summary>Create a <see cref="GetInstalledPackageResultStatus"/></summary>
+    /// <param name="status"></param>
+    /// <param name="packages"></param>
+    /// <returns></returns>
+    public static GetInstalledPackagesResult CreateGetInstalledPackagesResult(GetInstalledPackageResultStatus status, IReadOnlyCollection<NuGetInstalledPackage> packages);
+}
```
AArnott: While looking forward to C# 9 record types is admirable, requiring a factory to create what is a simple serializable data type is _not_ API friendly, nor is it in anyway consistent with our other remotable service interface data types defined elsewhere. 
We do use immutable data types, but we define a public constructor with a parameter for each property. I am not sure, but I suspect records may offer similar public API. But even if it doesn't, a public constructor instead of a factory method on some other random type is very difficult to discover.

designs/NuGetExtensibility/INuGetProjectServices.md
AArnott: (mind the typo in @nkolev92's prescribed name).

AArnott (changes requested):
I *love* how thoroughly you xml doc comment your APIs here, including examples and explanations for why/when certain values would occur. Very well done.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: `PackageIdentity` is not in the BCL, and we've had problem is in the past where other components ship NuGet dlls in Visual Studio. Resharper had to resign NuGet's dlls with a different strong name key to get around this issue. Therefore, we should never use types defined in any NuGet assembly in our VS extensibility APIs.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: This is an implementation question, whereas this document is the API design. Resolving as out of scope (but it's a very relevant question for the PR that implements this API).

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: I don't believe so. The assets file is NuGet's "internal" representation of whatever we need to know (well, the .NET SDK knows about the format too). But callers of this extensibility API shouldn't need to know. Plus it would mean that each time we change the assets file we'd need to create new APIs here, and maintaining backwards compatibility with the older APIs becomes significant technical debt.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: The RPC contracts best practices document (sadly not public, so I won't link it) says:

> Use `IReadOnlyCollection<T>` for unordered collections or `IReadOnlyList<T>` for ordered collections instead.

Order isn't important for NuGet package definitions, hence the reason I chose that type.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: I understand and agree with your comment. I think discoverability can be addressed to some extent with XMLdoc that points to the factory (see new commit once I push it). This might be insufficient for types that API customers need to create. But at this time, the two classes defined are both return-value only. Meaning the most likely reason customers might use it is for tests, assuming they don't have their own abstraction of our result (Roslyn does, so is unlikely to create instances).  We can make constructors public later without it being a breaking change, but we can't take them away, so I'm still leaning towards being optimistic about record types at this time.

donnie-msft (approved):

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Non-Goals
+
+* This will not be a 1:1 replacement for any of NuGet's existing `IVs` interfaces.
+
+## Solution
+
+```cs
+/// <summary>Service to interact with projects in a solution</summary>
+/// <remarks>This interface should not be implemented. New methods may be added over time.</remarks>
```
loic-sharma, line 31: `This interface should not be implemented. New methods may be added over time.`

Isn't this what default interface methods are for? 😄

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Who are the customers
+
+Visual Studio component and extension developers.
+
+## Goals
+
+* Can be called on any thread (ideally background thread, but work if called on UI thread).
+* Do not block caller (implement async properly, return incomplete task instead).
+* Works with Visual Studio Codespaces, and hopefully Live Share.
```
loic-sharma, line 21: Can I make a console application that leverages this API? Should this be a goal to support this scenario? If not, why not? Just curious :)

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: This assembly is the contract for NuGet's Visual Studio extensibility. When we add a method like `InstallPackageAsync`, there's no way to add a default implementation that is meaningful, other than throw new not implemented exception.

This remark is a signal that anyone using this interface should not implement a wrapper by "inheriting" the interface, particularly in production code. I'd also suggest they don't implement the interface in a test project to use as a mock for their own code.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: No, this is the contract for extensibility that we provide within Visual Studio.

loic-sharma: @zivkan answered more of my questions offline, closing this off now

designs/NuGetExtensibility/INuGetProjectServices.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Future Work
+
+* Other required APIs: (incomplete list)
+  * InstallPackageAsync
+  * UninstallPackageAsync
+
+## Considerations
+
+* Use `System.Guid` for project identifiers
```
nkolev92, line 139: Looks like we're using guids :)

nkolev92 (approved):
The rest of it looks fine. 

I'm ok with either going the public constructor direction or the factory direction for the NuGetInstalledPackage type.

Personally I think everything is good if the type is not mutable, records would be nice, but we won't lose too much with that.

designs/NuGetExtensibility/INuGetProjectServices.md
zivkan: Yes, talking to project system people, and I got the same feedback from the RPC contracts api reviewers group, other identifiers, like "unique id" can cause problems when projects are renamed. So multiple people recommended we use project guid.

Someone from the RPC contracts api reviewers group also recommended using `System.Guid`, rather than string. The two points I listed below were basically copy-paste from their advice :)

designs/NuGetExtensibility/INuGetProjectServices.md
AArnott: OK. It sounds like you've considered all the angles I was thinking of, so I yield to your judgment.

Issue #9394, opened 2020-04-03
NuGet should support multiple config files in %APPDATA%\NuGet directory


debonte:
## Details about Problem
NuGet product used: VS UI
NuGet version: 5.6.0.6489
VS version: 16.7
Worked before?: No

## Detailed repro steps so we can see the same problem
1. Unzip [SomeOther.zip](https://github.com/NuGet/Home/files/4430294/SomeOther.zip)
2. Place SomeOther.config in %APPDATA%\NuGet. This config file declares a single package source named "IssueRepro"
3. Start VS and open/create some project
4. Open NuGet Package Manager
5. Expand "Package source" combo box

**Expected:** IssueRepro source is one of the options.

**Actual:** IssueRepro source is not shown as an option.

**Workarounds**:
1. Place the file in %ProgramFiles(x86)%\NuGet\Config instead. 
1. Merge the IssueRepro package source into %APPDATA%\NuGet\NuGet.Config's list of package sources.

Note that what I really care about is fallback folders, but I'm using package sources for ease of repro.

## Reasoning
The XAML and WinForms teams want to allow control library installers to add user-specific NuGet fallback folders by having their SDK installers drop config files in %APPDATA%\NuGet. On uninstall this file would be deleted.

We considered two alternatives which I mentioned as workarounds above:
1. Require SDK installers to drop config files in %ProgramFiles(x86)%\NuGet\Config instead. However, we've gotten feedback from control vendors that they would like the option of installing without elevation which prevents the installer from being able to write to this directory.
2. Require SDK installers to modify the one %APPDATA%\NuGet\NuGet.config file. However this seems error prone.

@rrelyea @zivkan @aortiz-msft @mgoertz-msft

nkolev92:
PR out https://github.com/NuGet/NuGet.Client/pull/3421

Issue #9574, opened 2020-05-18
[Test Failure] Closing invalid Nuget.config xml box hangs VS


v-luzh:
## Details about Problem
NuGet version: NuGet Client Dev\5.7.0.6612
VS Version: Master\30114.135
OS Version: Windows-10-Enterprise-RS5
## Detailed repro steps
1. Go to %appdata%\nuget\Nuget.Config
2. Open Nuget.Config and edit the file so that it contains invalid xml. In this case, mine contains multiple root level elements.
3. Close Nuget.Config and open VS
4. Create a new ASP.NET (NET core) web app.
5. The window below appears. Close it
## Expected
The window closed successfully without VS hanging.
## Actual
VS hangs as below screenshot.
![image](https://user-images.githubusercontent.com/20632625/82193957-b1dcca80-9928-11ea-9606-a2962cb41cf3.png)
## Dump:
\\\mlangfs1\public\v-masche\dumps\NugetConfig.dmp
## Note:
1.This bug can be repro on both 16.3 (D16.3\29303.96—repro rate: 100%) and 16.6 (VS Master\29905.127 + NuGet Client Dev\5.6.0.6489 – repro rate: 50%).
2.Move from internal bug 976738.

erdembayar:
@nkolev92  @aortiz-msft 
Is it compatibility tenet issue?

aortiz-msft:
@v-luzh - Does this repro still?

v-luzh:
It is fixed on Main\31820.317 + NuGet Client Dev\6.1.0.30. No crash happened. Close this issue now, thanks.

PR #9658, opened 2020-06-08
Design for additional user wide configuration

nkolev92:
Design for https://github.com/NuGet/Home/issues/9394. 

Implementation at: https://github.com/NuGet/NuGet.Client/pull/3421

### DIFF
--- a/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
+++ b/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
@@ -31,7 +31,7 @@ The solution to start detecting additional configs from the user specific direct
 Given that these are additional configs, they will be merged with lower priority than the default user specific config. The rest are merged deterministically in alphabetical order because they are not expected to confict.
 The determinism allows an out in case something really goes wrong.
 
-There 2 solutions that satisfy the requirement with the least amount of friction.
+There are 2 solutions that satisfy the requirement with the least amount of friction.
 
 ### Re-use the existing %APPDATA%\NuGet directory
 
--- a/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
+++ b/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
@@ -37,8 +37,8 @@ There 2 solutions that satisfy the requirement with the least amount of friction
 
 Pros:
 
-* Simple & intuitive. Windows: %appdata%\NuGet\NuGet.Config
-Mac/Linux: ~/.config/NuGet/NuGet.Config or ~/.nuget/NuGet/NuGet.Config.
+* Simple & intuitive. Windows: %appdata%\NuGet\*.Config
+Mac/Linux: ~/.config/NuGet/*.[C|c]onfig or ~/.nuget/NuGet/*.[C|c]onfig.
 
 Cons:
 
@@ -46,7 +46,8 @@ Cons:
 
 ### Define a new folder for additiona user-wide configuration
 
-The proposal is to use `%APPDATA%\NuGet\config`.
+The proposal is to use `%APPDATA%\NuGet\config\` on Windows and `~/.config/NuGet/config/` or `~/.nuget/NuGet/config/` on Mac/Linux
+
 
 Pros:
 
--- a/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
+++ b/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
@@ -26,28 +26,14 @@ All NuGet customers are affected. Primary consumers are the WPF and WinForms cus
 
 ## Solution
 
-The solution to start detecting additional configs from the user specific directory similar to the extensibility from machine wide configs.
+We will start detecting additional configs from the user specific directory similar to the extensibility from machine wide configs.
 
 Given that these are additional configs, they will be merged with lower priority than the default user specific config. The rest are merged deterministically in lexicographical order because they are not expected to conflict.
 The determinism allows an out in case something really goes wrong.
 
-There are 2 solutions that satisfy the requirement with the least amount of friction.
+### A new folder for additional user-wide configuration
 
-### Re-use the existing %APPDATA%\NuGet directory
-
-Pros:
-
-* Simple & intuitive. Windows: %appdata%\NuGet\*.Config
-Mac/Linux: ~/.config/NuGet/*.[C|c]onfig or ~/.nuget/NuGet/*.[C|c]onfig.
-
-Cons:
-
-* This is an existing folder. There are some known instances of unexpected NuGet.config files there. Examples include: NuGet_backup.config found by the OP of the linked issue. It's also likely there are some additional copies of NuGet.config in there.
-
-### Define a new folder for additional user-wide configuration
-
-The proposal is to use `%APPDATA%\NuGet\config\` on Windows and `~/.config/NuGet/config/` or `~/.nuget/NuGet/config/` on Mac/Linux
-
+Use %APPDATA%\NuGet\config\` on Windows and `~/.config/NuGet/config/` or `~/.nuget/NuGet/config/` on Mac/Linux
 
 Pros:
 
@@ -55,13 +41,11 @@ Pros:
 
 Cons:
 
-* Arguably not as intuitive as the first proposal, but it doesn't suffer from the same potential issues with unexpected configuration files. Given the difficulty one could have identifying which config brings in what source or fallback folder. The trade off might be good enough.
+* Arguably not as intuitive as some other proposals, but it doesn't suffer from the same potential issues with unexpected configuration files. Given the difficulty one could have identifying which config brings in what source or fallback folder. The trade off might be good enough.
 
 Worth to mention that while there have been instances of additional configs in the current user wide directory, it's impossible for us to determine the exact indicidence. We are likely talking about very few customers.
 
 ### Reading and writing priority
-
-In both solutions the reading and writing priority proposal is the same.
 
 Say we have 1 solution configuration file and 2 user-wide configuration files.
 
@@ -111,7 +95,8 @@ Given that the local file is the closest and the additional user configuration i
 ### Writing priority
 
 Conversely, when writing we write to the furthest available config file.
-Specifically if we were to write `key5`, it would be written to the `AdditionalUser` config file.
+In order to preserve backwards compatibility on the writing side, we will consider this additional user config as `read only`. The configuration commands will not write to the additional configuration files.
+Specifically if we were to write `key5`, it would be written to the `User` config file.
 
 ## Future Work
 
@@ -119,13 +104,20 @@ None
 
 ## Open Questions
 
-* Which of the 2 approaches do we prefer? Personally I was leaning toward re-using the same folder, but if we want to be risk averse, we can just take the new folder approach.
-* Do we like the writing priority? We can reverse the order to preserve writing compatibility to user-wide folder, but that will potentially allow for additional configs that are supposed to be installed by 3rd party components to overwrite the user wide selections.
-* A potential answer to this ask could be we do nothing.
-
 ## Considerations
 
-None
+The considered but ultimately not accepted proposal is detailed below. 
+
+### Re-use the existing %APPDATA%\NuGet directory
+
+Pros:
+
+* Simple & intuitive. Windows: %appdata%\NuGet\*.Config
+Mac/Linux: ~/.config/NuGet/*.[C|c]onfig or ~/.nuget/NuGet/*.[C|c]onfig.
+
+Cons:
+
+* This is an existing folder. There are some known instances of unexpected NuGet.config files there. Examples include: NuGet_backup.config found by the OP of the linked issue. It's also likely there are some additional copies of NuGet.config in there.
 
 ### References
 
--- a/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
+++ b/designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
@@ -59,6 +59,60 @@ Cons:
 
 Worth to mention that while there have been instances of additional configs in the current user wide directory, it's impossible for us to determine the exact indicidence. We are likely talking about very few customers.
 
+### Reading and writing priority
+
+In both solutions the reading and writing priority proposal is the same.
+
+Say we have 1 solution configuration file and 2 user-wide configuration files.
+
+* Local - solution file.
+* User - The default user wide file from %APPDATA%\NuGet
+* AdditionalUser - The additional user file from whichever the new location is.
+
+The contents of the files are as follows respectively:
+
+Local
+
+```xml
+<configuration>
+    <SectionName>
+        <add key="key1" value="local" />
+        <add key="key2" value="local" />
+    </SectionName>
+</configuration>
+```
+
+User
+
+```xml
+<configuration>
+    <SectionName>
+        <add key="key2" value="user" />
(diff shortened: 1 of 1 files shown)


designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,77 @@
+
+# Spec Name
```
zivkan: update? :)

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,77 @@
...
+The determinism allows an out in case something really goes wrong.
+
+There 2 solutions that satisfy the requirement with the least amount of friction.
+
+### Re-use the existing %APPDATA%\NuGet directory
+
+Pros:
+
+* Simple & intuitive. Windows: %appdata%\NuGet\NuGet.Config
+Mac/Linux: ~/.config/NuGet/NuGet.Config or ~/.nuget/NuGet/NuGet.Config.
```
zivkan: Did you mean?

```suggestion
* Simple & intuitive. Windows: %appdata%\NuGet\*.Config
Mac/Linux: ~/.config/NuGet/*.[C|c]onfig or ~/.nuget/NuGet/*.[C|c]onfig.
```

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,77 @@
...
+
+Cons:
+
+* This is an existing folder. There are some known instances of unexpected NuGet.config files there. Examples include: NuGet_backup.config found by the OP of the linked issue. It's also likely there are some additional copies of NuGet.config in there.
+
+### Define a new folder for additiona user-wide configuration
+
+The proposal is to use `%APPDATA%\NuGet\config`.
```
zivkan:
```suggestion
The proposal is to use `%APPDATA%\NuGet\config\` on Windows and `~/.config/NuGet/config/` or `~/.nuget/NuGet/config/` on Mac/Linux

```

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,77 @@
...
+
+Worth to mention that while there have been instances of additional configs in the current user wide directory, it's impossible for us to determine the exact indicidence. We are likely talking about very few customers.
+
+## Future Work
+
+None
+
+## Open Questions
+
+* Which of the 2 approaches do we prefer? Personally I was leaning toward re-using the same folder, but if we want to be risk averse, we can just take the new folder approach.
```
zivkan: Best case: customer doesn't have any other `*.config` file in the directory, so there's no problem
Middle case: customer has a backup config file which previously was not loaded, starts getting loaded and causes restore issues. For example, maybe a backup config has a package source that no longer exists, or the customer is switching config files when on and off their corporate network that has a private package source. Restore breaks until they rename/delete the backup file(s), but it won't be at all obvious what the cause is or what the fix is, all the customer did was update VS or install a newer .NET SDK.
Worst case: there's a config file with an XML syntax error. Not only does restore break, but VS might hang too: https://github.com/NuGet/Home/issues/9574

I also prefer re-using the same folder, but I'm concerned that the "harm" for the customers with additional files on their computers might be serious enough to warrant using a subdirectory.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: rip, I tried :D

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: Yeah, that's better :D

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: If you had to pick, I guess you are leaning option 2 correct? I'l consider a thumbs up as yes :)

nkolev92:
I added an additional question about the reading/writing priority.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+Pros:
+
+* Simple & intuitive. Windows: %appdata%\NuGet\*.Config
+Mac/Linux: ~/.config/NuGet/*.[C|c]onfig or ~/.nuget/NuGet/*.[C|c]onfig.
+
+Cons:
+
+* This is an existing folder. There are some known instances of unexpected NuGet.config files there. Examples include: NuGet_backup.config found by the OP of the linked issue. It's also likely there are some additional copies of NuGet.config in there.
+
+### Define a new folder for additiona user-wide configuration
```
dominoFire:
```suggestion
### Define a new folder for additional user-wide configuration
```

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Provide a way for 3rd party vendors to add new user-wide configuration files without elevation.
+
+## Non-Goals
+
+## Solution
+
+The solution to start detecting additional configs from the user specific directory similar to the extensibility from machine wide configs.
+
+Given that these are additional configs, they will be merged with lower priority than the default user specific config. The rest are merged deterministically in alphabetical order because they are not expected to confict.
```
dominoFire:
```suggestion
Given that these are additional configs, they will be merged with lower priority than the default user specific config. The rest are merged deterministically in lexicographical order because they are not expected to conflict.
```

dominoFire: - [ ] Reminder for Fernando: Make sure that ordering is culture invariant.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+The determinism allows an out in case something really goes wrong.
+
+There 2 solutions that satisfy the requirement with the least amount of friction.
```
dominoFire:
```suggestion
There are 2 solutions that satisfy the requirement with the least amount of friction.
```

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Goals
+
+* Provide a way for 3rd party vendors to add new user-wide configuration files without elevation.
+
+## Non-Goals
+
+## Solution
+
+The solution to start detecting additional configs from the user specific directory similar to the extensibility from machine wide configs.
```
dominoFire: Suppose you have a project with the following layout:

```
app.config
code.cs
nuget.config
project.csproj
solution.sln
```

Would `nuget.config` and `app.config` be parsed as NuGet.config files? If that's the case, could we adopt `.nuget.config` as a file extension to avoid parsing `app.config`?

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+
+Cons:
+
+* Arguably not as intuitive as the first proposal, but it doesn't suffer from the same potential issues with unexpected configuration files. Given the difficulty one could have identifying which config brings in what source or fallback folder. The trade off might be good enough.
+
+Worth to mention that while there have been instances of additional configs in the current user wide directory, it's impossible for us to determine the exact indicidence. We are likely talking about very few customers.
+
+## Future Work
+
+None
```
dominoFire, line 103: Would we detect `<clear/>` nodes in the installer's config file?

dominoFire (commented):
Some questions and typos.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
```diff
@@ -0,0 +1,78 @@
...
+* Issue: [9394](https://github.com/NuGet/Home/issues/9394) NuGet should support multiple config files in %APPDATA%\NuGet directory
+
+## Problem Background
+
+The XAML and WinForms teams want to allow control library installers to add user-specific NuGet fallback folders by having their SDK installers drop config files in %APPDATA%\NuGet. On uninstall this file would be deleted.
+
+We considered two alternatives which I mentioned as workarounds above:
+
+1. Require SDK installers to drop config files in %ProgramFiles(x86)%\NuGet\Config instead. However, we've gotten feedback from control vendors that they would like the option of installing without elevation which prevents the installer from being able to write to this directory.
+1. Require SDK installers to modify the one %APPDATA%\NuGet\NuGet.config file. However this seems error prone.
```
anangaur, line 15: Why is this error prone? Adding one line and then removing it upon uninstall should be simple IMO. Similar to adding and removing sources/feeds.

zivkan: My understanding is they're going to be adding a fallback folder, not package source. So if the user's config file doesn't already contain a fallback folders element, then it's 3 XML elements that needs to be added, and then there's the question whether the fallbackFolders element should be deleted if the last fallback folder is removed from that list.

If the fallbackFolders list contains multiple entries and the uninstaller has a bug, it might delete the wrong line in the file, or might delete all the fallback folders, not just its own. There's less risk that the uninstaller deletes the wrong file, or overwrites something it shouldn't on install.

Depending on what APIs the control vendor uses in their setup program, it might strip out XML comments from the file, or change whitespace. While this won't cause problems, it might not be what the customer wants, particularly for a file that is not version controlled.

Installer technologies are optimised for writing files and uninstall to delete files, so that's the easiest thing to do. Adding logic to edit existing files, while possible, increases the complexity for control vendors to use this feature.

It would be better for performance if we didn't load `*.config` from anywhere, but the risks and additional complexity for anyone who wants to use this feature doesn't feel worthwhile to me. We can minimise perf impact by loading in parallel if we're not already.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
zivkan: The proposal is to load `%appdata%\NuGet\config\*.config`, not to load `*.config` in the project directory.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
anangaur: How does a Fallback folder line look in the nuget.config? If we can provide an API to add and remove the fallback folder, given a unique name, should it still be a problem. My thinking is to abstain from adding more moving parts to the nuget config management.

srdjanjovcic, line 15: Another possible destination would be `%ProgramData%` folder (typically, `C:\ProgramData`). Non-admin users can write to it, and all users can read from it.

srdjanjovcic: What I do not know what would be Linux/OSX equivalent? Something under `/usr`?

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: What @zivkan said.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: The config merges works the same way it does today for any number of configs. 

This only affects the config discovery, so all `</clear>` tags will work as before :)

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92:
```xml
<fallbackPackageFolders>
   <add key="XYZ Offline Packages" value="C:\somePath\someFolder\"/>
</fallbackPackageFolders>
```

from https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#fallbackpackagefolders-section. 

Given that the installation of these configs is a 3rd party experience, they are free to use our libraries. We have them, but we don't know if the installer is written in .NET. 

Regarding the error prone part, editing onfigs has caused unexpected problems for customers in the past. 
There are some occasions in which NuGet itself trying to edit the config caused issues. 

@rrelyea might know more about the history of those issues.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
anangaur: What is the expected behavior if </clear> is mentioned in multiple files? I feel irrespective of the priority defined by us, its going to be confusing.

designs/UserWideConfiguration-AdditionalExtensibilityFolders.md
nkolev92: That's certainly an option. Familiarity was the key motivator for proposing AppData. 

The programdata equivalent on linux is `/etc/opt`

zivkan: It looks almost identical to the packageSources section: https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#fallbackpackagefolders-section.

Even if we create an API (presumably in .NET, although we technically already have APIs, in NuGet.Configuration.dll), I have no idea how installers would call the API. Can MSI's natively execute .NET assemblies, or would the MSI author have to jump through additional hoops to copy .NET assemblies to a temporary location, run, then clean up this temporary stuff after itself? What about installers that are not MSIs?

zivkan: /etc/opt on my Ubuntu on WSL is not user writable.

donnie-msft:
Mentioned offline: concern is that if I load one of these projects, and now have a nuget.config in this "special folder", any other project I load will utilize this special nuget.config even though it's irrelevant. 
That could mean slower Restores. Maybe different metadata for a given packageId (finding it on SpecialSource instead of nuget.org). Possibly other consequences...

nkolev92:
Notes from the meeting. 

•	We should be mindful of features that push configuration outside of the repository being worked on. We should maintain our focus on reproducibility as an end game. 
o	Given the current state of the art, we will still do this feature as it’s main benefits are discoverability. 
•	We prefer the new folder approach. Eliminates the need for any unexpected behaviors
•	On the reading side, the default user config has higher priority than the additional configs
•	On the writing side, the additional configs should be considered read only. 
•	Config discoverability, overwriting is not great in our product. We need to improve that. I’ll follow up to ensure we have the relevant issues tracked.
•	On the control vendor guidance side, @debonte will ensure that the key/valueness of fallback folders in configs is communicated. 

I'll incorporate all the feedback into the design doc.

nkolev92:
Updated the spec with all the details. 

I'll merge when someone approves it. 

I'll proceed with the implementation.

dominoFire (approved):

anangaur:
May be I missed the resolution for this multiple config files priority order. What would happen if there multiple config files with clear nodes followed by one or two sources (for example). Which sources will finally be effective in such cases? Will clear command work for each file or will all configs be consolidated at one level and then treated as one config?

nkolev92:
Clear works for each config.

Similar to machine wide configs we will recommend that no clears are ever added there. 

The priority is alphabetical. It's all convention based.

PR #9699, opened 2020-06-18
Add prerelease option to dotnet.exe add package command

martinrrm:
Design for new prerelease option in dotnet.exe add

### DIFF
--- a/designs/Prerelease-Option.md
+++ b/designs/Prerelease-Option.md
@@ -1,0 +1,55 @@
+# Spec Name
+
+* Status: In Review
+* Author(s): [Martin Ruiz](https://github.com/martinrrm)
+* Issue: [4699](https://github.com/NuGet/Home/issues/4699) dotnet add package - support for NuGet "Pre" version
+
+## Problem Background
+
+Customers want to add the latest version of a package to their project, even tough it is a prerelease version.
+
+## Who are the customers
+
+All dotnet.exe customers
+
+## Goals
+
+Add an option to the command `add package` enable installing latest prerelease version.
+
+## Non-Goals
+
+## Solution
+
+Add the option `--prerelease` in the command `add package`. This command should get the latest version of the packet available even if it is not a stable version.
+
+```package 2.0.0
+package 3.0.0
+package 3.3.1-preview.3
+```
+
+| Command | Result | Description |
+|---------|--------|--------------|
+| `dotnet.exe add package` | 3.0.0 | latest stable version of package |
+| `dotnet.exe add package --prerelease` | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add package --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
+
+```package 2.0.0
+package 3.0.0
+package 3.3.1-preview.3
+package 3.4.0
+```
+
+| Command | Result | Description |
+|---------|--------|--------------|
+| `dotnet.exe add package` | 3.4.0 | latest stable version of package |
+| `dotnet.exe add package --prerelease` | 3.4.0 | latest version of package |
+
+## Future Work
+
+## Open Questions
+
+## Considerations
+
+### References
+
+* <https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference>
--- a/designs/Prerelease-Option.md
+++ b/designs/Prerelease-Option.md
@@ -26,33 +26,14 @@ When searching for the latest version available, only listed packages will be co
 
 ### Scenarios
 
-Packages available
 
-```
-Contoso.Library 2.0.0
-Contoso.Library 3.0.0
-Contoso.Library 3.3.1-preview.3
-```
-
-| Command | Result | Description |
-|---------|--------|--------------|
-| `dotnet.exe add Contoso.Library` | 3.0.0 | latest stable version of package |
-| `dotnet.exe add Contoso.Library --prerelease` | 3.3.1-preview.3 | latest version of package |
-| `dotnet.exe add Contoso.Library --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
-
-Packages available
-
-```
-Contoso.Library 2.0.0
-Contoso.Library 3.0.0
-Contoso.Library 3.3.1-preview.3
-Contoso.Library 3.4.0
-```
-
-| Command | Result | Description |
-|---------|--------|--------------|
-| `dotnet.exe add Contoso.Library` | 3.4.0 | latest stable version of package |
-| `dotnet.exe add Contoso.Library --prerelease` | 3.4.0 | latest version of package |
+| Command | Available Versions | Output Version | Description |
+|---------|--------------------|------------------|-------------|
+| `dotnet.exe add Contoso.Library` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | 3.0.0 | latest stable version of package |
+| `dotnet.exe add Contoso.Library --prerelease` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add Contoso.Library --prerelease --version 3.0.0` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | error | The user cannot use the `--prerelease` and `--version` commands at the same time |
+| `dotnet.exe add Contoso.Library` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> Contoso.Library 3.4.0 <br> | 3.4.0 | latest stable version of package |
+| `dotnet.exe add Contoso.Library --prerelease` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> Contoso.Library 3.4.0 <br> | 3.4.0 | latest version of package |
 
 ## Future Work
 
--- a/designs/Prerelease-Option.md
+++ b/designs/Prerelease-Option.md
@@ -14,7 +14,7 @@ All dotnet.exe customers
 
 ## Goals
 
-Add an option to the command `add package` to enable installing latest prerelease version.
+Allow dotnet CLI users to install the latest version of a NuGet package, including prerelease versions. This can be accomplished by adding an option to the existing command `add package` that enables installing the last version, including prerelease versions.
 
 ## Non-Goals
 
@@ -29,30 +29,30 @@ When searching for the latest version available, only listed packages will be co
 Packages available
 
 ```
-package 2.0.0
-package 3.0.0
-package 3.3.1-preview.3
+Contoso.Library 2.0.0
+Contoso.Library 3.0.0
+Contoso.Library 3.3.1-preview.3
 ```
 
 | Command | Result | Description |
 |---------|--------|--------------|
-| `dotnet.exe add package` | 3.0.0 | latest stable version of package |
-| `dotnet.exe add package --prerelease` | 3.3.1-preview.3 | latest version of package |
-| `dotnet.exe add package --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
+| `dotnet.exe add Contoso.Library` | 3.0.0 | latest stable version of package |
+| `dotnet.exe add Contoso.Library --prerelease` | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add Contoso.Library --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
 
 Packages available
 
 ```
-package 2.0.0
-package 3.0.0
-package 3.3.1-preview.3
-package 3.4.0
+Contoso.Library 2.0.0
+Contoso.Library 3.0.0
+Contoso.Library 3.3.1-preview.3
+Contoso.Library 3.4.0
 ```
 
 | Command | Result | Description |
 |---------|--------|--------------|
-| `dotnet.exe add package` | 3.4.0 | latest stable version of package |
-| `dotnet.exe add package --prerelease` | 3.4.0 | latest version of package |
+| `dotnet.exe add Contoso.Library` | 3.4.0 | latest stable version of package |
+| `dotnet.exe add Contoso.Library --prerelease` | 3.4.0 | latest version of package |
 
 ## Future Work
 
--- a/designs/Prerelease-Option.md
+++ b/designs/Prerelease-Option.md
@@ -6,7 +6,7 @@
 
 ## Problem Background
 
-Customers want to add the latest version of a package to their project, even tough it is a prerelease version.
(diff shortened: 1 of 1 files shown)


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=9699) <br/>All CLA requirements met.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
+# Spec Name
+
+* Status: In Review
+* Author(s): [Martin Ruiz](https://github.com/martinrrm)
+* Issue: [4699](https://github.com/NuGet/Home/issues/4699) dotnet add package - support for NuGet "Pre" version
+
+## Problem Background
+
+Customers want to add the latest version of a package to their project, even tough it is a prerelease version.
```
nkolev92:
```suggestion
Customers want to add the latest version of a package to their project, even if the latest version is prerelease.
```

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Who are the customers
+
+All dotnet.exe customers
+
+## Goals
+
+Add an option to the command `add package` enable installing latest prerelease version.
```
nkolev92:
```suggestion
Add an option to the command `add package` to enable installing latest prerelease version.
```

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Non-Goals
+
+## Solution
+
+Add the option `--prerelease` in the command `add package`. This command should get the latest version of the packet available even if it is not a stable version.
```
nkolev92:
```suggestion
Add the option `--prerelease` in the command `add package`. This command should get the latest version of the package available even if it is not a stable version.
```

nkolev92: You want to call out that this command will get the latest `listed` package. 

For reference see: https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/publish-nuget-package

Specifically https://docs.microsoft.com/en-us/nuget/nuget-org/policies/deleting-packages.

The idea behind listed/unlist packages is that, I as an author have decided that the package version I published had some issues and do not want it to be discoverable. So I unlist it. 

A user can explicitly request and use an unlisted version, but the author does not want it to be appear in search, like when you go to nuget.org and search: See example: https://www.nuget.org/packages?q=mypackage

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Future Work
+
+## Open Questions
+
+## Considerations
+
+### References
+
+* <https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference>
```
nkolev92: Use a link like so [dotnet add package reference](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference)

```suggestion
* [dotnet add package reference](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference)
```

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+package 3.3.1-preview.3
+```
+
+| Command | Result | Description |
+|---------|--------|--------------|
+| `dotnet.exe add package` | 3.0.0 | latest stable version of package |
+| `dotnet.exe add package --prerelease` | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add package --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
+
+```package 2.0.0
```
nkolev92: nit: 
when you format put it on a new line.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+## Goals
+
+Add an option to the command `add package` enable installing latest prerelease version.
+
+## Non-Goals
+
+## Solution
+
+Add the option `--prerelease` in the command `add package`. This command should get the latest version of the packet available even if it is not a stable version.
+
```
nkolev92, line 24: I would add a ### Scenarios heading or something like that, to clearly call attention to the section.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,55 @@
...
+| Command | Result | Description |
+|---------|--------|--------------|
+| `dotnet.exe add package` | 3.4.0 | latest stable version of package |
+| `dotnet.exe add package --prerelease` | 3.4.0 | latest version of package |
+
+## Future Work
+
+## Open Questions
+
```
nkolev92, line 43: A potential open question should be whether we add a shorthand version for prereleae. 

See --framework for example from https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference#options.

zivkan:
Please improve the PR title, as when looking at the pull request list, I have no idea what this is adding prerelease option to.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,69 @@
...
+| `dotnet.exe add package` | 3.4.0 | latest stable version of package |
+| `dotnet.exe add package --prerelease` | 3.4.0 | latest version of package |
+
+## Future Work
+
+* When the design has been approved, we will create an issue with the [dotnet.exe team](https://github.com/dotnet/sdk)
+
+## Open Questions
+
+* Do we need to add a shorthand version for prerelease? `--p`
```
nkolev92, line 44: My vote here would be no, because `--p` makes me think property.

nkolev92 (approved):

nkolev92:
@chgill-MSFT @JonDouglas Can we get some PM eyes on this? 

@KathleenDollard @Dsplaisted @wli3 @marcpopMSFT some CLI eyes on this?

designs/Prerelease-Option.md
wli3: same here. I don't think we need a short hand. One letter shorthands are precious (need to work across all commandline).

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,69 @@
...
+Add the option `--prerelease` in the command `add package`. This command will get the latest version of the package available even if it is not a stable version.
+
+When searching for the latest version available, only listed packages will be considered.
+
+### Scenarios
+
+Packages available
+
+```
+package 2.0.0
```
wli3: Is 0.9 considered prerelease?
given versions:
0.8
0.9

with --prerelease -> 0.9
without -> fail ?

designs/Prerelease-Option.md
anangaur: A couple of questions/observations:
1. Using the —prerelease option should get the latest version including prerelease. Correct?
2. I find the name of the option a bit long. Can we also enable —pre may in addition to —prerelease? - a nit

designs/Prerelease-Option.md
anangaur: I don’t think so and @nkolev92 can confirm. The definition of prerelease is a version with “-“ in it.

nkolev92: Nope, as @anangaur mentioned, prerelease requires a `-` in the version. 
`0.9` is just a weird stable :D 

For reference: https://nugettoolsdev.azurewebsites.net/5.5.0-preview.2.6382/parse-version?version=0.9

designs/Prerelease-Option.md
nkolev92: 1. Yes. Similar to how the PM UI behaves. If you have the prerelease filter we show you the latest version if it's prerelease. 

2. We just opted for the full name cause I haven't seen option names not being the exact word. We can't really do --pre as  short hand. A short option must be one letter only. 
So we either have `--prerelease` or `--pre`.

designs/Prerelease-Option.md
marcpopMSFT: Is there any formatting that includes the - for a non-preview version or doesn't have the - for a preview version?  Is that formatting enforced in other places as well and documented?

nkolev92: > Is there any formatting that includes the - for a non-preview version or doesn't have the - for a preview version?

Nope, that's per the semver specification. 
Documented here: https://docs.microsoft.com/en-us/nuget/concepts/package-versioning

wli3: Looks like I can restore a package without 1.0 with *. So that should be fine

```
<PackageReference Include="xunit.analyzers" Version="*" />
```

wli3:
Is there anyway to get prerelease using csproj? Or dotnet tool still cannot get prerelease packages. (one of the top asks) Note, dotnet tool still uses a temp project which has something like

```
<PackageReference Include="my-tool" Version="*" />
```

chrisraygill:
> Is there anyway to get prerelease using csproj? Or dotnet tool still cannot get prerelease packages. (one of the top asks) Note, dotnet tool still uses a temp project which has something like
> 
> ```
> <PackageReference Include="my-tool" Version="*" />
> ```

In the csproj you can specify latest version including prerelease with \*-\*. Something like 5.\*-\* also works to get the latest version including prelease in the version 5 range.

Does that answer your question @wli3?

nkolev92:
What @chgill-MSFT said. 

See https://github.com/NuGet/Home/issues/912 was fixed in 5.6.

wli3:
ah, that was in already. Great!

nkolev92:
@chgill-MSFT @JonDouglas 

You guys good with this design?
Think the *only* topic of potential contention is the option name. @anangaur suggested it might be too long, but personally I think it's ok to be explicit.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,69 @@
...
+
+Customers want to add the latest version of a package to their project, even if the latest version is prerelease.
+
+## Who are the customers
+
+All dotnet.exe customers
+
+## Goals
+
+Add an option to the command `add package` to enable installing latest prerelease version.
```
chrisraygill: Just nitpick, but I think the goal should be framed as helping a customer accomplish a "job to be done." In this case the goal should be something like:

 "Allow dotnet CLI users to easily install the latest version of NuGet package, including prereleases. This can be accomplished by adding an option to the existing `add package` that enables installing prerelease versions."

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Non-Goals
+
+## Solution
+
+Add the option `--prerelease` in the command `add package`. This command will get the latest version of the package available even if it is not a stable version.
```
chrisraygill, line 23: Has there been any consideration to allow specification of prerelease-enabled floating versions with the `dotnet add package` command? for example:

`dotnet add package newtonsoft.json *-*`

Today `dotnet add package newtonsoft.json 12.*` will work, but `dotnet add package newtonsoft.json *-*` will return `error: '*-*' is not a valid version string.`

chrisraygill: Maybe you would need to add `--prerelease` so that would work?:
`dotnet add package --prerelease newtonsoft.json *-*`

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,69 @@
...
+| `dotnet.exe add package --prerelease` | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add package --prerelease --version 3.0.0` | error | The user cannot use this commands at the same time |
+
+Packages available
+
+```
+package 2.0.0
+package 3.0.0
+package 3.3.1-preview.3
+package 3.4.0
```
chrisraygill: These examples are kind of confusing because the full command is `dotnet add package <packageId>`. So using a an example package named "package" makes it hard to grok. Maybe using Newtonsoft.Json would work better?

chrisraygill:
> @chgill-MSFT @JonDouglas
> 
> You guys good with this design?
> Think the _only_ topic of potential contention is the option name. @anangaur suggested it might be too long, but personally I think it's ok to be explicit.

Personally, I'm fine with `--prerelease` especially since it's the wording we use elsewhere in NuGet so it'll be easy to remember. Added a relevant comment to the comment thread with Anand.

designs/Prerelease-Option.md
nkolev92: `dotnet add package Nuget.common --version "*-*"` works for me. 

Remember you need 3.1.300 for `*-*` support.

nkolev92:
@chgill-MSFT No precedent afaik. 
nuget.exe install uses the complete spelling, https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-install.

designs/Prerelease-Option.md
chrisraygill: Personally, I'd rather go with the options that is the most instinctual. So if I forgot what the options were but I wanted to enable prereleases, my instinct would be to add `--prerelease` not `--pre` even tho it's shorter. I think intuitiveness > saving a couple key strokes.

chgill-MSFT: I vote sticking with `--prerelease` because it's the most instinctive and is the wording used in other client tools such as PMUI and nuget.exe. If I forget what the options are and want to install a prerelease package, my instinct would be to try `--prerelease` before `--pre`.

designs/Prerelease-Option.md
chrisraygill: OOOHHHH that makes sense, thanks 👌

designs/Prerelease-Option.md
zivkan: Or maybe something like `Consoto.Library` to reduce confusion of versions of `Newtonsoft.Json` that don't actually exist on nuget.org.

designs/Prerelease-Option.md
```diff
@@ -0,0 +1,50 @@
...
+When searching for the latest version available, only listed packages will be considered.
+
+### Scenarios
+
+
+| Command | Available Versions | Output Version | Description |
+|---------|--------------------|------------------|-------------|
+| `dotnet.exe add Contoso.Library` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | 3.0.0 | latest stable version of package |
+| `dotnet.exe add Contoso.Library --prerelease` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | 3.3.1-preview.3 | latest version of package |
+| `dotnet.exe add Contoso.Library --prerelease --version 3.0.0` | Contoso.Library 2.0.0 <br> Contoso.Library 3.0.0 <br> Contoso.Library 3.3.1-preview.3 <br> | error | The user cannot use the `--prerelease` and `--version` commands at the same time |
```
chrisraygill, line 34: Should we error for `dotnet.exe add Contoso.Library --prerelease --version 3.0.0` or just ignore the --prerelease and let it work since the intent to install version 3.0.0 seems clear? I think that exact scenario is less likely, but I could potentially see someone doing  `dotnet.exe add Contoso.Library --prerelease --version 3.0.0-preview` thinking they need --prerelease to install it.

Erroring makes sense if that's the standard behavior of dotnet.exe in these kinds of scenarios.

nkolev92: I think it's ok to error when the options provided are confusing and do not make sense. 

The https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers command throws a bunch of, hey you can't combine these switches errors.

nkolev92 (approved):

nkolev92:
Last calls for feedback :) @chgill-MSFT @JonDouglas @anangaur

Issue #9704, opened 2020-06-22
Search CLI Tool for NuGet.exe


advay26-zz:
This project is a search tool available through the command line for NuGet.exe that allows users to query the NuGet API and see its search results.

- [x] Learning how to query a NuGet server endpoint

- [x] Writing a feature spec for this issue

- [x] Figuring out how to consolidate and display search results from the queries

- [x] Figure out how to handle search results from multiple sources

- [x] Refactoring code to fit into the codebase properly

- [x] Writing tests for the new feature

- [x] Writing relevant documentation

PR #9718, opened 2020-06-25
Created SearchCommandSpec.md

advay26-zz:
Created a design spec for the Search feature I'm implementing for NuGet.exe. [[Issue 9704](https://github.com/NuGet/Home/issues/9704)]

### DIFF
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -1,0 +1,98 @@
+# Search CLI Tool for NuGet.exe and Dotnet.exe
+
+* Status: **Implementing**
+* Author(s): [Advay Tandon](https://github.com/advay26)
+* Issue: [9704](https://github.com/NuGet/Home/issues/9704) - Search CLI Tool for NuGet.exe and Dotnet.exe
+
+## Problem Background
+
+Search functionality from the command line for NuGet.exe and the .NET Core CLI has been a feature request from the NuGet open source community for some time. Additionally, the NuGet API is already equipped to handle search requests, so an extension of this service to command line users would be both practical to implement and also highly valuable.
+
+## Who are the customers
+
+All NuGet.exe and .NET CLI users.
+
+## Goals
+
+* Providing search functionality for NuGet packages through the commandline
+* Displaying the search results to users in a way that maximizes the visibility of the most relevant results.
+
+## Solution
+
+Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+```
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+
+```
+====================
+Source: NuGet.org
+--------------------
+> Microsoft.Extensions.Logging.Abstractions | v5.0.0-preview.6.20305.6 | DLs: 345145935
+Logging abstractions for Microsoft.Extensions.Logging.
+
+Commonly Used Types:
+Microsoft.Extensions.Logging.ILogger
+Microsoft.Extensions.Logging.ILoggerFactory
+Microsoft.Extensions.Logging.ILogger&lt;TCategoryName&gt;
+Microsoft.Extensions.Logging.LogLevel
+Microsoft.Extensions.Logging.Logger&lt;T&gt;
+Microsoft.Extensions.Logging.LoggerMessage
+Microsoft.Extensions.Logging.Abstractions.NullLogger
+
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging | v5.0.0-preview.6.20305.6 | DLs: 243186566
+Logging infrastructure default implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.IdentityModel.Logging | v6.6.0 | DLs: 120177247
+Includes Event Source based logging support.
+
+--------------------
+> Microsoft.Extensions.Logging.Configuration | v5.0.0-preview.6.20305.6 | DLs: 63205816
+Configuration support for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.Console | v5.0.0-preview.6.20305.6 | DLs: 61796431
+Console logger provider implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.Debug | v5.0.0-preview.6.20305.6 | DLs: 58643995
+Debug output logger provider implementation for Microsoft.Extensions.Logging. This logger logs messages to a debugger monitor by writing messages with System.Diagnostics.Debug.WriteLine().
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.EventSource | v5.0.0-preview.6.20305.6 | DLs: 32121683
+EventSource/EventListener logger provider implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Serilog.Extensions.Logging | v3.0.2-dev-10280 | DLs: 43115462
+-- More  --
+```
+
+## Future Work
+
+* Enabling cross-platform functionality for the search feature
+* Adding optional arguments that allow users to customize their search queries
+
+## Open Questions
+
+* What represents the most effective design choice for displaying the search results, in terms of spacing, demarcation, truncating package descriptions, and so on?
+
+## Considerations
+
+### 1. Combining the results from different sources
+
+Initially, we considered consolidating all search results from the various sources and displaying them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
+
+### 2. Displaying results by printing the output stream to the console without pagination
+
+If we were to simply print the output to the console as a regular stream, the most visible results would be the ones at the end of the output stream, since these would be located right above the next command prompt in the terminal. Pagination, as proposed in this document, ensures that the most relevant results are shown to the users first.
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -35,18 +35,15 @@ Source: NuGet.org
 > Microsoft.Extensions.Logging.Abstractions | 5.0.0-preview.6.20305.6 | Downloads: 345,145,935
   Logging abstractions for Microsoft.Extensions.Logging.
 
-Commonly Used Types:
-Microsoft.Extensions.Lo...
-
+  Commonly Used Types:
+  Microsoft.Extensions.Lo...
 --------------------
 > Microsoft.Extensions.Logging | 5.0.0-preview.6.20305.6 | Downloads: 243,186,566
   Logging infrastructure default implementation for Microsoft.Extensions.Logging.
 When using NuGet 3....
-
 --------------------
 > Microsoft.IdentityModel.Logging | 6.6.0 | Downloads: 120,177,247
   Includes Event Source based logging support.
-
 --------------------
 > Microsoft.Extensions.Logging.Configuration | 5.0.0-preview.6.20305.6 | Downloads: 63,205,816
 ```
@@ -79,7 +76,7 @@ Verbosity Levels:
 
 * _quiet_ - Package ID, Version
 * _normal_ - Package ID, Version, Downloads, Preview of Description
-* _detailed_ - Package ID, Version, Downloads, Full Description, Other information regarding the arguments and filters applied to the search
+* _detailed_ - Package ID, Version, Downloads, Full Description, Other information such as the query URL, and the arguments and filters applied to the search
 
 ## Future Work
 
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -20,7 +20,7 @@ This targets all NuGet.exe and .NET users that want to be able to search for pac
 
 ## Solution
 
-The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results, sorted by relevance, one source after another. The command line interface is simple: __./NuGet.exe search \<query\>__.
 
 ```
 PS C:\> ./NuGet.exe search logging
@@ -33,19 +33,19 @@ Each search result displays the name, version, number of downloads, and a previe
 Source: NuGet.org
 --------------------
 > Microsoft.Extensions.Logging.Abstractions | 5.0.0-preview.6.20305.6 | Downloads: 345,145,935
-Logging abstractions for Microsoft.Extensions.Logging.
+  Logging abstractions for Microsoft.Extensions.Logging.
 
 Commonly Used Types:
 Microsoft.Extensions.Lo...
 
 --------------------
 > Microsoft.Extensions.Logging | 5.0.0-preview.6.20305.6 | Downloads: 243,186,566
-Logging infrastructure default implementation for Microsoft.Extensions.Logging.
+  Logging infrastructure default implementation for Microsoft.Extensions.Logging.
 When using NuGet 3....
 
 --------------------
 > Microsoft.IdentityModel.Logging | 6.6.0 | Downloads: 120,177,247
-Includes Event Source based logging support.
+  Includes Event Source based logging support.
 
 --------------------
 > Microsoft.Extensions.Logging.Configuration | 5.0.0-preview.6.20305.6 | Downloads: 63,205,816
@@ -66,11 +66,20 @@ Search provides users with some optional arguments that they can use to customiz
 
 | Name | Description | Usage |
 | ---  |     ---     |  :-:  |
-| PreRelease | `true` or `false` determining whether to include pre-release packages | -PreRelease `<true/false>` |
-| PackageTypes | The package type to use to filter packages. If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
+| PackageTypes | The package type to use to filter packages (if supported by the source). If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
+| PreRelease | Pre-release packages are not included by default, but can be included by using this argument | -PreRelease |
 | Source | Specific package source(s) to search instead of querying the default sources in __nuget.config__ | -Source `<Source URL>`|
-| Take | The number of results to return | -Take `<positive integer>` |
+| Take | The number of results to return. The default value is 20. | -Take `<positive integer>` |
+| Verbosity | The level of detail to display in the output. The default is _normal_. (See the note below)  | -Verbosity `<quiet\|normal\|detailed>` |
 | Help | Displays help information for the command | -Help |
+
+__NOTE__
+
+Verbosity Levels:
+
+* _quiet_ - Package ID, Version
+* _normal_ - Package ID, Version, Downloads, Preview of Description
+* _detailed_ - Package ID, Version, Downloads, Full Description
 
 ## Future Work
 
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -40,7 +40,7 @@ Source: NuGet.org
 --------------------
 > Microsoft.Extensions.Logging | 5.0.0-preview.6.20305.6 | Downloads: 243,186,566
   Logging infrastructure default implementation for Microsoft.Extensions.Logging.
-When using NuGet 3....
+  When using NuGet 3....
 --------------------
 > Microsoft.IdentityModel.Logging | 6.6.0 | Downloads: 120,177,247
   Includes Event Source based logging support.
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -79,7 +79,7 @@ Verbosity Levels:
 
 * _quiet_ - Package ID, Version
 * _normal_ - Package ID, Version, Downloads, Preview of Description
-* _detailed_ - Package ID, Version, Downloads, Full Description
+* _detailed_ - Package ID, Version, Downloads, Full Description, Other information regarding the arguments and filters applied to the search
 
 ## Future Work
 
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -87,6 +87,7 @@ Verbosity Levels:
 * Adding further optional arguments that allow users to customize their search queries
 * Using columns rather than `|` to display results
 * Consider using pagination to display the search results
+* Allow multiple/no query terms
 
 ## Open Questions
 
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -10,89 +10,87 @@ Search functionality from the command line for NuGet.exe and the .NET Core CLI h
 
 ## Who are the customers
 
-All NuGet.exe and .NET CLI users.
+This targets all NuGet.exe and .NET users that want to be able to search for packages from the command line, rather than using NuGet Gallery or VS PMUI. Also, while `nuget.exe list` allows users to search different package sources for packages with a search query, the results are returned sorted by the package Id, rather than by relevance, as proposed for this Search command. Further, the Search command would provide more information on the packages than List currently does.
 
 ## Goals
 
-* Providing search functionality for NuGet packages through the commandline
-* Displaying the search results to users in a way that maximizes the visibility of the most relevant results.
+* Providing search functionality for NuGet packages through the commandline for NuGet.exe
+* Implementing a prototype of this feature for dotnet.exe
+* Displaying the search results to users in a way that maximizes the visibility of the most relevant results
 
 ## Solution
 
-Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+
 ```
 PS C:\> ./NuGet.exe search logging
 ```
 
-Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+Each search result displays the name, version, number of downloads, and a preview of the description of the package.
 
 ```
 ====================
 Source: NuGet.org
 --------------------
-> Microsoft.Extensions.Logging.Abstractions | v5.0.0-preview.6.20305.6 | DLs: 345145935
+> Microsoft.Extensions.Logging.Abstractions | 5.0.0-preview.6.20305.6 | Downloads: 345,145,935
 Logging abstractions for Microsoft.Extensions.Logging.
 
 Commonly Used Types:
-Microsoft.Extensions.Logging.ILogger
-Microsoft.Extensions.Logging.ILoggerFactory
-Microsoft.Extensions.Logging.ILogger&lt;TCategoryName&gt;
-Microsoft.Extensions.Logging.LogLevel
-Microsoft.Extensions.Logging.Logger&lt;T&gt;
-Microsoft.Extensions.Logging.LoggerMessage
-Microsoft.Extensions.Logging.Abstractions.NullLogger
-
-When using NuGet 3.x this package requires at least version 3.4.
+Microsoft.Extensions.Lo...
 
 --------------------
-> Microsoft.Extensions.Logging | v5.0.0-preview.6.20305.6 | DLs: 243186566
+> Microsoft.Extensions.Logging | 5.0.0-preview.6.20305.6 | Downloads: 243,186,566
 Logging infrastructure default implementation for Microsoft.Extensions.Logging.
-When using NuGet 3.x this package requires at least version 3.4.
+When using NuGet 3....
 
 --------------------
-> Microsoft.IdentityModel.Logging | v6.6.0 | DLs: 120177247
+> Microsoft.IdentityModel.Logging | 6.6.0 | Downloads: 120,177,247
 Includes Event Source based logging support.
 
 --------------------
-> Microsoft.Extensions.Logging.Configuration | v5.0.0-preview.6.20305.6 | DLs: 63205816
-Configuration support for Microsoft.Extensions.Logging.
-When using NuGet 3.x this package requires at least version 3.4.
-
---------------------
-> Microsoft.Extensions.Logging.Console | v5.0.0-preview.6.20305.6 | DLs: 61796431
-Console logger provider implementation for Microsoft.Extensions.Logging.
-When using NuGet 3.x this package requires at least version 3.4.
-
---------------------
-> Microsoft.Extensions.Logging.Debug | v5.0.0-preview.6.20305.6 | DLs: 58643995
-Debug output logger provider implementation for Microsoft.Extensions.Logging. This logger logs messages to a debugger monitor by writing messages with System.Diagnostics.Debug.WriteLine().
-When using NuGet 3.x this package requires at least version 3.4.
-
---------------------
-> Microsoft.Extensions.Logging.EventSource | v5.0.0-preview.6.20305.6 | DLs: 32121683
-EventSource/EventListener logger provider implementation for Microsoft.Extensions.Logging.
-When using NuGet 3.x this package requires at least version 3.4.
-
---------------------
-> Serilog.Extensions.Logging | v3.0.2-dev-10280 | DLs: 43115462
--- More  --
+> Microsoft.Extensions.Logging.Configuration | 5.0.0-preview.6.20305.6 | Downloads: 63,205,816
 ```
+
+If a particular source returns no results for a query, it will display output that indicates this.
+
+```
+====================
+Source: dotnet-msbuild
+--------------------
+No results found.
+```
+
+### Optional Arguments
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PreRelease | `true` or `false` determining whether to include pre-release packages | -PreRelease `<true/false>` |
+| PackageTypes | The package type to use to filter packages. If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
+| Source | Specific package source(s) to search instead of querying the default sources in __nuget.config__ | -Source `<Source URL>`|
+| Take | The number of results to return | -Take `<positive integer>` |
+| Help | Displays help information for the command | -Help |
 
 ## Future Work
 
-* Enabling cross-platform functionality for the search feature
-* Adding optional arguments that allow users to customize their search queries
+* Integrating the Search feature into dotnet.exe
+* Adding further optional arguments that allow users to customize their search queries
+* Using columns rather than `|` to display results
+* Consider using pagination to display the search results
 
 ## Open Questions
 
 * What represents the most effective design choice for displaying the search results, in terms of spacing, demarcation, truncating package descriptions, and so on?
+* Would pagination (using the 'more' or 'less' commands) provide a better viewing experience for users? Would users prefer to pipe the search results to these commands themselves, or have Search do it for them? See _Considerations_ for more. 
+* What, if any, additional information should be included for the packages returned by Search?
 
 ## Considerations
 
 ### 1. Combining the results from different sources
 
-Initially, we considered consolidating all search results from the various sources and displaying them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
+Initially, we considered consolidating all search results from the various sources and interleaving them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
 
-### 2. Displaying results by printing the output stream to the console without pagination
+### 2. Displaying results using pagination
 
-If we were to simply print the output to the console as a regular stream, the most visible results would be the ones at the end of the output stream, since these would be located right above the next command prompt in the terminal. Pagination, as proposed in this document, ensures that the most relevant results are shown to the users first.
+Rather than simply printing the output to the console as a regular stream, using pagination (with the 'more' or 'less' commands) would ensure that users have access to the search results in order of decreasing relevance. They would see the most relevant results (at the top/beginning of the results) first. A single page of results would be shown at a time, and the user could look at the next page by pressing the __space__ key. If we were to simply print the output to the console as a regular stream, the most visible results would be the ones at the end of the output stream, since these would be located right above the next command prompt in the terminal.
--- a/designs/SearchCommandSpec.md
+++ b/designs/SearchCommandSpec.md
@@ -1,8 +1,8 @@
-# Search CLI Tool for NuGet.exe and Dotnet.exe
+# Search CLI Tool for NuGet.exe
 
 * Status: **Implementing**
 * Author(s): [Advay Tandon](https://github.com/advay26)
-* Issue: [9704](https://github.com/NuGet/Home/issues/9704) - Search CLI Tool for NuGet.exe and Dotnet.exe
+* Issue: [9704](https://github.com/NuGet/Home/issues/9704) - Search CLI Tool for NuGet.exe
 
 ## Problem Background
 
@@ -10,17 +10,18 @@ Search functionality from the command line for NuGet.exe and the .NET Core CLI h
 
 ## Who are the customers
 
-This targets all NuGet.exe and .NET users that want to be able to search for packages from the command line, rather than using NuGet Gallery or VS PMUI. Also, while `nuget.exe list` allows users to search different package sources for packages with a search query, the results are returned sorted by the package Id, rather than by relevance, as proposed for this Search command. Further, the Search command would provide more information on the packages than List currently does.
+This targets all NuGet.exe users that want to be able to search for packages from the command line, rather than using NuGet Gallery or VS PMUI. Also, while `nuget.exe list` allows users to search different package sources for packages with a search query, the results are returned sorted by the package Id, rather than by relevance, as proposed for this Search command. Further, the Search command would provide more information on the packages than List currently does.
 
 ## Goals
 
 * Providing search functionality for NuGet packages through the commandline for NuGet.exe
 * Implementing a prototype of this feature for dotnet.exe
 * Displaying the search results to users in a way that maximizes the visibility of the most relevant results
+* Eventually replacing `list` with `search`
 
 ## Solution
 
-The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results, sorted by relevance, one source after another. The command line interface is simple: __./NuGet.exe search \<query\>__.
+The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results, sorted by relevance, one source after another. The command line interface is simple: __./NuGet.exe search \<query terms\>__.
 
 ```
 PS C:\> ./NuGet.exe search logging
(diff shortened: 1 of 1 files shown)


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=9718) <br/>All CLA requirements met.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.EventSource | v5.0.0-preview.6.20305.6 | DLs: 32121683
+EventSource/EventListener logger provider implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Serilog.Extensions.Logging | v3.0.2-dev-10280 | DLs: 43115462
+-- More  --
```
zkat: You can probably cut this example down to 2-3 entries instead of the full actual results. This takes up a lot of space!

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+* Author(s): [Advay Tandon](https://github.com/advay26)
+* Issue: [9704](https://github.com/NuGet/Home/issues/9704) - Search CLI Tool for NuGet.exe and Dotnet.exe
+
+## Problem Background
+
+Search functionality from the command line for NuGet.exe and the .NET Core CLI has been a feature request from the NuGet open source community for some time. Additionally, the NuGet API is already equipped to handle search requests, so an extension of this service to command line users would be both practical to implement and also highly valuable.
+
+## Who are the customers
+
+All NuGet.exe and .NET CLI users.
```
zkat: Those who want to search without going to the nuget gallery or using VS PMUI, maybe?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Serilog.Extensions.Logging | v3.0.2-dev-10280 | DLs: 43115462
+-- More  --
+```
+
+## Future Work
+
+* Enabling cross-platform functionality for the search feature
```
zkat: I think this is a core feature. It's not hard to add multi-platform pager support. We can work on this together.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+* Adding optional arguments that allow users to customize their search queries
```
zkat: There are some optional arguments that we should definitely include in the current iteration. This spec should propose a few (I don't think we need to deal with skip/take for now, though, but being able to control prerelease and package type would be good, and not hard to add right out the door

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+## Goals
+
+* Providing search functionality for NuGet packages through the commandline
+* Displaying the search results to users in a way that maximizes the visibility of the most relevant results.
+
+## Solution
+
+Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+```
+PS C:\> ./NuGet.exe search logging
```
zkat, line 27: How is the experience different for dotnet users? The way parameters are passed in to both CLIs is different, for example.

Should we, though, do dotnet support as "future work"?

zkat (commented):
This is a great first draft! Let's iterate on it a bit :)

I'd love to get feedback from others on the team, too.

designs/SearchCommandSpec.md
loic-sharma: Why do we want to support `NuGet.exe` customers? Is `nuget.exe list` not enough? It may be helpful to describe the benefits and differences between `list` and `search`

designs/SearchCommandSpec.md
loic-sharma, line 27: Do we want to expose any options? For example, exclude prerelease packages, filter to a specific package type?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+```
+
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+
+```
+====================
+Source: NuGet.org
+--------------------
+> Microsoft.Extensions.Logging.Abstractions | v5.0.0-preview.6.20305.6 | DLs: 345145935
```
loic-sharma: How did we choose the abbreviation "DLs"? Is that clear to customers? Could we change this to `Downloads`? Also, can we format the downloads (e.g. `345,145,935`)?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+Logging abstractions for Microsoft.Extensions.Logging.
+
+Commonly Used Types:
+Microsoft.Extensions.Logging.ILogger
+Microsoft.Extensions.Logging.ILoggerFactory
+Microsoft.Extensions.Logging.ILogger&lt;TCategoryName&gt;
```
loic-sharma: Should we HTML decode the description?

designs/SearchCommandSpec.md
loic-sharma: Oh never mind, @zkat already mentioned this below. Closing this comment as it is duplicate.

designs/SearchCommandSpec.md
loic-sharma: Also, what about a `source` parameter to use a different NuGet server?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Open Questions
+
+* What represents the most effective design choice for displaying the search results, in terms of spacing, demarcation, truncating package descriptions, and so on?
+
+## Considerations
+
+### 1. Combining the results from different sources
+
+Initially, we considered consolidating all search results from the various sources and displaying them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
```
loic-sharma: I definitely agree with this section. That said, NuGet normally uses your `nuget.config` file to discover package sources to read from. Users may be confused since this search experience won't do that. Is this something we want to address somehow?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+All NuGet.exe and .NET CLI users.
+
+## Goals
+
+* Providing search functionality for NuGet packages through the commandline
+* Displaying the search results to users in a way that maximizes the visibility of the most relevant results.
+
+## Solution
+
+Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
```
donnie-msft:
```suggestion
The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
```

donnie-msft: What if I want to search a specific source, can we have an option for  `-source` to specify one? (eg, I don't want to see a bajillion "logger" results from nuget.org, but I want to browse my org's internal ones.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+```
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
```
donnie-msft: Can sort be specified? eg, `-sort downloads`, or `-sort name`? 

What's the default sort? Should it be relevancy, download count, or name? 
I think some API's are Server dependent (either registration or search), so depending on which you use, it might be nice to tell the user what order they're about to see, and if they can change that.

donnie-msft: What about other tidbits like deprecation, renamed packages, or known vulnerabilities?

designs/SearchCommandSpec.md
donnie-msft: Where will it get its sources, then?

designs/SearchCommandSpec.md
donnie-msft: Which functionality are we talking about? dotnet.exe implies cross-plat.

donnie-msft (commented):
😎 going to be awesome!

designs/SearchCommandSpec.md
joelverhagen: Agreed on `source`, this aligns with many of the other nuget commands.

designs/SearchCommandSpec.md
joelverhagen: I took this to mean the sources still come from config, but they are not interleaved.

If there is one "section" per source, how will they be sorted? lex by name? lex by URL? same order as config?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Considerations
+
+### 1. Combining the results from different sources
+
+Initially, we considered consolidating all search results from the various sources and displaying them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
+
+### 2. Displaying results by printing the output stream to the console without pagination
+
+If we were to simply print the output to the console as a regular stream, the most visible results would be the ones at the end of the output stream, since these would be located right above the next command prompt in the terminal. Pagination, as proposed in this document, ensures that the most relevant results are shown to the users first.
```
joelverhagen: How will the pager be implemented? IIRC git uses `less` by default but it is configurable. Writing our own might be a little tricky right? you have to figure out terminal dimensions maybe?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+
+```
+====================
+Source: NuGet.org
+--------------------
+> Microsoft.Extensions.Logging.Abstractions | v5.0.0-preview.6.20305.6 | DLs: 345145935
+Logging abstractions for Microsoft.Extensions.Logging.
+
+Commonly Used Types:
```
zivkan: Is commonly used types going to be output at default verbosity?  Do we have any customer feedback that this type of information is commonly desired? My personal preference would be package names only, and I'll explicitly ask for more details if I want them.  But I acknowledge that I'm not at all representative of the average, or beginner, .NET developer.

designs/SearchCommandSpec.md
zivkan: As someone who is somewhat used to using Linux, I don't want the tool itself to do pagination, I'll pipe the output to `more` or `less` myself. Was the proposal to do this within nuget because of customer feedback about existing features, or trying to be pro-active?

As a developer on NuGet, I'm worried about having to detect when the command is interactive or being piped to a text file, or being caputed by another process, and therefore should not pause for user-input.

designs/SearchCommandSpec.md
zivkan: Also consider different package types. A commonly used types list makes sense for "dependency" package types (packages that either explicitly list dependency as the package type, or do not specify any package type). But .NET tools, MSBuildSDK, template package types do not bring in compile time dependencies, and therefore "commonly used types" doesn't make sense as a concept.

designs/SearchCommandSpec.md
zkat: My plan is to use `more` by default on Windows (it's built into `cmd.exe /c more`). This is what the current prototype is using. We would use `less` by default on *nix systems. Additionally, I think as a stretch goal, we should support the [standard-ish `$PAGER` env var](https://en.wikibooks.org/wiki/Guide_to_Unix/Environment_Variables#PAGER).

designs/SearchCommandSpec.md
zkat: This is literally just the description for the project, as sent back from the API

designs/SearchCommandSpec.md
loic-sharma: NuGet search protocol does not support sorting today. I'd suggest keeping this out of the MVP

designs/SearchCommandSpec.md
joelverhagen: Perhaps having the proposed usage text for both .net cli and nuget.exe is the best way to communicate that? it's succinct and well-understood. just an idea, it should be communicated somehow

designs/SearchCommandSpec.md
joelverhagen: The only sort provided by our official API is relevance. /cc @t-mog-msft who is plumbing some of this through. downloads and name are not supported by API and there is no concrete plan for this yet.

designs/SearchCommandSpec.md
zivkan: I thought this was related to the roslyn analyzer/server team work that suggests packages to install when a type cannot be found in an assembly referenced by the project :)

designs/SearchCommandSpec.md
joelverhagen: I agree, it adds complexity and should only be part of the MVP if it's really the goal. it feels very different from other nuget commands. Do other .NET CLI commands have pager behavior on by default?

designs/SearchCommandSpec.md
zivkan: above comment not relevant given that the commonly used types is actually part of the package description, not a proposed feature of search.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.Console | v5.0.0-preview.6.20305.6 | DLs: 61796431
+Console logger provider implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.Debug | v5.0.0-preview.6.20305.6 | DLs: 58643995
+Debug output logger provider implementation for Microsoft.Extensions.Logging. This logger logs messages to a debugger monitor by writing messages with System.Diagnostics.Debug.WriteLine().
```
joelverhagen: Is there a limit of the description shown? IIRC we support up to 4000 characters on nuget.org but there is no limit from the protocol

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+> Microsoft.IdentityModel.Logging | v6.6.0 | DLs: 120177247
+Includes Event Source based logging support.
+
+--------------------
+> Microsoft.Extensions.Logging.Configuration | v5.0.0-preview.6.20305.6 | DLs: 63205816
+Configuration support for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging.Console | v5.0.0-preview.6.20305.6 | DLs: 61796431
```
joelverhagen: Will this show normalized version or full version (includes build metadata)? API returns full version.

designs/SearchCommandSpec.md
joelverhagen: This should be done by the API unless the description is double-encoded. Ideally the client does minimal manipulation on the returned description

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+
+Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+```
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+
+```
+====================
```
joelverhagen, line 33: What will the behavior be if there are no results? Some text?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.Extensions.Logging | v5.0.0-preview.6.20305.6 | DLs: 243186566
+Logging infrastructure default implementation for Microsoft.Extensions.Logging.
+When using NuGet 3.x this package requires at least version 3.4.
+
+--------------------
+> Microsoft.IdentityModel.Logging | v6.6.0 | DLs: 120177247
```
joelverhagen: Just a question, is it normal that we show the `v` in front of the package version? Do we do that elsewhere? I'm used to just seeing the version without the `v` prefix but maybe I am totally wrong

designs/SearchCommandSpec.md
nkolev92: For a popular term it could lead to way too many results which can be unusable :)

designs/SearchCommandSpec.md
nkolev92, line 27: The one thing never mentioned is the `nuget.exe list` command. 

The list command is really the old V2 "search". 
That command already exists and it has a certain set of options and parameters. 

We **might** want to deprecate list first or at least in the same timeframe (potentially @dominoFire can resurrect his PR).
https://github.com/NuGet/NuGet.Client/pull/2772
https://github.com/NuGet/NuGet.Client/pull/2772

designs/SearchCommandSpec.md
nkolev92: I like the approach of `dotnet list package` for showing results. 

So have columns rather than `|`. 

This ofc becomes challenging when you factor in for pagination.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,98 @@
...
+Th search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
+```
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and description of the package. Additionally, Search uses pagination to provide a clean viewing experience for users as they look at the results of their search. This ensures that users have access to the search results in order of decreasing relevance. A single page of results is shown at a time, and the user can look at the next page by pressing the __space__ key.
+
+```
+====================
+Source: NuGet.org
```
nkolev92, line 34: I think description might be too verbose to include in the most minimal output.

nkolev92 (commented):
What the options this command can take? 

See: https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-list for inspiration. 

IncludeDelisted is not relevant there.

designs/SearchCommandSpec.md
chrisraygill: What's the real drawback from a customer experience perspective to making pagination the default behavior? I understand that it might be inconsistent with other elements of the dotnet CLI, but consistency is only a priority when it serves to improve the customer experience, which I don't think it does here.

I'm not sure many people will have the knowledge or motivation to pipe the output for pagination themselves, I wouldn't. The default experience without pagination ends the execution of the list/search command with the _least_ relevant results showing. In the case of there being many results, scrolling to the top for the most relevant results could be a chore and scrolling too quickly could cause you to easily miss/pass the desired package.

Another potential solution is to list results in reverse relevance order, but my concern would be that that would be difficult to communicate clearly and inherently counterintuitive.

The ultimate difference will be that with the current behavior, a user probably has to _look for_ their desired package among the results. With the proposed behavior, it's much more likely they'll just see it immediately with most relevant packages being surfaced in the first "page."

> As a developer on NuGet, I'm worried about having to detect when the command is interactive or being piped to a text file, or being caputed by another process, and therefore should not pause for user-input.

I'm not sure what you mean by "having to detect" here. Could you elaborate?

zivkan: > What's the real drawback from a customer experience perspective to making pagination the default behavior?

Shouldn't we be justifying why to spend effort on something, rather than why not to?

The feature is already built into every operating system or shell. What's being proposed (space to scroll 1 screen) has less features than what's built-in (space to scroll 1 screen, or enter to scroll 1 line on Windows. Linux's `more` also has text search built-in). We'd just be spending development resources duplicating features that already exist and are already available for all customers, using a technique that works for all programs that simply output text to the console. Composing excellent tools together often has a better outcome than a single development team trying to re-invent the most basic features of everyone else's wheels into one integrated product. But this is probably being proposed to help people who are not like me.

I don't know about Windows or Mac, but on Linux I can use `dir | less` to get scroll up and well as down, either one line at a time, or one screen at a time, plus text search forwards or backwards, and [a whole lot of other features](https://man7.org/linux/man-pages/man1/less.1.html). I can use `dir | grep ...`, `dir | sed ...` to do more advanced scripting. Learning how to `dir | more` opens the door to learning about all these scripting techniques. Building it into an app, and therefore hiding away the entire scripting world from a beginner harms their education, even if it's marginally easier to use. And my opinion is that `| more` is trivial to learn, so the benefit is really marginal.

I was going to write that if we want to invest in a command line search that is interactive, we should consider the command line UI library that Loic used recently and build something much more unique (although Loic showed me that something like this already exists for npm at least). But then when I found Loic's app, [it's already a command line NuGet search tool](https://github.com/loic-sharma/nugs).

> I'm not sure many people will have the knowledge or motivation to pipe the output for pagination themselves,

I disagree. My hypothesis is that anyone who has spent much time in any command line will have learnt about piping output to `more`, since this works for all command line apps on all operating systems. Any web search about how to scroll one screen at a time on the command line shows many blog posts, stack overflow questions & answers, and other resources that introduce `| more`.

> I'm not sure what you mean by "having to detect" here. Could you elaborate?

With this proposal if I run `nuget search whatever`, NuGet is going to print one screen's worth of text, tell me to press space to continue, then stop and wait for input before outputting more.  If I run `nuget search whatever > results.txt`, the text telling me to press space will not be output to the terminal, so NuGet just hangs waiting for input that will never come because I don't know it's waiting for me to press anything. What about if I run `nuget search term1 | grep -i  term2`, is that going to hang waiting for input?

designs/SearchCommandSpec.md
advay26-zz: I've made this change in my latest commit.

designs/SearchCommandSpec.md
advay26-zz: I've added this to the 'Who are the customers' section.

designs/SearchCommandSpec.md
advay26-zz: We have tabled the idea for the pager for now, and also decided to leave dotnet.exe functionality to future work

designs/SearchCommandSpec.md
advay26-zz: I have added an 'Optional Arguments' section in 'Solution' that introduces Pre-Release, Package Type, Source, Take and Help as arguments available to users.

designs/SearchCommandSpec.md
advay26-zz: In the latest commit, I have put down creating a prototype for dotnet.exe as a goal, and integrating Search into dotnet.exe as something for 'Future Work'

designs/SearchCommandSpec.md
advay26-zz: I have attempted to address this in the 'Who are the customers' section. Let me know where I've fallen short in my explanation and I'd be happy to amend it further!

designs/SearchCommandSpec.md
advay26-zz: I've taken both suggestions on board and adapted the example in 'Solution'

designs/SearchCommandSpec.md
advay26-zz: This section was meant to convey that instead of interleaving the results from the various sources in nuget.config, we display the results for each source as a separate section, followed by the results for the next source, and so on. As far as I'm aware, the NuGet API's Search only returns results sorted by relevance. I have also added a -Source optional argument in the 'Optional Arguments' section in 'Solution' if a user wants to specify a source for the Search to target.

designs/SearchCommandSpec.md
advay26-zz: I have added this in the 'Optional Arguments' section at the end of 'Solution'.

designs/SearchCommandSpec.md
advay26-zz: I haven't added these pieces of information yet, but in 'Open Questions', I've addressed the fact that we may want to consider additional/alternative information for the packages returned by Search.

designs/SearchCommandSpec.md
advay26-zz: For now I have put the pager idea in 'Future Work' and we can consider building it in later if we have a consensus.

advay26-zz: Since we don't currently have a consensus on the pager idea, I have put it in 'Future Work' for now.

designs/SearchCommandSpec.md
advay26-zz: I have built in a character limit in my latest commit. The description is truncated if it exceeds a certain character limit. You can see an example of this in the first result in the 'Solution' section.

designs/SearchCommandSpec.md
advay26-zz: Looking at some specific examples, like NuGet.Protocol, the API seems to return the normalized version already, but feel free to correct me if I'm wrong.

designs/SearchCommandSpec.md
advay26-zz: The 'Solution' section now includes an example of what is shown if a particular source returns no results.

designs/SearchCommandSpec.md
advay26-zz: I've included this change in the latest commit.

designs/SearchCommandSpec.md
advay26-zz: We've left pagination to 'Future Work' for now because of a lack of consensus on it. To help with the problem of too many results potentially being returned, I have also added an optional argument for -Take, which limits the number of results to display.

designs/SearchCommandSpec.md
advay26-zz: I have addressed the comparisons between list and search in 'Who are the customers' in the latest commit, but if this section is incomplete or incorrect, please let me know!

designs/SearchCommandSpec.md
advay26-zz: For now, I have put this down in 'Future Work' rather than an immediate goal.

designs/SearchCommandSpec.md
joelverhagen: The NuGet.Protocol API returns whatever the HTTP endpoint returns. Per protocol specification, this should be the full version:
https://docs.microsoft.com/en-us/nuget/api/search-query-service-resource#search-result

I don't know the specific API you are using in client, but if a `NuGetVersion` is involved instead of a raw string, you can choose via one of these two methods:
https://github.com/NuGet/NuGet.Client/blob/088004379cb99890f766e9b7a0a29f7b080f888b/src/NuGet.Core/NuGet.Versioning/SemanticVersionBase.cs#L13-L30

designs/SearchCommandSpec.md
advay26-zz: I have introduced a character limit to the description field in the latest commit, and you can see an example of this in the first result in the 'Solution' section.

designs/SearchCommandSpec.md
advay26-zz: Ok, I'll make to sure to incorporate this into how we display results.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+### Optional Arguments
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PreRelease | `true` or `false` determining whether to include pre-release packages | -PreRelease `<true/false>` |
+| PackageTypes | The package type to use to filter packages. If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
```
joelverhagen: Adoption of package type filtering on the server side is likely very sparse. AFAIK the only implementor is nuget.org. Other feeds like Azure DevOps and GitHub probably don't support it today. Therefore, we should be careful about the wording and graceful failures. There is a way to check if a server supports package type filtering so I would recommend spec'ing here what the behavior is if the server does not support package type filtering. It's up to you, but maybe say "Package type filtering is not supported on this package source." for each source not support it instead of showing any search results at all.

BTW, you detect package type filtering capabilities via the `SearchQueryService/3.5.0` type in the service index, e.g. just search for that string here https://api.nuget.org/v3/index.json and you'll see.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+| Source | Specific package source(s) to search instead of querying the default sources in __nuget.config__ | -Source `<Source URL>`|
+| Take | The number of results to return | -Take `<positive integer>` |
```
joelverhagen: nit: Perhaps specify the default `take` in the description?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+No results found.
+```
+
+### Optional Arguments
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PreRelease | `true` or `false` determining whether to include pre-release packages | -PreRelease `<true/false>` |
```
joelverhagen: nit: other commands like `install` and `list` have prerelease off by default and included with just a switch like `-Prerelease`. AFAIK you don't pass in `true` or `false` but I may be wrong here. (I'm just looking at `nuget.exe list -help` and `nuget.exe install -help`)

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Goals
+
+* Providing search functionality for NuGet packages through the commandline for NuGet.exe
+* Implementing a prototype of this feature for dotnet.exe
+* Displaying the search results to users in a way that maximizes the visibility of the most relevant results
+
+## Solution
+
+The search feature leverages the search service provided by the NuGet API. Search targets multiple sources and displays all of their results one after the other. The command line interface is simple: __./NuGet.exe search \<query\>__.
```
joelverhagen: Consider also mentioned whether the query term is required (`nuget.exe list` does not require one but it's up to you), how to specify multiple terms (in quotes, most likely), and the query syntax (determined by server implementation, e.g. nuget.org has [this](https://docs.microsoft.com/en-us/nuget/consume-packages/finding-and-choosing-packages#search-syntax).

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and a preview of the description of the package.
+
+```
+====================
+Source: NuGet.org
+--------------------
+> Microsoft.Extensions.Logging.Abstractions | 5.0.0-preview.6.20305.6 | Downloads: 345,145,935
```
joelverhagen, line 36: I really like this lines format. It lends itself to grep'ing very well since there is a well-known character `>` at the beginning of the line. Consider indenting the description lines by at least one space so that a description containing `>` is not a false positive. Also, the indenting might look nice a la `dotnet list package`

designs/SearchCommandSpec.md
donnie-msft: So the client won't be sorting, gotcha.
I'd still like to see some note to the user, like `(sorted by relevance according to the source)` ? 
We get so many questions about lack of sort/filter, so somehow stating it might be valuable. At least consider it for `-verbosity detailed`.

designs/SearchCommandSpec.md
donnie-msft: Consider full/more characters when using `-verbosity detailed`.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PreRelease | `true` or `false` determining whether to include pre-release packages | -PreRelease `<true/false>` |
+| PackageTypes | The package type to use to filter packages. If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
+| Source | Specific package source(s) to search instead of querying the default sources in __nuget.config__ | -Source `<Source URL>`|
+| Take | The number of results to return | -Take `<positive integer>` |
+| Help | Displays help information for the command | -Help |
```
joelverhagen, line 71: Riffing off what @donnie-msft said, what about `-Verbosity` here?

When you increase verbosity on `nuget.exe list`, it also shows the API URL that is being hit. Might be nice, a la https://github.com/NuGet/Home/pull/9630 (shameless plug)

designs/SearchCommandSpec.md
donnie-msft: Consider keeping the full text for `-verbosity detailed`

designs/SearchCommandSpec.md
donnie-msft:
```suggestion
| PackageTypes | The package type to use to filter packages (if supported by the source). If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
```

donnie-msft: Should indicate whether the Filter actually got applied.
I think in my ideal world, we wouldn't simply return empty, but instead output:
(at least consider for `-verbosity detailed`)
- "Applying Filters"
- "Ignoring Filters (not supported by source)"
- Error: "Invalid Filter"

designs/SearchCommandSpec.md
donnie-msft: +1 to specifying the default.

designs/SearchCommandSpec.md
donnie-msft: You may find some option flags requiring true/false in nuget.exe, which is a bug (so don't copy everything you see there 😄). 
Maybe even rename this to `IncludePreRelease` ? Implying that the default is false.

designs/SearchCommandSpec.md
loic-sharma: Looks great!

donnie-msft (approved):
Left a few more comments, but looking good!

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,96 @@
...
+* Author(s): [Advay Tandon](https://github.com/advay26)
+* Issue: [9704](https://github.com/NuGet/Home/issues/9704) - Search CLI Tool for NuGet.exe and Dotnet.exe
+
+## Problem Background
+
+Search functionality from the command line for NuGet.exe and the .NET Core CLI has been a feature request from the NuGet open source community for some time. Additionally, the NuGet API is already equipped to handle search requests, so an extension of this service to command line users would be both practical to implement and also highly valuable.
+
+## Who are the customers
+
+This targets all NuGet.exe and .NET users that want to be able to search for packages from the command line, rather than using NuGet Gallery or VS PMUI. Also, while `nuget.exe list` allows users to search different package sources for packages with a search query, the results are returned sorted by the package Id, rather than by relevance, as proposed for this Search command. Further, the Search command would provide more information on the packages than List currently does.
```
loic-sharma: This is great! Another common problem that is addressed by your spec are customers that use `nuget.exe` on a private NuGet server that only supports the NuGet V3 protocol but not V2 (in case you're interested, see https://github.com/loic-sharma/BaGet/issues/187). Today, these customers will get the following error if they use the `nuget list` command:

> This version of nuget.exe does not support listing packages from package source 'http://get.xxxxxxxx.com/v3/index.json'.

Your spec gives these customers an escape hatch if the server supports the search protocol. We could improve the error message to say something like:

> This version of nuget.exe does not support listing packages from package source 'http://get.xxxxxxxx.com/v3/index.json'. Consider using the nuget.exe search command instead.

What do you think?

designs/SearchCommandSpec.md
advay26-zz: I've added Donnie's disclaimer in the PackageTypes description, and _-Verbosity detailed_ can provide additional information on the filters being used and whether they were applied or not.

designs/SearchCommandSpec.md
advay26-zz: I've changed -PreRelease to just be a switch. I haven't renamed it to -IncludePreRelease simply to maintain consistency with other nuget commands like _list_.

designs/SearchCommandSpec.md
advay26-zz: I've included this change, but the PrintJustified method I used in my prototype needs some adapting to fit my needs here so I'm going to continue working on this.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,106 @@
...
+Each search result displays the name, version, number of downloads, and a preview of the description of the package.
+
+```
+====================
+Source: NuGet.org
+--------------------
+> Microsoft.Extensions.Logging.Abstractions | 5.0.0-preview.6.20305.6 | Downloads: 345,145,935
+  Logging abstractions for Microsoft.Extensions.Logging.
+
+Commonly Used Types:
```
joelverhagen: nit: wouldn't this be indented too?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,106 @@
...
+No results found.
+```
+
+### Optional Arguments
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PackageTypes | The package type to use to filter packages (if supported by the source). If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
```
joelverhagen: nit, maybe a better user experience is saying "package type filtering is not supported on this feed" instead of the same message that occurs if they have too specific of parameters and no results are found. Said another way, this problem of no package type filtering cannot be resolved by the user but no results probably can.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,106 @@
...
+| Verbosity | The level of detail to display in the output. The default is _normal_. (See the note below)  | -Verbosity `<quiet\|normal\|detailed>` |
+| Help | Displays help information for the command | -Help |
+
+__NOTE__
+
+Verbosity Levels:
+
+* _quiet_ - Package ID, Version
+* _normal_ - Package ID, Version, Downloads, Preview of Description
+* _detailed_ - Package ID, Version, Downloads, Full Description, Other information regarding the arguments and filters applied to the search
```
joelverhagen: nit: can the query URLs be printed out at this verbosity level too? Not a big deal but it can help in some cases to diagnose problems.

joelverhagen (approved):

designs/SearchCommandSpec.md
advay26-zz: Yes, that's the aim. There was a bug in PrintJustified, which is the Console Method that prints indented text to the terminal for commands like list. I fixed that and created another [PR](https://github.com/NuGet/NuGet.Client/pull/3489) for it , so once that's approved the indentation should work fine. I'll fix it in the spec in the meanwhile.

donnie-msft (approved):

loic-sharma (approved):

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,103 @@
...
+Search functionality from the command line for NuGet.exe and the .NET Core CLI has been a feature request from the NuGet open source community for some time. Additionally, the NuGet API is already equipped to handle search requests, so an extension of this service to command line users would be both practical to implement and also highly valuable.
+
+## Who are the customers
+
+This targets all NuGet.exe and .NET users that want to be able to search for packages from the command line, rather than using NuGet Gallery or VS PMUI. Also, while `nuget.exe list` allows users to search different package sources for packages with a search query, the results are returned sorted by the package Id, rather than by relevance, as proposed for this Search command. Further, the Search command would provide more information on the packages than List currently does.
+
+## Goals
+
+* Providing search functionality for NuGet packages through the commandline for NuGet.exe
+* Implementing a prototype of this feature for dotnet.exe
```
chrisraygill, line 18: What does a prototype entail here? Are there areas where the experience between nuget.exe and dotnet.exe would diverge?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,103 @@
...
+
+```
+PS C:\> ./NuGet.exe search logging
+```
+
+Each search result displays the name, version, number of downloads, and a preview of the description of the package.
+
+```
+====================
+Source: NuGet.org
```
chrisraygill, line 34: How are we deciding the display order of the sources? Response order, alphabetical, # of results?

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Optional Arguments
+
+Search provides users with some optional arguments that they can use to customize their search queries:
+
+| Name | Description | Usage |
+| ---  |     ---     |  :-:  |
+| PackageTypes | The package type to use to filter packages (if supported by the source). If the provided package type is not a valid package type as defined by the [Package Type document](https://github.com/NuGet/Home/wiki/Package-Type-%5BPacking%5D), an empty result will returned. | -PackageType `<Package Type>`|
+| PreRelease | Pre-release packages are not included by default, but can be included by using this argument | -PreRelease |
```
chrisraygill, line 67: This is a nitpick, but I think options might typically be lowercased https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet. I'm not sure if the casing here may confuse anyone.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,103 @@
...
+
+* What represents the most effective design choice for displaying the search results, in terms of spacing, demarcation, truncating package descriptions, and so on?
+* Would pagination (using the 'more' or 'less' commands) provide a better viewing experience for users? Would users prefer to pipe the search results to these commands themselves, or have Search do it for them? See _Considerations_ for more. 
+* What, if any, additional information should be included for the packages returned by Search?
+
+## Considerations
+
+### 1. Combining the results from different sources
+
+Initially, we considered consolidating all search results from the various sources and interleaving them together, rather than separating them based on their source. However, users may wish to know where a particular package is located. Also, combining the search results for different sources leads to the complex problem of how to combine these results most effectively.
```
chrisraygill, line 99: Don't we combine results in Visual Studio? I imagine we should be able to use the same method to combine results here. My instinct is that I like has results from each source displayed separately, but I wonder how often customers may be using some configuration their team set up and actually aren't aware what source a particular package they need is coming from. @loic-sharma any input here?

chrisraygill (approved):

designs/SearchCommandSpec.md
advay26-zz: Other NuGet commands like [`list`](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-list) seem to use uppercase options similar to the ones used here, so I thought it made sense to do this in the interest of consistency

designs/SearchCommandSpec.md
advay26-zz: Unless specific sources are specified, the sources are queried in the order they appear in the nuget.config file (as far as I know).

designs/SearchCommandSpec.md
chrisraygill: Oh that's interesting. It makes sense to be consistent with nuget.exe docs since that's what we're focusing on 👍

designs/SearchCommandSpec.md
advay26-zz: So far my focus has been on the NuGet.exe experience, and I haven't been able to explore what the dotnet.exe experience would look like. Maybe @zkat could answer this better?

designs/SearchCommandSpec.md
nkolev92: We are using Lucene.NET.dll to combine the search results. 
We cannot use Lucene.NET in dotnet.exe because of source-build concerns. See https://github.com/dotnet/source-build#net-core-source-build. 

From their repo: 


> The key goal of this repository is to satisfy the official packaging rules of commonly used Linux distributions, such as Fedora and Debian. Many Linux distributions have similar rules. These rules tend to have two main principles: consistent reproducibility, and source code for everything.

We basically would need to implement client side merging with .NET runtime only dependencies. 
Not sure if we have anything readily available.

joelverhagen: Also merging search results falls apart when one feed is very slow and another one is fast on the command line. For the VSUI you can adjust the ordering dynamically as slow results come in. For a command line once you write a search result it can't be (easily) undone so you need to wait for all search pages to return before showing anything. This can lead to a subpar experience when you have a slow feed in the mix.

designs/SearchCommandSpec.md
nkolev92: I think we should drop PackageTypes from the initial implementation. 

1. It increases the scope of the feature.
1. We don't have super well understood use cases for package types yet. `dotnet tool search` is an obvious example but that has it's use right now. 

I'd prefer to add that at a later point based on community feedback.

designs/SearchCommandSpec.md
```diff
@@ -0,0 +1,103 @@
...
+* _normal_ - Package ID, Version, Downloads, Preview of Description
+* _detailed_ - Package ID, Version, Downloads, Full Description, Other information such as the query URL, and the arguments and filters applied to the search
+
+## Future Work
+
+* Integrating the Search feature into dotnet.exe
+* Adding further optional arguments that allow users to customize their search queries
+* Using columns rather than `|` to display results
+* Consider using pagination to display the search results
+* Allow multiple/no query terms
```
nkolev92: Deprecating list maybe? 
Should list point to the new search command? Was that something we considered?

designs/SearchCommandSpec.md
nkolev92: Only thing I'd suggest is calling out the list deprecation plan. 

We want to write a feature that people will use. Deprecating list and pointing to search will be something that should increase the usage of the command.

nkolev92 (commented):
Looks great. 
A few logistics comments. 

I think we should not add PackageTypes support in V1. Beyond that fix the title of the spec maybe? Given that dotnet.exe is not part of V1. 

I do have some concerns about shipping new features in nuget.ee but not in dotnet.exe but if PMs are fine, I won't insist.

designs/SearchCommandSpec.md
zivkan: nuget.exe uses powershell `-PascalCase` style arguments, whereas dotnet uses `--posix` or `--snake-case` style arguments.

nkolev92 (approved):
Well done! 💯

Issue #3788, opened 2016-10-27
PackageReference with a star in version is displayed under the updates tab/incorrectly represented in the installed tab


emgarten:
### Steps
1. File > New Project > NETCore console app
2. Add a package reference with a star
3. Open the NuGet UI and go to the installed tab
### Expected

No updates displayed, and the correct package info is shown
### Actual

An update is displayed and no package info

``` xml
    <PackageReference Include="Newtonsoft.Json">
      <Version>9.0.*</Version>
    </PackageReference>
```

![capture](https://cloud.githubusercontent.com/assets/6381138/19752314/3b52d558-9bb1-11e6-9326-19bc09ffa8ee.PNG)

Another customer report: 
https://developercommunity.visualstudio.com/content/problem/277887/nuget-package-manager-doesnt-respect-package-refer.html

Another scenario: https://github.com/NuGet/Home/issues/3101

[SPEC]
https://github.com/NuGet/Home/wiki/%5BSpec%5D-NuGet-Build-Integrated-Project-handling-of-floating-versions-and-ranges-in-Visual-Studio-Client

hkoelewijn:
It would be nice if VS allowed you to specify wildcards in the package manager in stead of us having to open the proj file directly

pmiossec:
I think that's linked but when, in the "Package Manager Console", when you list the packages installed (with `Get-Package` command), you get the last number at "0".

Example: 
For: `<PackageReference Include="Newtonsoft.Json" Version="9.0.*" />`
you will always get (even if `9.0.3` is really installed) : `Newtonsoft.Json                     {9.0.0}` 

That's surely why "Nuget Package Manager` see it as "Should be updated"...

Is there some update on this issue? Seems to be postponed since v4.0... :cry:

Neme12:
I found this after searching and trying out different things for 20 minutes thinking I must be doing something wrong with the wildcard :confused:

tobiasschulz:
Yeah I really don't get why this has not been fixed yet.

rfcdejong:
Any ETA?

pmiossec:
That's the type of issue, that completely prevent the use of wildcard (I had to fallback to hardcoded version 😢 ) and so to put in place an update strategy that let me think that the paket ( https://github.com/fsprojects/Paket ) community driven project is a lot better solution to manage dependencies in a .net project....

[...]

PR #9773, opened 2020-07-09
Create FloatingVersions-In-PMUI.md

martinrrm:
Issue: https://github.com/NuGet/Home/issues/3788


designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Floating Version Display in Solution View
+
+We will add a new column to show the requested version to match the behavior of `dotnet list package` command. Show correct highest resolved version in the solution in the package list view.
+
+![Solution View](../resources/FloatingVersionsInPMUI/SolutionView.png)
+
+## Future Work
+
+* Indicate the resolved version in the solution explorer.
```
JonDouglas: I believe this should be included in this spec as it completes it. After talking to users regarding floating versions, not many understand the wildcard syntax even when they use it. We should be very clear with what version is `requested` and what version is `resolved`.

designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,104 @@
...
+![SolutionExplorer](../resources/FloatingVersionsInPMUI/ProjectFile&SolutionExplorer.png)
+
+#### Project Manager UI
+
+![Package UI](../resources/FloatingVersionsInPMUI/PackageManagerUI.png)
+
+![Project UI](../resources/FloatingVersionsInPMUI/PackageManagerForSolution.png)
+
+```
+The requested and installed versions are not correct because dotnet list package shows a different installed version.
```
JonDouglas: We should remain as true as we can to the `dotnet` cli as possible. `Requested` and `Resolved` would be better words than what we have today (Installed / Version).

designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,104 @@
...
+
+![Solution View](../resources/FloatingVersionsInPMUI/SolutionView.png)
+
+## Future Work
+
+* Indicate the resolved version in the solution explorer.
+
+## Open Questions
+
+Should we show available updates for packages that use custom version ranges or floating versions that don’t resolve to the latest version?
```
JonDouglas, line 111: What's an example of this?

designs/FloatingVersions-In-PMUI.md
martinrrm: When we were doing the initial design spec, VS had a bug and this has been fixed, the solution explorer does show the resolved version. I will update the spec

designs/FloatingVersions-In-PMUI.md
martinrrm: https://github.com/NuGet/Home/blob/dev-martinrrm-FloatingVersionsInPMUI/designs/FloatingVersions-In-PMUI.md#example

In this example the user requested for a version range, but the newest version is bigger than the range that the user specified.

designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,103 @@
+
+# Floating Versions in the Package Manager UI
+
+* Status: In Review
+* Author(s): [Martin Ruiz](https://github.com/martinrrm)
+* Issue: [3788](https://github.com/NuGet/Home/issues/3788) PackageReference with a star in version is displayed under the updates tab/incorrectly represented in the installed tab
+
+## Problem Background
+
+Currently the user can specify a version range or a floating version, but the current behavior of the package manager UI has poor support for this, displaying an incorrect installed version. If the user indicates a floating version in any package (eg `5.*`) the PM UI will display the minimum version possible (`5.0.0`) as installed even though the resolved version could be different of the resolved version in the command `dotnet list package`. Furthermore the user cannot specify a version range free hand.
```
nkolev92: Change from: `resolved version could be different of the`  `to resolved version could be different from the `
A couple of small suggestions:

```suggestion
In PackageReference the user can specify a version range or a floating version, but the package manager UI has poor support for this, displaying an incorrect installed version. 
If the user indicates a floating version for any package (eg `5.*`) the PM UI will display the minimum version possible (`5.0.0`) as installed even though the resolved version could be different of the resolved version in the command `dotnet list package`. 
Furthermore the user cannot specify a version range free hand.
```

designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Current behavior
+
+### Updates Tab
+
+With the current behavior even if the user indicated a floating version, such as `*`, for the package version, the UI will tell the user that there is an update available which is misleading.
+
+#### Project File and Solution Explorer
+
+![SolutionExplorer](../resources/FloatingVersionsInPMUI/ProjectFile&SolutionExplorer.png)
```
nkolev92, line 20: I think we should update this picture. 

The solution explorer actually updates to the resolved version.

designs/FloatingVersions-In-PMUI.md
```diff
@@ -0,0 +1,103 @@
...
+* Author(s): [Martin Ruiz](https://github.com/martinrrm)
+* Issue: [3788](https://github.com/NuGet/Home/issues/3788) PackageReference with a star in version is displayed under the updates tab/incorrectly represented in the installed tab
+
+## Problem Background
+
+Currently the user can specify a version range or a floating version, but the current behavior of the package manager UI has poor support for this, displaying an incorrect installed version. If the user indicates a floating version in any package (eg `5.*`) the PM UI will display the minimum version possible (`5.0.0`) as installed even though the resolved version could be different of the resolved version in the command `dotnet list package`. Furthermore the user cannot specify a version range free hand.
+
+### Current behavior
+
+### Updates Tab
```
nkolev92: The spec below talks about more than just the updates tab. 

I think the headers need a bit of work. 

* Project level PM UI
  * Installed tab
  * Updates tab

* Solution level PM UI
  * Installed tab
  * Updates tab.

I don't think you have the updates tab for solution level PM UI. 
Consider adding it, but it's very similar to the installed tab.

nkolev92: Ofc the `#### Project File and Solution Explorer` sub-section is good an useful.

nkolev92 (commented):
Looks great. 

Some minimal feedback. 

As far as the solution goes, I think it'll be nice to see *all* display scenarios. 

So Installed/Update tab for each solution and PM UI.

zivkan:
nitpick: when viewing the markdown file on github, github limits the width displayed, even though I have a nice, large, widescreen monitor. The images are downscaled to fit within this width, and have become unreadable.  I don't know if it's worthwhile re-creating the screenshots, but in the future consider cropping the screenshot to a much smaller section, rather than showing the whole window/screen. Not only does this reduce downscaling, but if there's only a small part of teh screenshot that is relevant to the spec, it's clear that that section is, rather than having the reader hunt around the entire screenshot to figure it out (you had to use green boxes in one to highlight because there was too much non-informational screenshot).

nkolev92 (approved):

Issue #9752, opened 2020-07-06
Advay's summer internship project 2020


advay26-zz:
This is a tracking issue for all the tasks that @advay26 will take on as part of his internship summer project.

zkat:
This is all done! :)

chrisraygill:
Hey, can this issue be closed since you're an FTE now @advay26? 😁

PR #9832, opened 2020-07-22
Created a tech spec for Embedded Readmes

advay26-zz:
Created an implementation spec for the Embedded Readmes feature [[Issue 9752](https://github.com/NuGet/Home/issues/9752)].

### DIFF
--- a/designs/EmbeddedReadmesTechSpec.md
+++ b/designs/EmbeddedReadmesTechSpec.md
@@ -22,7 +22,7 @@ NuGet customers that use the commandline to publish their packages, and want to 
 ## Non-Goals
 
 * Server side functionality
-* Parsing the readme as a valid markdown file
+* Parsing the readme as a valid markdown file on the Client side
 
 ## Solution
 
@@ -50,53 +50,35 @@ After this, a `ManifestMetadata` object is used to create the output *nuspec* fi
 AddElementIfNotNull(elem, ns, "readme", metadata.Readme);
 ```
 
-**If no Readme property is present in the source file, but a readme.md file exists in the base directory**
-
-Once the properties and files have been processed, if the PackageBuilder object's Readme property is null, we need to look through the base directory for a "readme.md" file. If such a file exists, we must include it in the package and and populate the Readme property with "readme.md".
-
-```
-public void CheckForReadme(string basePath)
-{
-    string readmeName = "readme.md";
-    List<PhysicalPackageFile> searchFiles = ResolveSearchPattern(basePath, readmeName, string.Empty, false).ToList();
-
-    if (searchFiles.Any())
-    {
-        Readme = readmeName;
-
-        if (!Files.Contains(searchFiles[0]))
-        {
-            Files.Add(searchFiles[0]);
-        }
-    }
-}
-```
-
-This CheckForReadme method in [PackageBuilder.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs) is called [here](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L725) and [here](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Commands/CommandRunners/PackCommandRunner.cs#L726), based on whether the package is being built using a *nuspec* or *csproj* file.
-
 **Validation and Errors**
 
 To validate the Readme property and the corresponding file when `pack` is called, the ValidateReadmeFile method in [PackageBuilder.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs) checks that:
 * The file extension matches a Markdown file
 * The file specified in the Readme property exists in the package's files
-* The file can be opened
 * The file size does not exceed 1 MB
 * The file is not empty
 
 Each of these errors has a corresponding log code (NU5038 - NU5042) and error message defined in [NuGetLogCode.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs) and [NuGetResources.resx](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.resx) respectively, and the appropriate support must be added in [NuGetResources.Designer.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.Designer.cs).
 
+**NOTE:** The Server team will parse and validate the markdown contents of the Readme file when it is uploaded.
+
 ## Future Work
 
-* Adding readme support through the VS Project Properties UI
-* Adding a `View Documentation` link to the package readme in the PM UI
-
+* Adding readme support through the VS Project Properties UI (Project System)
+* Adding a `View README` link in the PM UI that points to the package's readme file
 ## Open Questions
 
-* Should we warn users if the Readme property is empty?
 * Are there any other validation checks to do when packing a Readme?
 
 ## Considerations
 
+### 1. Should we warn users if the Readme property is empty?
+
+Based on feedback from both the NuGet team and customers, we should not do this. The goal of encouraging user adoption of Embedded Readmes can be pursued as part of a separate Package Quality endeavor.
+
+### 2. Automatically packing a Readme file from the base directory
+
+Similar to the above point, this functionality is not something that we will look to add right now. Relying on explicit user input for the metadata is more desirable, and avoids springing any surprises on customers.
 
 ### References
 
--- a/designs/EmbeddedReadmesTechSpec.md
+++ b/designs/EmbeddedReadmesTechSpec.md
@@ -22,6 +22,7 @@ NuGet customers that use the commandline to publish their packages, and want to 
 ## Non-Goals
 
 * Server side functionality
+* Parsing the readme as a valid markdown file
 
 ## Solution
 
--- a/designs/EmbeddedReadmesTechSpec.md
+++ b/designs/EmbeddedReadmesTechSpec.md
@@ -28,7 +28,7 @@ NuGet customers that use the commandline to publish their packages, and want to 
 
 ## Solution
 
-**NOTE:** To see the code described in this section, you can look at the edited files [here](https://github.com/NuGet/NuGet.Client/pull/3573/files).
+**NOTE:** To see the code described in this section, you can look at the edited files [here](https://github.com/NuGet/NuGet.Client/pull/3575/files).
 
 First, we need to add a `readme` string element to the [*nuspec* schema](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Packaging/compiler/resources/nuspec.xsd) to ensure that the readme property is recognized and parsed from the *nuspec* file. We also need to ensure that the `PackageReadmeFile` property is parsed from a *csproj* file by editing [NuGet.Build.Tasks.Pack.targets](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L198) to include this line:
 
--- a/designs/EmbeddedReadmesTechSpec.md
+++ b/designs/EmbeddedReadmesTechSpec.md
@@ -24,6 +24,7 @@ NuGet customers that use the commandline to publish their packages, and want to 
 
 * Server side functionality
 * Parsing the readme as a valid markdown file on the Client side
+* Enabling readme selection through the VS Project Properties UI (Project System)
 
 ## Solution
 
@@ -69,6 +70,7 @@ Each of these errors has a corresponding log code (NU5038 - NU5042) and error me
 
 * Adding readme support through the VS Project Properties UI (Project System)
 * Adding a `View README` link in the PM UI that points to the package's readme file
+
 ## Open Questions
 
 * Are there any other validation checks to do when packing a Readme?
--- a/designs/EmbeddedReadmesTechSpec.md
+++ b/designs/EmbeddedReadmesTechSpec.md
@@ -15,8 +15,9 @@ NuGet customers that use the commandline to publish their packages, and want to 
 ## Goals
 
 * Implement the Client side functionality for Embedded Readmes:
-    * nuget pack support
-    * dotnet pack support
+    * `nuget pack` support
+    * `dotnet pack` support
+    * `msbuild /t:pack` support
     * Validation of readme files
 
 ## Non-Goals
@@ -26,7 +27,7 @@ NuGet customers that use the commandline to publish their packages, and want to 
 
 ## Solution
 
-**NOTE:** To see the code described in this section, you can look at the edited files [here](https://github.com/NuGet/NuGet.Client/compare/dev-advay26-readme).
+**NOTE:** To see the code described in this section, you can look at the edited files [here](https://github.com/NuGet/NuGet.Client/pull/3573/files).
 
 First, we need to add a `readme` string element to the [*nuspec* schema](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Packaging/compiler/resources/nuspec.xsd) to ensure that the readme property is recognized and parsed from the *nuspec* file. We also need to ensure that the `PackageReadmeFile` property is parsed from a *csproj* file by editing [NuGet.Build.Tasks.Pack.targets](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L198) to include this line:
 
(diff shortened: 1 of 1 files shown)


designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,102 @@
...
+## Goals
+
+* Implement the Client side functionality for Embedded Readmes:
+    * nuget pack support
+    * dotnet pack support
+    * Validation of readme files
+
+## Non-Goals
+
+* Server side functionality
```
dominoFire, line 25: Include as a non-goal: parse readme as a markdown, because:
- we don't have built-in libraries to parse markdown, and adding one would take some time
- There's multiple Markdown implementations (github, daring fireball etc.)

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,102 @@
...
+
+After this, a `ManifestMetadata` object is used to create the output *nuspec* file that is included in the nupkg. For this step, we need to edit the [ToXElement](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Packaging/PackageCreation/Xml/PackageMetadataXmlExtensions.cs#L32) method in PackageMetadataXmlExtensions.cs to include this line:
+
+```
+AddElementIfNotNull(elem, ns, "readme", metadata.Readme);
+```
+
+**If no Readme property is present in the source file, but a readme.md file exists in the base directory**
+
+Once the properties and files have been processed, if the PackageBuilder object's Readme property is null, we need to look through the base directory for a "readme.md" file. If such a file exists, we must include it in the package and and populate the Readme property with "readme.md".
```
nkolev92: I'd caution against this.

There's no precedent to this approach, and there are features with similar behavior, see licenses, icons. 

Users being explicit about metadata is good!

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,102 @@
...
+
+```
+public void CheckForReadme(string basePath)
+{
+    string readmeName = "readme.md";
+    List<PhysicalPackageFile> searchFiles = ResolveSearchPattern(basePath, readmeName, string.Empty, false).ToList();
```
nkolev92: We already glob a couple of times during pack, we don't need an additional glob.

designs/EmbeddedReadmesTechSpec.md
nkolev92: There are really 3 pack scenarios. 

1. dotnet pack(msbuild pack). All the content that needs included in the package will be included through Content/None items through MSBuild. We should not do additional searches. We should not pull in things accidentally that might not be part of the project. 
2. nuget.exe pack with a nuspec - without a files element...this basically will include all files under the nuspec directory.
3. nuget.exe pack with a nuspec - with a files element - we will include only the files that the files element has specified. 

In neither of the above 3 cases, do we arbitrarily glob looking for a file to include as metadata.

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+This CheckForReadme method in [PackageBuilder.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs) is called [here](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L725) and [here](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Commands/CommandRunners/PackCommandRunner.cs#L726), based on whether the package is being built using a *nuspec* or *csproj* file.
+
+**Validation and Errors**
+
+To validate the Readme property and the corresponding file when `pack` is called, the ValidateReadmeFile method in [PackageBuilder.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs) checks that:
+* The file extension matches a Markdown file
+* The file specified in the Readme property exists in the package's files
+* The file can be opened
```
nkolev92: Can you clarify what opening a file means in this case?

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,102 @@
...
+Each of these errors has a corresponding log code (NU5038 - NU5042) and error message defined in [NuGetLogCode.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs) and [NuGetResources.resx](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.resx) respectively, and the appropriate support must be added in [NuGetResources.Designer.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.Designer.cs).
+
+## Future Work
+
+* Adding readme support through the VS Project Properties UI
+* Adding a `View Documentation` link to the package readme in the PM UI
+
+## Open Questions
+
+* Should we warn users if the Readme property is empty?
```
nkolev92: If they don't have a readme at all? 
I'd vote no for now. 
Lots of customer elevate warnings to errors and while a good quality package will have a readme, I'd rather that get added during the actual high quality package effort (@chgill-MSFT, @JonDouglas will know more).

nkolev92 (commented):
This is a great write-up @advay26. 
It's very detailed. Given that I'm not gonna be able to make the meeting, adding my comments now.

designs/EmbeddedReadmesTechSpec.md
advay26-zz: I've removed this functionality, and the references to it within this spec, based on feedback from the review meeting and Chris' interviews with customers.

designs/EmbeddedReadmesTechSpec.md
advay26-zz: This was meant to cover any other exceptions thrown while attempting to open the embedded readme file, but I've now removed it from the spec in order to be more clear about the validation goals.

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,85 @@
...
+    * Validation of readme files
+
+## Non-Goals
+
+* Server side functionality
+* Parsing the readme as a valid markdown file on the Client side
+
+## Solution
+
+**NOTE:** To see the code described in this section, you can look at the edited files [here](https://github.com/NuGet/NuGet.Client/compare/dev-advay26-readme).
```
dominoFire: Will the link be alive once this spec is merged? Right now it points to a new page.

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,85 @@
...
+## Who are the customers
+
+NuGet customers that use the commandline to publish their packages, and want to be able to include Readme files through the `pack` process, rather than having to do it manually after the package has been published.
+
+## Goals
+
+* Implement the Client side functionality for Embedded Readmes:
+    * nuget pack support
+    * dotnet pack support
+    * Validation of readme files
```
dominoFire: Add a bullet point for MSBuild, since you're already adding support in the NuGet Pack targets.

```suggestion
    * `nuget pack` support
    * `dotnet pack` support
    * `msbuild /t:pack` support
    * Validation of readme files
```

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,85 @@
...
+* [PackageBuilder](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L25)
+
+In order to transmit the `Readme` value from the source file (*nuspec* or *csproj*) to the final output *nuspec* file, the appropriate assignment statements are needed at the relevant locations.
+
+After this, a `ManifestMetadata` object is used to create the output *nuspec* file that is included in the nupkg. For this step, we need to edit the [ToXElement](https://github.com/NuGet/NuGet.Client/blob/64ed0cb4054226f6060752757d29c50287b312b3/src/NuGet.Core/NuGet.Packaging/PackageCreation/Xml/PackageMetadataXmlExtensions.cs#L32) method in PackageMetadataXmlExtensions.cs to include this line:
+
+```
+AddElementIfNotNull(elem, ns, "readme", metadata.Readme);
+```
+
```
dominoFire, line 56: Describe briefly how are you reading the new readme property. Tip: Look at NuspecReader class

dominoFire (approved):
Minor comments

designs/EmbeddedReadmesTechSpec.md
advay26-zz: I've added a reference to how the Readme property is read by the `GetReadme` method in the `NuspecReader` class. I imagine this is what you were referring to, but feel free to correct me if I'm wrong.

designs/EmbeddedReadmesTechSpec.md
advay26-zz: I've changed the link to point to my (Draft) PR instead of the branch I was working on so that the link will still be functional  even once the PR is merged and the branch is deleted.

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,88 @@
+# Embedded Readmes Technical Spec
+
+* Status: Implementing
+* Author(s): [Advay Tandon](https://github.com/advay26)
+* Issue: [6873](https://github.com/NuGet/Home/issues/6873) README.md (markdown description) support for NuGet packages
+
+## Problem Background
+
+Users of NuGet want to be able to include Readme files as part of the package when they `pack` and `push` their project through the commandline. The [design document](https://github.com/NuGet/Home/wiki/Packaging-READMEs-within-the-nupkg) outlines the work needed for this feature and the intended results.
```
dominoFire: The link to `https://github.com/NuGet/Home/wiki/Packaging-READMEs-within-the-nupkg` points to a new wiki page. Is it intentional? Are you going to create a new wiki page once this PR is merged?

dominoFire (approved):

designs/EmbeddedReadmesTechSpec.md
advay26-zz: It looks like the name of the document within the wiki was changed at some point. I've updated the link to point to the correct place now. Thanks for pointing this out!

advay26-zz:
I've made changes to the tech spec based on feedback I've received, specifically addressing concerns about the automatic packing of READMEs (now disabled) and the Validation strategy (who validates the actual markdown file). I wasn't able to request reviews from some of you through the Reviewers tab, so I'm tagging you here to request comments/feedback/approvals.

@zkat @zivkan @sbanni @erdembayar @rrelyea @srdjanjovcic @cristinamanum @heng-liu @donnie-msft

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,88 @@
...
+```
+AddElementIfNotNull(elem, ns, "readme", metadata.Readme);
+```
+
+**Validation and Errors**
+
+To validate the Readme property and the corresponding file when `pack` is called, the ValidateReadmeFile method in [PackageBuilder.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs) checks that:
+* The file extension matches a Markdown file
+* The file specified in the Readme property exists in the package's files
+* The file size does not exceed 1 MB
```
donnie-msft: These validations are mainly for nuget.org, right? If we eventually want to have looser restrictions for readmes for local or private feeds, we feel this is a minimum set of validations going forward? eg, In future iterations of this feature, I could have a 5GB readme on my private feed... who cares? But I can't publish to nuget.org with that. 

I think the answer here is "yes", so may want to mention that in the description how we feel this validation set could evolve without causing backwards compatibility issues.

If you don't like the idea of having Pack validation eventually removed/less restrictive, then I'd question why we need to validate on Pack and not on Push?...
```
Pack -> warn that nuget.org won't accept >1MB
Push -> error from nuget.org (server doesn't accept > 1MB)
(in the future)
Push  (NOT nuget.org) -> accept/error depending on server behavior
Push (local feed) -> local feed doesn't care
```

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Future Work
+
+* Adding readme support through the VS Project Properties UI (Project System)
+* Adding a `View README` link in the PM UI that points to the package's readme file
+## Open Questions
+
+* Are there any other validation checks to do when packing a Readme?
+
+## Considerations
```
donnie-msft, line 77: I agree with these. It's opt-in for now.

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,88 @@
...
+* The file extension matches a Markdown file
+* The file specified in the Readme property exists in the package's files
+* The file size does not exceed 1 MB
+* The file is not empty
+
+Each of these errors has a corresponding log code (NU5038 - NU5042) and error message defined in [NuGetLogCode.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs) and [NuGetResources.resx](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.resx) respectively, and the appropriate support must be added in [NuGetResources.Designer.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.Designer.cs).
+
+**NOTE:** The Server team will parse and validate the markdown contents of the Readme file when it is uploaded.
+
+## Future Work
```
donnie-msft, line 68: 👍

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Goals
+
+* Implement the Client side functionality for Embedded Readmes:
+    * `nuget pack` support
+    * `dotnet pack` support
+    * `msbuild /t:pack` support
+    * Validation of readme files
+
+## Non-Goals
```
donnie-msft, line 23: May want to make it clear that this excludes the Visual Studio property page selection of a Readme.

designs/EmbeddedReadmesTechSpec.md
advay26-zz: @chgill-MSFT I was hoping you could clarify this

chrisraygill: This is a good question!

Advay and I had a chat about this. It looks like we don't validate for size on license file today. In general we don't make any assumptions on pack that a package isn't meant for NuGet.org today. Not only that, but the likelihood that a README will be larger than 1mb is very low.

Ultimately, we came to the conclusion that is makes sense to remove the size check on pack and only validate README file size on push/upload to NuGet.org.

donnie-msft: Sounds good! I think keeps our options open in the long run. 
Are we planning on changing `pack` to `push` in this paragraph?

donnie-msft: Nevermind, yes the basic checks will remain on `pack`. I'm guessing `push` validations are out of scope? May want to indicate that in **Future**?

advay26-zz: Since we won't be doing the validation on our side and the size check will only be done by nuget.org, should we say we validate size on `push`, or should we just add a note saying that "NuGet.org will only accept README files that do not exceed 1 MB"?

advay26-zz: I'll add the 1 MB size validation to the "Considerations" section and explain why we won't be doing it here. I'll include a reference to NuGet.org's 1 MB size limit there, and mention that this would result in an error when someone uses `push` to upload a package to NuGet.org. Does that seem fair?

advay26-zz: Done!

donnie-msft (approved):
Awesome! 👏

JonDouglas:
**Open Questions**
Are there any other validation checks to do when packing a Readme?

> When a user packs a NuGet package without a README, there should be a warning.

**Considerations**
1. Should we warn users if the Readme property is empty?
Based on feedback from both the NuGet team and customers, we should not do this. The goal of encouraging user adoption of Embedded Readmes can be pursued as part of a separate Package Quality endeavor.

> I believe we should set a standard with this feature today. READMEs are a critical component for someone to not have to navigate away from NuGet.org or Visual Studio in the sense that they can learn about the package & get started with it quickly. Many customers said they would prefer to not have to go to GitHub to view the README if possible.

2. Automatically packing a Readme file from the base directory
Similar to the above point, this functionality is not something that we will look to add right now. Relying on explicit user input for the metadata is more desirable, and avoids springing any surprises on customers.

> Does this mean adding a README by default? I don't think we need to do that for the scope of this spec.

3. Enforcing a size limit on Readme files on pack
Originally, we intended to enforce a 1 MB size limit on Embedded Readme files both while running pack on the Client-side and while using push to upload to NuGet.org. However, since we do not want to impose such size restrictions on users looking to upload packages to private or local feeds, we decided against throwing an error for this on pack. Now, the user can run pack to create packages with Readme files over 1 MB, but will receive an error if they try to upload such a package to NuGet.org, which enforces a strict 1 MB limit.

> The follow-up question here is that most readmes should generally be in the KB range & all resources for the README is hosted somewhere else like github content or another location. Will embedded readmes support this similar concept where outside resources can be accessed or will users need to embed those resources as well?

chrisraygill:
I followed up with Jon in chat regarding the above, but here is the response I gave for transparency:

>1.  I believe we should set a standard with this feature today. READMEs are a critical component for someone to not have to navigate away from NuGet.org or Visual Studio in the sense that they can learn about the package & get started with it quickly. Many customers said they would prefer to not have to go to GitHub to view the README if possible.

As @zivkan has brought in past discussions, warnings can be problematic for some customers who elevate warnings to errors. Since a README isn't strictly necessary for private/internal package use, this could be a frustrating experience for those who don't need it. 

The alternative is that we just print a "suggestion" or message as this would increase feature awareness without being overly aggressive. I decided not to do this for the initial release for a few reasons:
1. Customer feedback indicated interest in an overall package quality auditing experience which is aligned with @JonDouglas package quality indicators work. "Suggestions" will likely be apart of that work.
1. Our README support on NuGet.org is not at full parity with GitHub and may have some friction for customers to get their READMEs "NuGet.org ready."  We should understand and remove these friction points as well as we can before pinging customers for not adopting READMEs.
1. Most client/pack validations today are private/public package agnostic. Customer's who are creating private packages that don't need READMEs have no way of disabling these suggestions.

> 2. Does this mean adding a README by default? I don't think we need to do that for the scope of this spec.

👍

> 3. The follow-up question here is that most readmes should generally be in the KB range & all resources for the README is hosted somewhere else like github content or another location. Will embedded readmes support this similar concept where outside resources can be accessed or will users need to embed those resources as well?

We will not support embedded resources for README images in this initial release. Allowlist --> proxy service is a more likely evolution, so size shouldn't be much of an issue. 1mb is a lot for text.

aortiz-msft (approved):

designs/EmbeddedReadmesTechSpec.md
```diff
@@ -0,0 +1,93 @@
...
+* The file specified in the Readme property exists in the package's files
+* The file is not empty
+
+Each of these errors has a corresponding log code and error message defined in [NuGetLogCode.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Common/Errors/NuGetLogCode.cs) and [NuGetResources.resx](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.resx) respectively, and the appropriate support must be added in [NuGetResources.Designer.cs](https://github.com/NuGet/NuGet.Client/blob/0f8ad8263539cb9bc69c441569453c1da98fb4cc/src/NuGet.Core/NuGet.Packaging/PackageCreation/Resources/NuGetResources.Designer.cs).
+
+**NOTE:** The Server team will parse and validate the markdown contents of the Readme file when it is uploaded.
+
+## Future Work
+
+* Adding readme support through the VS Project Properties UI (Project System)
```
nkolev92: It might be nice to provide links to the future work.

nkolev92 (approved):
This is one of the most detailed technical specs I have seen in a while :) 
Looks great @advay26!

designs/EmbeddedReadmesTechSpec.md
advay26-zz: Done!

Issue #9850, opened 2020-07-28
Surface basic vulnerability metadata in PMUI view models


drewgillies:
Part of the work to display vulnerability information in the package graph on `Installed tab of Package Manager Dialog (PMUI).

Spec: https://github.com/NuGet/Home/wiki/Flag-vulnerable-packages

- Ingress registration metadata for vulnerabilities and augment view model metadata

PR #9896, opened 2020-08-11
Add package vulnerability design one-pager

drewgillies:
Re https://github.com/NuGet/Home/issues/9850, I'm submitting this one-pager spec for discussion. This PR: https://github.com/NuGet/NuGet.Client/pull/3540 is a candidate for design discussion.

### DIFF
--- a/designs/PackageVulnerability/PMUIVulnerability.md
+++ b/designs/PackageVulnerability/PMUIVulnerability.md
@@ -1,0 +1,55 @@
+
+# PMUI Display vulnerability information in the package graph
+
+* Status: [In Review] / Reviewed / Implementing / Implemented
+* Author(s): [drewgillies](https://github.com/drewgillies)
+* Issues: 
+[1](https://github.com/NuGet/Home/issues/9850) Surface basic vulnerability metadata in PM UI view models
+[2](https://github.com/NuGet/Home/issues/8755) Display vulnerability info on Installed tab of PMUI
+
+## Problem Background
+
+The [Flag Vulnerable Packages spec](https://github.com/NuGet/Home/wiki/Flag-vulnerable-packages) covers a number of MVP initiatives of which this is one. The problem being solved is feature parity with existing package services (npm) couples with a need to service our customers' security needs.
+
+## Who are the customers
+
+All publishers of NuGet.org packages, all consumers of NuGet.org packages.
+
+## Goals
+
+When a customer browses their installed packages in the PMUI, the MVP is for them to see notifications on their installed tab for vulnerable packages. 
+
+## Non-Goals
+
+Notifications on the Browse tab
+Restore warnings
+
+## Solution
+
+The package vulnerability data is now available in registration metadata, alongside data for deprecations which is currently in use.
+
+The plan is the extend the design used for surfacing deprecation notifications in the past, to also include vulnerability information. This includes models, view models and views. The work is likely to be very similar in all areas. The strengths of this approach are a reasonably fast time to market, and low risk.
+
+## Future Work
+
+Next steps are also to show transitive vulnerabilities, and beyond that we can continue to extend functionality into the Browse tab and other locations as customer engagement and general strategy determines.
+
+## Open Questions
+
+There is an open question on the design employed by deprecations, which is well described in the thread on [this PR](https://github.com/NuGet/NuGet.Client/pull/3540), and the main points are:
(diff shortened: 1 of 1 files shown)


designs/PackageVulnerability/PMUIVulnerability.md
```diff
@@ -0,0 +1,55 @@
...
+- there are performance concerns in the case where a user has disabled http caching because of repeated calls to NuGet.Protocol
+These are critiques of the design employed by the deprecation work (also pre-existing it) and this work extends that design.
+
+## Considerations
+
+An approach consistent with answering the above concern is under consideration but is likely to be rejected at this stage due to:
+- the work amounts to large refactoring in the codebase--cost and risk 
+- it will mean a subtractive and therefore breaking change to the public API on existing functionality, and if existing APIs are preserved and added to only, the original design is largely preserved as well
+- it may be worth returning to this question and refactoring the client codebase in favor of a deprecation approach to the outgoing APIs at some stage, but this is likely to delay the feature work beyond that which is desired
+
```
aortiz-msft, line 51: Would it be possible to get the cost estimate for both designs? I'm leaning towards moving to the right design now (and any needed refactoring) and not create additional technical debt.

zkat (approved):

designs/PackageVulnerability/PMUIVulnerability.md
```diff
@@ -0,0 +1,55 @@
...
+## Non-Goals
+
+Notifications on the Browse tab
+Restore warnings
+
+## Solution
+
+The package vulnerability data is now available in registration metadata, alongside data for deprecations which is currently in use.
+
+The plan is the extend the design used for surfacing deprecation notifications in the past, to also include vulnerability information. This includes models, view models and views. The work is likely to be very similar in all areas. The strengths of this approach are a reasonably fast time to market, and low risk.
```
zivkan, line 31: As discussed briefly in the PR, but mostly in Teams, we've (very) recently started paying a lot more attention to the [NuGet SDK](https://github.com/NuGet/NuGet.Client/blob/dev/docs/nuget-sdk.md) than did in the past. I have no concerns from the customer experience point of view.

However, the existing deprecations implementation added a `GetDeprecationMetadataAsync` on `IPackageSearchMetadata` that's defined in `NuGet.Protocol.dll`, which is part of the NuGet SDK. Customers calling these methods in their own apps **always** get `null`, even though when they check the package on nuget.org in a web browser, it shows the information exists. This is just bad API design.

Considering the only implementation of `GetVersionsAsync` and `GetDeprecationMetadataAsync` that do not return null exist in some PM UI assembly, this is where the interface that defines the methods should exist, in my opinion. My suggestion, and I have no idea how feasible it actually is, is to add any new `GetVulnerabilityMetadataAsync` method to an interface/class/whatever in PM UI, not in `NuGet.Protocol`.

Anything that's not in the NuGet SDK, and therefore doesn't require changes to any `PublicAPI.*.txt` file to compile, is fine, because we're free to refactor as we wish. Any API listed in `PublicAPI.*.txt` might be used by customers (external to Microsoft, but also teams in Microsoft, such as the .NET CLI/SDK or VS for Mac), and therefore we should do a better job of designing APIs.

designs/PackageVulnerability/PMUIVulnerability.md
drewgillies: I'm about to start work on a proof of concept for the revised design. Extending the existing (deprecations) design to include vulnerabilities is already done (https://github.com/NuGet/NuGet.Client/pull/3540). I'll get back to you when I have enough info for a meaningful estimate.

drewgillies: I've created a prototype of the synchronous approach and pushed a commit here: https://github.com/NuGet/NuGet.Client/pull/3540. This makes me optimistic that we can take this approach. Not sure if any edge cases will fall out as we bash on it though, but I'll take this spec to a design review on the strength of it.

designs/PackageVulnerability/PMUIVulnerability.md
drewgillies: I've created a prototype of the synchronous approach and pushed a commit here: https://github.com/NuGet/NuGet.Client/pull/3540. This makes me optimistic that we can take this approach. Not sure if any edge cases will fall out as we bash on it though, but I'll take this spec to a design review on the strength of it.

PR #9945, opened 2020-08-26
spec for dotnet signature-related subcommands

zkat:
cc @dtivel @heng-liu @JonDouglas @kartheekp-ms @KathleenDollard

### DIFF
--- a/designs/DotNetSignCommands.md
+++ b/designs/DotNetSignCommands.md
@@ -50,7 +50,7 @@ USAGE:
     dotnet nuget sign [OPTIONS] <package-path>
 
 OPTIONS:
-        --output-directory              Directory where the signed package should be saved. By default, the original package is overwritten by the signed package.
+    -o, --output* <output-directory>    Directory where the signed package should be saved. By default, the original package is overwritten by the signed package.
         --certificate-path              File path to the certificate to be used while signing the package.
         --certificate-store-name        Name of the X.509 certificate store. (...)
         --certificate-store-location    Name of the X.509 certificate store to use to search for the certificate.
@@ -64,6 +64,7 @@ OPTIONS:
     -h, --help                          Prints help information.
     -v, --verbosity* <LEVEL>            Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
         --interactive*                  Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 
 EXAMPLES:
 
@@ -74,10 +75,10 @@ EXAMPLES:
 
 Notable deviations from `nuget.exe` version:
 
-- No `-ConfigFile` equivalent. `dotnet.exe` doesn't seem to do this at all.
 - No `-ForceEnglishOutput` equivalent. `dotnet.exe` doesn't seem to do this at all.
 - Use MSBuild verbosity levels for `--verbosity` instead of NuGet levels.
 - Use `--interactive` instead of `--non-interactive`, because that's how `dotnet.exe` does it.
+- Use `--output` instead of `--output-directory` for consistency with `dotnet.exe`.
 
 ### `dotnet nuget verify`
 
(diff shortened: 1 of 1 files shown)


designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,140 @@
...
+    add-author <name> <author> [<package>...]
+    add-service-index <name> <index>
+    add-certificate <name> <fingerprint>
+    remove <name>
+    sync <name>
+```
+
+## Open Questions
+
+Is there any precedent at all in `dotnet.exe` for a `--config-file` option? Anywhere?
```
zivkan, line 137: `dotnet nuget add source` and `dotnet nuget add client-cert` (although it uses `--configfile`, which I think seems incorrect, but I don't know if we can change now).

Also, not `--config-file` specifc, but `dotnet sln [path\to\file.lsn] add path\to\project.csproj` is semantically the same, as is `dotnet add [path\to\project.csproj] package package_id`. Just as `dotnet add` and `dotnet sln add`  modify files and allow dotnet to either automatically detect the file, or allow you to specify the target, `dotnet nuget trusted-signer` is all about modifying a config file.

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,140 @@
...
+### `dotnet nuget sign`
+
+```
+Signs a NuGet package at <package-path> with the specified certificate.
+
+USAGE:
+    dotnet nuget sign [OPTIONS] <package-path>
+
+OPTIONS:
+        --output-directory              Directory where the signed package should be saved. By default, the original package is overwritten by the signed package.
```
JonDouglas: dotnet CLI has a tendency to do `[-o|--output <OUTPUT_DIRECTORY>]`. Do we want to follow that here?

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,140 @@
...
+### `dotnet nuget verify`
+
+```
+Verifies a signed NuGet package at <package-path>.
+
+USAGE:
+    dotnet nuget verify [OPTIONS] <package-path>...*
+
+OPTIONS:
+        --all-verifications*                    Specifies that all verifications possible should be performed on the package(s). By default, only signatures are verified.
```
JonDouglas: Should we simplify to `--All`? 

`dotnet nuget verify --all-verifications` is a few extra keystrokes if we believe "verifications" can be inferred by `--All` or `--all`.

JonDouglas (approved):
Thank you for writing this all up! I only had minor feedback regarding some consistency/keystroke saving considerations.

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,141 @@
...
+
+Options marked with `*` indicate a notable change in behavior/naming.
+
+### `dotnet nuget sign`
+
+```
+Signs a NuGet package at <package-path> with the specified certificate.
+
+USAGE:
+    dotnet nuget sign [OPTIONS] <package-path>
```
heng-liu, line 50: Shall we change the position of [OPTIONS] in the middle for dotnet command? (NuGet sign command puts the [OPTIONS] in the end.) 
If so, the example below seems need to be changed as well.

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,141 @@
...
+- Use `--interactive` instead of `--non-interactive`, because that's how `dotnet.exe` does it.
+- Use `--output` instead of `--output-directory` for consistency with `dotnet.exe`.
+
+### `dotnet nuget verify`
+
+```
+Verifies a signed NuGet package at <package-path>.
+
+USAGE:
+    dotnet nuget verify [OPTIONS] <package-path>...*
```
heng-liu, line 89: Same here.

heng-liu:
Thanks for writing the spec up @zkat !
Just a few inconsistency between the examples and the command. Others look good :)

designs/DotNetSignCommands.md
zkat: The convention in dotnet itself is to put `[options]` _before_ positional arguments (see any non-nuget command). I'm trying to move us towards consistency with POSIX and the dotnet CLI itself.

designs/DotNetSignCommands.md
zkat: See above

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,141 @@
...
+    -h, --help                          Prints help information.
+    -v, --verbosity* <LEVEL>            Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive*                  Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+
+EXAMPLES:
+
+    dotnet nuget sign MyPackage.nupkg --timestamper https://foo.bar
+
+    dotnet nuget sign ../MyPackage.nupkg --timestamper https://foo.bar --output-directory ../signed
```
heng-liu, line 73: This example has the options in the end. Seems need to be changed?
And pls also change` --output-directory` to `--output`

zkat: Yeah, they can be in either position. That's pretty normal. I'll reorder them to avoid confusion, though.

And thanks! Nice catch on `--output-directory` 👍🏻

designs/DotNetSignCommands.md
heng-liu: I see. Thanks for clarifying!

designs/DotNetSignCommands.md
heng-liu: This example has the options in the end. Seems need to be changed to the middle?
And pls change` --output-directory` to `--output`.

zkat: done!

heng-liu (approved):

designs/DotNetSignCommands.md
```diff
@@ -0,0 +1,141 @@
+# Spec Name
```
herkogabor1985, line 1: #9945

PR #9992, opened 2020-09-04
Update dotnet cli vulnerable spec following spec review

drewgillies:
A few pieces have been updated here:
- `dotnet list package` default behavior is now unchanged
- `dotnet list package --vulnerable` (the new report) now has a colorized `Severity` column
- `dotnet list package --deprecated` now has `(O)` and `(V)` markers
- `dotnet list package --outdated` now has `(D)` and `(V)` markers
- The `--offline` option is removed
- I've moved markers from `Requested` to `Resolved` columns, as its resolved versions that we assess
- Made language for `--include-transitive` advisories in footers consistent, and put them in all footers
- General cleanup and disambiguation

### DIFF
--- a/designs/PackageVulnerability/DotnetListPackageVulnerable.md
+++ b/designs/PackageVulnerability/DotnetListPackageVulnerable.md
@@ -35,7 +35,7 @@ Combining the `--vulnerable` command option with the following flags will not be
 * `--outdated`
 * `--deprecated`
 
-The following flags will are only relevant to the `--outdated` command option, or any option that potentially marks versions with an `(O)` marker. They will continue to be ignored, without erroring, when using the `--deprecated` option, which is existing behavior. They will now also be ignored without erroring when using the `--vulnerable` option:
+The following flags will are only relevant to the `--outdated` command option. They will continue to be ignored, without erroring, when using the `--deprecated` option, which is existing behavior. They will now also be ignored without erroring when using the `--vulnerable` option:
 
 * `--include-prerelease`
 * `--highest-minor`
@@ -116,7 +116,6 @@ Project 'MySampleProject' has the following package references
    > My.Deprecated.Package         1.0.0       1.0.0
    > My.Vulnerable.Package         1.0.1       1.0.1
 
- > To see transitive packages, add the `--include-transitive` option.
 (A): Auto-referenced package(s).
 ```
 
@@ -136,23 +135,18 @@ The following sources were used:
    nuget.org - https://api.nuget.org/v3/index.json
    Local - C:\NuGet\NuGetLocal
 
-3 packages need your attention - 2 outdated, 1 deprecated, 1 vulnerable.
-
-Package                Requested   Resolved    Latest
-EntityFramework        6.1.2       6.1.2       6.2.0
-NUnit                  2.4.0       2.6.4 (D)   3.8.1  
-My.Sample.Pkg          2.1.3       4.1.0 (V)   4.1.0
-
- > To see transitive packages, add the `--include-transitive` option.
-(D): Deprecated package(s). Use 'dotnet list package --deprecated' for more info.
-(V): Vulnerable package(s). Use 'dotnet list package --vulnerable' for more info.
+Project `ClassLibrary1` has the following updates to its packages
+   [netcoreapp2.0]:
+   Top-level Package      Requested   Resolved   Latest
+   > EntityFramework      6.1.2       6.1.2      6.2.0
+   > NUnit                2.4.0       2.6.4      3.8.1  
+   > My.Sample.Pkg        2.1.3       4.1.0      4.1.0
 ```
+
 Notes:
 * *Important! Installed packages that are vulnerable or deprecated but not outdated will not be part of this command's output.*
 
-* If any of the listed packages is deprecated, a `(D)` marker is printed next to the version indicating the package version is deprecated. A description is printed to the console explaining what the `(D)` marker means, but only if the `(D)` marker is used.
-
-* If any of the listed packages is vulnerable, a `(V)` marker is printed next to the version indicating the package version is known to be vulnerable. A description is printed to the console explaining what the `(V)` marker means, but only if the `(V)` marker is used.
+- The `(D)` marker will be removed from this report.
 
 *Combining the `--vulnerable` option with the `--outdated` option is (currently) not supported. When doing so, the following error message will be presented:*
 
@@ -189,22 +183,15 @@ Project `ClassLibrary1` uses the following deprecated packages
    [netcoreapp2.0]:
    Top-level Package              Requested  Resolved    Reason                 Alternative
    > My.Legacy.Package            2.0.0      2.0.0       Legacy                 My.Awesome.Package >= 3.0.0
-   > My.Buggy.Package             1.1.0      1.1.0 (O)   Critical Bugs          My.NotBuggy.Package >= 2.0.0
+   > My.Buggy.Package             1.1.0      1.1.0       Critical Bugs          My.NotBuggy.Package >= 2.0.0
    > My.Deprecated.Package        3.2.1      3.2.1       Other                  My.NotBuggy.Package >= 2.0.0
-   > My.CompletelyBroken.Package  0.9.0      0.9.0 (V)   Legacy, Critical Bugs  My.Awesome.Package >= 1.0.0
-
- > To see transitive packages, add the `--include-transitive` option.
-(O): Outdated package(s). Use 'dotnet list package --outdated' for more info.
-(V): Vulnerable package(s). Use 'dotnet list package --vulnerable' for more info.
+   > My.CompletelyBroken.Package  0.9.0      0.9.0       Legacy, Critical Bugs  My.Awesome.Package >= 1.0.0
 ```
+
 Notes:
 * *Important! Installed packages that are vulnerable or outdated but not deprecated will not be part of this command's output.*
 
 * The table output for `--deprecated` should also include a new `Requested` column to improve clarity and for consistency with `--outdated` and `--vulnerable` commands.
-
-* If any of the listed packages is outdated, a `(O)` marker is printed next to the version indicating the package version is known to be outdated. A description is printed to the console explaining what the `(O)` marker means, but only if the `(O)` marker is used.
-
-* If any of the listed packages is vulnerable, a `(V)` marker is printed next to the version indicating the package version is known to be vulnerable. A description is printed to the console explaining what the `(V)` marker means, but only if the `(V)` marker is used.
 
 *Combining the `--deprecated` option with the `--vulnerable` option is (currently) not supported. When doing so, the following error message will be presented:*
 
@@ -242,14 +229,10 @@ Project `ClassLibrary1` uses the following vulnerable packages
    Top-level Package              Requested  Resolved    Severity   Vulnerability
    > My.Vulnerable.Package        2.0.0      2.0.0       Low        https://URL-to-vuln-details
    > My.Fixed.Package             1.1.0      1.1.0       Critical   https://URL-to-vuln-details
-
- > To see transitive packages, add the `--include-transitive` option.
 ```
 
 Notes:
 * *Important! Installed packages that are deprecated or outdated but not vulnerable will not be part of this command's output.*
-
-* None of the markers `(D)`, `(O)` or `(V)` are used in this report.
 
 * The Severity values will be colored default for Low, yellow for Moderate, and red for High and Critical (note that we can't represent this in GitHub markdown above).
 
--- a/designs/PackageVulnerability/DotnetListPackageVulnerable.md
+++ b/designs/PackageVulnerability/DotnetListPackageVulnerable.md
@@ -1,7 +1,7 @@
 # Package Vulnerability Support in DotNet CLI
 
 * Status: **Reviewing**
-* Authors: [Xavier Decoster](https://github.com/xavierdecoster)
+* Authors: [Xavier Decoster](https://github.com/xavierdecoster), [Drew Gillies](https://github.com/drewgillies)
 
 ## Issue
 
@@ -11,7 +11,7 @@
 
 Developers take dependency on a set of packages directly or indirectly (through transitive dependencies) and have no way to understand if any of their dependencies bring in any known vulnerabilities.
 
-NuGet should be able to flag vulnerable packages used in projects/solutions. To support this capability from the commandline, a new `dotnet.exe` CLI option will be introduced.
+NuGet should be able to flag vulnerable packages used in projects/solutions. To support this capability from the command line, a new `dotnet.exe` CLI option will be introduced.
 
 This CLI command provides a way to verify *already installed* (direct and indirect) dependencies against known vulnerabilities.
 
@@ -34,13 +34,12 @@ Combining the `--vulnerable` command option with the following flags will not be
 
 * `--outdated`
 * `--deprecated`
-* `--offline`
 
-The following flags will be ignored (but not error) when combined with the `--vulnerable` command option:
+The following flags will are only relevant to the `--outdated` command option, or any option that potentially marks versions with an `(O)` marker. They will continue to be ignored, without erroring, when using the `--deprecated` option, which is existing behavior. They will now also be ignored without erroring when using the `--vulnerable` option:
 
-* `--include-prerelease` (currently requires `--outdated` option)
-* `--highest-minor` (currently requires `--outdated` option)
-* `--highest-patch` (currently requires `--outdated` option)
+* `--include-prerelease`
+* `--highest-minor`
+* `--highest-patch`
 
 An informational message will be printed when the ignored command options are used:
 
@@ -53,31 +52,27 @@ The command option(s) '--include-prerelease', '--highest-minor', and '--highest-
 We'll introduce a new `--vulnerable` command option to the existing `dotnet list package` command.
 This command option will offer a detailed view of already installed packages with known vulnerabilities.
 
-The default behavior for `dotnet list package` will be to add a marker next to the packages that have known vulnerabilities (as well as those that are known to be deprecated or outdated). This is a change in behavior, as the default `dotnet list package` command will now try to reach servers online to get additional metadata.
-
-To continue supporting the old behavior, a new `--offline` command option will be introduced.
+The default behavior for `dotnet list package` will be unchanged.
 
 ### DotNet CLI
 
 The `dotnet` CLI repository must be updated to:
 
 * forward the `--vulnerable` command option for the `dotnet list package` command,
-* forward the `--offline` command option for the `dotnet list package` command,
-* handle the invalid combinations of command options, such as `--vulnerable --outdated`, `--vulnerable --deprecated`, and any combination of the new `--offline` option with either `--outdated`, `--deprecated`, or `--vulnerable` options.
+* handle the invalid combinations of command options, such as `--vulnerable --outdated`, `--vulnerable --deprecated`, and `--outdated --deprecated` (this last case is existing behavior).
 
 ### NuGet xplat CLI
 
 The `nuget` xplat CLI code must be updated to:
 
 * handle the `--vulnerable` command option for the `dotnet list package` command,
-* handle the `--offline` command option for the `dotnet list package` command,
-* handle the invalid combinations of command options, such as `--vulnerable --outdated`, `--vulnerable --deprecated`, and any combination of the new `--offline` option with either `--outdated`, `--deprecated`, or `--vulnerable` options.
+* handle the invalid combinations of command options, such as `--vulnerable --outdated`, `--vulnerable --deprecated`, and `--outdated --deprecated` (this last case is existing behavior).
 
 ### Command Output
 
 #### Help
 
-The `--help` output must be updated to include the new `--vulnerable` and `--offline` options. Other existing options should refer to the `--vulnerable` and `--offline` options as needed.
+The `--help` output must be updated to include the new `--vulnerable` option. Other existing options should refer to the `--vulnerable` option as needed.
 
 ```shell
 Usage: dotnet list <PROJECT | SOLUTION> package [options]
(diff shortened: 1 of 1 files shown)


drewgillies:
Question: should we have similar "this option was ignored" warnings when `--source` and `--config` are used in an offline scenario (e.g. `dotnet list package`'s default usage)? We already have this for `--include-prerelease`, `--highest-minor` and `--highest-patch` when not using `--outdated` or anything requiring `(O)`.

drewgillies:
Question: I've removed the `--include-prerelease` advisory from the `--vulnerable` footer because it's an irrelevant option for a `--vulnerable` report (it's ignored).
Does it make sense to put this advisory in an `--outdated` report or any report that uses an `(O)` marker?
If so, why are we singling out `--include-prerelease` while not mentioning `--highest-minor` or `--highest-patch` (also pertinent only to `--outdated`/`(O)`)? 
I'd recommend leaving them all out.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -248,21 +239,19 @@ The following sources were used:
...
- > To see prerelease packages, add the `--include-prerelease` option.
 ```
 
-*Combining the `--vulnerable` option with the `--offline` options is not supported. When doing so, the following error message will be presented:*
+Notes:
+* *Important! Installed packages that are deprecated or outdated but not vulnerable will not be part of this command's output.*
 
-```shell
-> dotnet list package --vulnerable --offline
+* None of the markers `(D)`, `(O)` or `(V)` are used in this report.
```
JonDouglas: If we use markers in the other commands (deprecated / outdated), what harm is there to add (D) & (O) to the `--vulnerable` command? This is especially helpful for those who have outdated packages that may contain a security vulnerability. Authors who may know about a vulnerability may then deprecate their package so users are unlikely to use it.

i.e.

**Outdated**
```
dotnet list package --outdated
Lists which installed packages have newer versions available.

> dotnet list package --outdated

The following sources were used:
   nuget.org - https://api.nuget.org/v3/index.json
   Local - C:\NuGet\NuGetLocal

3 packages need your attention - 2 outdated, 1 deprecated, 1 vulnerable.

Package                Requested   Resolved    Latest
EntityFramework        6.1.2       6.1.2       6.2.0
NUnit                  2.4.0       2.6.4 (D)   3.8.1  
My.Sample.Pkg          2.1.3       4.1.0 (V)   4.1.0

 > To see transitive packages, add the `--include-transitive` option.
(D): Deprecated package(s). Use 'dotnet list package --deprecated' for more info.
(V): Vulnerable package(s). Use 'dotnet list package --vulnerable' for more info.
```

**Deprecated**
```
dotnet list package --deprecated
Lists which installed packages have been deprecated.

> dotnet list package --deprecated

The following sources were used:
   nuget.org - https://api.nuget.org/v3/index.json
   Local - C:\NuGet\NuGetLocal

Project `ClassLibrary1` uses the following deprecated packages
   [netcoreapp2.0]:
   Top-level Package              Requested  Resolved    Reason                 Alternative
   > My.Legacy.Package            2.0.0      2.0.0       Legacy                 My.Awesome.Package >= 3.0.0
   > My.Buggy.Package             1.1.0      1.1.0 (O)   Critical Bugs          My.NotBuggy.Package >= 2.0.0
   > My.Deprecated.Package        3.2.1      3.2.1       Other                  My.NotBuggy.Package >= 2.0.0
   > My.CompletelyBroken.Package  0.9.0      0.9.0 (V)   Legacy, Critical Bugs  My.Awesome.Package >= 1.0.0

 > To see transitive packages, add the `--include-transitive` option.
(O): Outdated package(s). Use 'dotnet list package --outdated' for more info.
(V): Vulnerable package(s). Use 'dotnet list package --vulnerable' for more info.
```

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -248,21 +239,19 @@ The following sources were used:
...
 
-Invalid command. Combining the '--vulnerable' and '--offline' options is not supported.
-```
+* The Severity values will be colored default for Low, yellow for Moderate, and red for High and Critical (note that we can't represent this in GitHub markdown above).
 
 *Combining the `--vulnerable` option with the `--deprecated` option is (currently) not supported. When doing so, the following error message will be presented:*
```
skofman1, line 239: > the following error message will be presented:

where is the error message?

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: I'm happy to put all markers everywhere (except the default `dotnet list package` experience which is offline)--this would be consistent across report types, and IMO consistency removes confusion. My implementation of the earlier spec included this approach, which received pushback here: https://github.com/NuGet/NuGet.Client/pull/3615#issuecomment-682330207. But let's discuss here and arrive at a decision on whether D/O/V markers should be:
- nowhere
- in some places as tracked by this spec
- in some places with adjustments to this spec
- everywhere (except the default `dotnet list package` experience)

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Following this line is this shell text:

```shell
> dotnet list package --vulnerable --outdated

Invalid command. Combining '--vulnerable' and '--outdated' options is not supported.
```

That's the error referred to by this comment.

drewgillies:
@JonDouglas please note my 2 questions toward the top (`--source`/`--config` and `--include-prerelease`).

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -34,13 +34,12 @@ Combining the `--vulnerable` command option with the following flags will not be
...
+The following flags will are only relevant to the `--outdated` command option, or any option that potentially marks versions with an `(O)` marker. They will continue to be ignored, without erroring, when using the `--deprecated` option, which is existing behavior. They will now also be ignored without erroring when using the `--vulnerable` option:
 
-* `--include-prerelease` (currently requires `--outdated` option)
-* `--highest-minor` (currently requires `--outdated` option)
-* `--highest-patch` (currently requires `--outdated` option)
+* `--include-prerelease`
+* `--highest-minor`
+* `--highest-patch`
 
 An informational message will be printed when the ignored command options are used:
```
loic-sharma, line 44: For the information message below, can/should we mention why the options were ignored? Consider:

```shell
	The command option(s) '--include-prerelease', '--highest-minor', and '--highest-patch' were ignored as they are only relevant to the '--outdated' command option.
```

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -165,42 +140,43 @@ The following sources were used:
...
-My.Sample.Pkg          2.1.3 (V)   4.1.0       4.1.0
+NUnit                  2.4.0       2.6.4 (D)   3.8.1  
+My.Sample.Pkg          2.1.3       4.1.0 (V)   4.1.0
 
+ > To see transitive packages, add the `--include-transitive` option.
 (D): Deprecated package(s). Use 'dotnet list package --deprecated' for more info.
 (V): Vulnerable package(s). Use 'dotnet list package --vulnerable' for more info.
 ```
+Notes:
+* *Important! Installed packages that are vulnerable or deprecated but not outdated will not be part of this command's output.*
```
loic-sharma, line 147: Can we clarify this in the `dotnet list package --outdated` output? The current output may lead folks to mistakenly believe this included all deprecated/vulnerable packages.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -113,23 +107,17 @@ Lists the package references for a project or solution.
...
    > Microsoft.ML                  0.11.0      0.11.0
    > Microsoft.NETCore.App   (A)   [2.1.0, )   2.1.0
-   > My.Outdated.Package           0.1.0       0.1.0 (O)
-   > My.Deprecated.Package         1.0.0       1.0.0 (D)
-   > My.Vulnerable.Package         1.0.1       1.0.1 (V)
+   > My.Outdated.Package           0.1.0       0.1.0
+   > My.Deprecated.Package         1.0.0       1.0.0
+   > My.Vulnerable.Package         1.0.1       1.0.1
 
+ > To see transitive packages, add the `--include-transitive` option.
```
loic-sharma: Should we include hints for the `--vulnerable`, `--outdated`, and `--deprecated` options? Something like:

```suggestion
 > To see transitive packages, add the `--include-transitive` option.
 > To see outdated packages, add the `--outdated` option.
 > To see deprecated packages, add the `--deprecated` option.
 > To see vulnerable packages, add the `--vulnerable` option.
```

Note that this may cause users to try and mix and match options like `--deprecated --vulnerable`. If this is a concern, consider adding language warning against this.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -165,42 +140,43 @@ The following sources were used:
 
 Package                Requested   Resolved    Latest
 EntityFramework        6.1.2       6.1.2       6.2.0
-NUnit                  2.4.0 (D)   2.6.4       3.8.1  
-My.Sample.Pkg          2.1.3 (V)   4.1.0       4.1.0
+NUnit                  2.4.0       2.6.4 (D)   3.8.1  
+My.Sample.Pkg          2.1.3       4.1.0 (V)   4.1.0
 
+ > To see transitive packages, add the `--include-transitive` option.
```
loic-sharma: Just curious, why does this line start with a `>`? This looks a little weird to me due to the inconsistency with the markers on the following lines. Should we remove the `>` from this line for consistency? Or, should we add a `>` to the markers' lines as well?

designs/PackageVulnerability/DotnetListPackageVulnerable.md
```diff
@@ -165,42 +140,43 @@ The following sources were used:
...
 ```
+Notes:
+* *Important! Installed packages that are vulnerable or deprecated but not outdated will not be part of this command's output.*
+
+* If any of the listed packages is deprecated, a `(D)` marker is printed next to the version indicating the package version is deprecated. A description is printed to the console explaining what the `(D)` marker means, but only if the `(D)` marker is used.
 
-*Combining the `--outdated` option with the `--offline` options is not supported. When doing so, the following error message will be presented:*
+* If any of the listed packages is vulnerable, a `(V)` marker is printed next to the version indicating the package version is known to be vulnerable. A description is printed to the console explaining what the `(V)` marker means, but only if the `(V)` marker is used.
+
+*Combining the `--vulnerable` option with the `--outdated` option is (currently) not supported. When doing so, the following error message will be presented:*
```
loic-sharma, line 151: Just curious, why don't we support combining these options? Engineering cost?

drewgillies: The main reason would be that the format of the reports changes from type to type --the columns we see in `--vulnerable` are different from `--outdated` etc. If we wanted to combine them, we could--it wouldn't be a huge engineering cost--we'd just have to decide how to format the output. It might be a good question to ask if we decide to add different output formats, for example.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: If we were to do this, we would need to make equivalent changes to `--deprecated` and `--vulnerable` reports. I think this is a good addition, but I'm also wondering just how "chatty" we want to make our output. And perhaps the answer is "quite chatty". :) If these are helpful tips, perhaps they should be included.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Another thought could be to collapse all of this into:
- For details on including transitive dependencies or to view outdated/deprecated/vulnerable reports, see `dotnet list package --help`

nkolev92: fwiw, I like @drewgillies's latest suggestion.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
nkolev92: `(O)` as an outdated marker is super weird tbh. 
I think using `O` for anything other than `output` is misleading, legends be damned. 

That being said, I don't have a different suggestion.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
loic-sharma: Definitely agreed. Maybe we should participate in an UX review on the messaging? @chgill-MSFT knows a "word smith" (I don't know his official title, but he specializes on crafting clear yet concise messaging).

My suggestions:
1. Consider changing `Use 'dotnet list package --deprecated' for more information.` to  `Use 'dotnet list package --deprecated' to list all deprecated packages.`.
2. Reconsider this phrasing: `3 packages need your attention - 2 outdated, 1 deprecated, 1 vulnerable`. The messaging indicates the counts are exact, which is not true for deprecated and vulnerable in this example. Maybe use the phrasing `3 outdated packages need your attention` instead.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
loic-sharma: Yup I prefer @drewgillies's suggestion too!

designs/PackageVulnerability/DotnetListPackageVulnerable.md
loic-sharma: I see. Could we support combining the options by displaying just the common columns with deprecated/vulnerable/outdated markers?

designs/PackageVulnerability/DotnetListPackageVulnerable.md
loic-sharma: > everywhere (except the default dotnet list package experience)

This would be my preference, but I'm not sure I fully understand the trade-offs.

> `(O)` as an outdated marker is super weird tbh.
> I think using `O` for anything other than `output` is misleading, legends be damned.

Do we know CLI UX experts? There may be a common pattern we're unaware of that we can leverage. One alternative may be UT8 symbols like:

* Vulnerable: ⚠
* Deprecated: ⏰
* Outdated: ⬆

nkolev92: > Do we know CLI UX experts?

Not if anyone has that role but I think Kathleen might be as close as we can get to that :)

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: I guess. It would mean that we would just see package + requested & resolved version though (or package + resolved version for the transitive portion).

designs/PackageVulnerability/DotnetListPackageVulnerable.md
JonDouglas: @loic-sharma Not necessarily a common pattern, but I think it's one that may want to be considered in the future.

i.e. https://www.npmjs.com/package/npm-check

Just looking at the experiences we're trying to match here, markers really don't make sense in this context. The goal is that if a user typed in any command `--vulnerable`, `--deprecated`, or `--outdated`, that they would ultimately see the relevant information plus a footer of "How to fix it / start to fix it" via an `audit` and a fast way of `audit fix` to fix it for them.

Right now, we're trying to just keep the user typing these commands until they understand what is all vulnerable, deprecated, and outdated. The goal really is to get them to run our `audit` command, and ultimately fix everything in one command with an `audit fix` command.

I'm in favor of just removing markers given the contention on them entirely & removing the footnotes as well. Thoughts?

drewgillies: That all makes sense to me--let's go with that plan. If we want markers and footers later, we can add them after agreeing on what we want. For now, let's remove the blockers. Thanks. 👍

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Per @JonDouglas's comment above, we'll remove footers for now and let customers use `--help` for the full help experience.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Per @JonDouglas's comment above, we'll remove footers for now and let customers use `--help` for the full help experience.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Per @JonDouglas's comment above, we'll remove footers for now and let customers use `--help` for the full help experience.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: Per @JonDouglas's comment above, we'll remove footers for now and let customers use `--help` for the full help experience.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: In the spirit of @JonDouglas's comment above (even though this scenario isn't explicitly called out), we'll go with the lowest common denominator (the current mutually exclusive experience) for now and look at adding combinations in later versions.

drewgillies:
So, reiterating, and to close this loop, per @JonDouglas's comment above, we'll remove footers for now and let customers use `--help` for the full help experience. If we want to add/re-add markers and footers later, these can be separate v2+ discussions. For now, let's unblock v1 with the features we have consensus on: a new `vulnerable` report, no markers (except `(A)`), no footers.

designs/PackageVulnerability/DotnetListPackageVulnerable.md
JonDouglas: Here's a thought of what that looks like:

## Vulnerable

dotnet list package --vulnerable
Lists which installed packages have known vulnerabilities.

```
> dotnet list package --vulnerable

The following sources were used:
   nuget.org - https://api.nuget.org/v3/index.json
   Local - C:\NuGet\NuGetLocal

Project `ClassLibrary1` uses the following vulnerable packages
   [netcoreapp2.0]:
   Top-level Package              Requested  Resolved    Severity   Vulnerability
   > My.Vulnerable.Package        2.0.0      2.0.0       Low        https://URL-to-vuln-details
   > My.Fixed.Package             1.1.0      1.1.0       Critical   https://URL-to-vuln-details

 > Run `dotnet nuget audit fix` to fix them, or `dotnet nuget audit` for more details.
```

## Deprecated

dotnet list package --deprecated
Lists which installed packages have been deprecated.

```
> dotnet list package --deprecated

The following sources were used:
   nuget.org - https://api.nuget.org/v3/index.json
   Local - C:\NuGet\NuGetLocal

Project `ClassLibrary1` uses the following deprecated packages
   [netcoreapp2.0]:
   Top-level Package              Requested  Resolved    Reason                 Alternative
   > My.Legacy.Package            2.0.0      2.0.0       Legacy                 My.Awesome.Package >= 3.0.0
   > My.Buggy.Package             1.1.0      1.1.0       Critical Bugs          My.NotBuggy.Package >= 2.0.0
   > My.Deprecated.Package        3.2.1      3.2.1       Other                  My.NotBuggy.Package >= 2.0.0
   > My.CompletelyBroken.Package  0.9.0      0.9.0       Legacy, Critical Bugs  My.Awesome.Package >= 1.0.0

 > Run `dotnet nuget audit fix` to fix them, or `dotnet nuget audit` for more details.
```

## Outdated

dotnet list package --outdated
Lists which installed packages have newer versions available.

```
> dotnet list package --outdated

The following sources were used:
   nuget.org - https://api.nuget.org/v3/index.json
   Local - C:\NuGet\NuGetLocal

Project `ClassLibrary1` uses the following outdated packages
   [netcoreapp2.0]:
   Top-level Package        Requested   Resolved    Latest
   > EntityFramework        6.1.2       6.1.2       6.2.0
   > NUnit                  2.4.0       2.6.4       3.8.1  
   > My.Sample.Pkg          2.1.3       4.1.0       4.1.0

 > Run `dotnet nuget audit fix` to fix them, or `dotnet nuget audit` for more details.
```

drewgillies:
@JonDouglas I've updated the spec to reflect the above decisions. Once approved, I'll merge and can proceed with the engineering changes.
/cc @skofman1 @aortiz-msft @anangaur

designs/PackageVulnerability/DotnetListPackageVulnerable.md
drewgillies: @JonDouglas --thanks. Note though that the transitive footers are still in these samples, so they can all be considered gone as well, right?

JonDouglas: @drewgillies Ah sorry, yes it's late. Let me remove those.

JonDouglas: I don't think there's any warrant in having them without markers for the moment. Anyone else have thoughts?

drewgillies: >I don't think there's any warrant in having them without markers

Could you please clarify this comment--we want the reports without markers (except `(A)`--we can leave those in), right?

JonDouglas: To clarify, we don't need the `--include-transitive` footer as it'll be in the `--help` section. It doesn't really add any value being in each footer.

zivkan (approved):
I'm disappointed that the `--highest-*` options will be silently ignored. If it was intentional that deprecated didn't error out, I think that was a mistake then, and we're repeating it now. If it was an oversight for deprecated, I think we can do better now.

Otherwise I'm good with this as a minimal increment.

drewgillies:
@zivkan 
> I'm disappointed that the --highest-* options will be silently ignored.

There is something to clarify here, actually. There is an inconsistency between how these options are *currently* treated by the dotnet CLI and the NuGet Xplat CLI, and we should iron that out:

### For the Xplat CLI, as discussed, we currently have silent ignoring. To verify what we already know (for comparison)...
#### This method is the only place we read the options:
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs#L680
#### This is the only logic calling this filtering (which in the existing code means standard `dotnet list` and `--outdated` reports):
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/PackageReferenceCommands/ListPackage/ListPackageCommandRunner.cs#L459

...so:
- no message to console
- silent ignoring of the options in a `--deprecated` scenario, 
- our current spec currently extends this behavior to `--vulnerable`, adding an ignored message. 
- (FWIW I'm changing the logic to also exclude the standard `dotnet list` scenario because it's meaningless there--this only makes sense for `--outdated`. Essentially these are "outdated how?" options).

### For the dotnet CLI, we are *currently throwing* if these options are present. 
#### For example, the prerelease option is tested here:
https://github.com/dotnet/sdk/blob/master/src/Cli/dotnet/commands/dotnet-list/dotnet-list-package/ListPackageReferencesCommand.cs#L54
#### And the throwing happens here:
https://github.com/dotnet/sdk/blob/master/src/Cli/dotnet/commands/dotnet-list/dotnet-list-package/ListPackageReferencesCommand.cs#L96

Note that not only is it throwing, it's allowing these options through for `--deprecated` reports, which is meaningless. In the NuGet Xplat CLI, as you can see above, they're ignored for deprecated reports.

### So, here's what I propose, as since we should fix this inconsistency, I will be making changes anyway:
- Let's use @zivkan's approach here and throw for all invalid scenarios.
- For this scenario it means throwing for dotnet CLI and bailing in NuGet Xplat CLI when these 3 options (`--include-prerelease `, `--highest-patch`, `--highest-minor`) are used in any report that doesn't use test for outdated (per the current spec this would be `--outdated` reports only, but if we ever were to reintroduce markers, it would include scenarios where we use `(O)`).
- No ignoring, no ignore messages.

Thoughts?

/cc @JonDouglas @skofman1 @aortiz-msft

JonDouglas:
I think we are having so much resistance to this feature because we're expecting all experiences in `dotnet list package` to be compliant with others.

Ignoring these three flags is perfectly fine to me as it's typical in CLI to just ignore options that won't do anything. It's also perfectly fine if we error them out saying it's an invalid flag/scenario. Outdated is the only area I can think of that makes sense to have these flags as it needs to check against the upstream to see if a package is outdated or not.

Issue #5887, opened 2017-09-14
PackRef does not show transitive dependencies - install tab & update tab behavior?


akamyshanov:
When using "legacy"-style `.csproj` project files we have a separate `packages.config` file where all dependencies are listed, including transitive ones. This allows a use case when one installs a package with dependencies and then decides which transitive dependencies can be manually updated. So, the benefits are:

- Dependencies are easily identifiable due to presence of a flat list
- Fine-grain control over all dependency versions

There are drawdowns of course: the flat list can blow up to a huge one.

E.g., after installing Autofac.WebApi2.Owin from NuGet, we have a picture like this:

<img width="873" alt="whajk" src="https://user-images.githubusercontent.com/1358330/30423164-b3059094-994a-11e7-8d0c-86d723aa4913.png">

Transitive dependencies which are clearly viewable can be manually updated very easily.

When using the new Sdk-style .csproj projects NuGet references are added as `<PackageReference/>` to the project file itself and transitive dependencies are resolved by MSBuild automatically:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net462</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Autofac.WebApi2.Owin" Version="4.0.0" />
  </ItemGroup>
</Project>
```

So, to update transitive dependencies, one would have to

1. Identify them (e.g. via obj/project.assets.json)
2. Add all of them explicitly to the project
3. Perform updates

And this has to be done after each update and for every (!) transitive dependency in the project which is nearly impossible.

This seems like a "regression" to me.

The best possible resolution would be to show transitive dependency updates in NuGet GUI or via CLI. If the user decides to update one of the dependencies, it would get explicitly added to the project. (similar to https://github.com/NuGet/Home/issues/3159)

Maybe there is some workaround to achieve the specified behaviour?

davidfowl:
I'm curious, why do you need this? Once you update the transitive dependency do you expect it to be hoisted into your project? Packages can remove dependencies in future versions so why does it matter that the dependency in transitive in the first place? 

Why not just add the ones you care about explicitly updating manually to your project?

akamyshanov:
@davidfowl 
Thank you for the answer.

> why do you need this?

* Bug/security fixes in the dependencies
* performance improvements
* other reasons

> Once you update the transitive dependency do you expect it to be hoisted into your project?

[...]

Issue #5553, opened 2017-07-08
Allow users to determine package resolution strategy during package restore - direct or transitive


Wil73:
I am not going to debate if highest or lowest should be the default: leave the actual behavior up to the consumer.

Why on earth should Microsoft FORCE users to adopt the "safest is best" strategy in the first place?

Is there any reason why we cannot add a switch to nuget.exe and NPM that allows the user to decide their own package restore strategy?

This way consumers of nuget can align nuget behavior to their own business strategies, be they conservative or aggressive.

What I don't understand is why Microsoft has to decide this for consumers in the first place.

To me this is a straight up consumer decision based on private business strategy and policy.

It makes absolutely no sense why a tool vendor should allow themselves to unilaterally decide business strategy for all nuget consumers on the planet without an option to override that.

It's been two years and there's no apparent movement on this issue.

Leave lowest version as the default strategy if you want Microsoft, but for goodness sakes offer a switch to allow consumers to make their own business decisions.

davidfowl:
> I am not going to debate if highest or lowest should be the default: leave the actual behavior up to the consumer.

Lowest should be the default, but that's just my opinion. It's the best way to get a "working set of packages". Everyone in the graph gets what they compiled against and things only roll forward minimally if sibling packages need higher versions.

> Why on earth should Microsoft FORCE users to adopt the "safest is best" strategy in the first place?

It's more of a "safest is default strategy" because people want to get a working set of packages by default. In a system where you *need* binary compatibility between components to in order to make things function things have a better chance of working if you stay closer to what components build, and test against.

> Is there any reason why we cannot add a switch to nuget.exe and NPM that allows the user to decide their own package restore strategy?

This is already possible when using packages.config based projects by picking the dependency behavior on install:

![image](https://user-images.githubusercontent.com/95136/27986627-2694d382-63b6-11e7-8174-3d4c21fd8a21.png)

We don't own NPM, that's the node package manager, not sure if you mean something else.

To properly do this with PackageReference, a lock file needs to be introduced (which has been discussed anyways).

Wil73:
I am not going to debate what the default ought to be.  I really am not concerned over that.  Technically this is NOT a default when you can't configure it otherwise. It's only a default if you can change it.  We can't.

My company is frustrated that Microsoft will not allow businesses to DECIDE Nuget resolution behavior for themselves.

It makes absolutely no sense why Microsoft should have a monopoly on the business policies of the private sector: this is in effect what Microsoft does here.

[...]

PR #9993, opened 2020-09-04
Add flag vulnerable packages spec

drewgillies:
Let's add this spec here in its current form so we can track changes and discussions, e.g. regarding the MVP.

### DIFF
--- a/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
+++ b/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
@@ -1,4 +1,4 @@
-# Flag vulnerable packages
+# Flag vulnerable packages - Stage 1
 
 * Status: **Reviewing**
 * Author(s): [Anand Gaurav](https://github.com/anangaur) ([@adgrv](https://twitter.com/adgrv)), [Xavier Decoster](https://github.com/xavierdecoster) ([@xavierdecoster](https://twitter.com/xavierdecoster))
@@ -48,6 +48,7 @@ NuGet will not support self-reporting of vulnerabilities at this time and instea
 
 * For Visual Studio:
   * **Display vulnerability information** in the package graph on `Installed` tab of Package Manager Dialog (PMUI)
+  * **Display vulnerability information** for *transitive packages* in PMUI (additional UX complexity) - note that this is dependent on transitive packages being displayed in general, which is to be specced separately
 
 * For nuget.org:
   * **Automatically ingest and refresh NuGet package vulnerability data from GitHub**
@@ -62,55 +63,7 @@ Some enhancements to this feature are out of scope for the initial implementatio
 
 * Enable `restore` to Inform or Warn (see Stage 2)
 * Expose vulnerability metadata in nuget.org *Search* service (not required for MVP)
-* Display vulnerability information for *transitive packages* in PMUI (additional UX complexity)
 * Expose vulnerability information *publicly* on nuget.org's package details page and version history (disclosure policies TBD)
-
-### Stage 2
-
-Stage 2 of this feature will focus on *highlighting known vulnerabilities when the package is already installed*. This involves introducing new validations at package `restore` time, and new ways of dealing with them.
-
-#### ✔️ In scope (Stage 2)
-
-* For CLI:
-  * For `dotnet restore`, vulnerability reporting is opt-out. By default vulnerabilities are reported during a restore operation.
-  * The same vulnerability output for `dotnet restore` will appear when running `dotnet add package` (which actually does a *restore*), unless `-n|--no-restore` is specified.
-  * `dotnet nuget audit` to output a full list of packages indicating outdated, deprecated and/or vulnerable status.
-  * `dotnet nuget audit fix` to apply fixes based on rules yet TBD.
-
-* For Visual Studio:
-  * Inform consumers about known vulnerabilities at time of installation or restore:
-    * Warn (⚠️) for `critical` and `high` impact vulnerabilities (minimal info + command/way to uncover detailed information)
-    * Info (ℹ️) for lower impact vulnerabilities (minimal info + command/way to uncover detailed information)
-  * Display vulnerability information for *transitive packages* in PM UI - note that this is dependent on transitive packages being displayed in general, which is to be specced separately and is not part of the vulnerability feature work.
-  * New settings to enable/disable warnings for the various impact levels of vulnerabilities
-
-* For nuget.org:
-  * Show vulnerability information in packages to their publishers on nuget.org
-    * Show the vulnerability info on the package page and version history
-  * Users should be able to see any reported vulnerabilities at NuGet.org/advisories. This data can be directly pulled from https://github.com/advisories?query=ecosystem%3Anuget or through our API integration pieces. The user should be able to see & filter based on the severity of the advisory (low, moderate, high, critical) and sort based on the newest, oldest, recently updated, and least recently updated.
-    * Upon selecting an advisory the user should see an advisory detail screen that provides the title, published date, CVE identifier, severity type, affected package, affected versions, patched versions (if known), a description of the vulnerability, and references to the known vulnerability.
-    * From a glance, a user should be able to see:
-      1.	The Advisory Title
-      2.	The Affected Package Name
-      3.	The Severity of the Vulnerability
-      4.	The Date of the Advisory
-      5.	The Status of the Advisory.
-    * Additionally, the server should show a banner on a package with regards to the vulnerability and the 5 details above with a link to the advisory URL either on nuget.org/advisories or directly to github.com/advisories.
-
-
-#### ❓ Open questions (Stage 2)
-
-* Disclosure policies if and when we start consuming non-disclosed vulnerability data? (e.g. under review by package author) - GH API currently doesn't surface these
-
-| Feature | State |
-|:--- |:--- |
-| Show vulnerability information as part of search results | Discuss |
-| Show vulnerability information publicly for each package like deprecation info on the package details | Discuss |
-
-Legend:
-
-* *Backlog: Follow-up feature to be implemented after the current one is implemented*
-* *Discuss: Feature that needs further discussion*
 
 ## Solution (Stage 1)
 
--- a/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
+++ b/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
@@ -54,14 +54,13 @@ NuGet will not support self-reporting of vulnerabilities at this time and instea
   * **Automatically ingest and refresh NuGet package vulnerability data from GitHub**
   * **Surface package vulnerability information on-demand**:
     * as part of package metadata in `v3` registration blobs (`JSON`)
-    * by showing vulnerability information to package owners (only!) on package details page and version history
-  * Basic (e-mail) notification to notify package publishers for any new vulnerability detected in one or more of their packages
+    * by showing vulnerability information on package details page and version history
 
 #### ❌ Out of scope (Stage 1)
 
 Some enhancements to this feature are out of scope for the initial implementation of this feature because there are several unknowns still to be clarified. See also the *Open Questions* section below.
 
-* Enable `restore` to Inform or Warn (see Stage 2)
+* Enable `restore` to Inform or Warn
 * Expose vulnerability metadata in nuget.org *Search* service (not required for MVP)
 
 ## Solution (Stage 1)
--- a/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
+++ b/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
@@ -63,7 +63,6 @@ Some enhancements to this feature are out of scope for the initial implementatio
 
 * Enable `restore` to Inform or Warn (see Stage 2)
 * Expose vulnerability metadata in nuget.org *Search* service (not required for MVP)
-* Expose vulnerability information *publicly* on nuget.org's package details page and version history (disclosure policies TBD)
 
 ## Solution (Stage 1)
 
@@ -85,7 +84,7 @@ Clients should just not display the severity if they don't understand the value 
 
 ### List Vulnerabilities using a CLI command
 
-[See client spec for `dotnet list package --vulnerable`.](https://github.com/NuGet/Home/wiki/dotnet-list-package---vulnerable)
+[See client spec for `dotnet list package --vulnerable`.](https://github.com/NuGet/Home/blob/dev/proposed/2020/PackageVulnerability/DotnetListPackageVulnerable.md)
 
 ### Display Vulnerabilities in Visual Studio PMUI
 
--- a/proposed/archive/PackageVulnerability/FlagVulnerablePackages.md
+++ b/proposed/archive/PackageVulnerability/FlagVulnerablePackages.md
@@ -73,7 +73,10 @@ Stage 2 of this feature will focus on *highlighting known vulnerabilities when t
 #### ✔️ In scope (Stage 2)
 
 * For CLI:
-  * Output warnings or errors when running `dotnet add package` (which actually does a *restore*)
+  * For `dotnet restore`, vulnerability reporting is opt-out. By default vulnerabilities are reported during a restore operation.
+  * The same vulnerability output for `dotnet restore` will appear when running `dotnet add package` (which actually does a *restore*), unless `-n|--no-restore` is specified.
+  * `dotnet nuget audit` to output a full list of packages indicating outdated, deprecated and/or vulnerable status.
+  * `dotnet nuget audit fix` to apply fixes based on rules yet TBD.
 
 * For Visual Studio:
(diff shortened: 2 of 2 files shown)


designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+  * Display vulnerability information for *transitive packages* in PMUI
+  * New settings to enable/disable warnings for the various impact levels of vulnerabilities
+
+* For nuget.org:
+  * Show vulnerability information in packages to their publishers on nuget.org
+    * Show the vulnerability info on the package page and version history
+
+#### ❓ Open questions (Stage 2)
+
+* For `restore`, should it be opt-in or opt-out (preferred for security features)?
```
JonDouglas: It should be default / opt-out. Security should be ran on every-single restore just like npm install.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+* For `dotnet add package`, define behavior (info, warn, error?) when encountering packages known to be vulnerable.
```
JonDouglas: None. We don't do this for other audits & generally we are just advising people on restore to not use these packages & providing them a resolution. Everything should be in dotnet restore instead.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+* Need for another command (e.g. `dotnet nuget audit`) to create an audit report for deprecated, vulnerable and may be outdated packages. And a way to fix the vulnerabilities with a single command/click.
```
JonDouglas: Proposed `dotnet nuget audit` & `dotnet nuget audit fix` for resolution.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+* Disclosure policies if and when we start consuming non-disclosed vulnerability data? (e.g. under review by package author) - GH API currently doesn't surface these
+
+| Feature | State |
+|:--- |:--- |
+| A way to specify another source for vulnerability information | Backlog |
+| `why` command to understand why a package is shown as a dependency for my project | Backlog |
```
JonDouglas: This is a separate feature entirely.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+#### ❓ Open questions (Stage 2)
+
+* For `restore`, should it be opt-in or opt-out (preferred for security features)?
+* For `dotnet add package`, define behavior (info, warn, error?) when encountering packages known to be vulnerable.
+* Need for another command (e.g. `dotnet nuget audit`) to create an audit report for deprecated, vulnerable and may be outdated packages. And a way to fix the vulnerabilities with a single command/click.
+* Disclosure policies if and when we start consuming non-disclosed vulnerability data? (e.g. under review by package author) - GH API currently doesn't surface these
+
+| Feature | State |
+|:--- |:--- |
+| A way to specify another source for vulnerability information | Backlog |
```
JonDouglas: I don't believe we have any other sources in mind today & likely a couple years from another standard. I don't think we need this on a backlog unless we plan to commit to it relatively soon.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+| `why` command to understand why a package is shown as a dependency for my project | Backlog |
+| Ability to report a vulnerability on nuget.org | Discuss |
```
JonDouglas: NuGet package page requires a "Report a vulnerability" feature that is triaged by someone on the NuGet team.

Otherwise we need to provide instructions for https://docs.github.com/en/github/managing-security-vulnerabilities/creating-a-security-advisory

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+| An API that lists all vulnerable nuget.org packages with the vulnerability information | Discuss |
```
JonDouglas: This is included in my design doc, maybe we can weave it in here.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,142 @@
...
+| Show vulnerability information as part of search results | Discuss |
```
JonDouglas: If anything, it should show as an icon similar to the deprecated package in the package details.

designs/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: I'll remove this.

designs/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: I'll remove this too :)

designs/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: I think the second option is more robust as it gives a single point of truth. In a scenario where we act as a moderator, there's more opportunity for lag before a new vulnerability is surfaced in our clients.

designs/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: I'll adjust the spec to reflect this, but I want to be sure that we're on the same page. `dotnet add package` includes a `dotnet restore` unless the `--no-restore` option is included (https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-package). So, we will have `dotnet restore` (above) defaulting to indicating vulnerabilities, and `dotnet add package` will have the same behavior as `dotnet restore` except when `--no-restore` is used. Right?

designs/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: I've replaced this open question with the relevant excerpt from your doc.

drewgillies:
@JonDouglas There's one piece of the MVP spec that we need to address: **Show transitive packages for `PackageReference` type projects** in PMUI (https://github.com/NuGet/Home/pull/9993/files#diff-54d6789f8b93ec3d4be67ca1e1a44334R51). 

My concern is that we are wanting to include general PM UI transitive dependency strategy in a vulnerability spec. Transitive dependency functionality for `PackageReference` projects doesn't exist yet in _any_ form in the PM UI. 

To go back a step, we have unresolved discussions around how we want to see these dependencies surfaced in PM UI, and whether there should be more user control on dependency resolution (along with whether there should be promotion-to-top-level and a related prune command):
* https://github.com/NuGet/Home/issues/5887
* https://github.com/NuGet/Home/issues/5553

I propose that we remove all PM UI transitive dependency vulnerability work from the vulnerability spec (or at least punt to stage 2) and let PM UI transitive dependencies work be spec'd and added to the product. Our decisions on how to implement vulnerability reporting here are downstream from the above questions. So, IMO it would be good to:

- Spec the PM UI transitive dependencies work in general. If we can get that spec completed and signed off (i.e. added _here_ to home/designs), then we have a starting point. If we do indeed have a firm spec, let's get it on the table in a separate PR here. If we don't, this should be our first step.
- Get this change in-product. We can explore the possibility of doing the vulnerability inclusions at the same time if this seems expedient (and we have the below spec)
- Ensure that we do indeed have a firm spec signed off (i.e. added _here_ to home/designs) that describe how we wish to surface vulnerabilities in the context of an understood and signed off (and possibly under construction) PM UI transitive dependencies spec.

With this in mind it seems to me that we shouldn't have anything related to PM UI transitive dependencies work in our MVP. I feel we run the risk of diluting the vulnerability work here.

/cc @aortiz-msft @skofman1

JonDouglas:
@drewgillies I agree with you and also brought this up in various email threads. I believe this spec is simply too large for a MVP considering the VS PM UI needing to support an entirely new UI concept that hasn't been specced by itself outside of various feature specs that reference it. That to me is scope creep at it's finest.

I do not believe the right approach is to simply lump this feature into existing feature(s), as it requires involvement by PM, UX, and Engineering to bring into Visual Studio, and then for existing features to leverage it to their needs at a later stage.

Given vulnerability is not the only spec that references this work, we should consider this a piece of work that these specs "depend on" & not necessarily part of the spec itself.

anangaur:
Irrespective of how we do it, IMO the vulnerability feature Phase 1 will be incomplete unless we are able to show all vulnerabilities in the full package graph for a project/solution in Visual Studio.

JonDouglas:
@anangaur Referring back to the original spec:

✔️ In scope (Stage 1)
> Show transitive packages for PackageReference type projects in PMUI

❌ Out of scope (Stage 1)
> Display vulnerability information for transitive packages in PMUI (additional UX complexity)

There does not exist a spec/design of the "in scope" work for showing transitive packages in PMUI. This is troublesome for many reasons because it affects the core Installed & Update experiences in PMUI. The [screenshots on the original spec](https://github.com/NuGet/Home/wiki/Flag-vulnerable-packages#display-vulnerabilities-in-visual-studio-pmui) I do not believe are sufficient enough for the larger body of work of showing transitive packages in PMUI.

For example:

- Can a user interact with transitive packages like they do top-level packages?
- Can a user see how it was implicitly referenced?
- What about selecting a transitive package? Do we show package details or not?
- Can a user install the implicit package explicitly to "promote" to a top-level package?
- etc

We should consider https://github.com/NuGet/Home/issues/5887 & https://developercommunity.visualstudio.com/content/idea/538119/add-dependency-tree-view-for-packagereferences.html as requirements to fully complete this feature work. This work I believe needs it's own design & spec itself given the body of work.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,147 @@
...
+* Expose vulnerability information *publicly* on nuget.org's package details page and version history (disclosure policies TBD)
+
+### Stage 2
+
+Stage 2 of this feature will focus on *highlighting known vulnerabilities when the package is already installed*. This involves introducing new validations at package `restore` time, and new ways of dealing with them.
+
+#### ✔️ In scope (Stage 2)
+
+* For CLI:
+  * For `dotnet restore`, vulnerability reporting is opt-out. By default vulnerabilities are reported during a restore operation.
```
nkolev92: I think this was brought up when the original design and I don't want to block/try to answer it here, but given the registration approach of reporting, reporting vulnerabilities at restore will incur a performance cost. 

We'll need a detailed design when we do get to phase 2. Personally I think we sohuld call it out the same way we're calling out the transitive dependency UX challenges.

designs/PackageVulnerability/FlagVulnerablePackages.md
nkolev92: automatic reporting during restore will have a cost. 
Whether we're willing to pay for that cost is important to discuss.

designs/PackageVulnerability/FlagVulnerablePackages.md
JonDouglas: Agreed. I'll make a note here given it is our eventual goal at what cost though? :)

JonDouglas:
@drewgillies I will do my best to see if there's an existing spec anywhere. I've looked and I don't think there is one. I may have to write one & include it as another proposal. I don't believe we should treat this as blocking however, we can always just show Top-Level dependencies vulnerabilities in the UI to start & add vulnerability information for transitive dependencies in Stage 2.

> Spec the PM UI transitive dependencies work in general. If we can get that spec completed and signed off (i.e. added here to home/designs), then we have a starting point. If we do indeed have a firm spec, let's get it on the table in a separate PR here. If we don't, this should be our first step.

designs/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,147 @@
...
+#### ❌ Out of scope (Stage 1)
+
+Some enhancements to this feature are out of scope for the initial implementation of this feature because there are several unknowns still to be clarified. See also the *Open Questions* section below.
+
+* Enable `restore` to Inform or Warn (see Stage 2)
+* Expose vulnerability metadata in nuget.org *Search* service (not required for MVP)
+* Display vulnerability information for *transitive packages* in PMUI (additional UX complexity)
+* Expose vulnerability information *publicly* on nuget.org's package details page and version history (disclosure policies TBD)
+
+### Stage 2
```
aortiz-msft: It would be good if we could move Stage 2 to another design spec that we can review when we are ready to take on that work. As part of this spec, we should only mention potential enhancements in Stage 2 that have an impact on the initial design.

JonDouglas (approved):
Approving for Stage 1 work. Stage 2 needs a separate spec.

anangaur:
We should also update this [announcement ](https://github.com/NuGet/Announcements/issues/39)or put another new one?

drewgillies:
@JonDouglas if stage 2 needs a separate spec, we should probably pull it out of this one. Agreed?

aortiz-msft:
> @JonDouglas if stage 2 needs a separate spec, we should probably pull it out of this one. Agreed?

Yes please :)

drewgillies:
@JonDouglas @aortiz-msft Done. I've moved the vulnerable package transitive dependencies work into the PM UI for stage 1 since that seems to be becoming the thinking.

aortiz-msft:
@drewgillies - anything needed here?

aortiz-msft:
@drewgillies - Is there work left for this issue?

JonDouglas:
Hello!

Actions to take:

drewgillies:
@aortiz-msft -- apologies for missed messages on this. I'll review personally shortly to see if any services/nuget.org/dotnet CLI pieces need adjusting, and then I'll submit this for an additional review cycle per @JonDouglas's recommendation.

drewgillies:
I've cleaned this up a little (removed non-goal of publishing vulnerabilities publicly on nuget.org, corrected a spec link) and moved this spec to the proposed directory. Re-requesting for final review per above. 
Easy link for viewing updated spec: https://github.com/NuGet/Home/blob/dev-drewgil-add-flagvulnerablepackages-spec/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
FYI @JonDouglas @aortiz-msft

proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,99 @@
...
+  * **Surface package vulnerability information on-demand**:
+    * as part of package metadata in `v3` registration blobs (`JSON`)
+    * by showing vulnerability information to package owners (only!) on package details page and version history
+  * Basic (e-mail) notification to notify package publishers for any new vulnerability detected in one or more of their packages
+
+#### ❌ Out of scope (Stage 1)
+
+Some enhancements to this feature are out of scope for the initial implementation of this feature because there are several unknowns still to be clarified. See also the *Open Questions* section below.
+
+* Enable `restore` to Inform or Warn (see Stage 2)
```
nkolev92: nit: Stage 2 is not defined, but referred to :)

I imagine you just want to say it'll be considered at a future time.

nkolev92 (approved):

proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
```diff
@@ -0,0 +1,99 @@
...
+
+* For Visual Studio:
+  * **Display vulnerability information** in the package graph on `Installed` tab of Package Manager Dialog (PMUI)
+  * **Display vulnerability information** for *transitive packages* in PMUI (additional UX complexity) - note that this is dependent on transitive packages being displayed in general, which is to be specced separately
+
+* For nuget.org:
+  * **Automatically ingest and refresh NuGet package vulnerability data from GitHub**
+  * **Surface package vulnerability information on-demand**:
+    * as part of package metadata in `v3` registration blobs (`JSON`)
+    * by showing vulnerability information to package owners (only!) on package details page and version history
```
anangaur: Are we still planning to show this info to authors only? /cc: @JonDouglas

proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: Thanks, @nkolev92 --this is removed now.

proposed/2020/PackageVulnerability/FlagVulnerablePackages.md
drewgillies: Good catch, @anangaur --I've removed this now, as well as the email bullet.

drewgillies:
If there are no further concerns about this, I'll go ahead and merge it. @aortiz-msft @nkolev92 @anangaur @JonDouglas

drewgillies:
Thanks everyone for your contributions! Merging.

PR #9997, opened 2020-09-04
CPVM Project Package Version Override

rconard:
Users have indicated that Central Package Version Management (CPVM) needs a mechanism to override the centrally managed versions of packages on a project-by-project basis. Project Package Version Override allows individual projects to import different versions of NuGet packages organized in the same, centralized location.

This design doc is based on input from NuGet users at Microsoft and from around the world.

Thank you to all of the engineers who have contributed!


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=9997) <br/>All CLA requirements met.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+2. Provide a means for organizing multiple package versions together.
+3. Manage the package versions of transitive dependencies alongside direct dependencies.
+4. Follow the MSBuild approach with minimal changes to existing XML patterns.
+
+# Feature Behavior
+
+- All `<PackageVersion>` should be defined within `Directory.Packages.props`, and `Version` may not be defined on `<PackageReference>` within projects.
+    - Specifying a `Version` on a `<PackageReference>` will throw a build error.
+    - Specifying a `<PackageVersion>` within a project file is an anti-pattern, and it will generate a build time warning. It will not throw an error, and a developer may choose to suppress the warning by specifying `<AllowProjectPackageVersion>true</AllowProjectPackageVersion>` within a `<PropertyGroup>`.
+- A `<PackageVersion>` may have a `CentralManagementGroup` defined which allows a `<PackageReference>` to refer to this version using the `CentralManagementGroup` attribute.
```
anangaur, line 17: `CentralManagementGroup` sounds odd. What about the original `VersionGroup` or `CentralVersionGroup`. I think just `Group` also makes sense in the context of NuGet/PackageReference.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - When a `Version` to be imported by a `<PackageReference>` needs to be different than the global group defined in `Directory.Packages.props`, a `CentralManagementGroup` can be specified on both the `<PackageReference>` in a project file and a corresponding `<PackageVersion>` in `Directory.Packages.props` that specifies the needed `Version`.
+    - A `<PackageReference>` may only have one `CentralManagementGroup` defined.
+    - A project can include references to multiple `CentralManagementGroup`.
+    - When `CentralManagementGroup` is not specified, both `<PackageVersion>` and `<PackageReference>` reference a global `CentralManagementGroup` that is automatically created.
```
anangaur, line 21: `Global` --> `Default`? Also may be call out explicitly that this name for the group is reserved and cannot be created by customers as their custom group.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
anangaur: My suggestion would be to avoid the word "central" for this property as the feature itself is about overriding the "central" behavior :)

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+- Transitive dependencies are resolved based on finding a common level of specificity between direct imports.
+    - If a transitive dependency is imported by a package defined within the global group, then the `<PackageVersion>` for the transitive dependency from the global group applies.
+    - If a transitive dependency is imported by a package in multiple `CentralManagementGroup` that both include a `<PackageVersion>` for that transitive dependency, then the `<PackageVersion>` for the transitive dependency from the global group applies.
```
anangaur, line 24: Couldn't parse this usecase. Will look at the example to understand more.. An inline example for each of these cases might have helped.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - If a transitive dependency is initiated by multiple packages within multiple, different `CentralManagementGroup` but **only one** of the `CentralManagementGroup` specifies a `<PackageVersion>` for the transitive dependency, then that version is used.
+    - If a transitive dependency is initiated by multiple packages within multiple, different `CentralManagementGroup` and multiple `CentralManagementGroup` specify **the same** `<PackageVersion>` for the transitive dependency, then the common version is used.
+    - If a transitive dependency is initiated by multiple packages within multiple, different `CentralManagementGroup` and multiple `CentralManagementGroup` specify **different** `<PackageVersion>` for the transitive dependency, then the project will throw a build error due to a [cousin dependency](https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#cousin-dependencies) conflict.
```
anangaur, line 28: Why error? This seems like regular NuGet resolution. In a non CPVM scenario, this is one of the easiest resolution to do. I am concerned that we may be looking at many such errors making the feature unusable.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+- A `<PackageVersion>` may have a `CentralManagementGroup` defined which allows a `<PackageReference>` to refer to this version using the `CentralManagementGroup` attribute.
+    - When a `Version` to be imported by a `<PackageReference>` needs to be different than the global group defined in `Directory.Packages.props`, a `CentralManagementGroup` can be specified on both the `<PackageReference>` in a project file and a corresponding `<PackageVersion>` in `Directory.Packages.props` that specifies the needed `Version`.
+    - A `<PackageReference>` may only have one `CentralManagementGroup` defined.
+    - A project can include references to multiple `CentralManagementGroup`.
+    - When `CentralManagementGroup` is not specified, both `<PackageVersion>` and `<PackageReference>` reference a global `CentralManagementGroup` that is automatically created.
+- Transitive dependencies are resolved based on finding a common level of specificity between direct imports.
```
anangaur, line 22: Have we established that folks need to refer to multiple groups for their packages in a project? How rigid is this requirement? The reason I ask this is because of the resolution use-cases explained below. It makes the override feature too complicated and the package resolution logic even more. Try explaining to customers before they pull their hair 😄.
What does `paket` do here? Can we sync up with them to understand their learnings in this space? I can provide contacts.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+
+1. Even though package versions need to be specified for specific projects, keep the package versions centralized. Avoid situations where solution-level and project-level package version management are mixed in the same repository.
+2. Provide a means for organizing multiple package versions together.
+3. Manage the package versions of transitive dependencies alongside direct dependencies.
+4. Follow the MSBuild approach with minimal changes to existing XML patterns.
+
+# Feature Behavior
+
+- All `<PackageVersion>` should be defined within `Directory.Packages.props`, and `Version` may not be defined on `<PackageReference>` within projects.
+    - Specifying a `Version` on a `<PackageReference>` will throw a build error.
```
cristinamanum, line 15: This requirement is not related with version override. Should it moved to the Main spec? The error is on Restore.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - Specifying a `<PackageVersion>` within a project file is an anti-pattern, and it will generate a build time warning. It will not throw an error, and a developer may choose to suppress the warning by specifying `<AllowProjectPackageVersion>true</AllowProjectPackageVersion>` within a `<PropertyGroup>`.
```
cristinamanum, line 16: This requirement is a new requirement not related with the Version Override. 
I am not sure if it can be detected the location of the items; in addition the PV items can imported from custom props files if developers need to go around this rule. I would suggest not have this rule.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+- A `<PackageVersion>` may have a `CentralManagementGroup` defined which allows a `<PackageReference>` to refer to this version using the `CentralManagementGroup` attribute.
+    - When a `Version` to be imported by a `<PackageReference>` needs to be different than the global group defined in `Directory.Packages.props`, a `CentralManagementGroup` can be specified on both the `<PackageReference>` in a project file and a corresponding `<PackageVersion>` in `Directory.Packages.props` that specifies the needed `Version`.
+    - A `<PackageReference>` may only have one `CentralManagementGroup` defined.
+    - A project can include references to multiple `CentralManagementGroup`.
```
cristinamanum, line 20: Allowing this behavior adds complexity that I am not sure it is needed. Is there user data supporting the necessity of this scenario?

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - When `CentralManagementGroup` is not specified, both `<PackageVersion>` and `<PackageReference>` reference a global `CentralManagementGroup` that is automatically created.
+- Transitive dependencies are resolved based on finding a common level of specificity between direct imports.
+    - If a transitive dependency is imported by a package defined within the global group, then the `<PackageVersion>` for the transitive dependency from the global group applies.
```
cristinamanum, line 23: It would be great to add samples for each of these cases. It is difficult to follow them.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - If a transitive dependency is imported by a package in multiple `CentralManagementGroup` that both include a `<PackageVersion>` for that transitive dependency, then the `<PackageVersion>` for the transitive dependency from the global group applies.
+    - If a transitive dependency is imported by a `<PackageReference>` with a `CentralManagementGroup` specified, then the `<PackageVersion>` for the transitive dependency from the `CentralManagementGroup` applies. If the transitive dependency does not have a matching `<PackageVersion>` within the same `CentralManagementGroup`, then the global group applies.
+    - If a transitive dependency is initiated by multiple packages within multiple, different `CentralManagementGroup` but **only one** of the `CentralManagementGroup` specifies a `<PackageVersion>` for the transitive dependency, then that version is used.
```
cristinamanum, line 26: Related with one of the above comments for supporting multiple groups inside of a project - in addition to the fact that this adds a great deal of complexity, the data for all the groups that a project needs need to carried to the NuGet core algorithms to  resolve the transitive dependencies. This could have performance impact.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    - If a transitive dependency is initiated by multiple packages within multiple, different `CentralManagementGroup` and multiple `CentralManagementGroup` specify **the same** `<PackageVersion>` for the transitive dependency, then the common version is used.
```
cristinamanum, line 27: Perf and complexity concerns as above.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
cristinamanum: I just observed a set of samples below, would be possible to add relevant sample(s) next to each verbatim requirement?

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+</Project>
+```
+
+This project uses a single `CentralManagementGroup="B"`.
+
+`baz` has three dependencies. These are transitive dependencies for the project.
+
+- `foo` is imported with version 3.3.3 from `CentralManagementGroup="B"` because `foo` has a `<PackageVersion>` defined for group `"B"`.
+- `qux` is imported with version 3.0.0 from `CentralManagementGroup="B"` because `qux` has a `<PackageVersion>` defined for group `"B"`.
+- `xyzzy` is imported with version 303.0.0 because `baz` has a direct dependency on `xyzzy` version 303.0.0. There is no `<PackageVersion>` defined for `xyzzy` in group `"B"`.
```
cristinamanum, line 195: I do not see any reason for this package to be pulled from group A.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+
+5*.csproj*
+
+```xml
+<Project>
+  <ItemGroup>
+    <!-- global group -->
+    <!-- Version 2.9.0 -->
+    <PackageReference Include="Serilog"/>
+    <!-- "A" group -->
```
cristinamanum, line 207: Isn't this Default group?

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+<Project>
+  <ItemGroup>
+    <!-- No entry for Newtonsoft.Json in the "B" group -->
+    <!-- ❗ Throws an Error - Project Will Not Build ❗ -->
+    <PackageReference Include="Newtonsoft.Json" CentralManagementGroup="B" />
+  </ItemGroup>
+</Project>
+```
+
+Because there is no entry for `Newtonsoft.Json` in `CentralManagementGroup="B"`, the project will throw an error and fail to build.
```
cristinamanum, line 333: ``the project will throw an error and fail to build'`` -> '``the project will fail restore'``

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+```xml
+<Project>
+	<PropertyGroup>
+    <AllowProjectPackageVersion>
+        true
+    </AllowProjectPackageVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+		<!-- Remove version defined in Directory.Packages.props, if any -->
```
cristinamanum, line 351: This is not necessary to be defined at the project level. A props file with this settings can be imported and not detected.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    <PackageVersion Remove="Newtonsoft.Json" /> 
+    <!-- Define PackageVersion which applies only to this project --> 
+    <PackageVersion Include="Newtonsoft.Json" Version="12.0.3" />
+    <!-- Version 12.0.3 -->
+    <PackageReference Include="Newtonsoft.Json" />
+  </ItemGroup>
+</Project>
+```
+
+Including a `<PackageVersion>` outside of the `Directory.Packages.props` file will ordinarily throw a build warning.
```
cristinamanum, line 361: I am not sure if MsBuild imports can be tracked. Same concern for the project data that we receive from ProjectSystem.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
```diff
@@ -0,0 +1,425 @@
...
+    </ItemGroup>
+</Project>
+```
+
+In this `Directory.Packages.props` file, there are four `CentralManagementGroup` of `<PackageVersion>` defined.
+
+- The global `CentralManagementGroup` is defined automatically and applies to version 1.1.1 of `foo` where no `CentralManagementGroup` is specified.
+- `"A"`, `"Anetcoreapp3.1"`, `"Anet472"` are three different `CentralManagementGroup`. They are named similarly as a convention, but they remain three independent groups. MSBuild does not understand any relationship between these groups.
+
+### Sample Project 11: Inline Multi-Targeting
```
cristinamanum, line 387: I think that multi-targeting should be a separate discussion.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
nkolev92: I think trying to restrict where a item is declared in MSBuild is largely a losing proposition. 

MSBuild's power comes from the fact that everything is processed in a predictable order.

designs/CentralNuGetManagement/CPVM-Project_Package_Version_Override.md
nkolev92: The transitive application approach sounds pretty complicated. Both from a technical + UX perspective. 

Predictability is one of the things that NuGet's dependency resolution strives for (assuming you don't use floating versions), and the transitive dependency version inference additions are pretty complex. 

I think my perspective is largely naive, but I can see myself wanting to change a version or two. Mixing many `groups` creates a difficult mental model.

nkolev92 (commented):
Thanks for the very detailed write-up @rconard! 
This is a complicated topic and you definitely captured as many scenarios as possible, so that the reader doesn't need to guess anything!

I left a few comments on the `how it works` part. 

I'll cover all the additional examples before the sync.

JonDouglas:
Hello!

Actions to take:

JonDouglas:
I am closing this PR due to staleness. If it's important to include in the repository, please re-open the PR and assign appropriate reviewers. Thanks!

Issue #9817, opened 2020-07-18
License and icon file existence check should always use a case-sensitive comparison


PathogenDavid:
When you attempt to create a NuGet package with a license file that doesn't exist, NuGet will complain with NU5030:

```
The license file 'LICENSE.txt' does not exist in the package.
```

However, [this check uses an OS-dependent string comparison](https://github.com/NuGet/NuGet.Client/blob/1f1213960012d6452b63a267607d1e237318025e/src/NuGet.Core/NuGet.Packaging/PackageCreation/Authoring/PackageBuilder.cs#L588).

This isn't correct as NuGet packages are zip files, which use case-sensitive paths. As a result, using the wrong case for the `license` element results in an invalid package.

--------------

Consider [this modified PackageLicenseFileExample](https://github.com/PathogenPlayground/Samples/blob/master/PackageLicenseFileExample/PackageLicenseFileExample.csproj) from NuGet/Samples. Namely the typo made to `PackageLicenseFile`:

```xml
<PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
```

Running `dotnet pack` results in a "successful" nupkg:

![image](https://user-images.githubusercontent.com/278957/87859745-13a3bb80-c8fd-11ea-8d1b-d2b438cb346c.png)

Attempting to upload this package to NuGet.org results in an error:

![image](https://user-images.githubusercontent.com/278957/87859778-464db400-c8fd-11ea-9d19-230c87285e92.png)

If you put the package into a local feed and attempt to view the license from Visual Studio 2019 (16.6.4) you'll get an error:

![image](https://user-images.githubusercontent.com/278957/87859796-7ac17000-c8fd-11ea-8953-64fc9606c831.png)

I also tried an old instance of Nexus Package Manager, which did accept the package. (However that may be more due to the fact that it doesn't understand embedded licenses.) I did not check how any other package servers reacted to the broken package.

--------------

## Versions and stuff from the issue template

NuGet product used: `dotnet` and `VS UI`.

NuGet version: Whatever `dotnet` is using under the hood. (The old `nuget.exe` I have kicking around in my environment is 4.8.1.5435)

dotnet.exe --version: 5.0.100-preview.6.20318.15

VS version: 16.6.4

OS version: Windows 10 v1909 (18363.959)

[...]

PR #10049, opened 2020-09-18
Design for warning on improper casing for PackageLicenseFile and PackageIcon

PathogenDavid:
The PR adds a design spec to address https://github.com/NuGet/Home/issues/9817

[Markdown view](https://github.com/PathogenDavid/NuGet-Home/blob/33f08de087d998d4041ef9c570737da68d024de2/designs/PackageValidation/FileNameCaseSensitivity.md)


designs/LicenseFileNameCaseSensitivity.md
```diff
@@ -0,0 +1,63 @@
+
+# Warn on improper license file name case
```
zivkan: This seems to be a pack validation, so I think it would be good to be in `designs/PackageValidation/`.

zivkan: I imagine the same problem exists for embedded icons, as the design is basically the same (nuspec property points to a file in the nupkg). I imagine the implementation could check for both at no higher cost than checking only one, so we should "expand" the scope.

designs/LicenseFileNameCaseSensitivity.md
```diff
@@ -0,0 +1,63 @@
...
+  * Not all package feeds can or will verify the license file exists. (IE: Local package feeds, Nexus Repository)
+* Fix the behavioral inconsistency between packing on Windows/macOS vs Linux.
+
+## Non-Goals
+
+Changing the behavior of packing on Linux is a non-goal.
+
+## Solution
+
+Introduce a new warning when attempting to pack a package with improper casing on the `PackageLicenseFile`:
```
zivkan: If we accept my proposal to make the warning apply to both embedded licenses as well as embedded icons, should we have one warning code per recourse type, or one warning code and make the message include sufficient information for the customer to understand where to fix the issue?

zivkan (approved):
I think this can be extended to work for both embedded icons, as well as licenses, at almost zero extra implementation cost. Otherwise I'm happy with the proposal.

designs/LicenseFileNameCaseSensitivity.md
nkolev92: How about we keep the same error code, but slightly change the message?

And do it for both license + icon.

nkolev92 (commented):
Thanks a lot for this @PathogenDavid. This is great. 

I think personally, I'd prefer this being an error with the same code as before, but an altered message. Let's pull in a few more people to get their thoughts!

Calling on @JonDouglas for help here! :)

designs/LicenseFileNameCaseSensitivity.md
PathogenDavid: Oops, I somehow didn't even notice the other folders. Moved in 2b7812b0e98437721a0187498b46790cd370926b

PathogenDavid: You're right, it does. I reworded the proposal to apply to package icons as well in https://github.com/NuGet/Home/pull/10049/commits/e19b3af5f919668d6712d3351f351a8d57f92e6b

image

designs/LicenseFileNameCaseSensitivity.md
PathogenDavid: If we don't re-use the same error codes like @nkolev92 is suggesting, I would imagine they should be separate error codes since [NU5030](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5030) and [NU5046](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5046) are already separate.

PathogenDavid:
I've updated the design to apply to package icons after @zivkan's suggestion since they have the same issue. I also moved it to the `PackageValidation` folder.

I also added @nkolev92's suggestion to the open questions section to ensure it isn't forgotten. I only gravitated towards the new warning to avoid breaking people and since @zkat [showed preference for it](https://github.com/NuGet/Home/issues/9817#issuecomment-670076280).

Ignoring compat concerns I think it should definitely be an error. So if the NuGet team isn't worried about breaking old, malformed packages, I'm more than happy to go along the error route.

designs/LicenseFileNameCaseSensitivity.md
zivkan: I forgot/didn't know about these existing errors. I very much like the idea of using the same codes and just changing the message when we detect the name with the wrong casing. 👍 

I don't want to add scope creep, but maybe we could also look for the filename in different directories in the nupkg, just in case the user didn't match the nupkg paths?

nkolev92 (approved):
This looks great so far, I think the only open question on NuGet side is warning or error. 

cc @JonDouglas @aortiz-msft 

Can we get someone to drive a quick internal discussion?

JonDouglas:
👍 to an error code. Given that if we warn for this scenario & a user can still be unsuccessful, an error is more proper.

designs/PackageValidation/FileNameCaseSensitivity.md
```diff
@@ -0,0 +1,74 @@
...
+NU0001: The icon file 'Icon.png' does not exist in the package. (Did you mean 'icon.png'?)
+```
+
+## Future Work
+
+A warning is specified to avoid breaking packages which previously packed successfully even though they were malformed. In a future .NET SDK release, this warning could become an error by default similar to [NU1605](https://github.com/dotnet/sdk/blob/bc5c514a45a1a900b71d39a3541196bc2cf1bb0d/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.CSharp.props#L19).
+
+## Open Questions
+
+Should this be an error that reuses the existing errors (NU5030 for licenses and NU5046 for icons) instead?
```
zivkan: I saw yes, it's reuse the existing error codes, and add new messages if we can find a case-insensitive match.

zivkan (commented):
@PathogenDavid, given Jon's approval, let's close out the last open question, and get this spec merged! 👍

designs/LicenseFileNameCaseSensitivity.md
PathogenDavid: The spec has been updated to use the existing error codes.

designs/PackageValidation/FileNameCaseSensitivity.md
PathogenDavid: Spec updated.

PathogenDavid:
@zivkan Sorry for the wait, finally got around to updating the spec.

> I don't want to add scope creep, but maybe we could also look for the filename in different directories in the nupkg, just in case the user didn't match the nupkg paths?

(This was in a review comment thread and we never talked about it.) It wouldn't be the hardest thing to add, but I'm inclined to think this is probably an exceedingly rare situation and is out-of-scope. I added it to the considerations and labeled it as out of scope, but I'm willing to add it if you think it's common enough to warrant doing.

Edit: As I started implementing the changes to NuGet.Client, the implementation I ended up with would make this easy to do without dirtying up `ValidateLicenseFile` or `ValidateIconFile`. So I'm actually less hesitant to do this now if you think it's valuable.

(Side note: Is there an easy way to run just a single test? Test Explorer didn't populate many of the tests for me and the `runTests.ps1` parameters don't seem to have much granularity. The tests added for this change are in [`NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs`](https://github.com/NuGet/NuGet.Client/blob/68b532ec076cb4a1fae6e1ab52399ab70d0e0136/test/NuGet.Core.FuncTests/Dotnet.Integration.Test/PackCommandTests.cs) since that's where similar tests live.)

zivkan:
> Is there an easy way to run just a single test? Test Explorer didn't populate many of the tests for me

You're probably aware, but Test Explorer can take some time to finish discovering all test assemblies. I've never seen it only discover a subset of tests from within an assembly. But it "works for me".

From the command line you can go to the right project directory and run "dotnet test --filter YourTestNameHere". If you've already compiled the project in Visual Studio, or elsewhere, it's a bit faster to add `--no-build` or pass the path of the assembly to test itself. For example, in NuGet.Packaging.Test `dotnet test .\bin\Debug\netcoreapp5.0\NuGet.Packaging.Test.dll --filter Convention`.

And thank you for updating the spec! Merging now.

PathogenDavid:
> You're probably aware, but Test Explorer can take some time to finish discovering all test assemblies. I've never seen it only discover a subset of tests from within an assembly. But it "works for me".

Ah, you're right. I thought it had finished but it actually hadn't (and/or the test was nested down further than I expected.) For whatever reason it can't actually run tests though, but `dotnet test` works so I'll just use that.

Thanks for the merge! PR to NuGet.Client should be coming soon.

PathogenDavid:
I was going to wait until CI finished running, but it looks like CI doesn't automatically run on community PRs or something along those lines. The PR is here: https://github.com/NuGet/NuGet.Client/pull/3754

zivkan:
yes, that's right. See step 4 from our contributing guide: https://github.com/NuGet/NuGet.Client/blob/dev/CONTRIBUTING.md#contributing-step-by-step

We're making slow steps towards being able to run CI automatically, and make CI logs public, but we have quite a way to go yet.

PathogenDavid:
Ah, makes sense. I read that at one point but forgot that bit. Thanks for the confirmation!

Issue #9986, opened 2020-09-03
Implement a pre-registration mechanism for legacy PR projects that call restore at solution open


nkolev92:
Say you have a mix of legacy + SDK projects, where the legacy projects are say extensibility which call NominateProjectASync without data. 

Given that NuGet has no clue that it's supposed to wait for these projects(unlike nomination),  we should implement a registration mechanism, that tells NuGet to expect an `I'm ready` call to avoid duplicate restores.

Design at https://github.com/NuGet/Home/pull/10064.

nkolev92:
Lowering the estimate. 

The design is out, and we need to go through the design first before scheduling this work.

nkolev92:
Lowering the estimate again.

Issue #9801, opened 2020-07-15
Ordered solution load and impact on NuGet restore performance


nkolev92:
NuGet PackageReference is transitive and as such additional information in children projects will affect the restore of the project being restored. 

Say we have a large solution among which the projects are:  A => B => C. 

Say the nominations come in the following order: A, B, then C.

If Nuget decides to restore after A & B have been nominated, whenever C gets nominated again, we run a sln restore, we'll need to invalidate A & B's restores and performance a full restore again. 

If the nominations come in the correct order, C => B=> A, no restore will be wasted.

erdembayar:
If we ever need to do this on our side instead of  expecting Project system to give nomination in correct order then I can take it. Of course it may not be perfect  but we can give try sometime in the future.

nkolev92:
Not a consideration at this point. 
Closing this issue for now. 

The focus is on running as few restores as possible.

PR #10064, opened 2020-09-23
Ensure legacy PackageReference projects, that trigger a restore through IVsSolutionRestoreService, participate in the solution load restore batching 

nkolev92:
Design for https://github.com/NuGet/Home/issues/9986. 

@davkean I'm relying on your expertise for feedback here and help deciding the right capability. 

fyi @panopticoncentral as this is related to solution load, but it's only mixed or legacy projects.

### DIFF
--- a/designs/SolutionLoad-CsprojPackageReferenceRestore.md
+++ b/designs/SolutionLoad-CsprojPackageReferenceRestore.md
@@ -56,7 +56,8 @@ While, there's many csproj.dll projects supporting PackageReference, we have so 
 
 ### What are the consequences
 
-Talk about the time wasted.
+Wasted resources. Reasoning about `solution load` is very difficult and because VS loads projects in parallel and in order in which they were declared in the solution, this means that not all legacy projects are guaranteed to nominate in that window, so we're likely to do restores that have no effective impact. 
+Note that this now means that we're running an extra optimized partial restore, but for legacy projects this can still be expensive.
 
 ## Who are the customers
 
--- a/designs/SolutionLoad-CsprojPackageReferenceRestore.md
+++ b/designs/SolutionLoad-CsprojPackageReferenceRestore.md
@@ -54,8 +54,7 @@ Given that VS loads projects in the order in which they are defined in the proje
 
 While, there's many csproj.dll projects supporting PackageReference, we have so far identified that Xamarin and VSIX projects are the ones that do call nomination on solution load.
 
-
-### What are the consequences 
+### What are the consequences
 
 Talk about the time wasted.
 
@@ -82,16 +81,23 @@ The reference data will still be read through the poll model. The reason for thi
 
 Project-systems have capability model. Whenever NuGet detects a project that's csproj.dll-based PackageReference, it can check for an additional capability that will indicate that a project is expected to call nominate at solution load.
 
-### What's the capability name
-??
+### What capability in particular
 
-Talk about the capability name, otions, considerations
+We *might* be able to re-use some existing capabilities, but we need to be able to correctly identify and bucketize the right groups of project.
+The particular capability name is up for discussion at this point.
+
+[See Open Questions](#open-questions)
 
 ## Future Work
+
+* Once the design has been agreed on, this will require changes in potentially multiple project-systems. The changes should be minimal and low effort.
 
 ## Open Questions
 
 * What should be the capability name?
+
+`PackageReferences` was something I considered, but unfortunately having a project-system declare `PackageReferences` would make a project end up in the NET Core/CPS based PackageReference codepaths, which rely on nomination for getting the reference data.
+These projects need to continue to work in the legacy codepath.
 
 ## Considerations
 
@@ -101,4 +107,5 @@ We can argue inaction as the extra restore are not going to happen consistently.
 
 ### References
 
-Add link to partial restore, capabilities docs etc.
+* [Visual Studio partial restore optimization](https://github.com/NuGet/Home/blob/756e168725e720491f3a1e013e30a73281b224c2/designs/VisualStudio-PartialRestoreOptimization.md)
+* [Project Capabilities](https://github.com/microsoft/VSProjectSystem/blob/master/doc/overview/project_capabilities.md)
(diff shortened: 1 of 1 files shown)


designs/SolutionLoad-CsprojPackageReferenceRestore.md
```diff
@@ -0,0 +1,110 @@
...
+
+Even if NuGet doesn't expect a nomination from certain projects, it will *still* batch the restores ran for that nomination call because they might've happened before the last nomination call for an SDK based project.
+
+Given that VS loads projects in the order in which they are defined in the project file, it happens almost consistently that the nomination calls from those 2 VSIX projects come after the last SDK based project nomination.
+
+While, there's many csproj.dll projects supporting PackageReference, we have so far identified that Xamarin and VSIX projects are the ones that do call nomination on solution load.
+
+### What are the consequences
+
+Talk about the time wasted.
```
zivkan: you didn't finish writing the spec?

designs/SolutionLoad-CsprojPackageReferenceRestore.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Open Questions
+
+* What should be the capability name?
+
+`PackageReferences` is a capability we can consider. I am not confident whether `PackageReference` suggests a communication method with NuGet (ie. push vs pull for getting the references). Is redefining a meaning of a capability acceptable?
+
+## Considerations
+
+No other solutions for batching were considered.
```
zivkan, line 104: What about ignoring nominations for project types we don't expect nominations from?  Although the XMLDoc for the nominate APIs explicitly says that the return task will be complete when the restore is complete. Therefore it's hard for us to claim that nominations aren't supposed to be a restore.

designs/SolutionLoad-CsprojPackageReferenceRestore.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Future Work
+
+* Once the design has been agreed on, this will require changes in potentially multiple project-systems. The changes should be minimal and low effort.
+
+## Open Questions
+
+* What should be the capability name?
+
+`PackageReferences` is a capability we can consider. I am not confident whether `PackageReference` suggests a communication method with NuGet (ie. push vs pull for getting the references). Is redefining a meaning of a capability acceptable?
```
zivkan, line 100: Given my recent sfproj experience, this worries me. They use CPS, just not PackageReference. I have no idea if they call NominateProjectAsync to get a restore. But there could be other CPS projects using packages.config who are calling NominateProjectAsync, and if we tell them to start advertising PackageReference, their projects will break, unless they make the full migration to PackageReference, including migrating existing customer projects on first load and possibly needing to implement different APIs to support modifying PackageReference items in the project file.

zivkan (commented):
sorry for taking 6 weeks to give feedback. 😞

designs/SolutionLoad-CsprojPackageReferenceRestore.md
nkolev92: I'll clean this up :)

designs/SolutionLoad-CsprojPackageReferenceRestore.md
nkolev92: The problem is that https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.SolutionRestoreManager.Interop/IVsSolutionRestoreService2.cs was added specifically with that intention. 

Even if the completion means `project is ready` we can't guarantee that because it's hard to differentiate between solution load and non-solution load scenarios.

designs/SolutionLoad-CsprojPackageReferenceRestore.md
nkolev92: packages.config based projects are not supposed to call nominate. The fake nominate call was added for legacy PackageReference.

They *should not* advertise PackageReference capability. This solution is not for them.

nkolev92:
🔔 @zivkan addressed feedback.

zivkan (approved):

designs/SolutionLoad-CsprojPackageReferenceRestore.md
```diff
@@ -0,0 +1,111 @@
...
+### Current implementation
+
+In 16.8, we made various improvements to allow us to run only 1 run on solution load. This required changing the batching logic for solution load.
+
+A bit of background around solution load and the activities happening there as of 16.8 P4.
+
+* At solution load, NuGet is able to identify and classify *all* projects in the solution (by listening to the SolutionEvents ProjectAdded).
+* NuGet now knows all projects in the solution that are expected to nominate.
+* When the first nomination comes in, NuGet starts waiting for *all projects* to be nominated.
+* In order to ensure that we don't wait for forever, NuGet has a 10s sliding window. This means that even if all the projects were *not* nominated yet, NuGet will still try to restore as long as 10s since the last nomination have passed.
```
erdembayar, line 41: Previously I saw issue which try to improve duplicate restore nomination nomination, but couldn't find it anymore.
How about we create tree graph of solution first then instead of waiting 10s can we let in bottom level projects and let them restore since they wouldn't be nominated/restored again. That way we might improve perf for large projects. 
If idea is feasible then you can assign that issue to me to investigate.

nkolev92: It's a loaded statement, but the general idea is something we've considered in the past. 

My best attempt to answer your individual comments. 

> How about we create tree graph of solution first then instead of waiting 10s can we let in bottom level projects and let them restore since they wouldn't be nominated/restored again. 

You are suggesting something close to https://github.com/NuGet/Home/issues/9801. 
The trick here is that we don't know the amount of work necessary to do a partial restore, while it could reduce individual restore times, but it has the potential to lead to slower overall experience. 

Note that in *most* cases, we don't drop out of the sliding window, but we get all nominations instead. 
The complexity of a partial approach is also something that needs to be accounted for, at this point our performance is `good enough` where such a rework might prove to be too costly.

erdembayar (approved):

PR #10156, opened 2020-10-19
Spec for PMUI Tab-Switch Refactoring

### LINKED ISSUE #10154: Spec for PMUI Tab-Switch Refactoring
@donnie-msft commented on [Wed Oct 14 2020](https://github.com/NuGet/Client.Engineering/pull/588)

---

@donnie-msft commented on [Thu Oct 15 2020](https://github.com/NuGet/Client.Engineering/pull/588#issuecomment-709643096)

> Thanks for the great write-up @donnie-msft 
> 
> Something that I was not able to find (or missed), when you open the PM UI and it ends up on the default tab, does the browse tab start loading? I assume the answer is no. 
> 
> How about the reverse? On the browse, does the installed tab equivalent data start loading? 
> I know that we do look at the existing installed packages in order to populate the number in the updates tab label, and I think this is one reason why initial PM UI loading feels heavy to users with machines that are not really powerful.

This spec has no effect on the existing behavior of automatically loading whichever tab is open when the window is first opened.

We always load the tab header counts, and perhaps could be more conservative with that. Do you feel it fits with this work or a future spec?

---

@nkolev92 commented on [Thu Oct 15 2020](https://github.com/NuGet/Client.Engineering/pull/588#issuecomment-709646231)

> This spec has no effect on the existing behavior of automatically loading whichever tab is open when the window is first opened.

Oh I didn't mean to suggest that. We read a setting currently and we default to that, I think that's great.

> We always load the tab header counts, and perhaps could be more conservative with that. Do you feel it fits with this work or a future spec?

I think it's ok for that to be a part of a future work item. 
Doesn't the tab header count update require fetching of all packages for that project/solution & checking whether they have any updates?

---

@donnie-msft commented on [Thu Oct 15 2020](https://github.com/NuGet/Client.Engineering/pull/588#issuecomment-709707000)

> Doesn't the tab header count update require fetching of all packages for that project/solution & checking whether they have any updates?

Yes, and Deprecation info!
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#L947-L950

---

@nkolev92 commented on [Fri Oct 16 2020](https://github.com/NuGet/Client.Engineering/pull/588#issuecomment-710624742)

> Yes, and Deprecation info!
> https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#L947-L950

Yuck :D 
So say you open the browse tab...wait for 20s, then open the installed tab, does it now get loaded quicker than just loading the installed tab directly? 
Basically, are we reusing the work that we did for figuring out the label count?

donnie-msft:
Fixes: https://github.com/NuGet/Home/issues/10154

### DIFF
--- a/designs/PMUITabSwitchRefactoring.md
+++ b/designs/PMUITabSwitchRefactoring.md
@@ -1,0 +1,133 @@
+# Package Manager UI Tab-Switch Refactoring
+
+* Status: In-process
+* Author(s): [Donnie Goodson](https://github.com/donnie-msft)
+* Issue: [10082](https://github.com/NuGet/Home/issues/10082) PMUI Package List Refactoring
+
+## Problem Background
+
+The NuGet Package Manager (PMUI) in Visual Studio (VS) has a design where navigating to a new tab will trigger a heavy reload of NuGet packages. At best, this is a perceivable flickering as the packages clear and repopulate. At worst, it can take 10's of seconds or even minutes. Reloading will reach out to the selected sources to retrieve an updated packages list. Nuget.org is generally quick, but some remote sources can be slow. 
+
+
+## Who are the customers
+
+Visual Studio users who may utilize the PMUI.
+
+## Goals
+
+* Eliminate the feeling of heavyweight navigation actions within the PMUI.
+* Refactor some XAML and backing code to improve performance and maintainability.
+
+## Non-Goals
+
+* Protocol performance tuning (eg, HTTP caching) is not part of this effort.
+
+## Solution
+I believe we unnecessarily refresh the package lists on every tab-switch. Users can refresh their package data by using the Refresh button or performing any package action (ie, Install, Update, change Sources). 
+
+Therefore, showing and hiding lists that are maintained in memory results in a much better experience.
+
+### My branch
+* My current branch (still actively working on it): https://github.com/NuGet/NuGet.Client/tree/dev-donnie-msft-prototypeTwoItemsLists
+* Current commit on branch as of this writing: https://github.com/NuGet/NuGet.Client/commit/420e1d12fa418089f1f299b981d1ed409ee59cc4
+* _(I've included links throughout this spec to specific lines)_
+
+### Core Concepts
+* From a user's point-of-view, there has traditionally been only one list of packages. My solution adds a **second Listbox** that I show and hide depending on selected tab. 
+
+* Additionally, a WPF [CollectionViewSource.Filter](https://docs.microsoft.com/en-us/dotnet/api/system.windows.data.collectionviewsource?view=netcore-3.1#events) ("UI Filtering") is applied to only show packages with Updates Available on the Updates tab.
+
+### High-level Logic
+`_listBrowse` - Represents packages on the Browse tab only
+
+`_listInstalled` - Represents packages on the Installed, Updates, and Consolidate tabs
+
+**Browse tab**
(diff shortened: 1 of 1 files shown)


designs/PMUITabSwitchRefactoring.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Installed tab**
+* `_listBrowse` is Collapsed
+* `_listInstalled` is Visibile
+* UI Filter set to null 
+
+**Updates tab**
+* `_listBrowse` is Collapsed
+* `_listInstalled` is Visibile
+* UI Filter set to only packages with an Update Available https://github.com/NuGet/NuGet.Client/blob/420e1d12fa418089f1f299b981d1ed409ee59cc4/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml.cs#L390
```
erdembayar, line 65: If I understand correctly instead of making new list we keep one  always in memory. Only part I worry is correctness/stale data issue. What if new version just pushed and available, so when we're going to see new latest updates? Even though current implementation is keep re-loading with every tab switch, still data is fresh and current.  Maybe we need another background task keep them fresh every 5 min or trigger it when UI is not busy.

designs/PMUITabSwitchRefactoring.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Consolidate tab**
+* `_listBrowse` is Collapsed
+* `_listInstalled` is Visibile
```
erdembayar, line 69: Above stale data problem applies here too.

designs/PMUITabSwitchRefactoring.md
```diff
@@ -0,0 +1,118 @@
...
+
+`_listInstalled` - Represents packages on the Installed, Updates, and Consolidate tabs
+
+**Browse tab**
+* `_listBrowse` is Visibile https://github.com/NuGet/NuGet.Client/blob/420e1d12fa418089f1f299b981d1ed409ee59cc4/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml#L379
+* `_listInstalled` is Collapsed https://github.com/NuGet/NuGet.Client/blob/420e1d12fa418089f1f299b981d1ed409ee59cc4/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/InfiniteScrollList.xaml#L388
+
+**Installed tab**
+* `_listBrowse` is Collapsed
+* `_listInstalled` is Visibile
```
erdembayar, line 59: If new package installed from console or by editing csproj file does this list reflect it?

designs/PMUITabSwitchRefactoring.md
donnie-msft: Right, we aggressively refresh, today. However, the worst case I can think of is a new Version of a package already in memory is available on a source, and that's not reflected in the UI. The natural user response to that is to press the "refresh" button.
I'm not sure if we really need a timer to automatically refresh.

designs/PMUITabSwitchRefactoring.md
donnie-msft: Yes, many events trigger a Refresh. Modifying the project is one of those.

nkolev92 (approved):

erdembayar (approved):

Issue #6339, opened 2017-12-19
NuGet Package Visualizer


nkolev92:
The NuGet Package Visualizer has been removed since VS 2015. 

This is a task to track the progress to bring it back and implement it for Package Reference.

rohit21agrawal:
more info about this?

nkolev92:
I think it's pretty clear, and this is as much information as I have at this point. 
Over a year back I built an initial prototype covering packages.config only, but there's plenty of design work  needed for the package reference scenario.

lovettchris:
My [GraphViewer](https://lovettsoftwarestorage.blob.core.windows.net/downloads/GraphViewer/GraphViewer.application) ClickOnce application which is based on Visual Studio DGML Editor can visualize NuGet package dependencies.  As an Example, I created a simple ConsoleApp and added "Azure.Storage.Blobs" and "Microsoft.Azure.Kusto.Cloud.Platform.Aad" nuget packages, then I dragged the .sln onto the GraphViewer canvas and repeatedly performed "select all" followed by "right click/Add Dependencies" until no more dependencies were found.  Attached is the complete graph.

[test.zip](https://github.com/NuGet/Home/files/9578541/test.zip)

You can then do some what-if scenarios, for example, if you load this graph then find the "System.Diagnostics.DiagnosticSource" node and press the letter "i" a few times it will select the incoming nodes all the way to the root then you can hide all unselected nodes and you get this nice graph explaining why that dependency is there.
 
![image](https://user-images.githubusercontent.com/18707114/190495874-b0d3bc8c-bbb2-4d9d-b8ea-a888892a264c.png)

nkolev92:
A hackathon built global tool can do this. 

![DependencyGraphExample](https://user-images.githubusercontent.com/2878341/200424859-47ca88a4-253a-4a1d-8d98-f8036d4c6e99.jpg)

harshsikhwal:
I have implemented my own version of dependency generator: 
[NuGet-Visualizer](https://github.com/harshsikhwal/nuget-visualizer)

An interactive python script which helps users to generate dependencies for both NuGet and .NET Solution.

Graph for Microsoft.CodeAnalysis.CSharp:
![codeanalysis graph](https://user-images.githubusercontent.com/52648233/200638325-1929fa83-0713-4202-98e7-38efa650433b.png)

Hope this helps!

nkolev92:
[...]

PR #10179, opened 2020-10-22
Add Transitive Dependencies Spec

JonDouglas:


designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+* Issue: [5887](https://github.com/NuGet/Home/issues/5887)
+
+## Description: What is it?
+
+When a developer uses NuGet in a CLI or IDE environment, they want to be able to see their project’s transitive dependencies and understand how certain dependencies are included in their package graph for PackageReference projects.
+
+## Problem: What problem is this solving?
+
+Developers are often confused with top-level & transitive dependencies in their project. A developer should be able to see all the transitive packages for their project or solution & understand where they originated from. A user should be able to promote a transitive dependency quickly & easily to a top-level dependency in the case there is a security concern on the package version depended on or if promoting a top-level package will resolve a conflict. 
+
```
drewgillies, line 15: "...promoting **_to_** a top-level package will resolve a conflict"?

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+![](../resources/TransitiveDependencies/TransitiveDotNetCLI.png)
+
+### Visual Studio
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will additionally be a header titled “Top-level Packages” where the current experience is today for top-level packages.
+
+**By default, a user’s transitive packages will be collapsed** as a project can have hundreds, if not thousands of transitive packages & would potentially impact performance displaying these many records on each load.
+
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
```
drewgillies, line 73: > what top-level package(s) are bringing it in.

If there are no top-level packages bringing it in, would the comment simply reflect this? e.g. "This transitive dependency is brought in by one or more transitive dependencies."

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+- Auditing/upgrading transitive dependencies. I can only do it manually by adding explicit packages.
+- Drilling down the dependency tree
+- In the package detail view, some way to explore dependencies of dependencies and then their dependencies and so on, any level deep, without installing the package.
+
+
+## Cross-cutting Concerns
+
+-	The words “top-level” & “transitive” are words most people must stop & think about.
+    -	Should we use “direct” & “indirect”? Or “explicit” & “implicit”?
+-	Should a transitive dependency have the “installed” text given it’s already technically installed? Should it be called “promote” instead?
```
drewgillies, line 124: I like the sound of "promote", as you're right--I think "installed" misleads people into thinking it's not installed already.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+- Auditing/upgrading transitive dependencies. I can only do it manually by adding explicit packages.
+- Drilling down the dependency tree
+- In the package detail view, some way to explore dependencies of dependencies and then their dependencies and so on, any level deep, without installing the package.
+
+
+## Cross-cutting Concerns
+
+-	The words “top-level” & “transitive” are words most people must stop & think about.
+    -	Should we use “direct” & “indirect”? Or “explicit” & “implicit”?
```
drewgillies, line 123: My preference would be "direct dependencies" and "indirect dependencies" here, if we're going for more simple language. I think this is descriptive enough without being too esoteric.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+- A: The scope today is project view first & seeing how complex including this in solution view would be in the engineering costing.
+
+- Q: Who should ultimately own this work?
+- A: The NuGet client team should own this work. They are familiar with the Visual Studio Package Manager UI & concepts of transitive dependencies.
+
+- Q: Should packages.config be considered in this work?
+- A: Since packages.config is based on top-level packages only, it will just use its existing experience by placing all packages in the top-level packages section.
+
+- Q: I want to know what transitive dependencies are being brought in by a package. How do I do that?
+- A: You install package X & view the transitive dependencies panel inside Visual Studio or within the project system’s NuGet reference node. If we do CLI work, you would be able to type “dotnet list package –include-transitive” to see a similar output.
```
drewgillies, line 139: This exists in CLI already.

drewgillies (approved):
A few minor tweaks/thoughts, but looks good, @JonDouglas! Thanks. :shipit:

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
+
+# Transitive Dependencies
+
+* Status: Reviewed
+* Author(s): [Jon Douglas](https://github.com/NuGet)
```
drewgillies: Note--this is a NuGet link. Is that intentional?

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+## Context and Scope
+
+-	https://github.com/NuGet/Home/issues/5887 
+-	https://developercommunity.visualstudio.com/content/idea/538119/add-dependency-tree-view-for-packagereferences.html 
+
+## Minimal Requirements
+
+-	Visual list of transitive dependencies.
+-	Ability to see a transitive dependency package details.
+-	Ability to determine where a transitive dependency originated.
```
nkolev92, line 55: Sounds like we are also trying to solve https://github.com/NuGet/Home/issues/6339 with this. 

Visual Studio 2015 had a Package Visualizer you can look into.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+We should consider a minor tweak to the existing experience of “dotnet list package --include-transitive” to provide the user with a sense of where the package came from. This could be a new column next to “Resolved” which says “Transitively Referenced” or “Referenced” and has a list of the top-level packages that requested the dependency.
+
+![](../resources/TransitiveDependencies/TransitiveDotNetCLI.png)
+
+### Visual Studio
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will additionally be a header titled “Top-level Packages” where the current experience is today for top-level packages.
+
+**By default, a user’s transitive packages will be collapsed** as a project can have hundreds, if not thousands of transitive packages & would potentially impact performance displaying these many records on each load.
```
nkolev92, line 70: Are we going to audit transitive packages automatically and display that in the updates/installed labels? 

That's what we're doing for the top level packages. 

Note that I'm not recommending this, merely just clarifying :)

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
+![](../resources/TransitiveDependencies/TransitiveVSPMUI.png)
+
+Finally, selecting a transitive dependency will provide the user the ability to “Install” the package as you normally would a top-level dependency. This will promote the transitive dependency to a top-level dependency, and the transitive dependency will now be placed in the top-level packages list instead.
+
+## Goals and Non-Goals
+
+### Goals
```
nkolev92, line 80: I found it weird that the goals are below the what. :) 

Maybe there was a meeting benefit here, but I think I was OOF that day.

designs/Transitive-Dependencies.md
nkolev92: I've seen the argument both directions, particularly top level vs transitive and direct vs indirect and in the end I think consistency matters.

My suggestion would be to double down to what we've already committed in top level and transitive. 

Note that other package managers such as paket, pub, npm already use top-level/direct & transitive. 

I've rarely seen indirect being used.

designs/Transitive-Dependencies.md
nkolev92: I agree with the promote suggestion here.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+- Q: I want to know all the packages that depend on a transitive package. How do I do that?
+- A: You would use the tooltip to hover over a transitive dependency & view the list of packages that bring it in. Otherwise you would use the dotnet CLI and view the “referenced” column.
+
+- Q: Any thought towards seeing a visual graph for debugging transitive dependencies?
+- A: Tools exist for this such as DGML, https://github.com/ThomasArdal/NuGetPackageVisualizer, & https://docs.microsoft.com/en-us/nuget/release-notes/nuget-1.4#package-visualizer. It is not in the scope of this feature & would likely need to be a part of “debugging dependencies” efforts.
+
+- Q: Any indicator of transitive dependencies in search? We show top-level packages that are installed.
+- A: No because these packages are not technically installed nor managed by the user. The user can only manage top-level dependencies. We want users to be able to promote to a top-level, but we do not need to let them know that something is already transitively brought in.
+
+- Q: Is this really a feasible solution for license auditing? If 99 transitive dependencies are brought in, am I supposed to click through them all?
```
nkolev92, line 150: When installing, licenses from *all* packages being installed, including transitive dependencies are displayed in the license dialog.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Visual Studio
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will additionally be a header titled “Top-level Packages” where the current experience is today for top-level packages.
+
+**By default, a user’s transitive packages will be collapsed** as a project can have hundreds, if not thousands of transitive packages & would potentially impact performance displaying these many records on each load.
+
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
+![](../resources/TransitiveDependencies/TransitiveVSPMUI.png)
```
nkolev92, line 74: It's not clear to me how the dependency path for a given transitive dependency would work based on this screenshot. 

I just 2 separate lists.

nkolev92 (commented):
I was OOF when the original meeting happened, so I might have some more basic questions :)

designs/Transitive-Dependencies.md
JonDouglas: @nkolev92 There's a tooltip control in the screenshot that will show the dependency relation of a given transitive dependency. This however is likely to be cut given it's not very accessible :/

designs/Transitive-Dependencies.md
JonDouglas: Thanks for the clarification! Nothing changes in this regard then.

JonDouglas: For "Promote" we are awaiting another UX workshop to go over these small details such as what language to be used. We want to remain consistent, but promote probably has a better connotation than "installing" it. Thanks to @skofman1 for the suggestion!

JonDouglas: Top-level and transitive are things I don't believe we can change although the UX team would like to review them as they are "stop-and-think" words.

designs/Transitive-Dependencies.md
JonDouglas: Haha, I just wanted to make sure people were reminded of the goals, but more importantly just to blabber non-goals 👍

designs/Transitive-Dependencies.md
JonDouglas: Mind clarifying a bit more on "audit transitive packages automatically"? Just so I understand further!

designs/Transitive-Dependencies.md
JonDouglas: I knew there was something... I think this is a separate thing entirely and can remain in that issue 💯

designs/Transitive-Dependencies.md
JonDouglas: UX Board will review this, but I have been a broken record of using language everyone already knows and is documented.

https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files for example uses top-level & transitive but it also uses direct. 🕺

JonDouglas: https://docs.microsoft.com/en-us/nuget/what-is-nuget also uses top-level & down-level

designs/Transitive-Dependencies.md
JonDouglas: 👍

designs/Transitive-Dependencies.md
JonDouglas: I'm not quite sure. How does that work exactly? Do we have an example where a transitive-only situation happens? It has to be brought in by at least one top-level package right?

designs/Transitive-Dependencies.md
JonDouglas: Copy / pasted a template and never updated it.

designs/Transitive-Dependencies.md
nkolev92: `down-level`
🤕

designs/Transitive-Dependencies.md
nkolev92: Ah, I missed that.

designs/Transitive-Dependencies.md
drewgillies: Good point. Ultimately it comes from a top-level package. I was thinking of the scenario like showing package D's details here:
[only toplevel package]A->B
B->C
C->M
M->N
M->D
N->D
From what you're saying, you'd list A as the top-level package bringing in D, and that's it. I was reading your text and thinking: since D _isn't_ brought in directly by a top-level package (directly being the operative word), what would you list? So you'd go with A here. Makes sense. The way a customer would use this information is--if I get rid of A, I get rid of D. 👍

JonDouglas: Yep! Just showing the root of the package that ultimately brought the transitive dependency in. (could be many!)

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+- Drilling down the dependency tree
+- In the package detail view, some way to explore dependencies of dependencies and then their dependencies and so on, any level deep, without installing the package.
+
+
+## Cross-cutting Concerns
+
+-	The words “top-level” & “transitive” are words most people must stop & think about.
+    -	Should we use “direct” & “indirect”? Or “explicit” & “implicit”?
+-	Should a transitive dependency have the “installed” text given it’s already technically installed? Should it be called “promote” instead?
+- Given that the tooltip is not accessible for .NET Framework workloads, we will have to come up with an alternative experience or cut the tooltip experience until we have an accessible experience.
```
donnie-msft, line 125: Can the reasons why be elaborated? It may have been mentioned in a meeting, but I can't recall and think the details should be here.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Scenarios and User Experience
+
+### CLI
+
+We should consider a minor tweak to the existing experience of “dotnet list package --include-transitive” to provide the user with a sense of where the package came from. This could be a new column next to “Resolved” which says “Transitively Referenced” or “Referenced” and has a list of the top-level packages that requested the dependency.
+
+![](../resources/TransitiveDependencies/TransitiveDotNetCLI.png)
+
+### Visual Studio
```
donnie-msft, line 66: Accessibility should be called out. If we have any input from the a11y team, let's add it as well. At a minimum, note here that we will run checks in Accessibility Insights, and ensure screen-readers can differentiate Top/Transitive, as well as read the tooltip content.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will additionally be a header titled “Top-level Packages” where the current experience is today for top-level packages.
```
donnie-msft, line 68: Is this spec committing to this specific implementation within PM UI? If not, the wording should be changed as such. Also consider whether there will be an iterative approach to bringing transitives to the UI, and what those phases might look like.

I think the collapsing panel is a fine paradigm, but did we consider alternatives for tops vs transitives? 
Perhaps a filter option to "Show transitives" would work just as well? 
The description in https://github.com/NuGet/Home/issues/5887 shows a screenshot with a highlight around a package that's "installed explicitly"/top. This made me think that a visual indicator (color/etc) and screen-reader property may work just as well. And much like our `show prerelease`, they can be filtered out.

What do we gain from a collapsible panel? If there's always going to be on the order of N=10 top-levels, and N=100's of Transitives, it's hard to envision why I would want the collapsing behavior.

designs/Transitive-Dependencies.md
nkolev92: Currently with the top level only approach, the updates tab will have a label telling you the amount of packages that have updates (this is pretty bad perf right now). 

When deprecation was added, a warning icon was equivalent to the number in the Updates tab on the Installed tab.
This means that when we look-up the metadata for the packages installed to determine if they have any newer versions, we also look at the metadata to determine whether they are deprecated. 

Are we going to do the equivalent for transitive dependencies?
Are we going to preemptively look for the metadata for the transitive dependencies to determine if they are deprecated/vulnerable? 

In particular I'm asking this because given the way the server protocol is written right now, this could lead to 100s of different network calls.

designs/Transitive-Dependencies.md
JonDouglas: Yes this spec is committing to this implementation as there was really no suggested alternative & people generally agreed this was a good approach. If we don't feel that way today, we should reiterate.

This work by itself is phased from the package vulnerability work. Because of the scope of this work, there's really no phased approach to this specific design.

The reason there is a collapsing concept is to allow users to have their screen real estate back if they do not care for transitive dependencies and would like to keep them collapsed while doing top-level management tasks.

The decision for two separate lists was primarily driven by many different constructs such as the existing dotnet CLI behavior for showing transitives, and the many eco systems that show top-level dependencies & transitive dependencies in separate lists. (gradle, ant, etc). Others show dependency trees, which was not designed previously and thus we're committing to a similar experience already completed in dotnet.

The jobs to be done are to look at top-level, transitive level, or both. There are many reasons one may look at each level & having collapsable controls helps users decide how they'd like their screen to be shown. Otherwise yes, we could do a filter, but the purpose was to differentiate the types of references clearly and not have them all in one list with a UI affordance to say "transitive" or "top-level" although you might be able to filter/order by that.

drewgillies:
@JonDouglas @donnie-msft @nkolev92 Conversation seems to have ceased on this spec. Shall we merge it and call it adopted?

nkolev92:
I think there was some feedback from me & @donnie-msft that might need addressing. 
I was just waiting for a ping to re-read when that's done (re-review notification basically).

designs/Transitive-Dependencies.md
JonDouglas: Resolving via new commits.

designs/Transitive-Dependencies.md
JonDouglas: We are keeping the language top-level / transitive to be consistent. If UX board believes we should change it at a craftsman review, then we can reconsider.

designs/Transitive-Dependencies.md
JonDouglas: Updating via commit

designs/Transitive-Dependencies.md
JonDouglas: What is the perf potential of 100s of network calls?

You would see the same behavior for any installed package whether top-level or transitive.

Installed tab would show the current version & the available version.

Update tab would show the current version & the available version.

We would only count top-level packages for the badge on the tabs given we don't want users to promote to top-level just for a new version.

designs/Transitive-Dependencies.md
JonDouglas: 1. .NET Core has accessibility properties for Tooltip, .NET Framework supposedly doesn't.
2. Tooltip doesn't scale well with listing potentially hundreds of packages that may bring in the transitive dependency.
3. UX Board wanted to introduce a new concept, but hasn't scheduled a workshop yet.

designs/Transitive-Dependencies.md
JonDouglas: Accessibility concerns should've been called out regarding the tooltip aspect. We will do a full compatibility pass when we're nearing the craftsmanship review of the UX board. It's likely we will have to cut the tooltip given we have no alternatives & need to unblock package vulnerability work for now.

designs/Transitive-Dependencies.md
donnie-msft: Was resolved but I don't see the details added to the spec

designs/Transitive-Dependencies.md
donnie-msft: Was resolved but I don't see the details added to the spec. If there will be a technical spec coming after this, let me know.

JonDouglas: I didn't add any details to this spec as it's a design document. Although accessibility is considered & called out in this spec already, we don't have any insight or clarity into further accessibility concerns right now & those may need to be considered in the technical implementation as these are more test/implementation scenarios than design scenarios.

JonDouglas: /cc @drewgillies if you were planning to write-up anything for technical spec to make note of accessibility during implementation.

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+### Non-Goals
+
+-	Complete management of transitive dependencies.
+-	Reinvent common UX patterns used in Visual Studio today for package management actions.
+
+## Design
+
+### Elements
+
+There are three primary elements that will be added as part of this design to the existing Visual Studio Package Manager UI.
```
donnie-msft, line 95: The spec feels a bit more PM level, but now we're getting specific with XAML controls. Are these just high level concepts of what we think we're changing in the UI?
When you say ScrollViewer... you expect a new scrollbar so that Top level and Transitive level have their own individual scrollbars?

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
+1.	ScrollViewer – To show transitive dependency nodes.
+2.	Expander – To collapse transitive dependency nodes by default & expand upon user action. To collapse top-level dependency nodes.
+3.	Tooltip – To show the transitive dependency’s origin.
```
donnie-msft, line 99: Mentioning the context from the other thread regarding a11y concerns, etc, would be good here.
Actually since you say below that design team has not yet decided this, we should indicate it as Optional for now?

designs/Transitive-Dependencies.md
```diff
@@ -0,0 +1,150 @@
...
+
```
donnie-msft, line 100:
```suggestion
The above elements will be checked against Accessibility Insights. We'll ensure screen-readers can differentiate Top versus Transitive dependencies, and the scrolling and expansion functionality will be accessible with the keyboard.
```

donnie-msft (changes requested):

designs/Transitive-Dependencies.md
donnie-msft: Moved this concern to the Elements section.

designs/Transitive-Dependencies.md
JonDouglas: I have no concept of what controls should be used. I defer to the individual/implementation to use the best controls. These are just three elements I'm aware of to accomplish the design today.

designs/Transitive-Dependencies.md
JonDouglas: Thanks for the commit. This should be deferred to the implementation & the a11y concerns for whatever controls are ultimately used.

designs/Transitive-Dependencies.md
donnie-msft: Ok, then I would think we shouldn't list them in the spec if they are being deferred.

donnie-msft (approved):
Looks good. Taking the specific implementation notes with a grain of salt. Suggested removing them as they seem to be authoritative.

designs/Transitive-Dependencies.md
nkolev92: We usually throttle based on the CPUs. 
The impact would be delayed population of the information in the UI. 

The number of network calls NuGet makes to load the information needed for UI is a problem today for the solution tab where the package count is large. A rough guestimate would be that the project level PM UI now needs as much information as the solution level one. 

If we don't want to shown badges for top level, then these concerns go away :) 

Note that these perf problems should be somewhat solvable, it's just a resourcing concern that'd likely take more effort than just displaying the transitive dependencies :)

nkolev92 (approved):
@donnie-msft brought up some good concerns, but beyond that + the thread I just replied on looks good!

designs/Transitive-Dependencies.md
JonDouglas: It's also completely fine to just not show updates available for transitive either. It would be a nice parity thing to reuse the current experience to show if a resolved transitive has an update with the hover experience. If it's complex perf wise, we can cut it.

designs/Transitive-Dependencies.md
JonDouglas: Added a sentence to help distinguish!

designs/Transitive-Dependencies.md
nkolev92: I think that's perfectly reasonable. 

I basically want us to make progress on the basics in this area without having to worry about perf implications now. 

If we want to have full parity later, we'll judge it on it's own merit.

drewgillies: The whole point of implementing transitive dependencies is for the metadata, right? Don't we want to see which transitive dependencies are vulnerable?

drewgillies: My understanding at this time, is that we're chiefly implementing transitive dependencies in PMUI *specifically* for the metadata, right? Isn't the North Star here being able to see which transitive packages have vulnerabilities?

drewgillies: (I do know that's not a part of this spec, but since it's come up in conversation, and we want a design that supports these future features, let's discuss it)

JonDouglas: I'm deferring to you both regarding what the complexities would be. The goal is to show transitive dependencies enough to unblock package vulnerability while keeping a consistent experience that we have today  🏃

nkolev92: The `perf` concern I have is if we aggressively fetch the metadata. 

ie, do we fetch metadata for transitive dependencies if the transitive dependencies are collapsed?

Right now, if you go to the browse tab, we'll fetch the metadata for all packages on that view...just so we can show the label.

drewgillies: I'm looking to take a step back and consider the intention driving this feature, and whether we're still achieving this objective by not fetching metadata for a collapsed list. If a customer is viewing the installed packages for their project and sees a collapsed transitive dependencies list, and we haven't yet parsed vulnerability or deprecation metadata for the dependency graph, the customer may be misled into thinking the project's only vulnerabilities (for example) are in the top-level packages, which brings us back to the product's trustworthiness. 

The way I see it, there are 3 paths forward:
- Leave transitive dependency metadata unfetched, transitive dependencies collapsed, therefore incomplete vulnerability/deprecation data for the customer (until the list is expanded). We could add a disclaimer on the collapsed list indicating that expanding the list is the only way to get a complete picture of project deprecation/vulnerability status, and that this may take some time to fetch. This will however give an inaccurate label on the tab if we wish to indicate vulnerability/deprecation status there.

- Fetch all the data for the graph up-front and take the perf hit.

- Rearchitect our services approach to provide a more performant (complete) picture up front.

JonDouglas: I'm going to merge this spec as per other approvals & we'll take up these questions at implementation or technical spec/design time.

PR #10258, opened 2020-11-12
Improving the experience when incompatible versions of the restore/build tooling are used

nkolev92:
Design for https://github.com/NuGet/Home/issues/10259.

### DIFF
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -12,7 +12,7 @@
 NuGet as a product has grown its importance in the .NET ecosystem over the years.
 What started with NuGet.exe & Visual Studio support has led to a first class experience in **the CLI**, dotnet.exe.
 
-With packages.config, NuGet was built on top of the build experience, NuGet functions as an add on.
+With packages.config, NuGet was built on top of the build experience, and functions as an add on.
 NuGet used concept such as `Reference` to add package dependencies. These references were written in the project file directly, with a relative path to a `known` packages folder.
 This meant that you can use NuGet.exe 3.4.4 with VS 2015 & VS 2013 as the build tooling.
 
@@ -46,7 +46,7 @@ On the CI, the story is a little bit different.
 
 #### Tooling options
 
-Starting with VS 2017 and now VS 2019, there 3 different tooling options on can use to run a restore on their CI.
+Starting with VS 2017 and now VS 2019, there 3 different tooling options one can use to run a restore on their CI.
 
 * MSBuild.exe
   * Ships with Visual Studio
@@ -68,7 +68,7 @@ Starting with VS 2017 and now VS 2019, there 3 different tooling options on can 
 
 GitHub Actions and Azure Pipelines are the most commonly used (and Microsoft run) automation services.
 Usually customers will have a `yaml build file` that specifies the list of steps to run, and the machines with the build tooling are often provided by these services. For example, Azure pipelines provides [Microsoft-hosted agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml), which handle maintenance and upgrade of these agents.
-In that maintenance, VS upgrades are on of those included things.
+In that maintenance, VS upgrades are one of those included things.
 
 Given all those tooling options, there are a few *common* ways customers build their repos.
 
@@ -221,4 +221,3 @@ Furthemore this increases the testing and support matrix significantly. It is no
 * Forward compatibility case [issue](https://developercommunity2.visualstudio.com/t/vsbuild-msbuild-task-failing-since-2020-11-10/1249432).
 * [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions)
 * Note that the complete implementation of the aliases feature in [NuGet/Home#5154](https://github.com/nuget/home/issues/5154) will definitely cause a problem if mismatched tooling is used. While we'd like the scenarios to work as often as possible, certain features do require big changes.
-
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -156,7 +156,7 @@ Given that the implied compatibility exists, the proposal is for the SDK build t
 * Warn when the restore output is from an older tool than anticipated.
 * Warn when the restore output is from a newer tool than anticipated.
 
-There are 2 technical options and each might have a different effect on the frequency in which this warning is raised.
+There are 3 technical options and each might have a different effect on the frequency in which this warning is raised.
 
 #### Option 1 - Use assets file version
 
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -78,47 +78,97 @@ In PackageReference, there's a nuance, because now you have a restore and build 
 
 Naturally this leads to 2 potential different problems.
 
-#### Old NuGet tooling, newer build tooling
+#### Old NuGet tooling, newer build tooling - NuGet forward compatibility
 
 In recent years, NuGet restore has added many capabilities to support .NET Core. From FrameworkReference in 5.3 to support for .NET 5 TFMs in 5.8.
 In *most* situations, using an older version of NuGet.exe *will not* cause any problems at build time. 5.8 and .NET 5 in particular are not such versions.
+[NuGet/Home#9756](https://github.com/NuGet/Home/issues/9756) introduces a change that was needed to support .NET 5 tfms and that change in itself required a significant change to how the build tasks interpret the `project.assets.json`.
 An example is the following Developer Community [issue](https://developercommunity2.visualstudio.com/t/NuGet-Restore-build-issues---projectass/1190427).
 
-#### New NuGet tooling, old build tooling
+#### New NuGet tooling, old build tooling - NuGet backward compatibility
+
+A newer NuGet version *might* generate an assets file that has more/different details than what the build tasks expect.
+A recent example is the following Developer Community [issue](https://developercommunity2.visualstudio.com/t/vsbuild-msbuild-task-failing-since-2020-11-10/1249432) which is not 100% understood yet.
+
+In this design, we are trying to improve the general mismatched tooling experience, but also consider alternatives for the particular issues customers are hitting with VS 16.8/NuGet 5.8/ SDK 5.0.100.
 
 ## Who are the customers
 
-Customers affected by this change.
+PackageReference customers
 
 ## Goals
 
-Specifically call out the goals for this proposed design.
+* Improve the experience for mismatched versions of restore/build tooling.
 
 ## Non-Goals
 
-Call out things that are not in consideration for this design. 
-
 ## Solution Overview
 
-Describe the proposed solution in as much detail as you deem necessary.
+In the 2 particular problems described above the experience when things go wrong is inconsistent. It's not *always* apparent why something goes wrong.
+This proposal covers improving that warning/error experience.
+
+### NuGet output versioning
+
+When PackageReference restore is run, NuGet writes out the 3 files mentioned above. 2 of those files have a `version`.
+
+* The `project.assets.json` has a version that has historically been used for `reading`, not tooling compatibility checks.
+* The `nuget.g.props` contains a `NuGetToolVersion` property.
+
+### Build Tasks warning (SDK)
+
+Given that the implied compatibility exists, the proposal is for the SDK build tasks raise a suppressible warning when the output is not in the expected shape.
+
+* Warn when the restore output is from an older tool than anticipated.
+* Warn when the restore output is from a newer tool than anticipated.
+
+There are 2 technical options and each might have a different effect on the frequency in which this warning is raised.
+
+#### Option 1 - Use assets file version
+
+NuGet increments the `project.assets.json` version every time there's a functional change in behavior. The build tasks in the SDK are built with a specific version of NuGet, and the SDK could raise a warning when the `project.assets.json` version is not the one it expects.
+
+Cons:
+
+* This version was not used with tool compatibility in mind, but rather  While unlikely, it's possible that other tooling depends on it.
+
+Pro:
+
+* We would only increment this when a functional restore change is made.
+
+Alternatively we could add a new version property, but that's probably unnecessary.
+
+#### Option 2 - Use NuGet tool version
+
+The nuget.g.props write out a `NuGetToolVersion` property which contains the version of the NuGet tooling that generated it.
+
+Cons:
+
+* This has the potential to be too noisy. Just because the version change that doesn't mean it's not compatible.
+
+Pros:
+
+* The flip side is that we prefer that customers use consistent versions of the tooling, so going this direction has the added benefit of consolidating tooling for our customers.
 
 ## Test Strategy
 
-Other than Unit Tests, which automated and manual integration, component, and end-to-end tests will be needed to validate the implementation? Is a special roll-out strategy (including kill-switch, opt-in flag, etc.) needed?
+These changes require integration test slightly different than the ones run today in both SDK and NuGet repo.
+The SDK is already considering adding tests for the cross version scenarios in the following [PR](https://github.com/dotnet/sdk/pull/14517).
 
 ## Future Work
 
-Declare any future work if applicable.
+None
 
 ## Open Questions
 
-Any open questions not specifically called out in the design.
-
 ## Considerations
 
-Any other solutions or approaches considered? 
-The rejected designs are often just as important as the proposed ones!
+* Should NuGet restore output always be backward/forward compatible?
+
+No. The build tooling shipped today, which NuGet is a part of, is self contained. NuGet.exe customers are the only ones likely to run this mismatched scenario and we have done work to help them migrate, through MSBuild.exe restore and packages.config support for restore.
+Furthemore this increases the testing and support matrix significantly. It is not something we have promised customers.
 
 ### References
 
 * NuGet restore build [issues](https://developercommunity2.visualstudio.com/t/NuGet-Restore-build-issues---projectass/1190427)
+* Making SDK tooling compatible with assets file generated by 5.7, [PR](https://github.com/dotnet/sdk/pull/14517).
+* Forward compatibility case [issue](https://developercommunity2.visualstudio.com/t/vsbuild-msbuild-task-failing-since-2020-11-10/1249432) 
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -3,7 +3,7 @@
 
 * Status: In Review
 * Author(s): [Nikolche Kolev](https://github.com/nkolev92)
-* Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
+* Issue: [TBD](https://github.com/NuGet/Home/issues/) No GH issue
 
 ## Problem Background
 
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -51,7 +51,8 @@ Starting with VS 2017 and now VS 2019, there 3 different tooling options one can
 * MSBuild.exe
   * Ships with Visual Studio
   * Supports PackageReference
-  * Starting with VS 2019, Update 5, packages.config scenarios in an opt-in fashion through -p:RestorePackagesConfig=true`
+  * Starting with VS 2019, Update 5, packages.config scenarios in an opt-in fashion through -p:RestorePackagesConfig=true`.
+    * We recommend you set `RestorePackagesConfig` to `true` in a [Directory.Build.Props](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets).
 
 * dotnet.exe
   * Ships with the .NET SDK
--- a/designs/RestoreAndBuildCompatibility.md
+++ b/designs/RestoreAndBuildCompatibility.md
@@ -1,5 +1,5 @@
 
-# Spec Name
+# Improving the restore and build compatibility experience
 
 * Status: In Review
 * Author(s): [Nikolche Kolev](https://github.com/nkolev92)
@@ -9,16 +9,16 @@
 
 ### NuGet packages.config
 
-NuGet as a product has grown it's importance in the .NET ecosystem over the years.
-What started with nuget.exe & Visual Studio support has led to a first class experience in **the CLI**, dotnet.exe.
+NuGet as a product has grown its importance in the .NET ecosystem over the years.
+What started with NuGet.exe & Visual Studio support has led to a first class experience in **the CLI**, dotnet.exe.
 
-With packages.config, NuGet was built on top of the build experience, NuGet functioned as an add on.
+With packages.config, NuGet was built on top of the build experience, NuGet functions as an add on.
 NuGet used concept such as `Reference` to add package dependencies. These references were written in the project file directly, with a relative path to a `known` packages folder.
 This meant that you can use NuGet.exe 3.4.4 with VS 2015 & VS 2013 as the build tooling.
 
-On the CI, the workflow usually is:
+The workflow for that usually is:
 
-* nuget.exe restore - which downloads a defined list of packages to a specific folder.
+* NuGet.exe restore - which downloads a defined list of packages to a specific folder.
 * build - with your tooling of choice. Said tooling doesn't need to be aware of the fact that NuGet is involved in any way.
 
 ### NuGet PackageReference
@@ -27,7 +27,7 @@ Since the introduction of project.json and PackageReference (2015 & 2017 respect
 Today NuGet's PackageReference is the tooling of choice for *all* .NET Core, now one unified .NET, projects.
 
 With PackageReference projects a restore is now *a must*.
-In addition to `nuget.exe`, you can restore PackageReference based projects with `msbuild /t:restore` and `dotnet.exe restore`.
+In addition to `NuGet.exe`, you can restore PackageReference based projects with `msbuild /t:restore` and `dotnet.exe restore`.
 
 The `restore` and `build` operations have a contract expressed through 3 files:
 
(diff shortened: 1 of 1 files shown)


nkolev92:
cc @marcpopMSFT, @dsplaisted

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
+
+# Spec Name
```
zivkan: spec name

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+
+The additional tooling options, the combination of project types in one repository now means that there are many ways people run the restore/build gestures on CI.
+
+1. dotnet.exe
+With dotnet.exe, the CI commands would normally look like one of the below options: 
+
+    * dotnet.exe restore & dotnet.exe build --no-restore
+    * dotnet.exe build (restore is implied)
+
+This is the recommended and full proof way if you have only projects that are supported by the [`.NET SDK`](https://github.com/dotnet/sdk).
```
zivkan:
```suggestion
This is the recommended and foolproof way if you have only projects that are supported by the [`.NET SDK`](https://github.com/dotnet/sdk).
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+dotnet.exe and it's respective SDK are a self-contained build tool. The restore and build contract is 100% understood and there are no compatibility concerns.
+
+2. MSBuild.exe
+
+With MSBuild.exe, the CI commands would normally look like one of the below options:
+
+* MSBuild.exe /t:restore & MSBuild.exe /t:build
+* MSBuild.exe /restore /t:build
+
+MSBuild.exe has had PackageReference restore support since VS 2017. Starting with VS 2019, update 5, MSBuild now has project specific packages.config support.
```
zivkan:
```suggestion
MSBuild.exe has had PackageReference restore support since VS 2017. Starting with VS 2019, update 5, MSBuild now has project specific packages.config support using `-p:RestorePackagesConfig=true`.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+Naturally this leads to 2 potential different problems.
+
+#### Old NuGet tooling, newer build tooling - NuGet forward compatibility
+
+In recent years, NuGet restore has added many capabilities to support .NET Core. From FrameworkReference in 5.3 to support for .NET 5 TFMs in 5.8.
+In *most* situations, using an older version of NuGet.exe *will not* cause any problems at build time. 5.8 and .NET 5 in particular are not such versions.
+[NuGet/Home#9756](https://github.com/NuGet/Home/issues/9756) introduces a change that was needed to support .NET 5 tfms and that change in itself required a significant change to how the build tasks interpret the `project.assets.json`.
+An example is the following Developer Community [issue](https://developercommunity2.visualstudio.com/t/NuGet-Restore-build-issues---projectass/1190427).
+
+#### New NuGet tooling, old build tooling - NuGet backward compatibility
```
zivkan: Is it worthwhile mentioning in either this backwards compat or the previous forwards compat sections, that the "full tfm alias" feature is probably going to need to break both backwards and forwards compat?

maybe this means we can't implement it until VS 17.0/.net 6

zivkan (commented):
I've only read half the document so far, but want to submit the feedback I have already.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+2. MSBuild.exe
+
+With MSBuild.exe, the CI commands would normally look like one of the below options:
+
+* MSBuild.exe /t:restore & MSBuild.exe /t:build
+* MSBuild.exe /restore /t:build
+
+MSBuild.exe has had PackageReference restore support since VS 2017. Starting with VS 2019, update 5, MSBuild now has project specific packages.config support.
+
+Using MSBuild.exe to restore and build the repo means that the restore and build contract is 100% understood.
```
rainersigwald:
```suggestion
Using MSBuild.exe to restore and build the repo means that the restore and build contract is 100% understood because the same tooling version is used for both operations.

```

designs/RestoreAndBuildCompatibility.md
rainersigwald: And maybe link to doc on `RestorePackagesConfig`?

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+1. dotnet.exe
+With dotnet.exe, the CI commands would normally look like one of the below options: 
+
+    * dotnet.exe restore & dotnet.exe build --no-restore
+    * dotnet.exe build (restore is implied)
+
+This is the recommended and full proof way if you have only projects that are supported by the [`.NET SDK`](https://github.com/dotnet/sdk).
+packages.config projects *are not* supported with dotnet.exe.
+
+dotnet.exe and it's respective SDK are a self-contained build tool. The restore and build contract is 100% understood and there are no compatibility concerns.
```
rainersigwald:
```suggestion
dotnet.exe and its respective SDK are a self-contained build tool. The restore and build contract is 100% understood and there are no compatibility concerns.

```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+* MSBuild.exe /restore /t:build
+
+MSBuild.exe has had PackageReference restore support since VS 2017. Starting with VS 2019, update 5, MSBuild now has project specific packages.config support.
+
+Using MSBuild.exe to restore and build the repo means that the restore and build contract is 100% understood.
+
+3. NuGet.exe
+
+NuGet.exe is the oldest tool that can be used to restore on CIs.
+It supports PackageReference and packages.config, but obviously it cannot build, so one still needs to build, usually using MSBuild.exe.
```
rainersigwald: nit: usually best to avoid "obviously"

```suggestion
It supports PackageReference and packages.config, but it cannot build, so one still needs to build, usually using MSBuild.exe.

```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+So the steps are usually:
+
+* NuGet.exe restore
+* MSBuild.exe /t:build
+
+In packages.config, this usually just works because of packages.config's architecture.
```
rainersigwald:
```suggestion
In packages.config, this usually just works because the build tooling can be oblivious to NuGet.

```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+In PackageReference, there's a nuance, because now you have a restore and build steps that are from potential different versions of the tooling.
```
rainersigwald:
```suggestion
In PackageReference, there's nuance, because now you have a restore and build steps that can use different versions of the tooling--`nuget.exe` will write assets that may differ from what the MSBuild-embedded tooling expects.

```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+
+* NuGet.exe restore
+* MSBuild.exe /t:build
+
+In packages.config, this usually just works because of packages.config's architecture.
+In PackageReference, there's a nuance, because now you have a restore and build steps that are from potential different versions of the tooling.  
+
+Naturally this leads to 2 potential different problems.
+
+#### Old NuGet tooling, newer build tooling - NuGet forward compatibility
```
rainersigwald: I think of this more as "build tooling backward compatibility" rather than "NuGet forward compat". It's not a promise that the outputs won't change, but a promise that new tooling will be able to understand older outputs.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+
+In recent years, NuGet restore has added many capabilities to support .NET Core. From FrameworkReference in 5.3 to support for .NET 5 TFMs in 5.8.
+In *most* situations, using an older version of NuGet.exe *will not* cause any problems at build time. 5.8 and .NET 5 in particular are not such versions.
```
rainersigwald:
```suggestion
In *most* situations, using an older version of NuGet.exe *will not* cause any problems at build time. 5.8 and .NET SDK 5.0.100 in particular _do_ cause problems.

```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+
+There are 2 technical options and each might have a different effect on the frequency in which this warning is raised.
+
+#### Option 1 - Use assets file version
+
+NuGet increments the `project.assets.json` version every time there's a functional change in behavior. The build tasks in the SDK are built with a specific version of NuGet, and the SDK could raise a warning when the `project.assets.json` version is not the one it expects.
+
+Cons:
+
+* This version was not used with tool compatibility in mind, but rather  While unlikely, it's possible that other tooling depends on it.
```
rainersigwald: "but rather" what? The "other tooling" thing makes me think there's a missing thought here.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+With MSBuild.exe, the CI commands would normally look like one of the below options:
+
+* MSBuild.exe /t:restore & MSBuild.exe /t:build
+* MSBuild.exe /restore /t:build
+
+MSBuild.exe has had PackageReference restore support since VS 2017. Starting with VS 2019, update 5, MSBuild now has project specific packages.config support.
+
+Using MSBuild.exe to restore and build the repo means that the restore and build contract is 100% understood.
+
+3. NuGet.exe
```
rainersigwald, line 94: I think it's worth a blurb here or elsewhere in this document about the tool-acquisition experience, since that's a major factor in breaks here.

Specifically I'd like to see mention of:

* `NuGet.exe` isn't bundled with the .NET SDK or Visual Studio.
* CI systems (Actions, Pipelines) often have an acquisition experience that allows specifying distinct/non-matching versions for NuGet and the SDK.
* CI system images can silently update Visual Studio and/or the SDK, causing mismatched versions _without user error_.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,179 @@
...
+The nuget.g.props write out a `NuGetToolVersion` property which contains the version of the NuGet tooling that generated it.
+
+Cons:
+
+* This has the potential to be too noisy. Just because the version change that doesn't mean it's not compatible.
+
+Pros:
+
+* The flip side is that we prefer that customers use consistent versions of the tooling, so going this direction has the added benefit of consolidating tooling for our customers.
+
```
rainersigwald, line 222: Is there an Option 3 to introduce a new version field somewhere, and potentially follow semver for it?

rainersigwald (commented):
I think getting a clear error when possible is definitely worth pursuing. I'd really like to emphasize the scenarios where the user didn't do anything even slightly wrong, though, to push toward "can we make the tooling backward compatible?"

designs/RestoreAndBuildCompatibility.md
nkolev92: It's very unfortunate that the `RestorePackagesConfig` switch is not documented. 
No better time than the present though.  https://github.com/NuGet/docs.microsoft.com-nuget/issues/1944

designs/RestoreAndBuildCompatibility.md
nkolev92: Great point. I have reworked the docs and added some extra details. 

My idea with this was to eventually use it as a reference doc to write good practice docs/reference docs and this is turning out even better :)

designs/RestoreAndBuildCompatibility.md
nkolev92: I get that :) Given that backward/forward compat is relative to the tooling, I'm remove the mentions and keep it as simple as `older` vs `newer`

designs/RestoreAndBuildCompatibility.md
nkolev92: I'll add it in references.

designs/RestoreAndBuildCompatibility.md
nkolev92: Added

designs/RestoreAndBuildCompatibility.md
nkolev92: Yep, great catch. Fixed.

nkolev92:
Thanks @zivkan and @rainersigwald!

I have addressed your feedback. It should hopefully be cleaner and more self-contained now.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+* Issue: [10259](https://github.com/NuGet/Home/issues/10259) Improving the experience when incompatible versions of the restore/build tooling are used
+
+## Problem Background
+
+### NuGet packages.config
+
+NuGet as a product has grown its importance in the .NET ecosystem over the years.
+What started with NuGet.exe & Visual Studio support has led to a first class experience in **the CLI**, dotnet.exe.
+
+With packages.config, NuGet was built on top of the build experience, NuGet functions as an add on.
```
dsplaisted:
```suggestion
With packages.config, NuGet was built on top of the build experience, and functions as an add on.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Restore, Build and Continue Integration
+
+When developing in Visual Studio or any other editor, the restore and build steps are integrated and things normally just work.  
+
+On the CI, the story is a little bit different.
+
+#### Tooling options
+
+Starting with VS 2017 and now VS 2019, there 3 different tooling options on can use to run a restore on their CI.
```
dsplaisted:
```suggestion
Starting with VS 2017 and now VS 2019, there 3 different tooling options one can use to run a restore on their CI.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+* MSBuild.exe
+  * Ships with Visual Studio
+  * Supports PackageReference
+  * Starting with VS 2019, Update 5, packages.config scenarios in an opt-in fashion through -p:RestorePackagesConfig=true`
```
dsplaisted: I would recommend putting RestorePackagesConfig in a Directory.Build.props file instead of having to pass it on the command line.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+* NuGet.exe
+  * NuGet.exe is the oldest tool that can be used to restore on CIs.
+  * NuGet.exe is *not* bundled with Visual Studio or dotnet.exe, as they both carry their own functionality.
+  * NuGet.exe supports packages.config & PackageReference
+
+#### Continious integration considerations
+
+GitHub Actions and Azure Pipelines are the most commonly used (and Microsoft run) automation services.
+Usually customers will have a `yaml build file` that specifies the list of steps to run, and the machines with the build tooling are often provided by these services. For example, Azure pipelines provides [Microsoft-hosted agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml), which handle maintenance and upgrade of these agents.
+In that maintenance, VS upgrades are on of those included things.
```
dsplaisted:
```suggestion
In that maintenance, VS upgrades are one of those included things.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+* Should the warnings/errors that the SDK raises could be context aware?
+For example, if the versions are missmatched, but the SDK is still able to find the assets, it can do a best effort and raise a warning. If the assets cannot be found, fail with an error *clearly* indicating the problem is the mismatched tooling versions.
+
+## Considerations
+
+* Should NuGet restore output always be backward/forward compatible?
+
+No. The build tooling shipped today, which NuGet is a part of, is self contained. NuGet.exe customers are the only ones likely to run this mismatched scenario and we have done work to help them migrate, through MSBuild.exe restore and packages.config support for restore.
+Furthemore this increases the testing and support matrix significantly. It is not something we have promised customers.
```
dsplaisted:
```suggestion
Furthermore this increases the testing and support matrix significantly. It is not something we have promised customers.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+* Status: In Review
+* Author(s): [Nikolche Kolev](https://github.com/nkolev92)
+* Issue: [10259](https://github.com/NuGet/Home/issues/10259) Improving the experience when incompatible versions of the restore/build tooling are used
+
+## Problem Background
+
+### NuGet packages.config
+
+NuGet as a product has grown its importance in the .NET ecosystem over the years.
+What started with NuGet.exe & Visual Studio support has led to a first class experience in **the CLI**, dotnet.exe.
```
zivkan, line 11: These two sentences can be moved to above the packages.config heading. They're generic nuget background info unrelated to packages.config. Particularly mentioning first class experience in dotnet.exe under a packages.config heading is... unexpected.

designs/RestoreAndBuildCompatibility.md
nkolev92: added.

nkolev92:
Addressed the feedback. 

@dsplaisted @rainersigwald @zivkan 

Any preference as far the approaches for warning go? 

cc @aortiz-msft

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+The `restore` and `build` operations have a contract expressed through 3 files:
+
+* `project.assets.json`
+* `projectName.ext.nuget.g.props`
+* `projectName.ext.nuget.g.targets`
+
+At build-time these files are loaded and interpreted as necessary.
+This means that the NuGet and build tooling version have an implied compatibility matrix.
+
+### Restore, Build and Continue Integration
```
zivkan:
```suggestion
### Restore, Build and Continuous Integration
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+On the CI, the story is a little bit different.
+
+#### Tooling options
+
+Starting with VS 2017 and now VS 2019, there 3 different tooling options one can use to run a restore on their CI.
+
+* MSBuild.exe
+  * Ships with Visual Studio
+  * Supports PackageReference
+  * Starting with VS 2019, Update 5, packages.config scenarios in an opt-in fashion through -p:RestorePackagesConfig=true`.
```
zivkan:
```suggestion
  * Starting with VS 2019, Update 5, packages.config scenarios in an opt-in fashion through `-p:RestorePackagesConfig=true`.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+    * We recommend you set `RestorePackagesConfig` to `true` in a [Directory.Build.Props](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#directorybuildprops-and-directorybuildtargets).
```
zivkan: we do? I know it's necessary for non-SDK style projects without any package references to get transitive packages (and I guess to avoid accidentally introducing an packages.config project). But it's completely unnecessary for a solution with only SDK style projects.

I know it's easier to say "always do this" rather than explain the nuance, but personally I never considered this as general advice.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+* dotnet.exe
+  * Ships with the .NET SDK
+    * Bundled in Visual Studio
```
zivkan:
```suggestion
    * Bundled in Visual Studio with the Cross Platform .NET Development workload
```

I don't know if it's now bundled with other .NET workloads, but in the past it was only installed if the .net core workload was explicitly selected by the customer.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+3. NuGet.exe
+
+NuGet.exe supports restore, but it cannot build, so one still needs to build, usually using MSBuild.exe.
+The steps are normally:
+
+* Acquire NuGet.exe
+* NuGet.exe restore
+* MSBuild.exe /t:build
+
+In packages.config, this usually just works because he build tooling can be oblivious to NuGet.
```
zivkan:
```suggestion
In packages.config, this usually just works because the build tooling is oblivious to NuGet.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+## Solution Overview
+
+In the 2 particular problems described above the experience when things go wrong is inconsistent. It's not *always* apparent why something goes wrong.
+This proposal covers improving that warning/error experience.
+
+### NuGet output versioning
+
+When PackageReference restore is run, NuGet writes out the 3 files mentioned above. 2 of those files have a `version`.
+
+* The `project.assets.json` has a version that has historically been used for `reading`, not tooling compatibility checks.
```
zivkan: I don't really understand what this means. If it wasn't used for compat, how was it used "for reading"?

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+There are 2 technical options and each might have a different effect on the frequency in which this warning is raised.
+
+#### Option 1 - Use assets file version
+
+NuGet increments the `project.assets.json` version every time there's a functional change in behavior. The build tasks in the SDK are built with a specific version of NuGet, and the SDK could raise a warning when the `project.assets.json` version is not the one it expects.
+
+Cons:
+
+* This version was not used with tooling/restore meaning compatibility in mind, but rather for writing/reading purposes. While unlikely, it's possible that other tooling depends on it.
```
zivkan: I don't know if this spec is designed to contain customer issues only. From an implementation point of view, it makes changes/insertions harder, because when we change NuGet and increment the assets file version, all the .NET SDK tests will fail, so we'll be blocked from insertions until the SDK is also updated to support it. Even when the assets file change is otherwise backwards compatible.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+#### Option 3 - Define a new compatibility version
+
+Pros:
+
+* We would only increment this when a functional restore change is made.
+
+Cons:
+
+* Both NuGet and SDK tooling require changes to add the initial warning.
```
zivkan: Like the existing assets file version number proposal, this makes any assets file change difficult to insert.

designs/RestoreAndBuildCompatibility.md
nkolev92: welp, don't know how that happened, lol.

designs/RestoreAndBuildCompatibility.md
nkolev92: I can do a consider. 

If you have packages.config project you want to restore with msbuild, I'd say DBP is recommended. 

Either way,that's not the point of this spec, so don't want to dwell too much :)

designs/RestoreAndBuildCompatibility.md
nkolev92: The version was incremented with schema changes. 

I fixed it. 

Note that it's been inconsistenly used, so we can change it to mean whatever we need it to mean if we want to :)

designs/RestoreAndBuildCompatibility.md
nkolev92: It should be a suppressible warning raised at `build time`. 
It should only be an error if the reference resolution cannot succeed. 

Does that address your concern?
SDK side tests should be *easy* to fix, because they'll depend on `LockFileFormat.Version` or `LatestVersion`, https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileFormat.cs#L21, so it'll be `fixed` by building.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+Say a customer has configured to get latest NuGet in their build tasks. While NuGet.exe and VisualStudio do ship on the same day, the rollout to all build machines is not atomic. The customers could end up in this mismatched situation without an explicit action.
+
+A newer NuGet version *might* generate an assets file that has more/different details than what the build tasks expect.
+A recent example is the following Developer Community [issue](https://developercommunity2.visualstudio.com/t/vsbuild-msbuild-task-failing-since-2020-11-10/1249432) which is not 100% understood yet.
+
+In this design, we are trying to improve the general mismatched tooling experience, but also consider alternatives for the particular issues customers are hitting with VS 16.8/NuGet 5.8/ SDK 5.0.100.
+
+## Who are the customers
+
+PackageReference customers
```
zkat:
```suggestion
PackageReference customers using NuGet.exe to restore separately from build tooling.
```

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+* The `nuget.g.props` contains a `NuGetToolVersion` property.
+
+### Build Tasks warning (SDK)
+
+Given that the implied compatibility exists, the proposal is for the SDK build tasks raise a suppressible warning when the output is not in the expected shape.
+
+* Warn when the restore output is from an older tool than anticipated.
+* Warn when the restore output is from a newer tool than anticipated.
+
+There are 2 technical options and each might have a different effect on the frequency in which this warning is raised.
```
zkat:
```suggestion
There are 3 technical options and each might have a different effect on the frequency in which this warning is raised.
```

zkat: Also, I feel like this document should pick one before being merged. Is that the intention?

designs/RestoreAndBuildCompatibility.md
```diff
@@ -0,0 +1,224 @@
...
+
+NuGet increments the `project.assets.json` version every time there's a functional change in behavior. The build tasks in the SDK are built with a specific version of NuGet, and the SDK could raise a warning when the `project.assets.json` version is not the one it expects.
+
+Cons:
+
+* Thas version that has historically been used for schema changes, not tooling compatibility checks. For example, aliases addition involved adding more information to the assets file, but not a schema change so this number was not incremented.
+
+Pros:
+
+* We would only increment this when a functional restore change is made.
```
zkat: Is a potential pro here that we would _already_ get a warning, for existing past versions of tooling?

zkat (commented):
I think I'd really like the decision between options to be made before this is merged.

designs/RestoreAndBuildCompatibility.md
nkolev92: Yes, my plan is to get opinions/ideas today & tomorrow, and decide by Monday (11/16/20) latest.

designs/RestoreAndBuildCompatibility.md
dsplaisted: Yeah, it was more a comment for the documentation that's going to be added for the property.

designs/RestoreAndBuildCompatibility.md
nkolev92: I think all of the approaches allow us to do that. This one might be the `hardest` to get a warning I think.

1. SDK change would need to look for `4`, since all the previous versions of the assets file wrote `3`. So we need a NuGet + SDK change.
1. SDK change only. The SDK can known the NuGet version it carries and react appropriately.
1. SDK change would need to look at a new property. Alternatively if the parser doesn't see that new property, the SDK could raise a warning. 

I could also be totally misunderstanding you :D

zkat: No, I think you got it. I was under the impression some kind of warning already existed for bumped version numbers.

nkolev92: Ahh...nope, unfortunately :(

designs/RestoreAndBuildCompatibility.md
nkolev92: I fixed the docs. I'll eventually update this spec if the docs make it to live sooner rather than later, but ideally this information leads to some more official docs.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -156,23 +175,38 @@ Given that the implied compatibility exists, the proposal is for the SDK build t
...
 
-There are 3 technical options and each might have a different effect on the frequency in which this warning is raised.
-
-#### Option 1 - Use assets file version
+### Assets file version
 
 NuGet increments the `project.assets.json` version every time there's a functional change in behavior. The build tasks in the SDK are built with a specific version of NuGet, and the SDK could raise a warning when the `project.assets.json` version is not the one it expects.
 
-Cons:
+This version that has historically been used for schema changes, not tooling compatibility checks. For example, aliases addition involved adding more information to the assets file, but not a schema change so this number was not incremented.
```
zivkan: Adding new properties into a JSON structure is a schema change, just as adding a new column to a SQL database table is a schema change. The change might be backwards compatible for most customers, and by not incrementing the schema version number, customers affected customers will get a compile error that doesn't make a lot of sense, rather than a clear "use a newer version of nuget" message.

For examples like package/assembly aliases, I think we do need to increment the schema number, but if we're concerned about forcing customers using old nuget.exe having to upgrade when it would otherwise work, then I think NuGet needs to do "feature detection" and when the feature (alias in this example) is not used, write the lower version number, and when the feature is used, write the higher version number. The .NET SDK will need to support both version numbers. But this is the only way I can see customers getting a good, actionable error message, rather than a bizzare compile time error.

However, the effort and complexity of implementing this may not be worthwhile compared to a strong push to get customers to move to msbuild and dotnet cli restore.

designs/RestoreAndBuildCompatibility.md
```diff
@@ -156,23 +175,38 @@ Given that the implied compatibility exists, the proposal is for the SDK build t
...
 
-* Thas version that has historically been used for schema changes, not tooling compatibility checks. For example, aliases addition involved adding more information to the assets file, but not a schema change so this number was not incremented.
+< TODO NK > Handle the complexity of an alternative .NET SDK?
```
zivkan: I look forward to this :)

designs/RestoreAndBuildCompatibility.md
nkolev92: There are ways to do it without a schema change. But yeah the `nuances` of the `break` cause trouble to say the least.

designs/RestoreAndBuildCompatibility.md
nkolev92: Me too, let me know when you figure something out :D

designs/RestoreAndBuildCompatibility.md
zivkan: My previous comment was a long way to say that I disagree with the markdown file when it says "not a schema change so the number was not incremented".  It was a schema change, but we didn't increment the number. It's misrepresenting (in my opinion) what actually happened and the reason.

JonDouglas:
Hello!

Actions to take:

JonDouglas:
Closing this PR for now due to the main issue also being closed & this being stale. Please re-open the PR & assign reviewers to merge if this should be included in the repository in the future. Thanks!

Issue #4413, opened 2017-01-27
[MacOS, Linux] NuGet writes the NuGet.Config in ~/.config folder, but dotnet restore reads from ~/.nuget


Iankodj:
When using NuGet to add a private feed, `NuGet Sources Add` writes in `~/.config/NuGet/NuGet.Config` file. But when using `dotnet restore` the `~/.nuget/NuGet/NuGet.Config` file is used to read the `packageSources` and `packageSourceCredentials`. 

This issue forces users to manually copy and paste all necessary information instead of having that generated automatically by the NuGet CLI. 

DotNetCore repo issue: https://github.com/dotnet/core/issues/453

zhili1208:
@rrelyea this is a known issue. It's a dupe of https://github.com/NuGet/Home/issues/4095
the problem here is ~/.config is defined by mono not us, we use .net framework Environment variable Appdata Folder for ,net framework(mono) and use ~/.nuget for .net core(this is defined in nuget).

we can't change mono to /.nuget, it will break mono restore for existing user.

My suggestion is not use mono and dotnet at the same time or use -configfile as a workaround.
mono nuget.exe sources add -configfile ~/.nuget/NuGet/NuGet.config

mrward:
So on Windows dotnet.exe and NuGet.exe both use AppData for their NuGet.Config files but on non-Windows platform they use different locations?

Iankodj:
@zhili1208 Thanks for the info. 

@mrward As far as I understand it is Mono that sets up the AppData variable with the `~/.config` path. And dotnet CLI uses directly the `~/.nuget` folder.

sartan:
For the time being, this inelegant workaround seems to do the trick:

`ln -s ~/.nuget/NuGet ~/.config/Nuget`

siberianguy:
Have the same problem in Jetbrains Rider (I suppose it's because they use both Mono and .net core). It was weird as the UI showed packages from private repositories but then did not install them. The symlink solution did not help as it blew up Rider (he was saying something about relative paths), so for now I fixed it with a Nuget.config in project's folder

fundead:
This is a footgun in userland i.e with the use case of needing to run `nuget setApiKey` (on MacOS in our case), as `dotnet nuget` doesn't currently have the `setApiKey` capability. I imagine many machines will get in this state due to installing tools like Rider above and Visual Studio for Mac alongside .NET Core. Would be awesome to see a consistent, non-surprising cross platform experience here.

[...]

Issue #5555, opened 2017-07-10
`UserSettingsDirectory` resolves incorrectly for dotnet on Mac


sandyarmstrong:
## Environment

`dotnet 2.0.0-preview3-006689`
OS X (and presumably Linux)

## The Problem

When using NuGet functionality in .NET Core on Mac (and presumably Linux), `NuGetFolderPath.UserSettingsDirectory` incorrectly resolves to `~/.nuget`, causing a different default `NuGet.config` to be used than Visual Studio for Mac or any non-Core user of `Nuget.Client`.

## Impact

On Windows, users expect to be able to modify their NuGet configuration via their IDE. They can add myget URLs, internal feeds, etc, all in Visual Studio, and those feeds will be available to `dotnet`. On Mac, users follow that same expectation and then cannot figure out why `dotnet` is not using the feeds they just configured.

I sincerely hope it's possible to fix this in time for .NET Core 2.0 final.

## Explanation

`NuGetFolderPath.UserSettingsDirectory` is [documented](https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior) to resolve to `%APPDATA%\NuGet`. On Mac and Linux, using Mono, `SpecialFolder.ApplicationData` has always followed the [freedesktop.org XDG Base Directory Spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html), and mapped to the value of `XDG_CONFIG_HOME` (or `~/.config` by default).

So, anyone using `NuGet.Client` on Mac or Linux would have `NuGetFolderPath.UserSettingsDirectory` map to `~/.config/NuGet` by default. This is depended on by Visual Studio for Mac, among other products.

In fact, `Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData` returns `~/.config` in .NET Core 2.0, just like Mono.

However, in `NuGetEnvironment`, which had to support earlier versions of .NET Core that lacked the `Environment.SpecialFolder` APIs, there is a [custom implementation](https://github.com/NuGet/NuGet.Client/blob/4.3.0.4273/src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs#L81) to resolve locations like `ApplicationData`.

[...]

PR #10364, opened 2020-12-08
Proposal of consolidating user wide NuGet.Config files on Linux/Mac

heng-liu:
Design for https://github.com/NuGet/Home/issues/4413

### DIFF
--- a/designs/Consolidate-NuGetConfig_OnLinux.md
+++ b/designs/Consolidate-NuGetConfig_OnLinux.md
@@ -1,0 +1,134 @@
+
+# Spec Name
+
+* Status: In Review
+* Author(s): [Heng Liu](https://github.com/heng-liu)
+* Issue: [4413](https://github.com/NuGet/Home/issues/4413) [MacOS, Linux] NuGet writes the NuGet.Config in ~/.config folder, but dotnet restore reads from ~/.nuget 
+
+## Problem Background
+
+On Linux/Mac, NuGet has inconsistent paths of user wide NuGet.Config file for different NuGet tools. Dotnet CLI uses hard coded ~/.nuget folder (~/.nuget/NuGet/NuGet.Config), while Mono uses the folder from environment variable "APPDATA", (by default, it's ~/.config/NuGet/NuGet.Config). This is confusing and we should consolidate the two into one location. 
+After discussing in NuGet team and checking with Visual Studio Mac/Mono team, we believe we should change the dotnet CLI path to match that of Mono’s, for the following reasons:
+ * On Linux/Mac, users usually get application data from environment variable APPDATA, which has the default value of ~/.config. Getting application data from ~/.nuget will cause inconsistency user experience between NuGet and other applications.
+ * Getting application data from environment variable APPDATA on Linux/Mac will make it consistent with the behavior on Windows.
+
+For the sake of simplicity, we will refer to ~/.nuget/NuGet/NuGet.Config as legacy path, and refer to ~/.config/NuGet/NuGet.Config as consolidated path in the following.
+
+## Who are the customers
+Dotnet CLI users who work on Linux and Mac.
+
+## Goals
+Make dotnet CLI user-wide config file path consistent with Mono on Linux/Mac and break as less dotnet CLI users as possible.
+
+## Non-Goals
+Completely get rid of the legacy path.
+
+## Background knowledge
+On Linux/Mac, user wide NuGet.Config file is created when trying to get the user wide NuGet.Config file at the beginning phase of running dotnet restore command(and many other dotnet commands with implicit restore if not specifying --no-restore option, like dotnet new, dotnet build, dotnet run, dotnet test, dotnet publish, and dotnet pack), or dotnet msbuild -t:restore command, and found the user wide NuGet.Config file does not exist. Then the user wide NuGet.Config will be created with package source of nuget.org, as the only default package source. 
+
+## Solution overview 
+Add following steps when trying to get user-wide config file path by running dotnet CLI on Linux/Mac:
+(If it's not on Linux/Mac, or it's not in .NET core code path, we will not add those steps. So that the mono and dotnet CLI on Windows will not be affected.)
+Check if the following conditions are satisfied.
+•	The legacy path does not exist.
+•	The consolidated path exists.
+If yes, NuGet will get user wide NuGet.Config path from consolidated path.
+If above condition is not satisfied, it should be in one of the following conditions:
+
+1.Neither legacy path nor consolidated exists.
+In this scenario, user has never called NuGet in dotnet CLI or Mono before.
+NuGet will create a default NuGet.Config file in consolidated path.
+NuGet will get user wide NuGet.Config file from consolidated path. 
+
+2.Legacy path exist, but consolidated path does not exist.
+In this scenario, user has called NuGet in dotnet CLI before, but has never called NuGet in Mono.
+NuGet will copy the NuGet.Config file from legacy path to consolidated path, then delete the one in legacy path. 
+NuGet will get user wide NuGet.Config file from consolidated path.
+
+3.Both legacy path and consolidated path exist.
+In this scenario, user has called NuGet in both dotnet CLI and Mono before.
+NuGet could not determine for users to consolidate the contents of the two NuGet.Config files, as there might be conflicting config.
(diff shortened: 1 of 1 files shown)


zkat (commented):
IMO:

* We should comply with the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for config directories, aka use the env var `$XDG_CONFIG_HOME`, which will default to `~/.config` **on Linux**.
* **For macOS**, we should conform to their [Standard Directories](https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/AppRuntime/AppRuntime.html) guidance (search for "Important App Directories"), which means we would use `~/Library/Application Support/NuGet/NuGet.Config`. That said, using XDG-style config storage _is_ possible in macOS but I'd rather comply with the OS's own guidance here.

zkat:
Other thoughts: We should add a `dotnet nuget config-path` command or something similar so customers scripting things around NuGet can portably get the correct config file on each platform (`$ cat $(dotnet nuget config-path)`)

JonDouglas:
@zkat I think the MacOS callout is a good one. Do you think this should apply for CLI/headless experiences given NuGet isn't really an installable app per-say for end users? There's a good section on hidden files & directories in [the same documentation about Simplifying the User Experience](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW7) which suggests that the Linux proposal should still be ideal for the CLI experience / software developers:

**UNIX-specific directories.** The directories in this category are inherited from traditional UNIX installations. They are an important part of the system’s BSD layer but are **_more useful to software developers than end users_**. Some of the more important directories that are hidden include:
- /bin—Contains essential command-line binaries. Typically, you execute these binaries from command-line scripts.
- /dev—Contains essential device files, such as mount points for attached hardware.
- /etc—Contains host-specific configuration files.
- /sbin—Contains essential system binaries.
- /tmp—Contains temporary files created by apps and the system.
- /usr—Contains non-essential command-line binaries, libraries, header files, and other data.
- /var—Contains log files and other files whose content is variable. (Log files are typically viewed using the Console app.)

Another proposal would be:

`XDG_CONFIG_HOME` would likely map to `~/Library/Preferences/` which would potentially pollute the user's app preference space for a CLI experience.

Java for example puts these preferences in the location above.

heng-liu:
For now, NuGet maintains a [SpecialFolder Enum](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs#L338) and complex logics to map different folders on different platforms in .NET Core code path (.NET Framework code path is different, it uses Environment.SpecialFolder from .NET Framework ):
(The code is : https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs#L86-L164)

The reason why NuGet maintains a SpecialFolder ourselves seems like there is [no such Enum in .NET core before](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Common/PathUtil/NuGetEnvironment.cs#L333-L337 ) 

But since .NET Core start to support Environment.SpecialFolder from netcoreapp2.0, shall we use [Environment.SpecialFolder Enum in .NET Core](https://docs.microsoft.com/en-us/dotnet/api/system.environment.specialfolder?view=net-5.0) instead of making a enum ourselves and maintain complex logics of mapping different folders on different platforms in .NET Core code path?

If we use that enum of .NET, we just let the .NET to decide which folder to pick on different platforms. 
e.g. NuGet could get user wide settings by picking Environment.SpecialFolder.ApplicationData folder and then check the subfolder named "NuGet".

nkolev92:
@heng-liu 

The special folder logic is hidden so it should be safe to refactor. 
Since netstandard2.0 added support for Environment.SpecialFolder, it should be safe to change to using that instead. 
We should just verify that the experience is consistent and that there are no surprises.

heng-liu:
Thanks @nkolev92 !
I'll compare the two and post the results here.

heng-liu:
Since only .NET Core code path will be changed, 
I compared `System.Environment.GetFolderPath(System.Environment.SpecialFolder._SomeFolder_)`
and `NuGetEnvironment.GetFolderPath(NuGetEnvironment.SpecialFolder._SomeFolder_)` on Windows/Linux/Mac for the .NET Core for all the 6 folders.

On Windows, everything is the same:
```
ProgramFilesX86 in .NET,  C:\Program Files (x86)
ProgramFilesX86 in NuGet,  C:\Program Files (x86)

ProgramFiles in .NET,  C:\Program Files
ProgramFiles in NuGet,  C:\Program Files

UserProfile in .NET,  C:\Users\henli
UserProfile in NuGet,  C:\Users\henli

CommonApplicationData in .NET,  C:\ProgramData
CommonApplicationData in NuGet,  C:\ProgramData

ApplicationData in .NET,  C:\Users\henli\AppData\Roaming
ApplicationData in NuGet,  C:\Users\henli\AppData\Roaming

LocalApplicationData in .NET,  C:\Users\henli\AppData\Local
LocalApplicationData in NuGet,  C:\Users\henli\AppData\Local
```
On Linux, **CommonApplicationData** and **ApplicationData** are different:

```
ProgramFilesX86 in .NET,
ProgramFilesX86 in NuGet,

ProgramFiles in .NET,
ProgramFiles in NuGet,

UserProfile in .NET,  /home/henli
UserProfile in NuGet,  /home/henli

CommonApplicationData in .NET,  /usr/share
CommonApplicationData in NuGet,  /etc/opt

ApplicationData in .NET,  /home/henli/.config
ApplicationData in NuGet,  /home/henli/.nuget

LocalApplicationData in .NET,  /home/henli/.local/share
LocalApplicationData in NuGet,  /home/henli/.local/share
```
On MacOS, **ProgramFiles**, **CommonApplicationData** and **ApplicationData** are different:
```
ProgramFilesX86 in .NET,
ProgramFilesX86 in NuGet,

ProgramFiles in .NET,  /Applications
ProgramFiles in NuGet,

UserProfile in .NET,  /Users/heng
UserProfile in NuGet,  /Users/heng

CommonApplicationData in .NET,  /usr/share
CommonApplicationData in NuGet,  /Library/Application Support

ApplicationData in .NET,  /home/henli/.config
ApplicationData in NuGet,  /home/henli/.nuget

LocalApplicationData in .NET,  /home/henli/.local/share
LocalApplicationData in NuGet,  /home/henli/.local/share
```
For all those differences, does .NET special folder value make sense? 
(.NET references document https://www.freedesktop.org/software/systemd/man/file-hierarchy.html)

The code from .NET is as following:
Unix : https://github.com/dotnet/runtime/blob/c97fafd09fe358129ee3bdeb4ba76cd1c111e8cb/src/libraries/System.Private.CoreLib/src/System/Environment.GetFolderPathCore.Unix.cs#L57-L144

We use ApplicationData for user wide NuGet.Config, and we're going to change the current home/.nuget folder, to use the value of env var XDG_CONFIG_HOME, if it's null, then we will get from home/.config. After this change, it will be the same with what we get from .NET: `System.Environment.GetFolderPath(System.Environment.SpecialFolder.ApplicationData)`

But how about CommonApplicationData ?

nkolev92:
`CommonApplicationData` seems to be used for MachineWide settings, so my gut instinct is we shouldn't change that. 
That one is consistent among mono and .NET Core, so that's all that matters imo.

heng-liu:
The implementation needs the symbolic link API, which is proposed at https://github.com/dotnet/runtime/issues/24271

lewing:
cc @sandyarmstrong re https://github.com/NuGet/Home/issues/5555

JonDouglas:
@heng-liu Would you be able to assign appropriate reviewers to this proposal so we can merge? There's a few individuals on this PR who can help review in addition. If we can't merge and should close this instead, feel free to close this out!

zivkan:
The dotnet/runtime repo finished their symbolic link APIs work, however, they didn't make it a package that we can use on .NET Framework/Mono.  Therefore, for this design to be implemented, NuGet would need to P/Invoke symlink APIs on relevant platforms. Otherwise, this feature need to be redesigned to not depend on symlink APIs in the runtime.

jeffkl:

Issue #5713, opened 2017-08-07
NuGet locals command should allow deleting specific packages from NuGet global packages folder


SupporterMan:
## Details about Problem
I have a nuget package that want to remove it from my project and I also want to clear its local global cache from userprofile\.nuget\package. According to the manage package document, I only could use "nuget locals global-packages -clear" command to clear all global caches, not a specific one. Otherwise, I need to navigate to the .nuget\packages path in file folder and delete it manually.

NuGet product used: NuGet.exe
NuGet version (4.1.0)

**edit by @nkolev92**

This issue should cover the VS side of things as well. https://github.com/NuGet/Home/issues/8204

anangaur:
cc @emgarten The following seem like repeated asks:

- nuget locals global-packages -clear **-packages {list of packages}**
- nuget locals global-packages -clear **-project {path to project file}**

In the 2nd case it clears all packages that are directly referenced in the project.

emgarten:
Looks very useful. It would also be nice to prune everything not in a solution, or older than a certain time.

neolib:
This win my vote. It should be very useful to remove older packages than just clear all of global cache.

mishra14:
@SupporterMan @neolib Thanks for reporting this request. We will consider this request, when possible.

CZEMacLeod:
Could really do with 
`nuget locals http-cache -clear -packages {list of packages}`
myself.
And/or the package to automatically be cleared from the http-cache when it is pushed.

gioce90:
There are news around this?
I asked something which somehow relates to this one [right here](https://github.com/NuGet/Home/issues/13918) (well, that's something that would benefit from the feature discussed here)

gioce90:
Deleting specific packages from the NuGet global packages folder will improve the inner-loop development of a NuGet package on the local machine. But this issue has been open since 2017, so in the meantime I created a solution to address this problem.

I recently decided to make my solution public by releasing a package called NuJet (https://github.com/gioce90/NuJet). It enhances the inner-loop development and testing process for NuGet packages, especially in local workflows.

Feel free to check it out—I’d love to hear your feedback!

PR #10427, opened 2021-01-06
Clearing an specific package from NuGet cache - Mini spec

dominoFire:
Mini spec for deleting a single package from cache.

This specs cover a part of the design described in https://github.com/NuGet/Home/issues/5713

### DIFF
--- a/designs/Commands/clear-package.md
+++ b/designs/Commands/clear-package.md
@@ -1,0 +1,92 @@
+# Clearing an specific package from NuGet cache
+
+* Status: **Incubating**
+* Author(s): [Fernando Aguilar](https://github.com/dominofire)
+
+The purpose is to add a feature to `nuget locals` command to delete a package from NuGet cache folder
+
+
+## Problem Background
+
+There's some interest on having features to delete an specific package instead of clearing the whole cache.
+
+- https://github.com/NuGet/Home/issues/5713 : Comments on deleting specific packages
+- https://github.com/NuGet/Home/issues/4980 : Comments on cache handling polices
+- https://github.com/chgill-MSFT/NuGetCleaner : A clean-up tool relying on last-acesss file metadata in NTFS file system
+
+In PackageReference world, the global packages folders are shared across all restored projects. 
+
+There's no current API or command to delete an specific package.
+
+
+## Who are the customers
+
+All `dotnet nuget locals` users.
+
+
+## Goals
+
+- Add command options to `dotnet nuget locals` to delete _one_ package from a NuGet cache folder, either _one_ or multiple versions.
+
+
+## Non-Goals
+
+- Cache expiration policies are not covered
+- Deleting cached packages from an specific folder are not covered 
+- `packages.config` projects are not covered. To delete a package, delete the pacakge folder from `packages` folder
+- `nuget.exe` commands are not covered
+
+
+## Solution Overview
+
+From ideas described in [this comment](https://github.com/NuGet/Home/issues/5713#issuecomment-320560636), the following command options will delete an specific package. Exmples are listed below.
+
+Delete a package from cache with an specific version:
+
+```
+dotnet nuget locals global-packages --clear --package a --version 1.3
+```
+
+Delete all package versions from cache:
+
+```
+dotnet nuget locals global-packages --clear --package a
+```
+
+Delete all versions of multiple packages:
+
+```
+dotnet nuget locals http-cache --clear --package a
+dotnet nuget locals http-cache --clear --package b
+dotnet nuget locals http-cache --clear --package c
+dotnet nuget locals http-cache --clear --package d
+```
+
+Delete all versions of multiple packages (version 2):
+
+```
+dotnet nuget locals global-packages -clear -packages a,b,c,d,e
+```
+
+### Implementation details
+
+1. Add command definitions at [src\NuGet.Core\NuGet.CommandLine.XPlat\Commands\LocalsCommand.cs](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/LocalsCommand.cs)
+1. Modify [src\NuGet.Core\NuGet.Commands\CommandRunners\LocalsCommandRunner.cs](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/CommandRunners/LocalsCommandRunner.cs) to delete an specific package 
+1. Update help in `dotnet nuget locals` command
+
+
+## Test Strategy
+
+Add a test in `Dotnet.Integration.Test` project
+
+
+## Future Work
+
+Follow up on:
+
+- https://github.com/NuGet/Home/issues/4980: A better clean-up cache policy
+- https://github.com/NuGet/Home/issues/8204: Delete package from Visual Studio
+
+### References
+
+- Clear one package on StackOverflow. [Link](https://stackoverflow.com/questions/49935118/clear-just-one-package-from-npm-cache)
--- a/designs/Commands/clear-package.md
+++ b/designs/Commands/clear-package.md
@@ -89,4 +89,4 @@ Follow up on:
 
 ### References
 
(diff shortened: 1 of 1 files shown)


donnie-msft (approved):
Looks good! Interested in the VS option down the road

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
+# Clearing an specific package from NuGet cache
```
nkolev92, line 1: It's not a cache, it's an installation directory. There's no updating, stale checks etc.

It's like program files.

Please refer to it as `global packages folder`.

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+* Status: **Incubating**
+* Author(s): [Fernando Aguilar](https://github.com/dominofire)
+
+The purpose is to add a feature to `nuget locals` command to delete a package from NuGet cache folder
```
nkolev92, line 6:
```suggestion
The purpose is to add a feature to `nuget locals` command to delete a package from NuGet global packages folder folder
```

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+## Problem Background
+
+There's some interest on having features to delete an specific package instead of clearing the whole cache.
```
nkolev92, line 11:
```suggestion
There's some interest on having features to delete an specific package instead of clearing the whole global packages folder.
```

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Goals
+
+- Add command options to `dotnet nuget locals` to delete _one_ package from a NuGet cache folder, either _one_ or multiple versions.
+
+
+## Non-Goals
+
+- Cache expiration policies are not covered
+- Deleting cached packages from an specific folder are not covered
```
nkolev92, line 35: Can you elaborate on this?

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+## Who are the customers
+
+All `dotnet nuget locals` users.
+
+
+## Goals
+
+- Add command options to `dotnet nuget locals` to delete _one_ package from a NuGet cache folder, either _one_ or multiple versions.
```
nkolev92, line 29:
```suggestion
- Add command options to `dotnet nuget locals` to delete _one_ package from a NuGet global packages folder, either _one_ or multiple versions.
```

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+- Deleting cached packages from an specific folder are not covered 
+- `packages.config` projects are not covered. To delete a package, delete the pacakge folder from `packages` folder
+- `nuget.exe` commands are not covered
+
+
+## Solution Overview
+
+From ideas described in [this comment](https://github.com/NuGet/Home/issues/5713#issuecomment-320560636), the following command options will delete an specific package. Exmples are listed below.
+
+Delete a package from cache with an specific version:
```
nkolev92, line 44:
```suggestion
Delete a package from the global packages folder with a specific version:
```

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+Delete all package versions from cache:
+
+```
+dotnet nuget locals global-packages --clear --package a
+```
+
+Delete all versions of multiple packages:
+
+```
+dotnet nuget locals http-cache --clear --package a
```
nkolev92, line 59: Is there an explicit ask for the http-cache scenarios?

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+From ideas described in [this comment](https://github.com/NuGet/Home/issues/5713#issuecomment-320560636), the following command options will delete an specific package. Exmples are listed below.
+
+Delete a package from cache with an specific version:
+
+```
+dotnet nuget locals global-packages --clear --package a --version 1.3
+```
+
+Delete all package versions from cache:
```
nkolev92, line 50:
```suggestion
Delete all package versions from global packages folder:
```

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+dotnet nuget locals http-cache --clear --package a
+dotnet nuget locals http-cache --clear --package b
+dotnet nuget locals http-cache --clear --package c
+dotnet nuget locals http-cache --clear --package d
+```
+
+Delete all versions of multiple packages (version 2):
+
+```
+dotnet nuget locals global-packages -clear -packages a,b,c,d,e
```
nkolev92, line 68: The approach adopted by other commands is multiple arguments instead of `,` delimited values.

Take `dotnet restore --source` for example.

nkolev92 (changes requested):
Meant to request changes. Apologies for the long overdue review.

designs/Commands/clear-package.md
```diff
@@ -0,0 +1,92 @@
...
+
+There's some interest on having features to delete an specific package instead of clearing the whole cache.
+
+- https://github.com/NuGet/Home/issues/5713 : Comments on deleting specific packages
+- https://github.com/NuGet/Home/issues/4980 : Comments on cache handling polices
+- https://github.com/chgill-MSFT/NuGetCleaner : A clean-up tool relying on last-acesss file metadata in NTFS file system
+
+In PackageReference world, the global packages folders are shared across all restored projects. 
+
+There's no current API or command to delete an specific package.
```
zivkan, line 19: There's no API to do it in a single call, but it's absolutely possible to chain together a few API calls to get the global packages folder from config, create a path resolver for that global packages folder, find the directory for a package, and then delete it.

Given this is the "problem background" section, it makes it sound like this is a problem (and neither the goals or non-goals sections below say whether this is in-scope of this design or not). I don't remember seeing any customers asking for an easier API in the NuGet SDK, so I suggest removing the mention of API here.

designs/Commands/clear-package.md
zivkan, line 68: Other examples have used double dash `--clear` and `--package`, rather than single dash `-clear` and `-packages`.

The dotnet CLI uses the GNU argument syntax where only single character arguments use a single dash and any multi-character argument must use double dash. I don't know if the dotnet CLI implements it, but programs using GNU's `getopt` treat single dash `-ab` the same as two arguments `-a -b`, which is why `git clean -xdf` is the same as `git clean -x -d -f`, but different to `git clean --xdf`

zivkan, line 68: Does this mean that `--version` is only valid with `--package` and not `--packages`? Should this be mentioned explicitly?

If it is valid, what's the expected behaviour? Is it possible to specify both multiple packages and multiple versions?

If we don't have customers asking to be able to delete multiple packages in a single command, I suggest it's better to drop this feature, and wait for customer feedback, with information to help us understand why it's difficult to delete one package at a time, like your example above. We can add features/arguments, but removing them is a breaking change, so much more difficult.

JonDouglas:
@dominoFire Can you assign appropriate reviewers & when you feel comfortable with merging this PR, go ahead and merge? I'm working to clear out stale PRs & this one has sat for some time.

designs/Commands/clear-package.md
kasperk81: you can use the same double colon syntax for version{s} as dotnet-new template. https://www.nuget.org/packages/NUnit3.DotNetNew.Template/ `dotnet new --install NUnit3.DotNetNew.Template::1.9.0`. it is ugly, it doesn't "fit" with anything else in the entire ecosystem... but (that didn't stopped it becoming part of dotnet-new so) at least you have a precedence there. :)

jeffkl:

designs/Commands/clear-package.md
CZEMacLeod: Yes - I need that!
I find that after pushing new packages, then building requiring the new version of the package, it fails as it gets the info for that package from the cache.
It should really flush the package from the cache when pushing automatically.

PR #10444, opened 2021-01-08
Reduce the verbosity of the issue template

### LINKED ISSUE #10441: The NuGet issue template is too verbose
Note that I have removed the issue template. 

This template is too verbose and honestly asks for a lot of information that's not very relevant.

Note that on 1080p screens, our template requires scrolling!!
That's excessive!

nkolev92:
Fixes: https://github.com/NuGet/Home/issues/10441

The current issue template is pretty verbose, contains information that we do not need/doesn't deviate usually and often time it leads to people deleting it altogether. 

These changes are meant to reduce the visual strain one has when looking at a new issue and improve issue quality by having something customers would be willing to follow instead of delete.

Note that this is a template tailored towards bugs. We could use a feature template, but that's something we don't have now, so I'm not yet introducing it. 

I came up with this template by looking at template from partner repos: 

https://github.com/microsoft/vscode/issues/new?assignees=&labels=&template=bug_report.md
https://github.com/dotnet/project-system/issues/new
https://github.com/dotnet/runtime/issues/new?assignees=&labels=&template=01_bug_report.md&title=

Note how all of those templates are significantly less verbose than ours.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,7 +12,7 @@ Is this a regression? Which version:
 
 ## Repro steps and/or sample project
 
-<!-- Please as much details as possible to help us to reproduce your problem. -->
+<!-- Please add as much details as possible to help us to reproduce your problem. -->
 
 ### Verbose Logs
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -8,7 +8,7 @@ NuGet product used (NuGet.exe | Visual Studio | MSBuild.exe | dotnet.exe):
 
 Product version: 
 
-Is this a regression? Which version: 
+Worked before? If so, with which NuGet version: 
 
 ## Repro steps and/or sample project
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,18 +4,16 @@
 
 ## Details about Problem
 
-NuGet tool used (NuGet.exe | VS | MSBuild | dotnet.exe):
+NuGet product used (NuGet.exe | VS | MSBuild | dotnet.exe):
 
-NuGet tool version: 
+Product version: 
 
-Worked before? If so, when:
+Is this a regression? Which version: 
 
 ## Repro steps and/or sample project
 
-1.
-
-1.
+<!-- Please as much details as possible to help us to reproduce your problem. -->
 
 ### Verbose Logs
 
-Please include verbose logs (NuGet.exe <COMMAND> -verbosity detailed | dotnet.exe <COMMAND> --verbosity diag | etc...)
+<!-- Please include verbose logs (NuGet.exe <COMMAND> -verbosity detailed | dotnet.exe <COMMAND> --verbosity diag | etc...) -->
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,41 +1,21 @@
-Please read the following information before posting the issue.
-
-## Before posting the issue...
-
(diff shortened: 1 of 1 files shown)


zkat (approved):
This is great! Thanks for doing this!

We could consider using multiple issue templates, which GitHub supports, to add a separate one for feature requests (although we should use Discussions for those going forward, imo)

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,41 +1,19 @@
...
 ## Details about Problem
 
-NuGet product used (NuGet.exe | VS UI | Package Manager Console | dotnet.exe):
-
-NuGet version (x.x.x.xxx):
-
-dotnet.exe --version (if appropriate):
-
-VS version (if appropriate):
+NuGet product used (NuGet.exe | VS | MSBuild | dotnet.exe):
```
donnie-msft: Maybe sort these from most likely to least likely? 

`NuGet.exe | dotnet.exe | VS | MSBuild`

donnie-msft (approved):

erdembayar (approved):

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,41 +1,19 @@
...
 
-OS version (i.e. win10 v1607 (14393.321)):
+Product version: 
 
-Worked before? If so, with which NuGet version:
+Is this a regression? Which version:
```
zivkan: Should we be more explicit in asking which version last worked?

zivkan (approved):

JonDouglas:
Wondering if we can just link people to our wiki / contribution too?

https://github.com/NuGet/Home/wiki/Submitting-Bugs-and-Suggestions#bug-reports

https://github.com/NuGet/Home/wiki/Contribute-to-NuGet

Additionally, first time contributors will see this:

image

So I created https://github.com/NuGet/Home/pull/10445 as I wrote most of this a few months ago :)

nkolev92:
@JonDouglas 

I think we can link https://github.com/NuGet/Home/wiki/Submitting-Bugs-and-Suggestions#bug-reports at the top, that's a good idea. 

Should we link to the wiki or should we wait your contributing PR to get merged first and link to that?

.github/ISSUE_TEMPLATE.md
nkolev92: Any suggestions?
The MO for this change is reducing verbosity and unnecessary clutter, so I wanted to make it shorter. 

To me, this reads just fine, but open to suggestions.

.github/ISSUE_TEMPLATE.md
nkolev92: I don't have a preference here.

I'm not sure I agree that order is the most likely on, I think there's more incoming VS issues than dotnet ones right now.  

I was strongly debating removing that, but I think it provides value of giving you a sense that you are filing issues in the right repo.

aortiz-msft: >VS [](start = 32, length = 2)

Visual Studio instead of VS since VS is not a common acronym. #Resolved

aortiz-msft: MSBuild.exe? #Resolved

.github/ISSUE_TEMPLATE.md
aortiz-msft: My recommendation: "Is this a regression? If yes, what's the last version it worked with:" #Resolved

aortiz-msft (changes requested):
:clock1:

.github/ISSUE_TEMPLATE.md
nkolev92: I'd like to keep this as short as possible. 

Visually, I think that recommendation takes too much space on the screen. 

I switched it back to the original text, wyt?

aortiz-msft (approved):
:shipit:

Issue #3324, opened 2016-08-16
NuGet and long file name support


AndreyAkinshin:
I often have the following exception during restore on Windows:

```
The specified path, file name, or both are too long. The fully qualified file name must be less than 260 characters, and the directory name must be less than 248 characters.
```

In the NuGet 2.x API, there is a nice interface `IFileSystem`. So, I could implement own file system which supports long paths. In the NuGet 3.x API, `IFileSystem` is legacy; the core logic works with the file system directly via `System.IO` (see [an example](https://github.com/NuGet/NuGet.Client/blob/3.5.0-rc1-final/src/NuGet.Core/NuGet.Common/PathUtil/FileUtility.cs)).

Is there a workaround which adds the long paths support in NuGet 3.x?

rrelyea:
We do not yet support long paths in NuGet.
https://channel9.msdn.com/Shows/On-NET/Jeremy-Kuhne-long-path-support has some details on new .net framework support for it. We'd like to support this in the future.

AndreyAkinshin:
@rrelyea, why did you drop the `IFileSystem` approach? We worked with NuGet 2.x + long paths for a long time. But we can't do it anymore because there is no `IFileSystem` in NuGet 3.x.

nerumo:
@rrelyea since Jenkins Branch Pipeline creates horribly long sln root folder names (and they aren't willing/capable to change that, so I can't shorten the path anymore. Is there a another workaround? Will this be addressed in 3.\* or in 4?

rrelyea:
We don't know the exact timeline when we'd be able to do this work. We know if won't fit into NuGet 4.0.0

NN---:
@nerumo You can try to shorten path with environment variables.

alpaix:
Bug 389573:[Feedback] Nuget restore fails because path too long

andymac4182:
Any update on this? 
I am running into this with a path that is long because of the system package names. `\runtime.opensuse.42.1-x64.runtime.native.system.security.cryptography.openssl\4.3.0\runtime.opensuse.42.1-x64.runtime.native.system.security.cryptography.openssl.4.3.0.nupkg.sha512`

AndreyAkinshin:
Any updates? It's a critical issue for us because we still have to use NuGet 2.x for some features; it's impossible to migrate to NuGet 4.x because of the lack of the long file name support.

PeterMurdoch:
[...]

PR #10460, opened 2021-01-15
INuGetPackageFileService spec

rrelyea:
associated code PR: https://github.com/NuGet/NuGet.Client/pull/3823


designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+* Author(s): [Rob Relyea](https://github.com/rrelyea)
+* Issue: [10151](https://github.com/NuGet/Home/issues/10151) INuGetRemoteFileService - Fetch Images and embedded licenses for codespaces and normal execution
+
+## Problem Background
+
+With Codespaces support added to NuGet client, our code needs to be able to handle all packages and communication with feeds happening on a server node. This is architected as async method calls to services that are remoted. In most cases, we also have chosen to have 1 code path, for normal execution and remoted execution. These services are still used in the normal execution, however, the service runs in the same process, as opposed to being remoted across the network.
+
+Package Icons can either come from:
+- a package feeds search service
+- (old style) or a URI to an arbitrary place on the web
```
dtivel: Where does the URI come from?  The package, a package feed, ...?

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+Defined a new interface for this service:
+
+namespace NuGet.VisualStudio.Internal.Contracts
+{
+    public interface INuGetRemoteFileService : IDisposable
+    {
+        ValueTask<Stream?> GetPackageIconAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
+        ValueTask<Stream?> GetEmbeddedLicenseAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
+    }
+}
```
dtivel: Move inside a markdown code block:

    ```C#
    
    ```

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+On the service side, in SearchObject.CacheBackgroundData(), beyond the packageSearchMetadata caching that was already done, we call a RemoteFileService instance to AddIconToCache() and to AddLicenseToCache().
```
dtivel: You can link to the current implementation to give curious readers more context:  [`SearchObject.CacheBackgroundData(...)`](https://github.com/NuGet/NuGet.Client/blob/b5b44526dea0379ebd6c8e51e8a041c06d5845ca/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/Services/SearchObject.cs#L218-L236)

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+On the client side, when trying to Fetch the Icon, we call RemoteFileService.GetPackageIconAsync, passing in the package identity.
+
+On the service side, we then retrieve the appropriate file from the appropriate location based on the entries in the MemoryCache.
```
dtivel: Although the introduction of the memory cache predates this feature, I think it would be meaningful to include in this spec general details about the memory cache as it pertains to this feature, to include:

*  current memory cache invalidation triggers (e.g.:  5 minutes (time-based), ...)
*  what the expected behavior is if the client requests a package icon or license and the memory cache does not contain an entry
*  how/when the memory cache gets populated with new/updated package icons/licenses.

Prior to Codespaces, package icon and license fetching did not use a `MemoryCache`.  There are probably some behavioral differences in this area between pre- and post-Codespaces clients.  It would be good to summarize those differences in this spec.

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Other Changes
+
+Several classes had a PackageReader property that was of type Func<PackageReader>.
+Func<PackageReader> wasn't designed to be remoted nicely. In serveral places, i replaced the PackageReader property with a PackagePath property, to be used when appropriate for loading files from a Package.
```
dtivel: *  serveral -> several
* i -> I
* Package -> package

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+None critical.
+
+## Open Questions
+
+Any open questions not specifically called out in the design.
+
+## Considerations
+
+At one point, the design of this interface passed in URIs...we moved to PackageIdentity and storing in the MemoryCache to tighten security of the feature.
```
dtivel: You can also call out:
* why package sources aren't passed in API calls
* performance concerns about making lots of client->service calls and are the concerns valid?

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Fetching Files
+Defined a new interface for this service:
+
+namespace NuGet.VisualStudio.Internal.Contracts
+{
+    public interface INuGetRemoteFileService : IDisposable
+    {
+        ValueTask<Stream?> GetPackageIconAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
+        ValueTask<Stream?> GetEmbeddedLicenseAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
```
dtivel: Do we need to distinguish between an embedded license not being present and a failure to get an embedded license?  If yes, how do we do that?

With your current design is it possible for a failure to obtain license information --- say, because one critical HTTP(S) call failed --- result in displaying to a user that there is no license information when in fact there may be?

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+On the service side, in SearchObject.CacheBackgroundData(), beyond the packageSearchMetadata caching that was already done, we call a RemoteFileService instance to AddIconToCache() and to AddLicenseToCache().
+
+On the client side, when trying to Fetch the Icon, we call RemoteFileService.GetPackageIconAsync, passing in the package identity.
```
dtivel: * Fetch -> fetch
* Icon -> icon
* 
What about when fetching license information?

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+On the service side, we then retrieve the appropriate file from the appropriate location based on the entries in the MemoryCache.
+
+### Other Changes
+
+Several classes had a PackageReader property that was of type Func<PackageReader>.
+Func<PackageReader> wasn't designed to be remoted nicely. In serveral places, i replaced the PackageReader property with a PackagePath property, to be used when appropriate for loading files from a Package.
+
+## Test Strategy
+
+Added unit tests for NuGetRemoteFileService and updated many existing tests for this new implemenation of icons/licenses.
```
dtivel: implemenation -> implementation

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
+
+# Spec Name
+
+* Status: In Review
+* Author(s): [Rob Relyea](https://github.com/rrelyea)
+* Issue: [10151](https://github.com/NuGet/Home/issues/10151) INuGetRemoteFileService - Fetch Images and embedded licenses for codespaces and normal execution
+
+## Problem Background
+
+With Codespaces support added to NuGet client, our code needs to be able to handle all packages and communication with feeds happening on a server node. This is architected as async method calls to services that are remoted. In most cases, we also have chosen to have 1 code path, for normal execution and remoted execution. These services are still used in the normal execution, however, the service runs in the same process, as opposed to being remoted across the network.
```
dtivel: Referring to standalone and Codespaces-connected clients as "normal" and "remoted" confusing, primarily over the use of the word "normal."

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+## Problem Background
+
+With Codespaces support added to NuGet client, our code needs to be able to handle all packages and communication with feeds happening on a server node. This is architected as async method calls to services that are remoted. In most cases, we also have chosen to have 1 code path, for normal execution and remoted execution. These services are still used in the normal execution, however, the service runs in the same process, as opposed to being remoted across the network.
+
+Package Icons can either come from:
+- a package feeds search service
+- (old style) or a URI to an arbitrary place on the web
+- embedded in a package
+
+Licenses can also come from a few places, including a license file embedded in a package.
```
dtivel: It would be good to list them all, like you did for package icons.

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Who are the customers
+
+All PM UI customers.
+
+## Goals
+
+Support remote and normal fetching of icons and licenses.
```
dtivel: Again, use of the word "normal" is confusing.

From a client perspective, how the client fetches package icons and license metadata should be consistent, whether the client is operating in standalone or Codespaces-connected mode.

From a service perspective --- whether the service is running locally or remotely ---, the service should support fetching package icons and license metadata from server-local data sources (e.g.:  global packages folder) or from server-remote data sources (like a remote package source).

designs/NuGet-RemoteFileService.md
```diff
@@ -0,0 +1,70 @@
...
+## Non-Goals
+
+## Solution Overview
+
+### Fetching Files
+Defined a new interface for this service:
+
+namespace NuGet.VisualStudio.Internal.Contracts
+{
+    public interface INuGetRemoteFileService : IDisposable
```
dtivel: I'm not in favor of `Remote` in the name.  The service abstracts away where files come from.  `INuGetFileService` seems more appropriate.

designs/NuGet-RemoteFileService.md
rrelyea: adding language to spec like:  (specified in the NuSpec via now deprecated iconUrl.)

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: switched to standalone and Codespaces-connected.  see next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: followed naming from other suggestion: standalone and Codespaces-connected. see next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: followed naming from other suggestion: standalone and Codespaces-connected. see next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: I don't think INuGetFileService is descriptive enough. We've had APIs for files before...ok with removing Remote...but want to think through.

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: package icon has always used a memory cache. Scott added the use for search with codespaces, and now this work is adding it for INuGetRemoteFileService. So we now have 3, we've had 1 for years.

will work on this feedback...

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-RemoteFileService.md
rrelyea: V2 had a IFileService which some customers want back - https://github.com/NuGet/Home/issues/3324

rrelyea: INuGetRemoteFileService - current spec
INuGetFileService - @dtivel suggestion
INuGetPackageFileService - better?
INuGetEmbeddedFileService - better?

thoughts? I lean towards INuGet*FileService

Talked: going with INuGetPackageFileService

dtivel: Bear in mind that this is an internal service and doesn't interfere with implementing a V2 file service separately.

rrelyea: good to know. yes.
do you still prefer your suggestion over the longer (hopefully more descriptive) names?

designs/NuGet-RemoteFileService.md
rrelyea: 1) don't think this API is where we should provide validation of package contents. Package quality like that should be enforced at several other times (pack/push/publish to nuget.org/etc...)

2) i'll think through the question about failing http(s) calls...

rrelyea: about 2) i'm assuming the http(s) failing call you are talking about is the remoted call to the server. is that generally handled by Wrapper class, which i don't have in this work currently? perhaps we have a call to talk this through.

designs/NuGet-RemoteFileService.md
rrelyea: updated spec PR with those details.

designs/NuGet-RemoteFileService.md
rrelyea: we discussed. Rob is going to go dig into more of the license code paths.
(note: NuGetProjectManagerService has some examples of remote exceptions in some of the methods - try/catch)

designs/NuGet-RemoteFileService.md
rrelyea: Talked: going with INuGetPackageFileService

rrelyea: Fixed in spec and in code PR.

designs/NuGet-RemoteFileService.md
rrelyea: SearchObject.cs .. along with search metadata and icon info, license info might be stored here on the service side

if packagePath != null && licenseMetadata != null
	store license uri in PackageIdentityToUri memoryCache


on the client side:  PackageLicenseUtilities.cs
   looks like this will create three different data model objects...based on metadata from nuspec file
	licenseText -- URL
        LicenseText -- expression
	LicenseFileText -- embedded license file 


These two code paths are in click handlers that are only attached to LicenseFileText objects shown in UI via databinding:

packagemetadatacontrol.xaml.cs
   void ViewLicense_Click(sender, e)
       calls PackageLicenseUtilities.GetEmbeddedLicenseAsync

LicenseAcceptanceWindow.xaml.cs might call:
  licensefiletext.cs -- LoadLicenseFile() which loads an embedded license

rrelyea: believe if there was a communication failure, an exception would be raised, and vs would crash. i've not added any other exception handling here. believe we are in a good place.
happy to talk through here or in code PR.

designs/NuGet-PackageFileService.md
```diff
@@ -0,0 +1,104 @@
...
+#### IdentityToUri Memory Cache notes
+
+The settings of the IdentityToUri memory cache that the NuGetPackageFileService class hosts and maintains are identical to the metadata MemoryCache that the search service uses as part of it codespaces rebuilding:
+- 4 MB memory limit
+- 0 physical memory limit percentage
+- 2min polling interval
+
+Given those settings, we don't expect to frequently not find an entry that should be available in the memory cache...we have a Telemetry call to record how often this does happen.
+
+Details of how the memory cache gets populated is overed in the "paragraph above that starts with "On the service side".
```
dtivel: *  overed -> covered
* "paragragh -> paragraph

designs/NuGet-PackageFileService.md
```diff
@@ -0,0 +1,104 @@
...
+On the service side, we then retrieve the appropriate file from the appropriate location based on the Uri stored in the MemoryCache and return the stream to the caller.
+
+#### IdentityToUri Memory Cache notes
+
+The settings of the IdentityToUri memory cache that the NuGetPackageFileService class hosts and maintains are identical to the metadata MemoryCache that the search service uses as part of it codespaces rebuilding:
+- 4 MB memory limit
+- 0 physical memory limit percentage
+- 2min polling interval
+
+Given those settings, we don't expect to frequently not find an entry that should be available in the memory cache...we have a Telemetry call to record how often this does happen.
```
dtivel: I had to read this a couple times.  I think the "... don't ... not ..." tripped me up.  Reword for clarity.  Suggestion:

We expect that failures to find an entry in the memory cache due to eviction will be rare, and telemetry will be added to verify this.

designs/NuGet-PackageFileService.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Open Questions
+
+Any open questions not specifically called out in the design.
+
+## Considerations
+
+At one point, the design of this interface passed in URIs...we moved to PackageIdentity and storing in the MemoryCache to tighten security of the feature.
+
+We also, chose to only key off of PackageIdentity, instead of PackageIdentity and a package source...staying consistent with our treatment of unique package idenities as the same, regardless of where they come from.
```
dtivel: *  also, -> also
*  idenities -> identities

dtivel (commented):
There's some feedback [here](https://github.com/NuGet/Home/pull/10460#discussion_r563932190) which I think might have been overlooked.

> Prior to Codespaces, package icon and license fetching did not use a MemoryCache. There are probably some behavioral differences in this area between pre- and post-Codespaces clients. It would be good to summarize those differences in this spec.

This requires a little deep thinking, but I feel that we're making a significant change here without fully understanding (or at least documenting) the impact to customer scenarios.  I'll make a few fake examples:

*  By introducing another level of caching (memory cache) beyond the existing HTTP cache, a potential behavioral change to package icon/license information in PM UI is ...
*  The overhead of remoting package icon / license information retrieval may adversely impact performance (e.g.:  delay in displaying the information to the user).
*  ...

designs/NuGet-PackageFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-PackageFileService.md
rrelyea: fixed in next iteration.

designs/NuGet-PackageFileService.md
rrelyea: fixed in next iteration.

rrelyea:
> There's some feedback [here](https://github.com/NuGet/Home/pull/10460#discussion_r563932190) which I think might have been overlooked.
> 
> > Prior to Codespaces, package icon and license fetching did not use a MemoryCache. There are probably some behavioral differences in this area between pre- and post-Codespaces clients. It would be good to summarize those differences in this spec.
> 
> This requires a little deep thinking, but I feel that we're making a significant change here without fully understanding (or at least documenting) the impact to customer scenarios. I'll make a few fake examples:
> 
> * By introducing another level of caching (memory cache) beyond the existing HTTP cache, a potential behavioral change to package icon/license information in PM UI is ...
> * The overhead of remoting package icon / license information retrieval may adversely impact performance (e.g.:  delay in displaying the information to the user).
> * ...

I've just added the following to the spec. Would love to know what other concerns you think may need to be documented/discussed.

#### Differences in behavior from current client

Icon fetching has a layer of indirection that may change timing a bit. In the past, the icon fetch would get a stream via http(s). Now, a service fetches that stream, and it is passed back over the service boundary. Given that we've enabled our icons to handle rendering in a more asyncronous way in the last few months, if there is a slight change in behavior, it is unlikely to be very impactful.

Embedded license fetching is used much less, only when somebody clicks on the link which shows the license in a window in the client. As such, it is unlikely to have any change in behavior.

designs/NuGet-PackageFileService.md
```diff
@@ -0,0 +1,112 @@
...
+#### Issue: should GetEmbeddedLicenseAsync be more generic. (question from zkat)
+    zkat - I guess this is fine for now, but I'd really like it if now-or-eventually, this is abstracted away into a set of supported "Well Known Files"--for example, if we want to add Changelog or docs support in the future.
+
+    rrelyea - thought a bit about it... i'd imagine we could pass an enum into a method... packageIdentity, wellKnownEmbeddedFile
+
+    where wellKnownEmbeddedFiles might be readme, license and what else?
+    maybe icon...but that isn't always embedded, can also be a url outside of the package.
+    i lean towards not doing this in a generic way yet...as we don't know the full set of needs yet...and since the nuspec is on the server, we cannot just take a relative path, etc...
+
+    would love to discuss, hear other thoughts.
```
zivkan, line 85: Whether we use an enum or new methods, they're both changes to the RPC contract that need to be versioned to support scenarios where the VS client uses a different version of NuGet to the Codespaces server.

Since this is an internal contract, it needs less API design than NuGet SDK APIs, since we don't need to consider what other customers might want to do. So, is there any reason why NuGet's VS extension would want to do something like this?

```cs
var wellKnownContent = new Dictionary<WellKnownContentType, byte[]>();

foreach (contentType in Enum.GetValues(typeof(WellKnownContentType)))
{
  bytes[] bytes;
  using (var stream = service.GetWellKnownContent(packageIdentity, contentType, cancellationToken))
  {
    bytes = stream.ReadAllBytes();
  }
  wellKnownContent[contentType] = bytes;
}
```

Additionally, there's nothing preventing someone doing `(WellKnownContentType)12345` to get a "strongly typed" `WellKnownContentType` instance, with an invalid value. Whereas you can't call methods that don't exist, so one method per well known content type provides better compile time guarantees. 

Finally, since different "well known content types" might have different data types, we have the opportunity to use more appropriate return values per method. By comparison, using `Stream` or `byte[]` for all content types feels like a leaky abstraction to me at worst, or just a lower level API at best.

Untimately, like I said in the first paragraph, the API defines the RPC contract that needs to be versioned to support server and client having different versions of NuGet. Therefore to me it's immaterial.

designs/NuGet-PackageFileService.md
```diff
@@ -0,0 +1,112 @@
...
+### Fetching Files
+Defined a new interface for this service:
+
+```C#
+namespace NuGet.VisualStudio.Internal.Contracts
+{
+    public interface INuGetPackageFileService : IDisposable
+    {
+        ValueTask<Stream?> GetPackageIconAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
+        ValueTask<Stream?> GetEmbeddedLicenseAsync(PackageIdentity packageIdentity, CancellationToken cancellationToken);
```
zivkan, line 44: Given that we expect licenses to be text, why not use a string return type.  If the license contents is very long, is there a perf benefit to using a stream?

Similarly, from an API design point of view, since we expect icons to be images, we could use that data type, but knowing that this is an RPC contract, I understand wanting to keep the byte[] -> image code out of the RPC converters.

JonDouglas:
I'm working to close out stale PRs. Can we get appropriate reviewers assigned to this PR & look for approval to merge the doc or close out the PR until a future date?

nkolev92:
Given that this is codespaces, I think we can close this one.

zivkan:
Closing as stale.

Issue #10316, opened 2020-11-21
dotnet nuget verify is too quiet


loic-sharma:
## Details about Problem

NuGet product used: dotnet.exe

dotnet.exe --version (if appropriate): 5.0.100

## Detailed repro steps so we can see the same problem

`dotnet nuget verify` does not output anything by default:

```
PS > dotnet nuget verify --all .\newtonsoft.json.12.0.3.nupkg
PS > dotnet nuget verify .\newtonsoft.json.12.0.3.nupkg
```

No output is given at all for these commands. This is unfortunate as the `verify` command's output provides useful information when configuring `trusted-signers` (see this [blog post](https://devblogs.microsoft.com/nuget/lock-down-your-dependencies-using-configurable-trust-policies/#configure-trusted-package-authors)).

However, command output is shown if I increase the verbosity to `normal`:

```
PS > dotnet nuget verify .\newtonsoft.json.12.0.3.nupkg --verbosity normal

Verifying Newtonsoft.Json.12.0.3
.\newtonsoft.json.12.0.3.nupkg
Signature Hash Algorithm: SHA256
Timestamp: 11/8/2019 4:56:46 PM
Verifying author primary signature's timestamp with timestamping service certificate:
  Subject Name: CN=TIMESTAMP-SHA256-2019-10-15, O="DigiCert, Inc.", C=US
  SHA1 hash: 0325BD505EDA96302DC22F4FA01E4C28BE2834C5
  SHA256 hash: 481F4373272D98586C5364B6C115E82425675AEBFD9FACF7ADC464FA2FFFB8F0
  Issued by: CN=DigiCert SHA2 Assured ID Timestamping CA, OU=www.digicert.com, O=DigiCert Inc, C=US
  Valid from: 9/30/2019 5:00:00 PM to 10/16/2030 5:00:00 PM
Signature type: Author
Verifying the author primary signature with certificate:
  Subject Name: CN=Json.NET (.NET Foundation), O=Json.NET (.NET Foundation), L=Redmond, S=wa, C=US, SERIALNUMBER=603 389 068
  SHA1 hash: 4CFB89FAA49539A58968D81960B3C1258E8F6A34
  SHA256 hash: A3AF7AF11EBB8EF729D2D91548509717E7E0FF55A129ABC3AEAA8A6940267641
  Issued by: CN=.NET Foundation Projects Code Signing CA, O=.NET Foundation, C=US
[...]

Issue #10047, opened 2020-09-18
add tests for VerifyCommandRunner


kartheekp-ms:
Add `unit` and `Funct` tests for [`VerifyCommandRunner`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/VerifyCommand/VerifyCommandRunner.cs)

Issue #8141, opened 2019-05-16
Avoid please in error messages


nkolev92:
Motivation from: https://github.com/dotnet/cli/issues/11336

https://docs.microsoft.com/en-us/windows/desktop/debug/error-message-guidelines

> Avoid the word "please". It can be interpreted to mean that a required action is optional.

We have ~139 please mentions in our resx files (about ~80 are not in the comment field). 
All of these need reviewed and potentially fixed.

Examples

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/Strings.resx#L393
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/Strings.resx#L435
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/Strings.resx#L606

StingyJack:
_EDIT: **TL;DR** -  Please should be avoided because its not necessary in the context of reporting an error or providing the user with a set of options to choose from.  A program is not a person (yet, see "corporations"), and does not have to display any humanizing features like politeness (or verbal abuse). "Please" has nothing to do with "optional", so that style guide is picking the wrong reason for excluding the word._

Kindness and/or courtesy is not a bug. The premise is.

Most uses of the word "please" do not communicate something being entirely optional, there is always a consequence. Usually "please" is followed with a direct or implied command. (EDIT: and implied or explicit consequence)

> Please put your tray tables up and your seats in an upright position. 

...or the flight will not land on time, and you will probably go to jail.

You always have a choice, even with a command given to you by the TSA agent pointing the taser at you. 

There are times when options have less consequence.

> Please prioritize this work after all other bugs have been fixed and feature gaps in package ref have been closed properly.

... Or else? 

"Or else" in this example has been shown to have negligible - but existing - consequence.

Rick-Anderson:
Per the Microsoft style guide:
*Avoid please except in situations where the customer is asked to do something inconvenient or the application or site is to blame for the situation.*

PR #10466, opened 2021-01-19
spec for dotnet nuget verify --verbosity

kartheekp-ms:
https://github.com/NuGet/Client.Engineering/issues/724

spec work for NuGet/Home#10316 issue.


designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,246 @@
...
+
+* Change the output of `dotnet nuget verify` command for various `--verbosity` options.
+
+## Non-goals
+
+* Add `--verbosity` option to `dotnet nuget verify` command.
+
+## Solution
+
+### Situation before this design
```
nkolev92: nit: Logically I think this belongs in the problem background section.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,246 @@
...
+`Certificate -> SHA1 hash`| ❌       | ❌          | ✔️         | ✔️         | ✔️   
+`Certificate -> Issued By`| ❌       | ❌          | ✔️         | ✔️         | ✔️   
+`Certificate -> Subject`| ❌       | ❌          | ✔️         | ✔️         | ✔️   
+`Package name being verified`                    | ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+`Type of signature (author or repository)`| ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+`Certificate -> SHA-256 hash`| ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+`Certificate -> Validity period`| ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+`Certificate -> Service index URL (If applicable)`| ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+
+> [!Note]
```
nkolev92: Not is not part of GH markdown, so it won't render like it does in the docs.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,246 @@
...
+#### Default verbosity | `--verbosity mimimal`
+
+```
+Verifying NuGet.Common.5.9.0-preview.2
+Signature type: Author
+Verifying the author primary signature with certificate:
+  SHA256 hash: 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
+  Valid from: 2/25/2018 4:00:00 PM to 1/27/2021 4:00:00 AM
+Timestamp: 11/17/2020 7:23:10 AM
+Verifying author primary signature's timestamp with timestamping service certificate:
```
nkolev92: This seems redundant imo. 

Is this a `progress` log message, or only at the end?

Either way, I don't think it's log level minimal.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,246 @@
...
+### Example output
+
+This is output of the invocation of `dotnet nuget verify nuget.common.5.9.0-preview.2.nupkg`
+
+#### Default verbosity | `--verbosity mimimal`
+
+```
+Verifying NuGet.Common.5.9.0-preview.2
+Signature type: Author
+Verifying the author primary signature with certificate:
```
nkolev92: Same as verifying author...etc. 

I don't think this belongs in minimal.

nkolev92 (commented):
Very neatly written, well done. 

Couple of suggestions.

mintninthiporn (changes requested):
removed comment as it looks like a spam content.

kartheekp-ms:
tagging @dtivel @loic-sharma @zkat @heng-liu for feedback on this pull request.

designs/Verify-Verbosity.md
loic-sharma: Redundant to what, the author signature? If so, they are different. Once the author signs their package, a trusted timestamp server then stamps the signature with a "Yup I saw this content at timestamp `11/17/2020 7:23:10 AM`". The trusted timestamp server has its own certificate with its own hash, expiration, and validity period.

I'd be okay with hiding the trusted timestamp's certificate info in the minimal log level. You can't add security policies on the trusted timestamp (the trusted timestamp is only used by NuGet to determine if a signature is valid or not). Please double check this with @dtivel though.

nkolev92: My only comment is about this specific line `Verifying author primary signature's timestamp with timestamping service certificate:`. 

That seems like a progress message, like the rest of the output is the actual result.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,240 @@
...
+  SHA256 hash: 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
+  Valid from: 2/25/2018 4:00:00 PM to 1/27/2021 4:00:00 AM
+Timestamp: 11/17/2020 7:23:10 AM
```
loic-sharma: It's not clear which this timestamp is. What do you think of this?

```suggestion
Author primary signature's timestamp: 11/17/2020 7:23:10 AM
```

Same comment for the repository signature timestamp below.

designs/Verify-Verbosity.md
loic-sharma: EDIT: I also like the idea of removing this progress message from minimal output.

<details>
<summary>Original comment...</summary>

Could you explain your reasoning? From my understanding, minimal is the default log level when using the `dotnet nuget verify` command. This output is useful when [configuring trusted signers](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#trustedsigners-section) as I need to know my package's author signature's signing certificate hash:

```xml
<trustedSigners>
    <author name="microsoft">
        <certificate fingerprint="3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
        <certificate fingerprint="AA12DA22A49BCE7D5C1AE64CC1F3D892F150DA76140F210ABD2CBFFCA2C18A27" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
    </author>
</trustedSigners>
```

</details>

designs/Verify-Verbosity.md
loic-sharma: Ah gotcha. I agree :)

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,240 @@
...
+`Verbose`, `Debug` | `d[etailed]`             | Low-importance Messages
+​                   | `diag[nostic]`           | Additional MSBuild-engine information
+
+### Example output
+
+This is output of the invocation of `dotnet nuget verify nuget.common.5.9.0-preview.2.nupkg`
+
+#### Default verbosity | `--verbosity mimimal`
+
+```
```
loic-sharma: What do you think of adding new lines between different sections? For example:

```
Verifying NuGet.Common.5.9.0-preview.2

Signature type: Author
Signing certificate:
  SHA256 hash: 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
  Valid from: 2/25/2018 4:00:00 PM to 1/27/2021 4:00:00 AM
Timestamp: 11/17/2020 7:23:10 AM
Timestamping service certificate:
  SHA256 hash: C474CE76007D02394E0DA5E4DE7C14C680F9E282013CFEF653EF5DB71FDF61F8
  Valid from: 12/22/2017 4:00:00 PM to 3/22/2029 4:59:59 PM

Signature type: Repository
nuget-v3-service-index-url: https://api.nuget.org/v3/index.json
nuget-package-owners: Microsoft, nuget
Signing certificate:
  SHA256 hash: 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
  Valid from: 4/9/2018 5:00:00 PM to 4/14/2021 5:00:00 AM
Timestamp: 12/9/2020 2:32:12 PM
Timestamping service certificate:
  SHA256 hash: C474CE76007D02394E0DA5E4DE7C14C680F9E282013CFEF653EF5DB71FDF61F8
  Valid from: 12/22/2017 4:00:00 PM to 3/22/2029 4:59:59 PM

Successfully verified package 'NuGet.Common.5.9.0-preview.2'.
```

designs/Verify-Verbosity.md
loic-sharma: Alternatively, you could separate out the sections. See: https://github.com/NuGet/Home/pull/10466/files#r564105499

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,240 @@
...
+------------------ | ------------------------ | --------------------
+`Error`, `Warning` | `q[uiet]`                | Errors and Warnings
+`Minimal`          | `m[inimal]`              | High-importance Messages
+`Information`      | `n[ormal]`               | Normal-importance Messages
+`Verbose`, `Debug` | `d[etailed]`             | Low-importance Messages
+​                   | `diag[nostic]`           | Additional MSBuild-engine information
+
+### Example output
+
+This is output of the invocation of `dotnet nuget verify nuget.common.5.9.0-preview.2.nupkg`
```
loic-sharma: Could you also include the output for different failure scenarios? For example, say a package has been tampered, a certificate does not chain to a trusted root, the signature doesn't have a trusted timestamp, etc...

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### Default verbosity | `--verbosity mimimal`
+
+```
+Verifying NuGet.Common.5.9.0-preview.2
+Signature type: Author
+Verifying the author primary signature with certificate:
+  SHA256 hash: 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
```
loic-sharma: Should we add subject names to the minimal output for human readability purposes?

designs/Verify-Verbosity.md
kartheekp-ms: I like the idea of separating out the sections as it improves readability.

designs/Verify-Verbosity.md
kartheekp-ms: sure, we can do that but I followed @dtivel suggestion https://github.com/NuGet/NuGet.Client/pull/3774#issuecomment-740216665

dtivel: It's totally fine with me to add subject name, but it needs to be on every signature and in both tools.  Given that a package can have up to 4 total signatures (i.e.:  author, author timestamp, repository countersignature, and repository countersignature timestamp), this will increase output by up to 4 lines per signed package.  Subject name is more human readable than a fingerprint, so it seems like an acceptable tradeoff for slightly less minimalistic logging.

designs/Verify-Verbosity.md
loic-sharma: ⚠ Given the Debian issue (https://github.com/NuGet/Home/issues/10491) please add this

kartheekp-ms: Thanks @loic-sharma. I added example output for couple of failure scenarios in the latest commit.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+#### `Verifying tampered package`
+
+<details>
+<summary>output</summary>
+
+```
+dotnet nuget verify tampered.12.0.1.nupkg -v -detailed
+error: NU3008: The package integrity check failed.
+error: Package signature validation failed.
+```
```
loic-sharma: Is this full the output you'd see even at the `detailed` level?

loic-sharma: Many folks on the Debian issue were confused due to the lack of information included in NuGet's error messages. Most folks thought the issue was a bug in NuGet when in reality the problem was due to Debian not trusting a certificate. Can we add more information in this error message?

For example:

```suggestion
dotnet nuget verify tampered.12.0.1.nupkg -v -detailed
error: NU3008: The package integrity check failed. Please ensure that the package has not been tampered since signing.
error: Package signature validation failed.
```

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+
+#### `Verifying tampered package`
+
+<details>
+<summary>output</summary>
+
+```
+dotnet nuget verify tampered.12.0.1.nupkg -v -detailed
+error: NU3008: The package integrity check failed.
+error: Package signature validation failed.
```
loic-sharma: Many folks on the Debian issue were confused due to the lack of information included in NuGet's error messages. Most folks thought the issue was a bug in NuGet when in reality the problem was due to Debian not trusting a certificate. Can we add more information in this error message?

For example:

```suggestion
dotnet nuget verify tampered.12.0.1.nupkg -v -detailed
error: NU3008: The package integrity check failed. Please ensure the package has not been tampered since it was signed.
error: Package signature validation failed.
```

designs/Verify-Verbosity.md
kartheekp-ms: That is the current output at the `detailed` level. I forgot to add new output in the spec. I will add it in the next commit.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+<summary>output</summary>
+
+```
+dotnet nuget verify "package.nupkg" -v n
+Verifying packageA.1.0.0
+C:\Users\kapenaga\Downloads\package.nupkg
+Signature Hash Algorithm: SHA256
+
+Signature type: Author
+Verifying the author primary signature with certificate:
```
loic-sharma: I recommend distinguishing between the signing and timestamping certificates wherever possible:

```suggestion
Verifying the author primary signature with signing certificate:
```

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+  Subject Name: CN=test
+  SHA1 hash: B0A2B3B1695AB8361B1D2B14A9F5D64136E26380
+  SHA256 hash: 89A2B6EB529E0AEBF0D11C8A18A846C7B8D1290791B6BF494BAFEC299F2EAAB2
+  Issued by: CN=test
+  Valid from: 1/29/2021 12:28:11 PM to 1/29/2021 1:28:11 PM
+error: NU3018: The author primary signature found a chain building
```
loic-sharma: Can we add a new line between the signing certificate's details and the list of errors?

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+issue: UntrustedRoot: A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.
```
loic-sharma: Most NuGet users likely don't understand chain building. This message includes a bunch of complicated info that is not very beginner friendly. What do you think of simplifying it to something like:

```suggestion
error: NU3018: The author primary signature's signing certificate is not trusted by the trust provider. 
```

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+error: NU3037: The author primary signature validity period has expired.
+warn : NU3027: The signature should be timestamped to enable long-term signature validity after the certificate has expired.
```
loic-sharma:
```suggestion
warn : NU3027: The signature should be timestamped to enable long-term signature validity after the signing certificate has expired.
```

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+
+```
+dotnet nuget verify tampered.12.0.1.nupkg -v -detailed
+error: NU3008: The package integrity check failed.
+error: Package signature validation failed.
+```
+
+</details>
+
+#### `Verifying author signed package with no timestamp and untrusted certificate root`
```
loic-sharma:
```suggestion
#### `Verifying author signed package with no timestamp and untrusted signing certificate root`
```

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+error: NU3018: The author primary signature found a chain building
+issue: UntrustedRoot: A certificate chain processed, but terminated in a root certificate which is not trusted by the trust provider.
+error: NU3037: The author primary signature validity period has expired.
+warn : NU3027: The signature should be timestamped to enable long-term signature validity after the certificate has expired.
+Finished with 2 errors and 1 warnings.
+
+error: Package signature validation failed.
+```
+
+</details>
```
loic-sharma: Can we also include the Debian case before and after the incident:

* Valid author signing certificate + untrusted timestamping certificate (you should see the `NU3027` warning in this case)
* Expired author signing certificate + untrusted timestamping certificate

loic-sharma: What are all the cases where a package was accepted on nuget.org but later fails package verification on a customer's machines? Scenarios I can think of:

* The machine distrusts the signing/timestamping certificate's root (Debian case)
* A certificate authority revokes a certificate in the signing/timestamping certificate's chain

Is there anything else? Please add example output for all of these scenarios

designs/Verify-Verbosity.md
kartheekp-ms: Thanks, Damon, for the feedback. The following fields are `included` for every certificate in the `minimal verbosity`.
- `Subject Name`
- `SHA256 hash`
- `Validity`

The following fields are `excluded` for every certificate in the `minimal verbosity`.
- `SHA1 hash`
- `Issued by`

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+The provided `--verbosity` value will include to NuGet log messages with the following levels:
+
+​             | `q[uiet]` | `m[inimal]` | `n[ormal]` | `d[etails]` | `diag[nostic]`
+------------- | --------- | ----------- | ---------- | ----------- | --------------
+`Error`       | ✔️        | ✔️         | ✔️         | ✔️         | ✔️   
+`Warning`     | ✔️        | ✔️         | ✔️         | ✔️         | ✔️   
+`Minimal`     | ❌        | ✔️         | ✔️         | ✔️         | ✔️   
+`Information` | ❌        | ❌         | ✔️         | ✔️         | ✔️   
+`Verbose`     | ❌        | ❌         | ❌         | ✔️         | ✔️   
+`Debug`       | ❌        | ❌         | ❌         | ✔️         | ✔️
```
zivkan: It's out of scope for this design spec, but this looks like a bug to me. Why shouldn't NuGet's "debug" logging be limited to msbuild's "diagnostic" logging level, just as what msbuild does itself?

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+`Certificate -> Service index URL (If applicable)`| ❌       | ✔️          | ✔️         | ✔️         | ✔️   
+```
+Once this spec has been implemented the output of `nuget.exe verify` command for various verbosity levels will change and be in sync with `dotnet nuget verify` command.
+```
+
+### Log level mapping - The following details are copied from [here](https://github.com/NuGet/Home/blob/dev/designs/Package-List-Verbosity.md#log-level-mapping). Thanks to [Joel Verhagen](https://github.com/joelverhagen) for the detailed information.
+
+The provided `--verbosity` value will include to NuGet log messages with the following levels:
+
+​             | `q[uiet]` | `m[inimal]` | `n[ormal]` | `d[etails]` | `diag[nostic]`
```
zivkan: Isn't it "detailed", not "details"?  The msbuild table below uses "detailed".

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+  Valid from: 12/22/2017 4:00:00 PM to 3/22/2029 4:59:59 PM
+
+Successfully verified package 'NuGet.Common.5.9.0-preview.2'.
+```
+
+</details>
+
+#### `--verbosity quiet`
+
+* No output unless there are any `errors` or `warnings`.
```
zivkan: How about outputting just :

> Successfully verified package 'NuGet.Common.5.9.0-preview.2'

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
+# `dotnet nuget verify --verbosity`
+
+* Status: **In Review**
+* Author: [Kartheek Penagamuri](https://github.com/kartheekp-ms)
+* Issue: [#10316](https://github.com/NuGet/Home/issues/10316) - dotnet nuget verify is too quiet
+
+## Problem background
+
+`dotnet nuget verify` command doesn't provide any output by default due to logical error in the implementation. If the user didn't pass value for `--verbosity` option, then the log level is set to [`LogLevel.Minimal`](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Commands/Signing/VerifyCommand.cs#L58). Customers need information about packages’ signatures to [manage package trust boundaries](https://docs.microsoft.com/en-us/nuget/consume-packages/installing-signed-packages).
```
zivkan: How is it a logical error that it defaults to `LogLevel.Mininal`?  That's the default verbosity for [`dotnet build`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-build), [`dotnet restore`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore), [`dotnet test`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-test), [`dotnet publish`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-publish). [`dotnet pack`'s docs don't specifiy the default verbosity](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-pack), and interestingly [`dotnet clean`](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-clean) defaults to normal verbosity.

As a dotnet cli power-user, it seems normal to me that the default logging level is mininal.

zivkan: Regarding the last sentence, customers using `dotnet nuget verify` to configure client trust policies, is this the primary or secondary/edge use case?

If it's the primary use case, it seems to me this is hard to discover and improving the command name to something like `get-signature` would help with discoverability, and expectations for what default output should be.

Otherwise, `dotnet nuget verify` sounds like the primary scenario is check if the package signature is valid or not, hence a single line of output per package would be the expected output at default verbosity.

I have no idea how customer are using `nuget.exe verify` or `dotnet nuget verify` "in the real world".

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+trace:       SHA256 hash: 51044706BD237B91B89B781337E6D62656C69F0FCFFBE8E43741367948127862
+trace:       Issued by: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:       Valid from: 10/22/2013 5:00:00 AM to 10/22/2028 5:00:00 AM
+trace: 
+trace:             Subject Name: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:             SHA1 hash: 0563B8630D62D75ABBC8AB1E4BDFB5A899B24D43
+trace:             SHA256 hash: 3E9099B5015E8F486C00BCEA9D111EE721FABA355A89BCF1DF69561E3DC6325C
+trace:             Issued by: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:             Valid from: 11/9/2006 4:00:00 PM to 11/9/2031 4:00:00 PM
+trace:
```
zivkan: Is the "trace:" prefix useful to customers?

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,302 @@
...
+trace:       SHA1 hash: 92C1588E85AF2201CE7915E8538B492F605B80C6
+trace:       SHA256 hash: 51044706BD237B91B89B781337E6D62656C69F0FCFFBE8E43741367948127862
+trace:       Issued by: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:       Valid from: 10/22/2013 5:00:00 AM to 10/22/2028 5:00:00 AM
+trace: 
+trace:             Subject Name: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:             SHA1 hash: 0563B8630D62D75ABBC8AB1E4BDFB5A899B24D43
+trace:             SHA256 hash: 3E9099B5015E8F486C00BCEA9D111EE721FABA355A89BCF1DF69561E3DC6325C
+trace:             Issued by: CN=DigiCert Assured ID Root CA, OU=www.digicert.com, O=DigiCert Inc, C=US
+trace:             Valid from: 11/9/2006 4:00:00 PM to 11/9/2031 4:00:00 PM
```
zivkan: Assuming this is showing a chain of 3 certificates, there is an empty line between the 2nd and 3rd certificate, but no empty line between the 1st and 2nd.

The "normal verbosity" example above doesn't use whitespace, except between the author signature and repo signature blocks. But each of those two blocks have "signing cert, timestamp, time server cert" all as one big block without whitespace.

So, it feels inconsistent to me to have an empty line between certs in the cert chain, especially since the cert chain is already represented by indenting the start of the line.

designs/Verify-Verbosity.md
zivkan: Specifically, can we give different errors depending if the signing certificate's chain is not trusted, compared to the timestamp's certificate chain not being trusted?

designs/Verify-Verbosity.md
nkolev92: Because there's no API for that. At least TaskLoggingHelper does not.

nkolev92: Also if you look at the description in the link below, it says that diagnostic is `Additional MSBuild-engine information`


https://docs.microsoft.com/en-us/visualstudio/msbuild/obtaining-build-logs-with-msbuild?view=vs-2019#verbosity-settings

designs/Verify-Verbosity.md
kartheekp-ms: Removed `Verifying author primary signature's timestamp with timestamping service certificate: `log message from `minimal` verbosity output.

designs/Verify-Verbosity.md
kartheekp-ms: Removed `Verifying the author primary signature with certificate:` log message from `minimal` verbosity output.

designs/Verify-Verbosity.md
kartheekp-ms: I like the idea of separating out the sections as it improves readability.

designs/Verify-Verbosity.md
kartheekp-ms: Added `Please ensure that the package has not been tampered since signing. ` to the error message.

designs/Verify-Verbosity.md
donnie-msft: I think `quiet` is supposed to default to empty output for success. I'd check similar commands to verify.

designs/Verify-Verbosity.md
kartheekp-ms: We have following error codes for different scenarios.

- [NU3018](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3018) for Author's primary signature chain not trusted.

- [NU3028](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028) for `Certificate chain building failed for the timestamp signature. The timestamp signing certificate is untrusted, revoked, or revocation information for the certificate is unavailable.`

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,306 @@
...
+Signature type: Repository
+nuget-v3-service-index-url: https://api.nuget.org/v3/index.json
+nuget-package-owners: Microsoft, nuget
+  Subject Name: CN=NuGet.org Repository by Microsoft, O=NuGet.org Repository by Microsoft, L=Redmond, S=Washington, C=US
+  SHA256 hash: 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
+  Valid from: 4/9/2018 5:00:00 PM to 4/14/2021 5:00:00 AM
+Timestamp: 12/9/2020 2:32:12 PM
+  Subject Name: CN=Symantec SHA256 TimeStamping Signer - G3, OU=Symantec Trust Network, O=Symantec Corporation, C=US
+  SHA256 hash: C474CE76007D02394E0DA5E4DE7C14C680F9E282013CFEF653EF5DB71FDF61F8
+  Valid from: 12/22/2017 4:00:00 PM to 3/22/2029 4:59:59 PM
```
loic-sharma: Do we need to show the trusted timestamp at the default logging level? Can we replace `nuget-v3-service-index-url:` with `Service index:` and `nuget-package-owners` with `Owners:`? Should we change `Subject Name` to `Subject name` to match the capitalization of the other fields?

designs/Verify-Verbosity.md
kartheekp-ms: added couple of failure cases related to debian issue to the spec.

kartheekp-ms: added couple of failure cases related to debian issue to the spec.

designs/Verify-Verbosity.md
kartheekp-ms: Fixed in the latest commit.

designs/Verify-Verbosity.md
kartheekp-ms: As per [MSBuild verbosity settings](https://docs.microsoft.com/en-us/visualstudio/msbuild/obtaining-build-logs-with-msbuild?view=vs-2019#verbosity-settings), only `warnings` or `errors` are displayed for `quiet` verbosity setting. IMO, the default verbosity setting `minimal` provides some output. If the user overrides `the verbosity to quiet` then they may prefer no output for success scenarios. @JonDouglas, any thoughts on this one?

designs/Verify-Verbosity.md
kartheekp-ms: The reason this is a logical error is explained here https://github.com/NuGet/Home/issues/10316#issuecomment-732359220. Please let me know the feedback. Happy to change the text accordingly.

designs/Verify-Verbosity.md
kartheekp-ms: > Can we replace nuget-v3-service-index-url: with Service index: and nuget-package-owners with Owners:? Should we change Subject Name to Subject name to match the capitalization of the other fields?

Fixed in the latest commit.

designs/Verify-Verbosity.md
kartheekp-ms: It looks the prefix "trace: " is configured [here](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/CommandOutputLogger.cs#L79). This was defined in mid of 2015 in this [commit](https://github.com/NuGet/NuGet.Client/commit/ee9ed7fea1f92d2131502af7e5062c1b080f1c2b). I would consider this as a out-of-scope for this spec because any change affects all the `dotnet` commands. Please let me know your thoughts. I will be happy to create a new issue if we would like to change this functionality.

designs/Verify-Verbosity.md
kartheekp-ms: removed "empty" line between certs when logging certificate chain information for `detailed` verbosity setting.

designs/Verify-Verbosity.md
kartheekp-ms: > Do we need to show the trusted timestamp at the default logging level?

As per @dtivel comment here https://github.com/NuGet/NuGet.Client/pull/3774#issuecomment-740216665, I think we can remove `timestamp` certificate details from `minimum` verbosity setting.

designs/Verify-Verbosity.md
kartheekp-ms: As you have mentioned above, the primary use case of this command is signature verification. Configuring client trust policies is the `secondary/edge use case`. I like your idea for a new command just to obtain signatures for configuring the client trust policies. copying @JonDouglas just in case if there has been customer ask around this request.

May not be relevant here but @dtivel recently added a tool to extract certificates from signed NuGet package https://github.com/NuGet/Entropy/pull/45

designs/Verify-Verbosity.md
dtivel: While Loic's comment was based on chain building error messages --- which can be difficult to interpret ---, his comment does apply generally.

In this case, NU3008 is not a chain building warning or error.  NU3008 indicates that the package has been tampered with since being signed, which seems like a relatively straightforward error when compared with [NU3028](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028).

The sentence you've added has a few problems:
1. `Please` should be avoided in error messages.  See https://github.com/NuGet/Home/issues/8141.
2. The new sentence tells a user to ensure that a package has not been tampered since signing, but [_that is a role of the signature --- to provide an integrity guarantee._](https://en.wikipedia.org/wiki/Digital_signature#Integrity)  How else would a typical user ensure this?

If you feel that the current error message is not helpful enough, you could add the sentence:  `The package has been tampered with since being signed.`

designs/Verify-Verbosity.md
dtivel: [NU3018](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3018) is for chain building issues with a primary signature.  The primary signature can be author, repository or "other".
[NU3035](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3035) is for chain building issues with a repository countersignature.
[NU3028](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028) is for chain building issues with a timestamp signature.

designs/Verify-Verbosity.md
dtivel: For any NU messages you update, be sure to update NU docs too.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,416 @@
...
+
+* Change the output of `dotnet nuget verify` command for various `--verbosity` options.
+
+## Non-goals
+
+* Add `--verbosity` option to `dotnet nuget verify` command.
+
+## Solution
+
+The details that should be displayed on each verbosity level are described below. Each level should display the same as the level below plus whatever is specified in that level. In that sense, `quiet` will be give the less amount of information, while `diagnostic` the most.
```
dtivel: less -> least

designs/Verify-Verbosity.md
kartheekp-ms: I have created three doc issues them and assigned them to [epic](https://github.com/NuGet/Client.Engineering/issues/814).

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,416 @@
...
+Timestamp: 2/3/2021 3:36:12 PM
+Verifying author primary signature's timestamp with timestamping service certificate:
+  Subject name: CN=NuGet Test Root Certificate Authority (40998d55-3d73-4a3b-a689-55e30c1fac3c), O=NuGet, L=Redmond, S=WA, C=US
+  SHA1 hash: 6B2378A3DC9CA185252BB66F24F262D129165B5B
+  SHA256 hash: 61B18DE3D814FA7960C6ED62DB20BEA6D0F8D65F678464D7D7C9227E7D5DEFBD
+  Issued by: CN=NuGet Test Root Certificate Authority (40998d55-3d73-4a3b-a689-55e30c1fac3c), O=NuGet, L=Redmond, S=WA, C=US
+  Valid from: 2/3/2021 3:36:11 PM to 12/31/2099 4:00:00 PM
+
+error: NU3018: The author primary signature's signing certificate is not trusted by the trust provider.
+error: NU3028: The author primary signature's timestamp signature's certificate is not trusted by the trust provider.
```
aortiz-msft: This is really hard to parse, and because of that, I anticipate that it will localize incorrectly for many languages. Can we make this easier to understand or separate into 2 sentences?

aortiz-msft (dismissed):

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,416 @@
...
+
+## Non-goals
+
+* Add `--verbosity` option to `dotnet nuget verify` command.
+
+## Solution
+
+The details that should be displayed on each verbosity level are described below. Each level should display the same as the level below plus whatever is specified in that level. In that sense, `quiet` will be give the least amount of information, while `diagnostic` the most.
+
+​                                  | `q[uiet]` | `m[inimal]` | `n[ormal]` | `d[etailed]` | `diag[nostic]`
```
JonDouglas: Great table, thank you so much for providing it!

The big question on this spec is..."What is considered the minimal amount of input for users to acknowledge verification?". I think @dtivel already have a proposal here which best matches the expectations:

https://github.com/NuGet/NuGet.Client/pull/3774#issuecomment-740216665

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,416 @@
...
+<details>
+<summary>After incident output</summary>
+
+```
+dotnet nuget verify "AuthorExpired.1.0.0.nupkg" -v n
+Verifying packageA.1.0.0
+C:\Users\kapenaga\Downloads\package.nupkg
+Signature Hash Algorithm: SHA256
+
+Signature type: Author
```
JonDouglas: Anyway to make this more human-readable? Some thoughts:

- Bold where needed
- Use color with intention
- ASCII art where it makes sense
- Symbols or emoji if they apply

JonDouglas: Main reason is that I think many people interpret this output as already very verbose even with minimal output. Bolding, spacing, and coloring the important parts I think would help with verbosities.

designs/Verify-Verbosity.md
kartheekp-ms: @aortiz-msft - How about `error: NU3028: The author primary signature is timestamped. However, the timestamp signature's certificate is not trusted by the trust provider.` message?

designs/Verify-Verbosity.md
kartheekp-ms: @JonDouglas - Thanks for the feedback, I really like your thoughts on how to make the log messages human-readable. TBH, I think we should definitely consider these suggestions for all the `dotnet` commands in another effort because `Logging framework is shared` and any change impacts other commands. We should discuss these ideas with `dotnet` team as they own the CLI experience. Please let me know what you think, I will be happy to engage in further conversations.

designs/Verify-Verbosity.md
kartheekp-ms: I followed Damon's proposal in the comment to come up the log messages displayed for `minimal` verbosity.

designs/Verify-Verbosity.md
kartheekp-ms: https://github.com/NuGet/Home/pull/10466#discussion_r593445534

designs/Verify-Verbosity.md
kartheekp-ms: I added few examples at the end in the spec with output for various verbosity levels.

designs/Verify-Verbosity.md
JonDouglas: Absolutely. This is a wonderful idea.

designs/Verify-Verbosity.md
kartheekp-ms: Updated the log for `normal` verbosity instead of `detailed` log.

designs/Verify-Verbosity.md
kartheekp-ms: > The machine distrusts the signing/timestamping certificate's root (Debian case)

I added example output for this case under `Debian case - Verifying expired author signed package with untrusted timestamping signing certificate` scenario.

> A certificate authority revokes a certificate in the signing/timestamping certificate's chain

This scenario falls under [NU3028](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3028) error code for handling timestamp signature certificate's chain building issues such as `untrusted, revoked, or revocation information for the certificate is unavailable`.

Damon added more details about various error codes for chain building issues in https://github.com/NuGet/Home/pull/10466#discussion_r591745903

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,454 @@
...
+Timestamp: 11/8/2019 5:28:02 PM
+Verifying repository countersignature's timestamp with timestamping service certificate:
+  Subject Name: CN=Symantec SHA256 TimeStamping Signer - G3, OU=Symantec Trust Network, O=Symantec Corporation, C=US
+  SHA1 hash: A9A4121063D71D48E8529A4681DE803E3E7954B0
+  SHA256 hash: C474CE76007D02394E0DA5E4DE7C14C680F9E282013CFEF653EF5DB71FDF61F8
+  Issued by: CN=Symantec SHA256 TimeStamping CA, OU=Symantec Trust Network, O=Symantec Corporation, C=US
+  Valid from: 12/22/2017 4:00:00 PM to 3/22/2029 4:59:59 PM
+
+Finished with 1 errors and 0 warnings.
+error: NU3008: The package integrity check failed. The package has been tampered with since being signed.
```
loic-sharma: This message is slightly different from what we have on our docs: https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu3008

While tampering is the correct security term, "The package has changed since it was signed" may be clearer for non-security folks. What do you think of this?

```suggestion
error: NU3008: The package integrity check failed. The package has changed since it was signed. Please ensure that the package has not been tampered.
```

donnie-msft (commented):
My feedback today about Documentation was regarding this page: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-verify#options

It does not in any way (AFAICT) indicate the implications of the output specific to the `verify` command.
As a user, I'd want to know what I'm sacrificing by lowering my verbosity. If this is somehow intuitive or natural compared to MSBuild CLI, then maybe we don't need to specify it (but that should be noted in spec).

Either including the whole table (below) in some format, or summarizing the deltas for each verbosity in written form would work. Examples: "d/diag show all output" "normal omits cert chain info", "minimal removes timestamping output", and so on..

image

Also, adding a Documentation section to the spec, and noting how this page will be changed (or not and why) would be great.

designs/Verify-Verbosity.md
kartheekp-ms: @loic-sharma proposed `The author primary signature's timestamping certificate is not trusted by the trust provider. ` error message.

designs/Verify-Verbosity.md
kartheekp-ms: @dtivel suggested `error: NU3008: The package integrity check failed. Please ensure that the package has not been tampered since signing.` message in https://github.com/NuGet/Home/pull/10466#discussion_r591740444. Happy to engage in further conversations if we need to tweak the error message.

designs/Verify-Verbosity.md
kartheekp-ms: @loic-sharma - I have updated the spec with Debian case before and after the incident. Please note that, `dotnet nuget verify` or `nuget.exe verify` command operates under [default policy](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Packaging/Signing/Verification/SignedPackageVerifierSettings.cs#L231-L246) and hence `NU3028: The author primary signature's timestamping certificate is not trusted by the trust provider` is considered as error.

Thanks to @heng-liu for helping me to repro Debian before incident case.

kartheekp-ms:
@donnie-msft - When I browsed docs for other `dotnet nuget` commands, I didn't find information about the differences in output for various verbosity levels. Given that this command outputs useful information for package consumers at various verbosity levels, I think It will be helpful to add the table you mentioned in the comment in some format to the docs.

I think I may not be able to add `Documentation` section to this spec unfortunately due to the following reasons.

- The timelines to complete the implementation of this spec.
- I am working with Jon and internal signing experts if we need to word smith the errors messages even further. This activity is WIP.

I have already created docs' issues and linked to the [tracking epic](https://github.com/NuGet/Client.Engineering/issues/814). The goal is complete all tasks in the epic this sprint except https://github.com/nuget/home/issues/10047. 

Please let me know what you think so that I can plan next steps accordingly.

kartheekp-ms:
The default `verbosity` for `dotnet restore` is `minimal`. I am considering `minimal` as default `verbosity` for `dotnet nuget verify` command as well unless team thinks otherwise.

//cc @zkat since this point came up during spec review meeting.

donnie-msft:
> I may not be able to add `Documentation` section to this spec unfortunately due to the following reasons

I'm a little confused since you say you can't add this, but you did create https://github.com/nuget/docs.microsoft.com-nuget/issues/2359.
I commented there.

Overall, it's you and the PM who will decide here, but I feel it's odd to have specific output variations undocumented, especially in the security space.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,410 @@
...
+Signature type: Author
+Verifying the author primary signature with signing certificate:
+  Subject name: CN=test
+  SHA1 hash: B0A2B3B1695AB8361B1D2B14A9F5D64136E26380
+  SHA256 hash: 89A2B6EB529E0AEBF0D11C8A18A846C7B8D1290791B6BF494BAFEC299F2EAAB2
+  Issued by: CN=test
+  Valid from: 1/29/2021 12:28:11 PM to 02/29/2021 1:28:11 PM
+Timestamp: 2/3/2021 3:36:12 PM
+Verifying author primary signature's timestamp with timestamping service certificate:
+  Subject name: CN=NuGet Test Root Certificate Authority (40998d55-3d73-4a3b-a689-55e30c1fac3c), O=NuGet, L=Redmond, S=WA, C=US
```
bartonjs: Maybe I missed something, but why is this the root?  (Is the NuGet test timestamp server signing with the root cert?)

kartheekp-ms: `NuGet Test Root Certificate Authority` is used as root for testing purposes.

kartheekp-ms:
@NuGet/nuget-client - Can I get approval on this spec so that I merge it into `dev` branch. FYI - This spec was implemented in PR https://github.com/NuGet/NuGet.Client/pull/3992 which was merged 3 days ago.

designs/Verify-Verbosity.md
kartheekp-ms: @bartonjs - Thank you for taking the time to review this spec. Please let us know if you have any additional questions so that I can address them ASAP.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,410 @@
...
+`Author/Repository Certificate -> Issued By`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Timestamp Certificate -> Issued By`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Timestamp Certificate -> SHA-256 hash`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Timestamp Certificate -> Validity period`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Timestamp Certificate -> SHA1 hash`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Timestamp Certificate -> Subject name`| ❌       | ❌          | ✔️         | ✔️         | ✔️
+`Author/Repository Certificate -> Subject name`| ❌       | ✔️          | ✔️         | ✔️         | ✔️
+`Author/Repository Certificate -> SHA-256 hash`| ❌       | ✔️          | ✔️         | ✔️         | ✔️
+`Author/Repository Certificate -> Validity period`| ❌       | ✔️          | ✔️         | ✔️         | ✔️
+`Author/Repository Certificate -> Service index URL (If applicable)`| ❌       | ✔️          | ✔️         | ✔️         | ✔️
```
dtivel: This should be:  Repository Certificate -> Service index URL

Author certificates won't have a service index URL and repository certificates MUST have this attribute; there's no "if applicable" --- it's always applicable.

designs/Verify-Verbosity.md
```diff
@@ -0,0 +1,410 @@
...
+`Package name being verified`                    | ❌       | ✔️          | ✔️         | ✔️         | ✔️
+`Type of signature (author or repository)`| ❌       | ✔️          | ✔️         | ✔️         | ✔️
+
+```
+Once this spec has been implemented the output of `nuget.exe verify` command for various verbosity levels will change and be in sync with `dotnet nuget verify` command.
+```
```
dtivel: I don't think this needs to be formatted as a code block.

designs/Verify-Verbosity.md
dtivel: Although "the Debian case" refers to an [incident from earlier this year](https://github.com/NuGet/Home/issues/10491) where Symantec CA distrust was first reported on Debian family distributions, we should not make the good name of Debian synonymous with the incident.  It would be better to label this the "Symantec CA distrust case."

dtivel (approved):

kartheekp-ms:
@aortiz-msft - This spec has been implemented and got approval Damon. I am unable to merge the PR because you have requested changes earlier. Please review the PR again and let me know your feedback so that I can merge it.

JonDouglas (approved):
Please merge when ready! @kartheekp-ms

PR #10496, opened 2021-01-28
Add project boards spec

drewgillies:
For discussion--note that all project boards in this discussion are private for now, until we land on an approach. All comments welcome.


designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
+# Using GitHub Project Boards for Customer Visibility and Tracking
+
+Our project boards will need to be created and accessed at the organizational level, not in any one repo (i.e. not in Home), because this is a restriction on project boards--if they're in a repo, they will be locked only to issues from that repo. Here's where ours will live: [https://github.com/orgs/NuGet/projects](https://github.com/orgs/NuGet/projects)
```
JonDouglas, line 3: Projects should be inside NuGet/Home so the public can see them.

designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
...
+
+We can use a simple or extended kanban column sets with automations (these are simply a template choice when a board is created), as we can see used in projects in other organizations:
+- https://github.com/dotnet/msbuild/projects
+- https://github.com/dotnet/roslyn/projects
```
JonDouglas, line 7: These are great examples of how projects should be used inside repos!

designs/ProjectBoards.md
JonDouglas: image

designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
...
+...and there's nothing stopping us from a complete rework of columns for any given project if that's helpful. Let's keep in mind simplicity where we can, and kanban columns are a pretty well-known approach. Here's a very simple example with automations, and additional status columns could be added if useful:
+
+![](../resources/ProjectBoards/kanban.png)
+
+Automations can extend to reviewing columns as well, e.g.:
+
+![](../resources/ProjectBoards/fullkanban.png)
+
+## A project per feature area
+Because issues can belong to multiple projects, and because automation in project boards will cause issues to move through various kanban columns, the overhead is low in tracking them separate locations.
```
JonDouglas, line 18: I would say this can be more simplified too.

Vulnerabilities Audit (PVA) is a single e2e feature that has components in VS, CLI, and NuGet.org. We can track that in a kanban board and have a single project?

designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
...
+
+![](../resources/ProjectBoards/featureareas.png)
+
+
+## Release Tracking
+We can track releases in projects too, adding issues to release that we intend to target.
```
JonDouglas, line 24: Love it!

NuGet 6.0 would be another release to track.

designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
...
+![](../resources/ProjectBoards/featureareas.png)
+
+
+## Release Tracking
+We can track releases in projects too, adding issues to release that we intend to target. 
+
+![](../resources/ProjectBoards/releaseprojects.png)
+
+## Epics, Wikis and Specs
+Our epics will continue to be our landing points for customers. In the case of PVA, it's [here](https://github.com/NuGet/Home/issues/8087) and will have links to project boards. Wikis can be linked to from epics also.
```
JonDouglas, line 29: Wikis are too static & should be removed entirely in my opinion in favor of designs/specs in the home repo.

designs/ProjectBoards.md
```diff
@@ -0,0 +1,32 @@
...
+
+Specs can continue to accumulate in the Home repo's [designs](https://github.com/NuGet/Home/tree/dev/designs) folder, and we should ensure that they are linked to by their respoective issues. This ties their discoverability to their relevance to a customer. If we wish to show a spec in a project board, we can add cards which display the link.
```
JonDouglas, line 31: Sounds good. Designs can link to epics and/or projects (if one exists)

designs/ProjectBoards.md
drewgillies: This won't work. If they're in Home they can only show Home repo items.

JonDouglas: ah crap, how do we make them public? Can we?

drewgillies: Yes--it's a setting: public/private. I just made this one public briefly to test: 
image

designs/ProjectBoards.md
drewgillies: We sure could. Splitting them up was discussed in the meeting which is why I did it. I can recombine - all tasks would be cobbled together.

designs/ProjectBoards.md
drewgillies: Works for me. Should we call this out in the spec?

drewgillies:
@JonDouglas I've collapsed the 3 PVA projects back into a single PVA project as discussed: https://github.com/orgs/NuGet/projects/4

drewgillies:
@JonDouglas I think creating a NuGet 6.0 project may need to fall to the client team as I don't have the context to know what issues to track there. When we adopt this model of public tracking though it would be good add it into the mix. FYI @anangaur @aortiz-msft @chgill-MSFT @jcjiang

JonDouglas:
This looks great, thanks again for kicking it off!

We'll likely add NuGet 6.0 & other projects after planning & start to fill issues into em & help bring more awareness to em in the public light.

designs/ProjectBoards.md
JonDouglas: Naw, I think we just start to get rid of them slowly as they are already outdated mostly.

JonDouglas (approved):
Approved as this looks like a path forward for release & feature tracking in a public light.

designs/ProjectBoards.md
donnie-msft: > Projects should be inside NuGet/Home so the public can see them.

I've checked this in the past with InPrivate browser, and I could see 3 projects. I still see 2 (not Drew's). 
Trying to figure out how Jon is seeing none.
image

JonDouglas: @donnie-msft at the time of my comment, all org projects are marked private. Drew mentioned that you can see them publicly by switching them over. Btw this is for org level not for home level. The limitation of home repo seems to be that we can't include gallery or client repos inside.

https://github.com/orgs/NuGet/projects

donnie-msft: Ok, this thread is confusing, but just wanted to point out Home projects have been public since at least October. But sounds like we wanted orgs for other reasons than that, which is fine

JonDouglas: Basically a home only project would only allow home issues.

An org project would allow issues from all repos (even private)

PR #10552, opened 2021-02-05
Cleanup Proposals

JonDouglas:
This PR cleans up the proposals from the original merge.

It includes fixed images, better explanations, and more.

It also includes a revamp to the proposal template to help describe the expectations of a proposal.

I think there's opportunity here to discuss what we'd like to see in a proposal template. I don't have a good answer, but gave it my best shot without making it too complex.


meta/template.md
```diff
@@ -1,48 +1,38 @@
...
+## How: How should it be implemented?
 
-Declare any future work if applicable.
+A design or implementation for bringing the proposal to life. 
 
-## Open Questions
+## When: When does this ship and what are the milestones?
 
-Any open questions not specifically called out in the design.
+A timeline including ideal dates to ship the proposal & any key milestones (if any).
```
loic-sharma: This proposal process is open to third-parties, right? How will they determine ship dates? Even for folks on the NuGet team, this seems like a hard section to fill out as part of a proposal. What do you think of removing this section?

meta/template.md
```diff
@@ -1,48 +1,38 @@
...
 
-## Considerations
 
-Any other solutions or approaches considered? 
-The rejected designs are often just as important as the proposed ones!
-
-### References
-
-References to related issues, features
```
loic-sharma, line 48: What do you think of keeping the "References" section? In my mind, I want to keep the previous sections to the point so that the reader can follow my thought process. I wouldn't want to overwhelm the reader with many links in the "Problem: ..." or "Why: ..." sections. On the other hand, the "References" section at the bottom is the perfect place to drop miscellaneous links for folks that want to drill deeper on the topic.

meta/template.md
```diff
@@ -1,48 +1,38 @@
...
+## Audience: Who are we building for?
 
-Describe the proposed solution in as much detail as you deem necessary.
+A brief description to help everyone understand who this proposal is for.
 
-## Test Strategy
+## What: What does this look like in the product?
 
-Other than Unit Tests, which automated and manual integration, component, and end-to-end tests will be needed to validate the implementation? Is a special roll-out strategy (including kill-switch, opt-in flag, etc.) needed?
+A low-fidelity mockup or pseduocode that shows how this proposal would look.
```
loic-sharma:
```suggestion
A low-fidelity mockup or pseudocode that shows how this proposal would look.
```

meta/template.md
JonDouglas: Great point. I think we should remove it then.

meta/template.md
JonDouglas: I looked at other design repos & didn't find any references to be needed.

https://github.com/rust-lang/rfcs/tree/master/text
https://github.com/npm/rfcs/tree/latest/accepted
https://github.com/dotnet/designs/tree/main/accepted/2020

I think the bottom may need a Q&A section though?

JonDouglas: Also, i just took a stab at the common things needed. Maybe there's a better combination from the various RFC & design processes already established.

loic-sharma: Q&A would work too. FYI, I don't feel strongly about this feedback so feel free to skip it :)

nkolev92: The `Considerations` and `References` sections were added as a response to a *need* in previous designs. 

In particular, the exact design is not the only thing that matters. Design that were *not* chosen, and the reasoning behind that is equally, if often times not more important. 

The `References` section is there to remind people that not everyone has the same context and that they should try to make the design as complete as possible. 

I think the `Considerations` section is especially valuable. I can see how `References` might not have as much value as others. You can always just add the references in the specific sub section they are necessary for.

meta/template.md
```diff
@@ -1,48 +1,37 @@
...
-# Spec Name
+# Title: A human-readable title for this proposal
 
-* Status: In Review / Reviewed / Implementing / Implemented
-* Author(s): [Author Name](https://github.com/NuGet)
-* Issue: [1](https://github.com/NuGet/Home/issues/1) Issue Title
+[Author Name](https://github.com/<username>) | [GitHub Issue](https://github.com/NuGet/Home/issues/<issue>)
 
-## Problem Background
+## Description: What is it?
```
nkolev92: Feels like the Description, Problem and Why sections are very closely related. 

I think one of the failures of the previous template was the number of sections and maybe we can compact that here?

meta/template.md
```diff
@@ -1,48 +1,37 @@
...
+## Problem: What problem is this solving?
 
-Customers affected by this change.
+A short background of the problem and how your solution will solve it.
 
-## Goals
+## Why: How do we know this is a real problem and worth solving?
 
-Specifically call out the goals for this proposed design.
+A few details on why we're doing this, and what pain points this solves. Use real, concrete examples to make your case.
```
nkolev92: Feels like the why and the problem section are very closely related, and can maybe be merged. 
I think we can provide `instructions` or guidelines as means of comments and not necessarily sections.

meta/template.md
```diff
@@ -1,48 +1,37 @@
...
 
-## Non-Goals
+## Success: How do we know if we've solved this problem?
```
nkolev92: This one is...awkward :D
Do we talk about metrics?

meta/template.md
```diff
@@ -1,48 +1,37 @@
...
 
-Call out things that are not in consideration for this design. 
+A couple ways on how we might measure the success or outcome of this proposal. 
 
-## Solution Overview
+## Audience: Who are we building for?
```
nkolev92: Feels like this belong somewhere with the problem statement.

nkolev92: Likewise, I think defining the scope of the changes is very important, hence why we used to have a goals/non-goals section. I'm not advocating for that, but ideally we have some sections that call out the non-goals in some way, shape or form.

meta/template.md
nkolev92: We tend to continue iterating on designs that have already been PRed, so an open questions section allows the reader to follow what the design covers and what it does not. 

Now, that section is definitely no ideal. The goal of most designs will be to have an empty open questions section, so I'm not particularly sure how this can be captured.

meta/template.md
```diff
@@ -1,48 +1,37 @@
...
+A brief description to help everyone understand who this proposal is for.
 
-## Test Strategy
+## What: What does this look like in the product?
 
-Other than Unit Tests, which automated and manual integration, component, and end-to-end tests will be needed to validate the implementation? Is a special roll-out strategy (including kill-switch, opt-in flag, etc.) needed?
+A low-fidelity mockup or pseudocode that shows how this proposal would look.
 
-## Future Work
+## How: How should it be implemented?
```
nkolev92: The `What` and `How` feel like sections that could be combined.

nkolev92 (commented):
I love that we keep iterating on this! This is making our repo more `presentable`. 

I do have some ideas about the template. 

I think our previous template suffered from having too many sections. Initial instinct is that maybe we're missing out on a few things. 

I do love the ehnanced descriptions of the sections in particular though.

meta/template.md
JonDouglas: There's a great section here on what type of questions are needed:

http://smallcultfollowing.com/babysteps/blog/2018/06/20/proposal-for-a-staged-rfc-process/#the-proposal-point-and-the-change-from-spitballing-to-designing (Rust RFC process)

My approach was to use a typical 1-pager process for a proposal. If anything is more than 1 page as a proposal, it might be hard to understand.

I think the `Considerations` or `Cross Cutting Concerns` / etc should be a section?

For `References`, that's how I was treating it as well. It would be part of the other areas or in the design itself.

meta/template.md
JonDouglas: They are definitely really closely related. 

Description -> Tell me a paragraph about it
Problem -> Tell me what problems it solves
Why -> Tell me why we should solve this problem

These three really help define the motivation & prioritization from the maintainer perspective in my opinion.

I think they can definitely be grouped, I just worry that we would imply people would know how to put that all under one section, so I broke it down like I usually do.

meta/template.md
JonDouglas: How about a `Motivation` section?

meta/template.md
JonDouglas: I think we get rid of it? I just think it helps people think about the "Why we're solving it" more deeply.

meta/template.md
nkolev92: We can maybe consider Header 3 sections that way it's communicated that they are related. 

> I think they can definitely be grouped, I just worry that we would imply people would know how to put that all under one section, so I broke it down like I usually do.

I don't think there's harm in that :) I think it makes sense to break them in this particular way to an extent. I just see this and see a lot of `1 line` why sections, because the problem kind of covered that. 

My proposal would be, include the breakdown, but not by means of section headings, rather comments (either `<!-- here be comment -->` or just regular inline comment.

meta/template.md
nkolev92: I really like `Motivation`!

meta/template.md
nkolev92: I think this particular one is super relevant for our own product processes, especially ones that involve a lot of customer development, but isn't necessarily something that I expect would be easy to propose/follow for external contributors. 

So, I like the gist of `think about this`, but not sure we'd benefit from a section for it.

JonDouglas:
Okay here's another proposal that is more rust inspired.

# Title

- Author Name
- Start Date
- GitHub Issue (If any)
- GitHub PR (If any)

## Summary

A concise, one-paragraph description of the proposal.

## Motivation 

- A short background of the problem and how your solution will solve it.
- A few details on why we're doing this, and what pain points this solves. Use real, concrete examples to make your case.
- A brief description to help everyone understand who this proposal is for.
- A couple ways on how we might measure the success or outcome of this proposal.

## Explanation

### Functional explanation

- A chance to introduce new concepts.
- A functional design explaining the proposal with examples.
- A low-fidelity mockup or pseudocode that shows how this proposal would look.

### Technical explanation

- A technical design explaining the proposal with implementation details.
- A clear interaction model with existing features.

## Drawbacks

- A consideration for not doing this.

## Alternatives / Cross Cutting Concerns

- A comparison of other possible designs.
- A considerations for other designs and the rationale of choosing them or not.
- An understanding of the impact of not doing this.

## Prior Art

- What features exist in other ecosystems and the experiences their communities have had with them?
- Has this been done by another community before?
- What lessons can we learn from other communities about it?
- Any great documentation, blogs, or papers that will help serve as a background?

## Unresolved Questions

- Bikeshedding

## Future Possibilities

- Does this unlock a future possibility or idea you have?

nkolev92:
I personally like that one a lot better.

meta/template.md
nkolev92: Sounds good to me.

zkat:
I like the Rust one because it's already gone through a lot of iteration and it's been proven to be pretty successful ;)

meta/template.md
JonDouglas: I'll include comments for the explanation sections in a new revision.

JonDouglas:
Okay, template is updated. Please tear apart the language (especially in the comments to describe to the poster)

meta/template.md
```diff
@@ -1,48 +1,52 @@
...
+## Motivation 
 
-## Who are the customers
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
 
-Customers affected by this change.
+## Explanation
 
-## Goals
+### Functional explanation
```
nkolev92, line 18: In the past we've never had a functional and technical separation, but I think it's worth a shot. 
Some NuGet features are almost *completely* technical. Others, not so much. 
This spec covers both.
Yay, trying out new things :)

meta/template.md
```diff
@@ -1,48 +1,52 @@
...
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevent to this proposal? -->
 
-## Open Questions
+## Unresolved Questions
 
-Any open questions not specifically called out in the design.
+<!-- What parts of the proposal do you expect to resolve before this gets accepeted? -->
```
nkolev92:
```suggestion
<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
```

nkolev92 (approved):
I like it!
**progress**!!

meta/template.md
JonDouglas: Maybe we should make a note that the explanation itself doesn't have to be both?

Like you can have a proposal purely technical or you can have one purely functional?

loic-sharma (approved):

meta/template.md
nkolev92: I don't mind that :)
Feel free to make a judgment call.

JonDouglas: 🚢it is my motto

JonDouglas:
2 approvals so far. Any further consideration before merge?

PR #10572, opened 2021-02-17
Fix broken relative image links in FloatingVersions-In-PMUI.md

donnie-msft:
Recent work moved resources in our repo. This broke image URLs. Fixing in a spec I'm currently looking at so it will actually render..


JonDouglas (approved):
Sorry for breaking ya!

PR #10573, opened 2021-02-17
Propose NuGet.org portions of TFM spec

drewgillies:
Quick link for viewing the spec: https://github.com/NuGet/Home/blob/dev-drewgil-add-nuget.org-tfms-spec/proposed/2021/NuGet.orgTFMs.md

Addresses the Nuget.org UX portion of: https://github.com/NuGet/NuGetGallery/issues/8387

In brief, we want to list TFMs on the Package Details page, and to arrive at consensus on how best to do that. My 2c: let's put this expander at the top above Dependencies, and show a flat list of friendly names in lexicographical order. Let's discuss! :)


proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
+# NuGet Target Frameworks on Package Details Page 
+
+- Author Name (https://github.com/jcjiang)
+- Start Date (2021-02-16)
+- GitHub Issue (https://github.com/NuGet/NuGetGallery/issues/8387)
+- GitHub PR ()
+
+## Summary
+
+When a developer opens the package details page or tab on NuGet.org or NuGet rooling, they will see what frameworks the package can support.
```
loic-sharma:
```suggestion
When a developer opens the package details page or tab on NuGet.org or NuGet tooling, they will see what frameworks the package can support. 
```

drewgillies: I think tooling shouldn't be here. I stripped out most of them and missed this one.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Summary
+
+When a developer opens the package details page or tab on NuGet.org or NuGet rooling, they will see what frameworks the package can support. 
+
+## Motivation
+
+### Problem: What problem is this solving? 
+
+Developers cannot use the package details page to determine what frameworks a NuGet package can support. In 2021, we live in a world where there are three evolutions of .NET coexisting on NuGet; .NET Framework, .NET Core, and .NET 5.
```
loic-sharma: There's also Mono, Unity, WebAssembly, NativeAOT. All of these runtimes have non-trivial usage and significant implications. I'd consider wording this in a more open-ended way:

```suggestion
Developers cannot use the package details page to determine what frameworks a NuGet package can support. In 2021, we live in a world where there are many evolutions of .NET coexisting on NuGet, such as .NET Framework and .NET 5.
```

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+Developers should be able to determine whether a package works for their project without having to first download it or rely on the author’s documentation. 
+
+### Why: How do we know this is a real problem and worth solving? 
+
+Finding a package that is compatible with your project is a challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today. One of the top NuGet errors issued in telemetry is [NU1202](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1202), which shows that we allow developers to install packages that are not compatible with their project or solution’s target framework. 
+
+### Success: How do we know if we’ve solved this problem? 
+
+- By measuring the change in the percentage of successful installs and lowering the amount of [NU1202](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1202) errors in telemetry.
```
joelverhagen: Does this make sense prior to the filtering work? The information will not be surfaced in VS in this phase so I imagine the VS behavior won't change significantly (not sure though)

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+### Audience: Who are we building for? 
+
+We are building this for the .NET developer who is browsing & making decisions about including a NuGet package into their project based on target framework & supported platform criteria. 
+
+## Explanation
+
+### Functional Explanation
+
+#### Context and Scope 
+When a developer accesses the details page for a NuGet package, they will see information on target frameworks in a dropdown menu. The framework information will be surfaced from the package. In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced.
```
loic-sharma: I'd consider changing `dropdown menu` to `expandable section` as dropdown menu has a slightly different connotation. For example: https://www.w3schools.com/howto/howto_css_dropdown.asp

```suggestion
When a developer accesses the details page for a NuGet package, they will see information on target frameworks in a expandable section. The framework information will be surfaced from the package. In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced. 
```

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Explanation
+
+### Functional Explanation
+
+#### Context and Scope 
+When a developer accesses the details page for a NuGet package, they will see information on target frameworks in a dropdown menu. The framework information will be surfaced from the package. In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced. 
+
+##### Minimal Requirements 
+- List of Frameworks on NuGet.org
```
joelverhagen: - What to do with non-Dependency type packages? e.g. .NET tools?
- What about `tools` assets in dependency packages? will `tools/net40` make the package appear as compatible with net40?

joelverhagen: I don't know if there are any differences between PackageReference and packages.config compatibility. Client team will know for sure. If there is, we should note that this is _PackageReference_ compatibiliy

proposed/2021/NuGet.orgTFMs.md
loic-sharma: > In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced. 

I'm guessing this is referring to target frameworks without a version? Should this be moved down to the `Technical Explanation` section? This is very a specific concern for the `Context and scope` section

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+##### Non-Requirements 
+- List of Platforms on NuGet.org
```
loic-sharma, line 45: I'd consider defining what platform means for folks that aren't on the NuGet team.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+  - Users interviewed did not express need for supported platform information 
+  - Cannot be surfaced directly from package data and would require additional layer of soft calculation from engineering (no clear mapping scheme currently exists between target frameworks and supported platforms) 
+  - Can be revisited in future 
+- Filters for Frameworks on NuGet.org
+    - Users interviewed were ambivalent on filters, requires further investigation. Can be revisited in future
```
loic-sharma: Does this mean users were ambivalent on searching by a specific target framework? This issue seems to indicate customers feel pretty strongly about filtering search results that are incompatible with my project: https://github.com/NuGet/NuGetGallery/issues/3098

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+Example: .NET Framework, .NET Core, .NET Standard, .NET 5 are target frameworks. 
+
+**Supported Platform** – A supported platform is a target that the APIs can be ran on. 
+
+Example: iOS, Android, Windows are platforms. 
+
+### Solution 
+
+- A .NET developer can view Frameworks of a package on NuGet.org: 
+![](../../meta/resources/NuGet.orgTFMs/PackageDetailsWithTFMs.png)
```
joelverhagen, line 54: Expanded by default? (dependencies are not)
Why isn't this the first section, even above dependencies?

proposed/2021/NuGet.orgTFMs.md
loic-sharma: Community members that read this may be sad that this doesn't mention Visual Studio. Are we planning on adding Visual Studio support in the future? Do we have a GitHub issue we can link to so that they can get updates on that?

proposed/2021/NuGet.orgTFMs.md
joelverhagen: If the section is collapsed by default on the UI, we can measure how often users expand the section. We also measure how often the copy command button is clicked. We could consider observing these metrics to gauge success. Just an idea.

proposed/2021/NuGet.orgTFMs.md
joelverhagen: What order are the frameworks in? Lex order? Or some order we predefine?

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Goals and Non-Goals 
+
+#### Goals 
+
+Make it clear to developers whether a package can be installed or updated for their project or solution based on a target framework.
```
joelverhagen, line 62: We should be clear whether we are listing all frameworks that are supported _or_ exact frameworks that have assets.

Consider a simple package with `lib/netstandard1.0/foo.dll`. Are we going to just say `netstandard1.0` or are we going to list all of the frameworks that support `netstandard1.0`.

proposed/2021/NuGet.orgTFMs.md
joelverhagen, line 54: Consider a display mimicking the API docus:
https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=net-5.0#applies-to
There's also Immo's https://apisof.net/catalog/e40ef6ad-bd6e-9f29-7864-b942e37be11f

joelverhagen, line 54: For new users, these may be rather opaque strings. Can we link to a "good" doc that the .NET team has (don't know which one) that explains these various frameworks.

Said, if I install the .NET SDK or Visual Studio and I want to hack away at a project, what good first document can I read to start understand this wonderful ✨ world of TFMs we have.

joelverhagen: I'm sorry, I now see `Unresolved questions`. This covers some of my questions.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+- .NET Standard (netstandard) 
+- .NET 5 (net5.0) 
+
+### Technical Explanation
+
+Packages whose layout follow our [guidelines](https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks) will be examined when uploaded and validated, prior to listing. Target Frameworks will be determined and displayed here.
+
+Note that packages which do not follow the guidelines may have inconsistent results.
+
+### Unresolved questions
```
joelverhagen, line 103: Could we have a general Q&A/FAQ section where these open questions get moved to Q & A when they are resolved? Perhaps there are common questions that the team or community will have about this spec that could be documented.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
+# NuGet Target Frameworks on Package Details Page 
+
+- Author Name (https://github.com/jcjiang)
+- Start Date (2021-02-16)
+- GitHub Issue (https://github.com/NuGet/NuGetGallery/issues/8387)
```
joelverhagen: There are a bajillion issues about package compatibility. Have we scoured them for clues on how to improve this limited-scope proposal? Many asks are about searching or VS UI which are out of scope, but I imagine some feedback can be considered right now.

joelverhagen: E.g. https://github.com/NuGet/NuGetGallery/issues/3098

joelverhagen: This issue have many links to and from it.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Goals and Non-Goals 
+
+#### Goals 
+
+Make it clear to developers whether a package can be installed or updated for their project or solution based on a target framework. 
+
+#### Non-Goals 
+
+Over-complicate what it means to be a compatible package. One target framework + one target platform = compatible package.
```
joelverhagen: This is spicy, lol 🤣

proposed/2021/NuGet.orgTFMs.md
drewgillies: _Do_ we consider `tools/net40` to be compatible with `net40`?

proposed/2021/NuGet.orgTFMs.md
JonDouglas: I believe https://github.com/NuGet/NuGetGallery/issues/4843 is the best issue to describe this scope.

proposed/2021/NuGet.orgTFMs.md
JonDouglas: I think this is largely an artifact of the old spec I wrote. @jcjiang or @drewgillies should be able to clean it up to outline the scope of the details page only for nuget.org.

JonDouglas: Another thought is % of users opening up dependencies vs. frameworks & doing the same you suggest.

proposed/2021/NuGet.orgTFMs.md
JonDouglas: > In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced.

These would be considered v0.0 and should show in the same way.

For unexpected scenarios, we should not show anything.

proposed/2021/NuGet.orgTFMs.md
JonDouglas: Yep. `tools/{TFM}` would represent that the tool is supported on that version of .NET. We should list it the same way in the sense that a tool can be compatible with net5.0 / netcoreapp3.1 / etc.

JonDouglas: @nkolev92 or @zivkan have any thoughts here?

JonDouglas: This should be listed in the "Future Possibilities" section.

proposed/2021/NuGet.orgTFMs.md
JonDouglas: Frameworks are really platforms pre .NET 5. .NET 5+ we can leverage platforms more. 

I think there's a hidden requirement here in which the "platform is implied by the TFM". At least from many people I've talked to, a TFM is synonymous with a platform today but will change in NET 6 drastically.

JonDouglas: In my eyes we do the following:

TFM -> Platforms -> Runtimes

proposed/2021/NuGet.orgTFMs.md
JonDouglas: framework precedence would be a great ordering.

JonDouglas: Looks great. Going back to framework precedence order it by "Latest TFM -> Oldest least compatible TFM"

JonDouglas: https://docs.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions in other words

JonDouglas: https://docs.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions imo

proposed/2021/NuGet.orgTFMs.md
JonDouglas: just netstandard1.0. we don't care about the whole list b/c that can be looked up by users. We should link to tables for any portable assets though like https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

proposed/2021/NuGet.orgTFMs.md
JonDouglas: lol even spicier is that I believe you defined this early on 🌶️

proposed/2021/NuGet.orgTFMs.md
drewgillies: How do we define framework precedence?

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Audience: Who are we building for? 
+
+We are building this for the .NET developer who is browsing & making decisions about including a NuGet package into their project based on target framework & supported platform criteria. 
+
+## Explanation
+
+### Functional Explanation
+
+#### Context and Scope
```
JonDouglas: I think this header should be removed & contents moved into the spec. The context & scope is already defined in the "Summary" & "Motivation" sections.

proposed/2021/NuGet.orgTFMs.md
joelverhagen: oh RIP ⚰️

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+We are building this for the .NET developer who is browsing & making decisions about including a NuGet package into their project based on target framework & supported platform criteria. 
+
+## Explanation
+
+### Functional Explanation
+
+#### Context and Scope 
+When a developer accesses the details page for a NuGet package, they will see information on target frameworks in a dropdown menu. The framework information will be surfaced from the package. In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced. 
+
+##### Minimal Requirements
```
JonDouglas: Requirements header can be dropped & consolidated into the functional design. Only include the minimal requirements, and not the non-goals.

proposed/2021/NuGet.orgTFMs.md
drewgillies: The logistics concern me there--we would need to maintain a dictionary, which could mean that we have the potential for it to contain errors or fall out of maintenance. Those who use abstracted APIs like netstandard would usually understand the abstraction, but we could provide a link to a conversion matrix a la https://docs.microsoft.com/en-us/dotnet/standard/net-standard, which given its location in docs should be kept up to date.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Functional Explanation
+
+#### Context and Scope 
+When a developer accesses the details page for a NuGet package, they will see information on target frameworks in a dropdown menu. The framework information will be surfaced from the package. In cases where versions are not specified (such as .net) or other unexpected scenarios, ‘Version Not Specified’ or similar messages will be surfaced. 
+
+##### Minimal Requirements 
+- List of Frameworks on NuGet.org
+
+##### Non-Requirements
```
JonDouglas: This is redundant & not in the scope of this proposal. We don't need non-requirements.

proposed/2021/NuGet.orgTFMs.md
zivkan: There are some differences, yes.

For example, packages.config will accept a package with `lib/foo.dll`, and this package will be deemed compatible with all TFMs. But PackageReference will completely ignore it, and only accept packages with `lib\<tfm>\foo.dll` (assuming the package doesn't have other assets.

I haven't yet read this spec, so I don't know if my comment makes sense in whatever context this section of the document is, but it's absolutely possible for a package to work only for packages.config, only for PackageReference, or work for both.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,96 @@
...
+
+**Target Framework** – A target framework is a specification of the APIs available to an application or library. 
+
+Example: .NET Framework, .NET Core, .NET Standard, .NET 5 are target frameworks. 
+
+**Supported Platform** – A supported platform is a target that the APIs can be ran on. 
+
+Example: iOS, Android, Windows are platforms. 
+
+### Solution
```
JonDouglas: The solution should be included in the "Functional Design" section. No extra headers are needed!

proposed/2021/NuGet.orgTFMs.md
JonDouglas: https://nugettoolsdev.azurewebsites.net/5.8.0/framework-precedence 

There's a few APIs in NuGet.Frameworks IIRC.

proposed/2021/NuGet.orgTFMs.md
drewgillies: We could move it to https://github.com/NuGet/Home/discussions/new ...

proposed/2021/NuGet.orgTFMs.md
JonDouglas: f in chat.

proposed/2021/NuGet.orgTFMs.md
joelverhagen: Your comment makes sense in whatever context this is 🥇

proposed/2021/NuGet.orgTFMs.md
JonDouglas: I think there's a few missing sections from the new template that would help out here further such as what was considered & what alternatives exist so there's less FAQ/QnA loops. Also the unresolved questions section should keep a tallied list of what comes up in PR reviews like this!

JonDouglas: https://github.com/NuGet/Home/blob/dev/meta/template.md if you're interested!

proposed/2021/NuGet.orgTFMs.md
loic-sharma: f

proposed/2021/NuGet.orgTFMs.md
JonDouglas: There's the other consideration that portable assets != platform assets & therefore we can't just list supported platforms as we can mislead users.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I made this a hybrid between the template and the doc spec, but I agree--it would be good to add those sections back in.

proposed/2021/NuGet.orgTFMs.md
zivkan: I think that framework precedence depends on project target framework. Or at least, with the current discussions about iOS/Mac TFMs in net6.0 it will be.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I love how these 4-year-old conversations just get peeled off in these discussions like they were sitting at top of mind for the encyclopedic amongst us. :-D

proposed/2021/NuGet.orgTFMs.md
drewgillies: Thanks @zivkan. So there may be a future-proofing concern here.
Just checking that we're not getting away from the intent of the spec a bit on this one?

proposed/2021/NuGet.orgTFMs.md
zivkan, line 103: I think `build/<packageid>.[props|targets]` poses a problem, because from NuGet's point of view this package is compatible with all TFMs, .NET, native/c++, or anything else. For packages like Nerdbank.GitVersioning, or SourceLink, this makes sense.  But many .NET packages incorrectly use this, rather than `build\<tfm>\<packageid>.[props|targets]`. I think it's worthwhile pointing this out and explaining what our plan is for these packages.

proposed/2021/NuGet.orgTFMs.md
drewgillies: This is pertinent to our TFM-assessment logic on package push. If we're looking to support `PackageReference`, we need to not fall back to nuspec reads when the package doesn't contain `lib\<tfm>\foo.dll`, and there's definitely no case for inspecting dlls for target framework metadata.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Removed.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I'll defer to @jcjiang on this.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Removed.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Both changed.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Added `- Runtime, Content Build assets and Tools packages will display TFM details.`

drewgillies: Added `- Support for PackageReference project restore compatibility only.`

joelverhagen: `tools/{tfm}/{rid}/{any}` is not used for dependency package compatibility ([source](https://github.com/NuGet/NuGet.Client/blob/97311b5cb48d38402b49fd41a6384b11db7dd453/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L431-L443)). If we choose to display it, people theoretically could still encounter failed installs with based on this information.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Moved.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I believe the question was about the priority of filtering, and I'll call this good for this iteration. My understanding is that this will be spec'd and implemented separately from the Package Details UX.

proposed/2021/NuGet.orgTFMs.md
JonDouglas: Naw, just that there's mostly a natural ordering we can still follow based on the net5.0 precedence defined today or shown similarly in the latest target framework docs!

drewgillies: I like the notion of not expanded by default to allow for tracking as you mentioned elsewhere. I've added this, but it can always be removed if this isn't where we choose to land: `- We will have the list collapsed by default in order to track popularity.` 
I'd like it before Dependencies also, and will defer to @jcjiang on this.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Added a Future Possibilities section with the following 
```
- Filtering NuGet.org search by TFM
- TM display in Visual Studio
```

proposed/2021/NuGet.orgTFMs.md
drewgillies: I've added this.

proposed/2021/NuGet.orgTFMs.md
drewgillies: That's an important distinction. I've added in the Technical Explanation section:
```
- When a portable TFM is being broken up and displayed (if feasible), we will need to flag it as portable, as portable assets are not the same as platform assets.
```

proposed/2021/NuGet.orgTFMs.md
drewgillies: Are we leaving our intended MVP scope behind a little? This would be impressive and might be a good v2 feature, but I'm concerned that we're throwing some cost in here where we don't need to. Can we backlog this one?

drewgillies: Great--I've added this to the Target Framework List section:
```
Let's link to an explanatory document for any frameworks we expose here, for example: https://docs.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions
```
@jcjiang--I'm not entirely sure how this would need to look in UI. I initially asked you for a mockup, but I think it would just be the first text we see when expanding. Might need your help on wording though, and anything else we might want in there.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I've added all but Prior Art back in.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I've moved @terrajobst's issue front and center in this spec, and will close the tracking issue I've been using and use Immo's.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Okay, this really doesn't need to be here, does it. I'll remove it. :)

proposed/2021/NuGet.orgTFMs.md
drewgillies: Done.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Done.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Removed.

proposed/2021/NuGet.orgTFMs.md
drewgillies: An artifact from the doc spec. Gone.

proposed/2021/NuGet.orgTFMs.md
drewgillies: I've moved this into unresolved questions, @zivkan.

proposed/2021/NuGet.orgTFMs.md
drewgillies: > Naw, just that there's mostly a natural ordering we can still follow based on the net5.0 precedence defined today or shown similarly in the latest target framework docs!

Fair enough. If there are APIs it should be easy enough to implement, and more importantly, it won't need maintenance.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Ah, that's interesting. Yes, that's the risk of including a pattern set that's not called out in the restore code's compat checks. This was a tenuous inclusion anyway, and I can remove it if we don't want this kind of misleading. Build assets could give us the same problem though, right?

proposed/2021/NuGet.orgTFMs.md
JonDouglas: I don't think there's extra cost here. Just deciding on whether a format like the existing `Dependencies` section or a table like `Version History` would represent this information better given NuGet.org is already quite heavy on vertical real estate.

Tables have pros in terms of saving vertical space at the cost of horizontal space.

Headers have pros in terms of saving horizontal space at the cost of vertical space.

/cc @jcjiang

proposed/2021/NuGet.orgTFMs.md
joelverhagen: Build assets are considered in the compat check:
https://github.com/NuGet/NuGet.Client/blob/97311b5cb48d38402b49fd41a6384b11db7dd453/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L441-L442

drewgillies: Not the same set, by the looks: https://github.com/NuGet/NuGetGallery/blob/dg-tfmbackfill/src/GalleryTools/Commands/BackfillTfmMetadataCommand.cs#L65-L67
For a purer experience, perhaps I should remove the transitive build pattern set.

Returning to tools, if we present a TFM based on the tools pattern set, and there is no compat checking on restore, isn't it more likely that the _NuGet.org_ experience is more restrictive? In other words, would we see TFMs on some tools packages that would always restore, but not all restorable tools packages would have TFMs on NuGet.org?

proposed/2021/NuGet.orgTFMs.md
drewgillies: Let's step back a moment. This seems predicated on us not just listing actual TFMs present in the archive (as originally floated), but on exploding this to every possible supported TFM. I want to be sure that we're intentionally adding this to the spec, because there's cost there. That's my cost concern--not &lt;br /&gt; vs. &lt;tr /&gt; tags.

drewgillies: If this is to be added to the spec, we'd need to call it out explicitly, and I'd suggest either:
- put it in as a nice-to-have, not an MVP item, or
- backlog it for the future

JonDouglas: If we put the technical cost aside since a user could multitarget the world & have every single TFM in their archive

i.e. Microsoft.CSharp

image

This gets rather unwieldy quickly in terms of vertical space.

JonDouglas: This example wouldn't help much with a table as only the `netstandard` items would likely save space. This one is just unique 😂

JonDouglas: Should just resolve this feedback & move on with the original mockup if it's easiest! Maybe put under a `Unresolved Questions` if @jcjiang wanted to dive more in the UX w/ the UX team here.

drewgillies: Yep, that works for me if @jcjiang is good with that approach. Good example! It's always good to look at nth degree cases like this.

proposed/2021/NuGet.orgTFMs.md
jcjiang: The change makes sense, always good to be more open-ended and accurate.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,94 @@
...
+Developers should be able to determine whether a package works for their project without having to first download it or rely on the author’s documentation. 
+
+### Why: How do we know this is a real problem and worth solving? 
+
+Finding a package that is compatible with your project is a challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today. One of the top NuGet errors issued in telemetry is [NU1202](https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1202), which shows that we allow developers to install packages that are not compatible with their project or solution’s target framework. 
+
+### Success: How do we know if we’ve solved this problem? 
+
+- By gauging excitement/disappointment after blogging about this feature & analyzing sentiment on Twitter/GitHub/DevCom/etc. 
+
```
jcjiang, line 28: - Compare the before and after numbers for error messages due to TFM incompatibility. Is there a noticeable decrease? If so, how long after the feature was rolled out for the decrease?

Not a perfect metric, but some data to show

proposed/2021/NuGet.orgTFMs.md
jcjiang: I also would like it before Dependencies - let's make the change. T

otally agree on not expanded by default to allow for tracking, could be another metric we can track for success. Are people even clicking open the section? And based on that, we could then toggle for automatic showing if that makes it easier for people to find the information. But I don't want to start off with automatic showing and not be sure if it's useful.

jcjiang: Agreed, let's put this under unresolved questions and I will schedule some time with the UX team when the time comes.

jcjiang: Makes sense. Ping me when you are messing with wording, happy to help there.

proposed/2021/NuGet.orgTFMs.md
jcjiang: Based on the convo below, how many people click open the section to expand.

jcjiang (changes requested):
LGTM, just needs some expanding on the success metrics (which we can add as we find them, currently added two more.)

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,94 @@
...
+Example: .NET Framework, .NET Core, .NET Standard, .NET 5 are target frameworks. 
+
+**Supported Platform** – A supported platform is a target that the APIs can be ran on. 
+
+Example: iOS, Android, Windows are platforms. 
+
+- A .NET developer can view Frameworks of a package on NuGet.org: 
+![](../../meta/resources/NuGet.orgTFMs/PackageDetailsWithTFMs.png)
+
+- We will have the list collapsed by default in order to track popularity.
```
nkolev92, line 56: At an earlier point in time (about ~1 year ago), we did some analysis on `what frameworks does a package really support` vs `what frameworks are declared in the dependencies section` and we found that for about 95% of packages, these are a 1 to 1 relationship. 

Have we consider making that more apparent instead? In the interest of saving some vertical space/introducing a new concept that's already covered in most scenarios.

JonDouglas: This is a good point. There's a challenge here where you run into the XUnit case.

image

To get to an actual "idea" of if the package supports a specific framework is layers deep to the dependencies defined such as:

image

So in these cases, we don't have a good answer without iterating dependencies.

nkolev92: That totally makes sense. 

I just thing integrating it differently, not in an additive way might save us some space. Right now there's *a lot*.

JonDouglas: We just need to consider a tab design for these pages as a "Future Possibility". `Dependencies`, `Frameworks`, `Version History`, `Dependents (Used By)`, etc really belong in their own individual tabs so everything isn't stacked ontop of each other. `README` for example will make scrolling to get to one of these a default experience.

/cc @jcjiang may be worth adding to this section!

proposed/2021/NuGet.orgTFMs.md
drewgillies: After discussion offline, we're looking to:
- keep build asset patterns but drop out transitive build assets in order to make this compat check consistent with nuget restore
- only assess tools pattern sets in the case of tools packages (also consistent with restore logic).

This is beyond the scope of a UX spec, of course, but relevant in making it happen.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Great--included this change.

drewgillies: I've added this as an unresolved question.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Added this metric to the spec.

proposed/2021/NuGet.orgTFMs.md
drewgillies: Discussed offline and we'll consider these to be universally compatible at this stage. I've added `- A design for meta packages` to future possibilities, and splitting expanders out to tabs may be called out here or in anothr spec. Up to you, @jcjiang. As we've nailed down the pertinent corner cases, for now perhaps we can merge this?

drewgillies:
Have discussed with @jcjiang (and mined from stats) offline, and regarding the outstanding unanswered question (on-screen format), we've elected to go with a flat list per original spec. The figures don't support the need for anything more complex at this stage (details of TFM numbers pushed to spec). 

So, awaiting approval to merge. FYI @jcjiang.

proposed/2021/NuGet.orgTFMs.md
```diff
@@ -0,0 +1,107 @@
...
+
+Q: Potentials for modelling our UX on:
+ - https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder?view=net-5.0#applies-to
+ - https://apisof.net/catalog/e40ef6ad-bd6e-9f29-7864-b942e37be11f
+ - https://docs.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions
+A: From a sample of 100,000 package/versions, we have:
+ - Only 445 have more than 5 supported TFMs
+ - Only 16 have more than 10
+ - Only 1 more than 15 (and it has 16)
+ So, we'll go with a flat list for now and look at other options for subsequent work if desired.
```
jcjiang, line 101: I can also work with folks on the UX side, accessibility of information here is more of a PM/UX question so happy to chase that down

jcjiang (approved):
Looks great!

drewgillies:
@joelverhagen @JonDouglas @skofman1 @jcjiang
Let's reopen a topic--TFM display and collapsed by default vs not collapsed by default. I was talking to @skofman1 and this notion came up from a related issue: https://github.com/NuGet/NuGetGallery/issues/4843#issuecomment-393821973 - i.e. do something similar to FuGet, which doesn't require expanding and gives a strong visual cue without the user needing to do anything. This also strikes me as super friendly and productive: 
image

Note the stats in the Q&A on the spec now. Only 16 in 100,000 package/versions had more than 10 TFMs. Fewer than 1% had more than 5. So putting color-coded tiles across the top like this would be tenable, visually appealing and not hog real estate. The color-coding makes for easy focusing in on relevant TFMs.

The sacrifice in this approach would be click data for determining success. But my 2c is that we may be compromising customer joy to achieve this even with the old display approach, and it may not be a good trade.

Thoughts?

JonDouglas:
The original proposal was both. The frameworks node as is and a banner next to the package ID showing the platforms supported like pub.dev & npm. I feel like this is a future opportunity and can be done after this implementation? I can provide some screenshots and mock-ups when I'm at a computer next.

JonDouglas:
Note: They would show platforms supported by the TFMs, not the frameworks themselves. We wanted to abstract people knowing what a tfm meant in terms of platforms.

drewgillies:
@JonDouglas 
> They would show platforms supported by the TFMs, not the frameworks themselves. We wanted to abstract people knowing what a tfm meant in terms of platforms.

That would definitely need further exploration then, as in a pre-.NET5 world we don't have ready access to that information. For example, what would we put in this list for platforms when all we have is the TFM `netstandard2.0`? 

What are thoughts on using a tiles approach for supported TFMs like we see in FuGet (above)? It sounds like you're leaning away from that idea.

JonDouglas:
I've done a bit of discovery here already. It comes down to portable and platform assets. For netstandard, it's a portable asset that can be used everywhere, in previous mock-ups that's known as simply supporting ".NET". For platform specific assets, you would then see the platform name associated. I.e Windows, macOS, Android, iOS, etc. This mostly gets difficult in terms of multiple TFMs that support a single platform with different frameworks, but it would have the fallback of the frameworks node to know exactly what.

JonDouglas:
Here are some mockups to help portray this idea:

image

image

image

image

Some other ecosystem examples:

image

image

image

image

image

I feel strongly that giving users a list of all the TFMs across a screen is unproductive. It's great if nothing else exists, but that's why there's a fallback section for those who want to see exact long names & versions supported.

Users barely know what TFMs are today let alone should they need to use a decoder table to understand if they can install and use it to target a platform. This story gets much better post .NET 5 with `TargetPlatformIdentifier` and more, but we can still categorize old generation TFMs into logical platforms by hand, or at least the most popular ones that we know are supported on a specific platform until there's more adoption of net5/net6 TFMs. The other choice is to only show platform badges for net5+ libraries.

JonDouglas:
I would propose that we keep this spec within the scope of just showing the frameworks node as is & we continue the discussion of "portable/platform asset badges" in a separate proposal as it largely has more to do with filtering & compatibility.

jcjiang:
Agreed with Jon that we should keep the spec within scope of the frameworks node. Platform asset badges will come with a lot of additional technical decisions to make so let's continue that discussion separately.

skofman1:
@JonDouglas , @jcjiang , agree that this spec should remain within scope of the frameworks node. However, my concern is that if we ship only this node, and not have tags as well, customers will perceive this as underwhelming.

anangaur:
What do you all feel about surfacing tags for .NET5/6 compat and not for all?

skofman1:
@anangaur , isn't the "big question" when it comes to compat is .NET Framework vs. Core?

anangaur:
May be that. I was hinting at doing a thin slice if possible for things that matter. Could be .NET5/6 compat or .NET Core vs. Framework, based on what's needed.

skofman1:
Agree that for tags it should be a thin list with the most important compat. The "Frameworks node" can be more detailed.

drewgillies:
Okay, it looks like we've come in to land on this:
- Platform tiles are in the spec. These will be basic .NET framework types until we have `TargetPlatformIdentifiers` in the `.NET 5.0` TFMs.
- Framework node is collapsed by default. Perhaps we could consider a note to users to expand it for more details, next to the tiles.

I've put this in the spec under Minimal Requirements:
- List of Target Frameworks on NuGet.org's package details page. The Frameworks node will be expanded by default.
- A series of tiles near package Id and version indicating supported platform--this will become more populated when .NET5.0's (TargetPlatformIdenitifiers)[https://github.com/NuGet/Home/issues/9240] are in the TFM (which we can display versionless with full details in Frameworks node), but for now we can have `.NET`, `.NET Core`, `.NET standard`, `Portable`, etc. TargetPlatformIdentifiers will give us iOS, Android, tvOS etc., perhaps grouped alongside .NET (which will be visibly 5.0 when the Frameworks node is viewed).

As we're now talking variants, I'm going to merge this as a `proposed` spec and submit a new PR to make this `approved`. We can continue any tweaks there before final approval. One tweak which I'd like to see is a mockup indicating what we'd like to see on screen given all of the above. There has been some evolution since the initial spec, which is good, but we need the spec to keep step. Thanks so much for this discussion and let's keep it going--this is going to be a sweet feature.

drewgillies:
New PR is here: https://github.com/NuGet/Home/pull/10605. I believe I have everyone on it as reviewers who's contributed here.

PR #10605, opened 2021-03-01
Approve NuGet.org TFMs spec

drewgillies:


JonDouglas (approved):

JonDouglas:
Approving as I don't see any reason to not accept the proposal in it's current state.

This requirement worries me as this is much more involved. If you're fine with including it in the scope of the original proposal, then by all means! Let's just make sure we have a functional explanation & details added to this spec post-merge. I was under the impression we'd be making a separate proposal for this work.

> A series of tiles near package Id and version indicating supported platform--this will become more populated when .NET5.0's (TargetPlatformIdenitifiers)[https://github.com/NuGet/Home/issues/9240] are in the TFM (which we can display versionless with full details in Frameworks node), but for now we can have .NET, .NET Core, .NET standard, Portable, etc. TargetPlatformIdentifiers will give us iOS, Android, tvOS etc., perhaps grouped alongside .NET (which will be visibly 5.0 when the Frameworks node is viewed).

Some thoughts:

- .NET 5 TPI exist today. We can grab the `TargetPlatformIdentifier`! - https://nugettoolsdev.azurewebsites.net/5.8.0/parse-framework?framework=net5.0-windows
- [There are three generations of frameworks & one shared framework](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#latest-versions): (Generation) .NET Framework, .NET Core, and .NET 5. (Shared) .NET Standard and .NET 5. There should be a consideration to not show the world here, but rather show these tiles only if there's at least one matching TFM in the package. (i.e. the presence of `netcoreappX.Y` would show `.NET Core`)
    - **Note: .NET 5 is a superset & thus is both generation & shared**
- There's the question of older generation TFMs prior to .NET 5. Should we show these? Some of the names may not make much sense to users like `netmf`, `monodroid`, or even `netcore` (don't ask it predates netcoreapp 😆). Because of this, it may need manual mapping to platforms.

There's a bit of customer development & UX review here left to ask/design what users would see near the package ID in terms of knowing at a glance whether a package is .NET Framework/.NET Core/.NET 5 supported & includes respective portable / platform assets. 

Edit: Here's an initial tweet to help the direction https://twitter.com/_JonDouglas/status/1366474225582825478

jcjiang:
I'll be doing some customer development on the details of the feature to have a more informed discussion on platforms, so l agree with Jon that a separate proposal would be best. Otherwise, this looks great.

jcjiang (approved):

drewgillies:
Okay, that's fine--I'll remove the target platform piece (and yes, I should have used present tense for the .NET 5 TFIs) and I'll go ahead and merge this. This will leave the TFM node collapsed by default (@skofman1 proposed expanding it but the tiles addressed the concerns voiced there), but we can leave this as an ongoing discussion/implementation detail. I'll put a not3e in unanswered question in the spec. Thanks, all!

Issue #8053, opened 2019-04-23
Signing: implement dotnet trusted-signers command


heng-liu:
Implement "dotnet trusted-signers" command according to NuGet.exe command "nuget trusted-signers".+
Spec is  here" https://github.com/NuGet/Home/pull/10628

erdembayar:
@heng-liu 
Is there spec for "nuget trusted-signers"? If not I just look at what "nuget trusted-signers" does and make parity with it would be enough for this issue?

heng-liu:
Hi @erdembayar , please refer to https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers

erdembayar:
@heng-liu @dtivel  @nkolev92 
Can I get some context? Where do I implement it? Is it in NuGet.Client repo or dotnet repo?

heng-liu:
Hi @erdembayar, the nuget.exe trusted-signers command is already implemented. This is to parity dotnet CLI with nuget.exe.
You may take nuget.exe verify command and dotnet nuget verify command as an example.
We implemented verify command for dotnet under src\NuGet.Core\NuGet.CommandLine.XPlat\Commands\Signing\VerifyCommand.cs

erdembayar:
@aortiz-msft @heng-liu @zkat @kartheekp-ms 
It looks spec for `dotnet nuget trusted-signer` not done due to complexity [mentioned here](https://github.com/NuGet/Home/blob/dev/proposed/archive/DotNetSignCommands.md#future-work). Do I need to create spec first and confirm with `dotnet` team first or not?

kartheekp-ms:
@erdembayar - `nuget.exe` trusted-signers command spec is archived [here](https://github.com/NuGet/Home/wiki/%5BSpec%5D-NuGet-Package-Signing-Client-Policy). Hope this helps you to create a spec for the new `dotnet` command.

nkolev92:
@erdembayar Remember to follow https://github.com/NuGet/NuGet.Client/blob/dev/docs/feature-guide.md for the process. 

You will probably need to update the template sued in the spec that @kartheekp-ms linked.

PR #10628, opened 2021-03-05
initial draft of dotnet nuget trusted-signers commands spec

zkat:
This spec is for #8053 

(note: I'm not going to be around until next Tuesday, so feel free to treat this as a first draft and make your own modifications if we need to land this asap.)

We'll also want Kathleen to take a look at this, but I don't remember her GH handle :(

### DIFF
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -1,18 +1,18 @@
-# Title
+# dotnet nuget trusted-signers command
 
 - Author Name [Kat Marchan (@zkat)](https://github.com/zkat)
 - Start Date 2021-03-04
-- GitHub Issue (GitHub Issue link)
-- GitHub PR (GitHub PR link)
+- GitHub Issue: https://github.com/NuGet/Home/issues/10634
+- GitHub PR: https://github.com/NuGet/Home/pull/10628
 
 ## Summary
 
-This specifies the syntax and behavior for the `dotnet.exe nuget trusted-signers` command.
+This specifies the syntax and behavior for the `dotnet nuget trusted-signers` command.
 
 ## Motivation
 
-This is part of a larger effort to reach full parity between `nuget.exe` and
-`dotnet.exe` for NuGet-related operations.
+This is part of a larger effort to reach full parity between `nuget` and
+`dotnet` for NuGet-related operations.
 
 ## Explanation
 
@@ -24,7 +24,7 @@ This is part of a larger effort to reach full parity between `nuget.exe` and
 Provides the ability to manage the list of trusted signers.
 
 USAGE:
-    dotnet nuget trusted-signer [OPTIONS] [COMMAND]
+    dotnet nuget trusted-signers [OPTIONS] [COMMAND]
 
 COMMANDS:
     list (default)
@@ -143,16 +143,16 @@ OPTIONS:
 ```
 Adds a trusted signer based on a given source.
 
-If only `<name>` is provided, and it's an existing configured source, this command will add a trusted signer with the same name as that source to the trusted list.
+If only `<name>` is provided without `<source-url>`, the package source from your NuGet configuration files with the same name will be added to the trusted list.
 
-If a `source-url` is provided, it MUST be a v3 source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
+If a `<source-url>` is provided, it MUST be a v3 package source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
 
 USAGE:
     dotnet nuget trusted-signers trust-source <name> [source-url]
 
 OPTIONS:
         --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
-        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository. Only valid when using the --repository option.
+        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository.
     -h, --help                          Prints help information.
     -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
         --interactive                   Allows command to stop and wait for user input or action.
@@ -162,7 +162,7 @@ OPTIONS:
 
 ### Technical explanation
 
-For the most part, these commands map pretty directly to their nuget.exe
+For the most part, these commands map pretty directly to their nuget
 counterparts, and most of their implementations should be reusable (removing
 `#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
 
@@ -170,7 +170,7 @@ The exception here is the `add` command, which has been split into the various
 `trust-*` commands that otherwise have the same behaviors.
 
 I don't know if there's any significant implementation above just remapping
-command invocations for dotnet.exe.
+command invocations for dotnet.
 
 ## Drawbacks
 
@@ -181,16 +181,16 @@ This whole feature is very complicated, but it's important for parity.
 I think the only one that might have a reasonable alternative here would be
 the `add` command. I found the command as a whole to be inscrutable, and thus
 decided splitting its behavior to be the best way forward. We could have, of
-course, mostly just copied the behavior from `nuget.exe trusted-signers add`.
+course, mostly just copied the behavior from `nuget trusted-signers add`.
 
 ## Prior Art
 
-This spec is based on `nuget.exe trusted-signers`.
+This spec is based on [`nuget trusted-signers`](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers).
 
 ## Unresolved Questions
 
-Is this the right UX? Does the naming make sense? This is a fairly complex/complicated feature and we want to make sure we
+Is this the right UX? Does the naming make sense? This is a fairly complex/complicated feature and we want to make sure we deliver something to customers that makes sense and gives an overall good experience.
 
 ## Future Possibilities
 
-I don't know.
+N/A
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -1,0 +1,197 @@
+# Title
+
+- Author Name [Kat Marchan (@zkat)](https://github.com/zkat)
+- Start Date 2021-03-04
+- GitHub Issue (GitHub Issue link)
+- GitHub PR (GitHub PR link)
+
+## Summary
+
+This specifies the syntax and behavior for the `dotnet.exe nuget trusted-signers` command.
+
+## Motivation
+
+This is part of a larger effort to reach full parity between `nuget.exe` and
+`dotnet.exe` for NuGet-related operations.
+
+## Explanation
+
+### Functional explanation
+
+#### `dotnet nuget trusted-signers`
+
+```
+Provides the ability to manage the list of trusted signers.
+
+USAGE:
+    dotnet nuget trusted-signer [OPTIONS] [COMMAND]
+
+COMMANDS:
+    list (default)
+    add-repository <name> <repository> [<package>...]
+    add-author <name> <author> [<package>...]
+    add-service-index <name> <index>
+    add-certificate <name> <fingerprint>
+    remove <name>
+    sync <name>
+```
+
+#### `dotnet nuget trusted-signers list`
+
+```
+Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], it means that certificate entry has allowUntrustedRoot set as true.
+
+USAGE:
+    dotnet nuget trusted-signers [OPTIONS] list
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+
+EXAMPLE:
+    $ dotnet nuget trusted-signers list
+    Registered trusted signers:
+
+    1.   nuget.org [repository]
+        Service Index: https://api.nuget.org/v3/index.json
+        Certificate fingerprint(s):
+            SHA256 - 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
+
+    2.   microsoft [author]
+        Certificate fingerprint(s):
+            SHA256 - 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
+            SHA256 - AA12DA22A49BCE7D5C1AE64CC1F3D892F150DA76140F210ABD2CBFFCA2C18A27
+
+    3.   myUntrustedAuthorSignature [author]
+        Certificate fingerprint(s):
+            [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
+```
+
+#### `dotnet nuget trusted-signers sync`
+
+```
+Requests the latest list of certificates used in a currently trusted repository to update the the existing certificate list in the trusted signer.
+
+Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.
+
+USAGE:
+    dotnet nuget trusted-signers sync [OPTIONS] <name>
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers remove`
+
+```
+Removes any trusted signers that match the given name.
+
+USAGE:
+    dotnet nuget trusted-signers remove [OPTIONS] <name>
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-package`
+
+```
+Adds a trusted signer with the given name to the config, based on one or more packages.
+
+USAGE:
+    dotnet nuget trusted-signers trust-package [OPTIONS] <name> <package>...
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --author                        Specifies that the author signature of the package(s) should be trusted. This option is mutually exclusive with --repository.
+        --repository                    Specifies that the repository signature or countersignature of the package(s) should be trusted. This option is mutually exclusive with --author.
+        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository. Only valid when using the --repository option.
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-certificate`
+
+```
+Adds a trusted signer with the given name to the config, based on a certificate fingerprint.
+
+Note: If a trusted signer with the given name already exists, the certificate item will be added to that signer. Otherwise a trusted author will be created with a certificate item from given certificate information.
+
+USAGE:
+    dotnet nuget trusted-signers trust-certificate [OPTIONS] <name> <fingerprint>
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --algorithm                     Specifies the hash algorithm used to calculate the certificate fingerprint. Defaults to SHA256. Values supported are SHA256, SHA384 and SHA512
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-source`
+
+```
+Adds a trusted signer based on a given source.
+
+If only `<name>` is provided, and it's an existing configured source, this command will add a trusted signer with the same name as that source to the trusted list.
+
+If a `source-url` is provided, it MUST be a v3 source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
+
+USAGE:
+    dotnet nuget trusted-signers trust-source <name> [source-url]
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository. Only valid when using the --repository option.
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+
+```
+
+### Technical explanation
+
+For the most part, these commands map pretty directly to their nuget.exe
+counterparts, and most of their implementations should be reusable (removing
+`#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
+
+The exception here is the `add` command, which has been split into the various
+`trust-*` commands that otherwise have the same behaviors.
+
+I don't know if there's any significant implementation above just remapping
+command invocations for dotnet.exe.
+
+## Drawbacks
+
+This whole feature is very complicated, but it's important for parity.
+
+## Rationale and alternatives
+
+I think the only one that might have a reasonable alternative here would be
+the `add` command. I found the command as a whole to be inscrutable, and thus
+decided splitting its behavior to be the best way forward. We could have, of
+course, mostly just copied the behavior from `nuget.exe trusted-signers add`.
+
+## Prior Art
+
+This spec is based on `nuget.exe trusted-signers`.
+
+## Unresolved Questions
+
+Is this the right UX? Does the naming make sense? This is a fairly complex/complicated feature and we want to make sure we
+
+## Future Possibilities
+
+I don't know.
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -28,10 +28,9 @@ USAGE:
 
 COMMANDS:
     list (default)
-    add-repository <name> <repository> [<package>...]
-    add-author <name> <author> [<package>...]
-    add-service-index <name> <index>
-    add-certificate <name> <fingerprint>
+    trust-package <name> <package>...
+    trust-source <name> [source-url]
+    trust-certificate <name> <fingerprint>
     remove <name>
     sync <name>
 ```
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -21,7 +21,7 @@ This is part of a larger effort to reach full parity between `nuget` and
 #### `> dotnet nuget trust`
 
 ```
-Manage the trusted signers. By default, NuGet accepts all authors and repositories. You can restrict this to just authors or repositories you trust using this command.
+Manage the trusted signers. By default, NuGet accepts all authors and repositories. These commands allow you to specify only a specific subset of signers whose signatures will be accepted, while rejecting all others.
 
 USAGE:
     dotnet nuget trust [COMMAND] [OPTIONS]
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -1,4 +1,4 @@
-# dotnet nuget trusted-signers command
+# dotnet nuget trust command
 
 - Author Name [Kat Marchan (@zkat)](https://github.com/zkat)
 - Start Date 2021-03-04
@@ -7,7 +7,7 @@
 
 ## Summary
 
-This specifies the syntax and behavior for the `dotnet nuget trusted-signers` command.
+This specifies the syntax and behavior for the `dotnet nuget trust` command.
 
 ## Motivation
 
@@ -18,43 +18,39 @@ This is part of a larger effort to reach full parity between `nuget` and
 
 ### Functional explanation
 
-#### `dotnet nuget trusted-signers`
+#### `> dotnet nuget trust`
 
 ```
 Provides the ability to manage the list of trusted signers.
 
 USAGE:
-    dotnet nuget trusted-signers [OPTIONS] [COMMAND]
+    dotnet nuget trust [COMMAND] [OPTIONS]
 
 COMMANDS:
-    list (default)
-    trust-package <name> <package>...
-    trust-source <name> [source-url]
-    trust-certificate <name> <fingerprint>
+    list
+    author <name> <package>...
+    repository <name> <package>...
+    source <name> [source-url]
+    certificate <name> <fingerprint>
     remove <name>
     sync <name>
 ```
 
-#### `dotnet nuget trusted-signers list`
+#### `> dotnet nuget trust list`
 
 ```
 Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], it means that certificate entry has allowUntrustedRoot set as true.
 
 USAGE:
-    dotnet nuget trusted-signers [OPTIONS] list
-
-OPTIONS:
-    -h, --help                          Prints help information.
-    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
-        --interactive                   Allows command to stop and wait for user input or action.
-        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+    dotnet nuget trust list [OPTIONS]
 
 EXAMPLE:
-    $ dotnet nuget trusted-signers list
+    $ dotnet nuget trust list
     Registered trusted signers:
 
     1.   nuget.org [repository]
         Service Index: https://api.nuget.org/v3/index.json
+        Owners: microsoft, aspnet, nuget
         Certificate fingerprint(s):
             SHA256 - 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
 
@@ -66,9 +62,15 @@ EXAMPLE:
     3.   myUntrustedAuthorSignature [author]
         Certificate fingerprint(s):
             [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 ```
 
-#### `dotnet nuget trusted-signers sync`
+#### `> dotnet nuget trust sync`
 
 ```
 Requests the latest list of certificates used in a currently trusted repository to update the the existing certificate list in the trusted signer.
@@ -76,7 +78,7 @@ Requests the latest list of certificates used in a currently trusted repository 
 Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.
 
 USAGE:
-    dotnet nuget trusted-signers sync [OPTIONS] <name>
+    dotnet nuget trust sync [OPTIONS] <name>
 
 OPTIONS:
     -h, --help                          Prints help information.
@@ -85,13 +87,13 @@ OPTIONS:
         --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 ```
 
-#### `dotnet nuget trusted-signers remove`
+#### `> dotnet nuget trust remove`
 
 ```
 Removes any trusted signers that match the given name.
 
 USAGE:
-    dotnet nuget trusted-signers remove [OPTIONS] <name>
+    dotnet nuget trust remove [OPTIONS] <name>
 
 OPTIONS:
     -h, --help                          Prints help information.
@@ -100,26 +102,47 @@ OPTIONS:
         --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 ```
 
-#### `dotnet nuget trusted-signers trust-package`
+#### `> dotnet nuget trust author`
 
 ```
-Adds a trusted signer with the given name to the config, based on one or more packages.
+Adds a trusted signer with the given name to the config, based on the author signature(s) of one or more packages. If <name> already exists in the configuration, the signature(s) will be appended.
 
 USAGE:
-    dotnet nuget trusted-signers trust-package [OPTIONS] <name> <package>...
+    dotnet nuget trust author [OPTIONS] <name> <package>...
+
+EXAMPLE:
+    dotnet nuget trust author Contoso ./contoso.library.nupkg
 
 OPTIONS:
         --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
-        --author                        Specifies that the author signature of the package(s) should be trusted. This option is mutually exclusive with --repository.
-        --repository                    Specifies that the repository signature or countersignature of the package(s) should be trusted. This option is mutually exclusive with --author.
-        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository. Only valid when using the --repository option.
     -h, --help                          Prints help information.
     -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
         --interactive                   Allows command to stop and wait for user input or action.
         --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 ```
 
-#### `dotnet nuget trusted-signers trust-certificate`
+#### `> dotnet nuget trust repository`
+
+```
+Adds a trusted signer with the given name to the config, based on the repository signature(s) or countersignature(s) of one or more packages. If <name> already exists in the configuration, the signature(s) will be appended.
+
+
+USAGE:
+    dotnet nuget trust repository [OPTIONS] <name> <package>...
+
+EXAMPLE:
+    dotnet nuget trust repository Contoso ./contoso.library.nupkg
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository.
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `> dotnet nuget trust certificate`
 
 ```
 Adds a trusted signer with the given name to the config, based on a certificate fingerprint.
@@ -127,7 +150,7 @@ Adds a trusted signer with the given name to the config, based on a certificate 
 Note: If a trusted signer with the given name already exists, the certificate item will be added to that signer. Otherwise a trusted author will be created with a certificate item from given certificate information.
 
 USAGE:
-    dotnet nuget trusted-signers trust-certificate [OPTIONS] <name> <fingerprint>
+    dotnet nuget trust certificate [OPTIONS] <name> <fingerprint>
 
 OPTIONS:
         --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
@@ -138,17 +161,19 @@ OPTIONS:
         --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
 ```
 
-#### `dotnet nuget trusted-signers trust-source`
+#### `> dotnet nuget trust source`
 
 ```
-Adds a trusted signer based on a given source.
+Adds a trusted signer based on a given package source.
 
 If only `<name>` is provided without `<source-url>`, the package source from your NuGet configuration files with the same name will be added to the trusted list.
 
-If a `<source-url>` is provided, it MUST be a v3 package source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
+If a `<source-url>` is provided, it MUST be a v3 package source URL (like https://api.nuget.org/v3/index.json). Other package source types are not supported.
+
+If <name> already exists in the configuration, the signature(s) will be appended.
 
 USAGE:
-    dotnet nuget trusted-signers trust-source <name> [source-url]
+    dotnet nuget trust source <name> [source-url]
 
 OPTIONS:
         --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
@@ -166,8 +191,9 @@ For the most part, these commands map pretty directly to their nuget
 counterparts, and most of their implementations should be reusable (removing
 `#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
 
-The exception here is the `add` command, which has been split into the various
-`trust-*` commands that otherwise have the same behaviors.
+The exception here is the `add` command, which has been split into separate
+commands that otherwise have the same behaviors, since options and semantics
+can vary significantly.
 
 I don't know if there's any significant implementation above just remapping
 command invocations for dotnet.
--- a/proposed/2021/DotnetNugetTrustedSigners.md
+++ b/proposed/2021/DotnetNugetTrustedSigners.md
@@ -44,47 +44,26 @@ Lists all the trusted signers in the configuration. This option will include all
 USAGE:
     dotnet nuget trust list [OPTIONS]
 
-EXAMPLE:
-    $ dotnet nuget trust list
-    Registered trusted signers:
-
-    1.   nuget.org [repository]
-        Service Index: https://api.nuget.org/v3/index.json
-        Owners: microsoft, aspnet, nuget
-        Certificate fingerprint(s):
-            SHA256 - 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
-
-    2.   microsoft [author]
-        Certificate fingerprint(s):
-            SHA256 - 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
-            SHA256 - AA12DA22A49BCE7D5C1AE64CC1F3D892F150DA76140F210ABD2CBFFCA2C18A27
-
-    3.   myUntrustedAuthorSignature [author]
-        Certificate fingerprint(s):
-            [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
-
 OPTIONS:
     -h, --help                          Prints help information.
-    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
-        --interactive                   Allows command to stop and wait for user input or action.
-        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+    -v, --verbosity <LEVEL>             Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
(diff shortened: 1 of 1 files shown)


proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
+# Title
```
nkolev92: You're missing an actual title :)

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+- Author Name [Kat Marchan (@zkat)](https://github.com/zkat)
+- Start Date 2021-03-04
+- GitHub Issue (GitHub Issue link)
```
nkolev92: And a GH issue link.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+decided splitting its behavior to be the best way forward. We could have, of
+course, mostly just copied the behavior from `nuget.exe trusted-signers add`.
+
+## Prior Art
+
+This spec is based on `nuget.exe trusted-signers`.
+
+## Unresolved Questions
+
+Is this the right UX? Does the naming make sense? This is a fairly complex/complicated feature and we want to make sure we
```
nkolev92: This sentence seems not finished.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Future Possibilities
+
+I don't know.
```
nkolev92: I'd add N/A here :D

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+## Rationale and alternatives
+
+I think the only one that might have a reasonable alternative here would be
+the `add` command. I found the command as a whole to be inscrutable, and thus
+decided splitting its behavior to be the best way forward. We could have, of
+course, mostly just copied the behavior from `nuget.exe trusted-signers add`.
+
+## Prior Art
+
+This spec is based on `nuget.exe trusted-signers`.
```
nkolev92: nit: I'd link the command.

nkolev92 (commented):
This all seems pretty reasonable to me :) 

I'm not the expert in the feature area, so I am curious what the SMEs & .NET CLI folks think here.

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: Created a GH issue: https://github.com/NuGet/Home/issues/10634

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Functional explanation
+
+#### `dotnet nuget trusted-signers`
+
+```
+Provides the ability to manage the list of trusted signers.
+
+USAGE:
+    dotnet nuget trusted-signer [OPTIONS] [COMMAND]
```
erdembayar: Missing plural `s` for `trusted-signer`.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+#### `dotnet nuget trusted-signers trust-package`
+
+```
+Adds a trusted signer with the given name to the config, based on one or more packages.
+
+USAGE:
+    dotnet nuget trusted-signers trust-package [OPTIONS] <name> <package>...
+
```
erdembayar, line 116: Currently `dotnet nuget verify` allows to pass more than 1 path even though `nuget verify` doesn't, how about here do we have same gesture behavior allow pass more than 1 paths or serviceindexes, fingerprints?
`dotnet nuget verify C:\Users\xxxx\source\repos\SigningDemo\TestSigned\TestSigned.1.0.0.nupkg C:\Users\xxxx\source\repos\SigningDemo\TestSigned\*.nupkg -v d`

erdembayar (commented):
@zkat @heng-liu 

1 question

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+```
+Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], it means that certificate entry has allowUntrustedRoot set as true.
+
+USAGE:
+    dotnet nuget trusted-signers [OPTIONS] list
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
```
erdembayar: It looks `verbosity` thing doesn't make sense here, I don't see it in nuget.exe implementation code.

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: That's a very good question.
My understanding is, the reason why we change the dotnet verify command to support multiple paths, is as following:
NuGet.exe only runs Windows(except on Mono) but dotnet.exe will run cross-platform
When executing the command in powershell, the xx\*.nupkg will be passed as one argument to nuget.exe,
but when executing the command in bash, the globbing will expand the xx/*.nupkg into multiple paths before passing to dotnet.exe command.  So if the dotnet.exe only accept one path, it will only get the first path but ignore the rest paths, and make it inconsistent with what we have on NuGet.exe. 

In one word, we change dotnet.exe to accept multiple paths because it makes dotnet acts more like NuGet.exe.

Previously when we only have NuGet.exe, we use ";" to separate elements in one option, if we have multiple serviceindexes, fingerprints.
When we have dotnet.exe later,  the ";" indicates the command terminates in bash. 
I agree we need to think it over and have more discussion if we want to change the separator. If we're going to deprecate NuGet.exe, we may have less impact on customers.
FYI @JonDouglas

proposed/2021/DotnetNugetTrustedSigners.md
zivkan: While nuget.exe did `nuget.exe <noun> <verb>` commands, like `nuget.exe source add`, when Rob ported sources to the dotnet cli, he changed the pattern to `dotnet nuget <verb> <noun>`, such as `dotnet nuget add source` and `dotnet nuget add client-cert`. This is in line with other dotnet CLI commands such as `dotnet add package`, `dotnet add reference`, `dotnet list reference`, `dotnet list package`.

Theredore, I think you should consider if the "trust" commands should follow the pattern with `dotnet nuget list trusted-signer`, `dotnet nuget add trusted-package`, `dotnet nuget add trusted-source`, `dotnet nuget remove trusted-???`.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+    3.   myUntrustedAuthorSignature [author]
+        Certificate fingerprint(s):
+            [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
+```
+
+#### `dotnet nuget trusted-signers sync`
+
+```
+Requests the latest list of certificates used in a currently trusted repository to update the the existing certificate list in the trusted signer.
```
zivkan: This only updates the trusted source, not trusted author, right?  Maybe `trusted-source` is better than `trusted-signers`, if it doen't also sync to the latest author certificates.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers remove`
+
+```
+Removes any trusted signers that match the given name.
```
zivkan: Is it confusing that when adding a trusted signer you need to specify either source or package, but when you remove you don't specify which 'type'?

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-package`
+
+```
+Adds a trusted signer with the given name to the config, based on one or more packages.
```
zivkan: I think `trust-owner` or `trust-author` would be better than `trust-package`. You don't trust an individual package.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+### Technical explanation
+
+For the most part, these commands map pretty directly to their nuget.exe
+counterparts, and most of their implementations should be reusable (removing
+`#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
+
+The exception here is the `add` command, which has been split into the various
+`trust-*` commands that otherwise have the same behaviors.
```
zivkan: Why's that?  In another comment I mentioned there are already multiple `dotnet add <noun>` and `dotnet nuget add <noun>`. Therefore, this spec is not following the convention of other commands in the dotnet cli

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: It seems the they have changed their mind after the implementation of source command. I remembered we had a discussion about it when reviewing dotnet verify command altogether. But I'm not 100% sure about the conclusion.

nkolev92:
```console
C:\Users\Roki2\Documents\Code\NuGet\NuGet.Client [dev-nkolev92-fixUpFrameworks ≡]> dotnet nuget source add                                                                                                                                   Specify --help for a list of available options and commands.
error: Unrecognized command or argument 'source'


Usage: dotnet nuget add [options] [command]

Options:
  -h|--help  Show help information

Commands:
  client-cert  Adds a client certificate configuration that matches the given package source name.
  source       Add a NuGet source.

Use "add [command] --help" for more information about a command.
C:\Users\Roki2\Documents\Code\NuGet\NuGet.Client [dev-nkolev92-fixUpFrameworks ≡]> dotnet nuget add source                                                                                                                                   error: Value cannot be null. (Parameter 'input')


Usage: dotnet nuget add source [arguments] [options]

Arguments:
  PackageSourcePath  Path to the package source.

Options:
  -n|--name                       Name of the source.
  -u|--username                   Username to be used when connecting to an authenticated source.
  -p|--password                   Password to be used when connecting to an authenticated source.
  --store-password-in-clear-text  Enables storing portable package source credentials by disabling password encryption.
  --valid-authentication-types    Comma-separated list of valid authentication types for this source. Set this to basic if the server advertises NTLM or Negotiate and your credentials must be sent using the Basic mechanism, for instance when using a PAT with on-premises Azure DevOps Server. Other valid values include negotiate, kerberos, ntlm, and digest, but these values are unlikely to be useful.
  --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used. For more information, see https://docs.microsoft.com/nuget/consume-packages/configuring-nuget-behavior.
  -h|--help                       Show help information
C:\Users\Roki2\Documents\Code\NuGet\NuGet.Client [dev-nkolev92-fixUpFrameworks ≡]>   
```

add source seems to be the command, although the experience is not great when the input doesn't contain anything.

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: According to the [code](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/TrustedSignersCommand/TrustedSignerActionsProvider.cs#L42-L68), yes you're right, it just sync certificate of repository signature.
I think trusted-signers means both author signers and repository signers. 
An example is： we add certificate of repository signature by running trusted-signers add command.
So the question is, it good enough to only call it out explicitly in the document, that author certs will not be synced?

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: According to the [code ](https://github.com/NuGet/NuGet.Client/blob/ac97d923e633f4ee65eec5ec690c4d99d7a922f3/src/NuGet.Core/NuGet.Commands/TrustedSignersCommand/TrustedSignerActionsProvider.cs#L243), there is a checking on duplicate names in trusted-signers section of config file when adding, to guarantee that there is no duplicate name, among all trusted-signers (both author and repository signers). So it might be good to just specifying the name, without 'type'.

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: What I remembered from last discussion is that dotnet want to change (they do not want to keep the same with dotnet add source any more). But I'm not 100% sure if that is the final decision.
Perhaps we can discuss with them when reviewing this document.

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: I think that make sense.
We don't have verbosity options in [dotnet nuget add source](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-add-source) either, which is also just editing the config file.

nkolev92: nuget.exe supports verbosity on every command. 
I think we should lean towards supporting verbosity for most commands in dotnet.exe as well.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+        Certificate fingerprint(s):
+            SHA256 - 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
+            SHA256 - AA12DA22A49BCE7D5C1AE64CC1F3D892F150DA76140F210ABD2CBFFCA2C18A27
+
+    3.   myUntrustedAuthorSignature [author]
+        Certificate fingerprint(s):
+            [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
+```
+
+#### `dotnet nuget trusted-signers sync`
```
loic-sharma: @chgill-MSFT what's the concern with `nuget.exe trusted-signers sync` again? I remember we chose not to advertise this command on this blog post: https://devblogs.microsoft.com/nuget/the-nuget-org-repository-signing-certificate-will-be-updated-as-soon-as-march-15th-2021/

Is there something we should do differently for dotnet.exe's command?

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+    2.   microsoft [author]
+        Certificate fingerprint(s):
+            SHA256 - 3F9001EA83C560D712C24CF213C3D312CB3BFF51EE89435D3430BD06B5D0EECE
+            SHA256 - AA12DA22A49BCE7D5C1AE64CC1F3D892F150DA76140F210ABD2CBFFCA2C18A27
+
+    3.   myUntrustedAuthorSignature [author]
+        Certificate fingerprint(s):
+            [U] SHA256 - 518F9CF082C0872025EFB2587B6A6AB198208F63EA58DD54D2B9FF6735CA4434
+```
```
loic-sharma, line 67: The `dotnet list package` describes the markers like `[U]` at the bottom of the command output. For example ([LINK](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-list-package#description)):

```
Project 'SentimentAnalysis' has the following package references
   ...
   > Microsoft.NETCore.App   (A)   [2.1.0, )   2.1.0

(A) : Auto-referenced package.
```

Personally I find that describing the markers at the bottom of the command output grabs my attention more than the end of the command's help description. What do you think of adding the markers at the bottom of the output too?

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: ~~+1 for `trust-author`, as it aligns with our "author signature" name~~

EDIT: I didn't realize this command could be used to trust a package's author or repository. I personally prefer `trust-package` now. That said, I do think a `trust-owner` command could be useful (see comment below)

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-package`
+
+```
+Adds a trusted signer with the given name to the config, based on one or more packages.
+
+USAGE:
+    dotnet nuget trusted-signers trust-package [OPTIONS] <name> <package>...
```
loic-sharma: I'm a little concerned that `name` and `package` may not click for customers right away. Would it be possible to add an example to the help output? For example: `dotnet nuget trusted-signers trust-package Contoso .\contoso.library.nupkg`

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: Is globbing files a scenario customers need when updating their trusted signers? Do you have an example when this would be useful?

The reason I ask is because this scenario scares me a little bit. Ideally customers should thoroughly review each and every author/source they choose to trust.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
```
loic-sharma: Can we guide customers against the `--allow-untrusted-root` option? I'd consider pushing this option to the bottom of the list and adding wording like:

```suggestion
        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root. This is not recommended.
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+If only `<name>` is provided, and it's an existing configured source, this command will add a trusted signer with the same name as that source to the trusted list.
+
+If a `source-url` is provided, it MUST be a v3 source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
+
+USAGE:
+    dotnet nuget trusted-signers trust-source <name> [source-url]
+
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository. Only valid when using the --repository option.
```
loic-sharma: It looks like this command doesn't have a `--repository` option. Consider:

```suggestion
        --owners                        Semi-colon separated list of trusted owners to further restrict the trust of a repository.
```

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: Thanks @nkolev92! Good to know the future trends.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+```
+
+#### `dotnet nuget trusted-signers trust-source`
+
+```
+Adds a trusted signer based on a given source.
+
+If only `<name>` is provided, and it's an existing configured source, this command will add a trusted signer with the same name as that source to the trusted list.
+
+If a `source-url` is provided, it MUST be a v3 source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
```
loic-sharma: What do you think of using the phrase "package source" wherever possible in descriptions? I _think_ this may improve clarity, but I don't feel strongly about this. For example:

```suggestion
Adds a trusted signer based on a given package source.

If only `<name>` is provided, and it's an existing configured package source, this command will add a trusted signer with the same name as that source to the trusted list.

If a `source-url` is provided, it MUST be a v3 package source URL (like https://api.nuget.org/v3/index.json). Other package source types are not supported.
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-source`
+
+```
+Adds a trusted signer based on a given source.
+
+If only `<name>` is provided, and it's an existing configured source, this command will add a trusted signer with the same name as that source to the trusted list.
```
loic-sharma: What do you think of this phrasing?

```suggestion
If only `<name>` is provided without `<source-url>`, the package source from your NuGet configuration files with the same name will be added to the trusted list.
```

This sentence is really hard! Maybe pull in a copywriter 😅

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: Consider adding < > to align with `<name>`:

```suggestion
If a `<source-url>` is provided, it MUST be a v3 source URL (like https://api.nuget.org/v3/index.json). Other source types are not supported.
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+OPTIONS:
+        --allow-untrusted-root          Specifies if the certificate for the trusted signer should be allowed to chain to an untrusted root.
+        --algorithm                     Specifies the hash algorithm used to calculate the certificate fingerprint. Defaults to SHA256. Values supported are SHA256, SHA384 and SHA512
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-source`
```
loic-sharma: Let's say I've already trusted 15 package owners for my mid-sized project. But now I want to add one more package that is owned by someone I did not previously trust. What would be my flow to append one more trusted owner? Do I need to provide all 16 package owners to the `trust-source` command? What would my flow be if I have a large project with many more owners?

The reason I ask is that I recently tried to set up trusted signers for the [NuGetGallery codebase](https://github.com/nuget/nugetgallery). This proved to be difficult using `nuget.exe`, so I resorted to manually editing the `nuget.config` file instead.

loic-sharma:
> We'll also want Kathleen to take a look at this, but I don't remember her GH handle :(

I believe her handle is @KathleenDollard :)

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+
+EXAMPLE:
+    $ dotnet nuget trusted-signers list
+    Registered trusted signers:
+
+    1.   nuget.org [repository]
+        Service Index: https://api.nuget.org/v3/index.json
+        Certificate fingerprint(s):
```
loic-sharma: Should we include trusted owners in this output?

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: One idea may be to add a `trust-owner` command that appends one or more owners to a trusted repository. See @zivkan comment here: https://github.com/NuGet/Home/pull/10628/files/5263069456da2276d0f3143fd5bf616e933b7809#r589669891

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: Also https://github.com/NuGet/Home/issues/8053

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
+# Title
+
+- Author Name [Kat Marchan (@zkat)](https://github.com/zkat)
+- Start Date 2021-03-04
+- GitHub Issue (GitHub Issue link)
+- GitHub PR (GitHub PR link)
+
+## Summary
+
+This specifies the syntax and behavior for the `dotnet.exe nuget trusted-signers` command.
```
JonDouglas: Should we remove the .exe terminology? Small nitpick, but this suggests Windows only. We removed this in docs a little ago after customers kept asking:

image

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: There's more on this here: https://docs.microsoft.com/en-us/dotnet/core/tools/?tabs=netcore2x#command-structure & https://github.com/dotnet/designs/blob/main/accepted/2018/cli-syntax-guideline.md

`$ dotnet [context] [action] [options] [context specification] [action type] [arguments]`

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+COMMANDS:
+    list (default)
+    trust-package <name> <package>...
+    trust-source <name> [source-url]
+    trust-certificate <name> <fingerprint>
+    remove <name>
+    sync <name>
+```
+
+#### `dotnet nuget trusted-signers list`
```
JonDouglas: Verb after context to match dotnet CLI guideline

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: I feel like if we use markers to describe the output, then we haven't done a good enough job explaining the output in the first place. I especially think describing `allowUntrustedRoot` is going to be difficult to explain in a single line at the bottom of the command.

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: > Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.

Maybe this? We should probably still implement it.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+    dotnet nuget trusted-signers remove [OPTIONS] <name>
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `dotnet nuget trusted-signers trust-package`
```
JonDouglas: dotnet nuget add trusted-signers [package] [other arguments]

Where package is an argument & the action is still `add`. Maybe?

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: Why is this not recommended? There seems to be valid cases where this is recommended.

proposed/2021/DotnetNugetTrustedSigners.md
zivkan: This command is used to trust the repository? I thought that's what the `trust-source` command was for?  What's the point of `trust-source` if `trust-package` does both?

Maybe this is what Kat meant when they said this feature is very confusing.

proposed/2021/DotnetNugetTrustedSigners.md
zivkan: I think it's fine to recommend against using it, but I feel that options should be sorted alphabetically in the help. Personally, I really find it much harder to find help on an option when the arguments are not sorted in the help. Searching console output is no where near as simple as searching a web page for an argument you see in your team's build scripts.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,196 @@
...
+
+```
+
+### Technical explanation
+
+For the most part, these commands map pretty directly to their nuget.exe
+counterparts, and most of their implementations should be reusable (removing
+`#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
+
+The exception here is the `add` command, which has been split into the various
```
JonDouglas: I personally think the add command would be better suited for being multi-purpose with different parameters. The syntax of `dotnet nuget add trusted-signers trust-*` seems a bit verbose compared to `dotnet nuget add trusted-signers <package(s)>` or `dotnet nuget add trusted-signers <service url>`.

In other words, it would be cool to just use one command and it sets up the trusted signers regardless of the parameter I pass in.

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: I think the biggest challenge is consistency throughout the dotnet CLI. I think decisions were made in the past that we have to adhere to in this spec as it might seem odd to users when using `nuget` contexts.

zivkan: We can't change how the settings are saved in the nuget.config file, but otherwise can we use this as an opportunity to rethink how to make all the commands easier to use?  Why do we have to have a 1:1 mapping between nuget.exe and dotnet cli commands? If we can learn from what we did in nuget.exe to make the dotnet cli commands easier to use, then that will be extra motivation for customers to use the better cli experience.

Or maybe the nuget.config schema is what you were referring to, not the commands.

JonDouglas (approved):
My main two areas of feedback are:

1. I think we have to remain consistent with the [dotnet CLI syntax guidelines](https://github.com/dotnet/designs/blob/main/accepted/2018/cli-syntax-guideline.md) in our parity efforts. So actions should follow the context.

2. I think the `add` action should simply support various parameters given to it such as `service index`, `certificate information`, and `package`. I think less is more with this command given typing "trusted" in twice seems odd.

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: I mostly meant in the sense of the CLI guidelines under the `nuget` [context umbrella](https://github.com/dotnet/designs/blob/main/accepted/2018/cli-syntax-guideline.md#form). I agree we should make things easier right now if there are opportunities to do so.

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: Agreed. We want to use the words "package source" as much as possible as that is what we use formally everywhere. 👍

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: Yeah definitely to alphabetical ordering. If we have compelling reasons to not recommend this, we should remove it instead of hiding it.

proposed/2021/DotnetNugetTrustedSigners.md
JonDouglas: There's a compelling argument here to break away from `trust-*` commands and just use `trust` with parameters. I personally say this is unneeded if we are already "adding" a "trusted-signer" which can be any of these items (package, source, etc)

proposed/2021/DotnetNugetTrustedSigners.md
zkat: This is already in that format? Am I missing something?

JonDouglas: The context is nuget here. So it would be 'dotnet nuget list trusted-signers'

proposed/2021/DotnetNugetTrustedSigners.md
zkat: I'm also working with Kathleen on specifics for this. It might end up having to be both ways, but this is what I'm doing for now.

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: > Why is this not recommended? There seems to be valid cases where this is recommended.

Normally, your operating system chooses which certificate authorities to trust or not. If a certificate authority isn't trusted, it's usually for a good reason (modulo the Debian incident). What valid cases do you have in mind @JonDouglas? In my opinion, this feature should rarely be used.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: idk, should we? I'm just emulating the output from nuget.exe here.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: oops, copy-paste fail. Thanks!

proposed/2021/DotnetNugetTrustedSigners.md
zkat: how do you differentiate packages from service URLs? I guess we're going to start detecting whether something looks like a URL and hope the regex does the right thing? What about for relative paths like `myprivatehost`? Is that a service index, or a local file?

proposed/2021/DotnetNugetTrustedSigners.md
zkat: There's a set of options that should be available in all commands. This is one of them, whether we take advantage of them or not.

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: I would, yes. For example:

```
    1.  nuget.org [repository]
        Service Index: https://api.nuget.org/v3/index.json
        Owners: microsoft, aspnet, nuget
        Certificate fingerprint(s):
            SHA256 - 0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D
```

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: Thanks for the good question.
Just checked the code.
In NuGet.exe command, the package path could be [wildcard](https://github.com/NuGet/NuGet.Client/blob/ac97d923e633f4ee65eec5ec690c4d99d7a922f3/src/NuGet.Core/NuGet.Commands/TrustedSignersCommand/TrustedSignersCommandRunner.cs#L93-L108), so there might be multiple paths after parsing.
But there is a checking on "Name" to reject any existing "Name" at https://github.com/NuGet/NuGet.Client/blob/ac97d923e633f4ee65eec5ec690c4d99d7a922f3/src/NuGet.Core/NuGet.Commands/TrustedSignersCommand/TrustedSignerActionsProvider.cs#L243
So when there are multiple packages, the first one will be added successfully, but the reset will fail for `A trusted signer 'Test' already exists.`
Apparently this is a bug. We should make the two consistent.
That is, do not accept multiple paths, or, accept multiple 'Name'.  
Raised issue at https://github.com/NuGet/Home/issues/10647

proposed/2021/DotnetNugetTrustedSigners.md
heng-liu: I have a similar question about a missing function in NuGet.exe:
After repo certificate rotation, we need to change the trustedSigners section to have two certs :
```
<trustedSigners>
  <repository name="nuget.org" serviceIndex="https://api.nuget.org/v3/index.json">
    <certificate fingerprint="0E5F38F57DC1BCC806D8494F4F90FBCEDD988B46760709CBEEC6F4219AA6157D" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
    <certificate fingerprint="5A2901D6ADA3D18260B9C6DFE2133C95D74B9EEF6AE0E5DC334C8454D1477DF4" hashAlgorithm="SHA256" allowUntrustedRoot="false" />
  </repository>
</trustedSigners>
```
But seems there is no command could do that. We have to edit config file manually.
Since we don't have that for NuGet.exe either, perhaps we can address that afterwards.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: I sat with @kathleendollard to clarify this, and the current entity ordering is correct. `dotnet nuget add source` is actually an anomaly and _not_ what we want to do, at least for now.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: see https://github.com/NuGet/Home/pull/10628#discussion_r591990207

proposed/2021/DotnetNugetTrustedSigners.md
zkat: see https://github.com/NuGet/Home/pull/10628#discussion_r591990207

proposed/2021/DotnetNugetTrustedSigners.md
zkat: See https://github.com/NuGet/Home/pull/10628#discussion_r591990207

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+    source <name> [source-url]
+    certificate <name> <fingerprint>
+    remove <name>
+    sync <name>
+```
+
+#### `> dotnet nuget trust list`
+
+```
+Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], it means that certificate entry has allowUntrustedRoot set as true.
```
loic-sharma, line 42: What do you think of this?

```suggestion
Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], NuGet will not reject the certificate if it chains to an untrusted root.
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+`dotnet` for NuGet-related operations.
+
+## Explanation
+
+### Functional explanation
+
+#### `> dotnet nuget trust`
+
+```
+Provides the ability to manage the list of trusted signers.
```
loic-sharma: What do you think of adding a little more detail? If I'm a NuGet newbie, I probably won't know what "trusted signers" are for. For example:

```suggestion
Manage the trusted signers. By default, NuGet accepts all authors and repositories. You can restrict this to just authors or repositories you trust using this command.
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+    dotnet nuget trust list [OPTIONS]
+
+EXAMPLE:
+    $ dotnet nuget trust list
+    Registered trusted signers:
+
+    1.   nuget.org [repository]
+        Service Index: https://api.nuget.org/v3/index.json
+        Owners: microsoft, aspnet, nuget
+        Certificate fingerprint(s):
```
loic-sharma: It looks like the name has an extra space in the output. Is that intentional?

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: I suggest we either:

1. Remove the `[U]` marker altogether, or...
1. Replace the marker with some text next to the SHA-256 fingerprint, or...
1. Define the `[U]` marker somewhere in this output. For the single line explanation, I suggest: `[U]: NuGet will not reject the certificate if it chains to an untrusted root.`. We can skip definition if no certs allow an untrusted root.

I'd vote for option 2 or 3 personally.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+#### `> dotnet nuget trust source`
+
+```
+Adds a trusted signer based on a given package source.
+
+If only `<name>` is provided without `<source-url>`, the package source from your NuGet configuration files with the same name will be added to the trusted list.
+
+If a `<source-url>` is provided, it MUST be a v3 package source URL (like https://api.nuget.org/v3/index.json). Other package source types are not supported.
+
+If <name> already exists in the configuration, the signature(s) will be appended.
```
loic-sharma: I'm not sure I understand this line. Is "signature(s)" supposed to be "certificate(s)"?

proposed/2021/DotnetNugetTrustedSigners.md
zkat: I'm not sure. I think it's good to make it clear that the `[U]` corresponds to an actual option in the config file, you know? What do you think, @heng-liu ?

proposed/2021/DotnetNugetTrustedSigners.md
zkat: nope :)

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+## Motivation
+
+This is part of a larger effort to reach full parity between `nuget` and
+`dotnet` for NuGet-related operations.
+
+## Explanation
+
+### Functional explanation
+
+#### `> dotnet nuget trust`
```
loic-sharma, line 21: I'm a huge fan of the new `trust` verb. Very clean 🥇

proposed/2021/DotnetNugetTrustedSigners.md
zkat: fix'd

proposed/2021/DotnetNugetTrustedSigners.md
zkat: kudos to @JonDouglas for the idea :)

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: Hm... IMHO ideally no one ever uses `allowUntrustedRoot`. If so, maybe we shouldn't bring up such an advanced topic in the command description? It is difficult to balance between useful information and verbosity. What do you think of only explaning `[U]` in command output if the user has `allowUntrustedRoot` enabled somewhere as per https://github.com/NuGet/Home/pull/10628#discussion_r592009496? If so, we can detail both the `allowUntrustedRoot` config plus give a quick human description of what that entails.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+
+```
+Manage the trusted signers. By default, NuGet accepts all authors and repositories. You can restrict this to just authors or repositories you trust using this command.
```
KathleenDollard: Does that mean if you trust any that makes all others untrusted? If so, I think this could say it more directly.

proposed/2021/DotnetNugetTrustedSigners.md
KathleenDollard: Just pointing out... you can use links to docs on more advanced topics. We anticipate doing this when we clean up all the CLI help near the end of .NET 6 or start of 7.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `> dotnet nuget trust sync`
+
+```
+Requests the latest list of certificates used in a currently trusted repository to update the the existing certificate list in the trusted signer.
```
KathleenDollard: As someone not familiar with this space, I found it difficult to understand this sentence. Is the most important part first? If not try that.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+
+Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.
```
KathleenDollard: Oh, wait. If this is what it does, why not just use this note as the description?

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
+```
+
+#### `> dotnet nuget trust author`
+
+```
+Adds a trusted signer with the given name to the config, based on the author signature(s) of one or more packages. If <name> already exists in the configuration, the signature(s) will be appended.
```
KathleenDollard: For all of these, what does "to the config" add for a normal use reading this. Suggest dropping it  in all these commands.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+
+```
+Lists all the trusted signers in the configuration. This option will include all the certificates (with fingerprint and fingerprint algorithm) each signer has. If a certificate has a preceding [U], it means that certificate entry has allowUntrustedRoot set as true.
+
+USAGE:
+    dotnet nuget trust list [OPTIONS]
+
+EXAMPLE:
+    $ dotnet nuget trust list
+    Registered trusted signers:
```
KathleenDollard: I do not like the example being included here. The user will see this output as soon as they run the command and it will push the usage off the screen if you do not maximize your terminal window. 

Also, will the list be long? Consider putting in your backlog doing a partial match, but talk to me about making your format consistent with the rest of the CLI direction if you do that (so maybe wait and see what people want).

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+
+Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.
+
+USAGE:
+    dotnet nuget trust sync [OPTIONS] <name>
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
+        --interactive                   Allows command to stop and wait for user input or action.
```
KathleenDollard: We usually include -i as abbrev for --interactive, unless you have a legacy conflict. Suggest adding for all commands.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+Requests the latest list of certificates used in a currently trusted repository to update the the existing certificate list in the trusted signer.
+
+Note: This gesture will delete the current list of certificates and replace them with an up-to-date list from the repository.
+
+USAGE:
+    dotnet nuget trust sync [OPTIONS] <name>
+
+OPTIONS:
+    -h, --help                          Prints help information.
+    -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
```
KathleenDollard: Update this in all commands please.

These are now the standard CLI/.NET Tools verbosities and MSBuild is just to be specific among ourselves. 

```suggestion
    -v, --verbosity <LEVEL>             Set the verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
```

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+        --interactive                   Allows command to stop and wait for user input or action.
+        --configfile                    The NuGet configuration file. If specified, only the settings from this file will be used. If not specified, the hierarchy of configuration files from the current directory will be used.
```
KathleenDollard: Look at this in the output and see how bad the wrapping is. I do not think the second sentence is helpful and this seems long and verbose for help. This is probably a description you use a lot across NuGet help (and the CLI), so let's see if we can find some nice concise consistent language.

proposed/2021/DotnetNugetTrustedSigners.md
KathleenDollard: I actually think this is nice. When you add a signer from a package you are not adding the package, but using it to find the certificate. When you remove the package, it does not matter where you got it. 

Not symmetric, but in this case the action is not symmetric

proposed/2021/DotnetNugetTrustedSigners.md
KathleenDollard: I suggest adding to the option description, and perhaps including a link to an explanation of why. Something like "Not recommended (link)." 

Experienced folks know they can break the rules, but hopefully consider.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+For `trust author` and `trust repository`, `<package>` may be a glob/wildcard.
+This must be consistent across platforms, so the Windows version must expand
+these wildcards when it receives them as arguments.
+
+#### `> dotnet nuget trust certificate`
+
+```
+Adds a trusted signer with the given name to the config, based on a certificate fingerprint.
+
+Note: If a trusted signer with the given name already exists, the certificate item will be added to that signer. Otherwise a trusted author will be created with a certificate item from given certificate information.
```
KathleenDollard: As someone unfamiliar with this space, I do not know what name means. Is it part of the certificate?

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+
+#### `> dotnet nuget trust author`
+
+```
+Adds a trusted signer with the given name to the config, based on the author signature(s) of one or more packages. If <name> already exists in the configuration, the signature(s) will be appended.
+
+USAGE:
+    dotnet nuget trust author [OPTIONS] <name> <package>...
+
+EXAMPLE:
```
KathleenDollard, line 92: I do not see this example adding much.

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
+# dotnet nuget trust command
+
```
KathleenDollard, line 2: Overall comment on help: 

System.CommandLine may be displaying your help (ask Sarah), in which case some of this will just be fixed. 

There are some inconsistencies in Help here with the rest of the CLI. For example, the section headers are not all caps in the CLI. Try a command like `dotnet build` (we have an issue on the wide first column)

proposed/2021/DotnetNugetTrustedSigners.md
```diff
@@ -0,0 +1,239 @@
...
+```
+
+### Technical explanation
+
+For the most part, these commands map pretty directly to their nuget
+counterparts, and most of their implementations should be reusable (removing
+`#if IS_DESKTOP` as needed, from the various TrustedSigners classes).
+
+The exception here is the `add` command, which has been split into separate
```
KathleenDollard, line 181: ? there is no longer an add command. Appears a couple of times in this section.

KathleenDollard (commented):
I think we landed in a good place. My comments are tweaks.

proposed/2021/DotnetNugetTrustedSigners.md
loic-sharma: To me `<name>` and `<package>` are vague terms. From this command alone, it's unclear how `<name>` will be used. Also, it's also unclear that `<package>` should be a path to a .nupkg file. The example does help clarify some of these questions.

Is it possible to add descriptions for the command arguments? Maybe that'd be better than an example.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: Yeah, that part is fine. The help format here is mostly for informative purposes. Things don't need to be pixel-perfect.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: This is referring to the `nuget trusted-signers add` command, which is what has been split up.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: The name in all of these refers to the identifier the trusted signer is saved under. The name of the entity itself, if you will. It's an arbitrary name that we use to key off of.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: I'm impartial. tbh just tell me which one y'all like best and I'll do that :)

proposed/2021/DotnetNugetTrustedSigners.md
zkat: I have a generalized concern that this feature is very complex and hard to explain in general, and I'm having trouble figuring out how to write these docs in a way that creates clarity. :(

proposed/2021/DotnetNugetTrustedSigners.md
zkat: It's the key we save the signer under, in the config file, and what you use to refer to it later.

proposed/2021/DotnetNugetTrustedSigners.md
zkat: That's exactly what this does. I'll try a different wording.

JonDouglas (approved):
I'm a big fan of the current rework. I think it meets all my concerns & really helps simplify this confusing area. Well done!

JonDouglas:
I did read through this blog last night & there were some UX items potentially to cover: https://haacked.com/archive/2019/04/03/nuget-package-signing/

1. When trusting a package author, we need a .nupkg. Can we additionally provide a UX to download the package for the user by specifying a package name / version as an alternative to supplying the package?

One work around is using an API which is a bit hidden to users: i.e. https://api.nuget.org/v3-flatcontainer/newtonsoft.json/12.0.1/newtonsoft.json.12.0.1.nupkg (We should do that for people if we can)

i.e. `dotnet nuget trust author Contoso Contoso.Library`

2. Additionally as a future thought, we have no way to just trust a package author through this method. We still are trusting a package in some fashion. Is there a way we can grab details from the package author to then trust the author? In other words, to say we trust "jdouglas" & any package they upload would be "trusted".

i.e. `dotnet nuget trust author jdouglas`

erdembayar:
@zkat
I can run below command `nuget trusted-signers Add -Name Nuget -config ..\nuget.config` and it works.
But how does it work for `dotnet nuget`?

erdembayar:
@zkat 
Also you can write example full line of code for each of cases?
`dotnet nuget trust repository`
`dotnet nuget trust certificate`
`dotnet nuget trust author`
`dotnet nuget trust source`

zkat:
@erdembayar The equivalent of `nuget trusted-signers Add -Name NuGet -config ..\nuget.config` is [`dotnet nuget trust source NuGet --configfile ..\nuget.config`](https://github.com/NuGet/Home/pull/10628/files#diff-144567c0cde83f8e17df62e525e71e33ee5da8a608bd8abb2e29aa742cbc1544R151-R173)

zkat:
as far as examples go: I got some feedback during code review that unless the example was fairly intricate, it added no value having such a section, so I removed them. Is there a different way I can help clear things up?

erdembayar:
> as far as examples go: I got some feedback during code review that unless the example was fairly intricate, it added no value having such a section, so I removed them. Is there a different way I can help clear things up?

Currently on there is only 1 example: `dotnet nuget trust author Contoso ./contoso.library.nupkg` 
I'm new to package signing, I still have doubt if I'm reading spec correctly, please verify and correct if following trusted-signers commands are valid so we can have them as unit tests in our implementation. 

```
dotnet nuget trust repository NugetOrg ./newtonsoft.json.13.0.1-beta1.nupkg
dotnet nuget trust certificate MyCompanyCert  CE40881FF5F0AD3E58965DA20A9F571EF1651A56933748E1BF1C99E537C4E039 --algorithm SHA256
dotnet nuget trust source NuGetTestTrust https://apidev.nugettest.org/v3-index/index.json --allow-untrusted-root true --owners "NugetTestOrg;MSTestOrg;TrustedTesOwners" 
dotnet nuget trust sync NugetOrg
dotnet nuget trust remove NugetOrg
dotnet nuget trust list -v:d
```

zkat:
Some corrections below:

`--allow-untrusted-root` does not take an argument. It's a flag:
```
dotnet nuget trust source NuGetTestTrust https://apidev.nugettest.org/v3-index/index.json --allow-untrusted-root --owners "NugetTestOrg;MSTestOrg;TrustedTesOwners"
```

Don't separate the verbosity with a `:`. Just put a space:
```
dotnet nuget trust list -v d
```

The others look fine to me!

PR #10842, opened 2021-05-07
Initial draft of technical spec for the Transitive Dependencies feature in the PMUI

### LINKED ISSUE #10747: Write technical spec for transitive dependencies feature in PMUI

jebriede:
Fixes: https://github.com/nuget/home/issues/10747

### DIFF
--- a/proposed/2021/TransitiveDependenciesInPMUI.md
+++ b/proposed/2021/TransitiveDependenciesInPMUI.md
@@ -1,0 +1,95 @@
+# Transitive Dependencies in the Visual Studio Package Manager UI - Technical Spec
+
+* Status: **In Review**
+* Author(s): [Jean-Pierre Briedé](https://github.com/jebriede)
+* Issue: [10184](https://github.com/NuGet/Home/issues/10184) Show transitive dependencies in PMUI for PackageReference projects
+* Type: Feature
+
+## Problem Background
+
+Developers can only see their top-level packages in their PackageReference project in the Package Manager UI. Transitive dependencies are not visible in the NuGet Package Manager. This makes it difficult for a developer to fully understand all of the NuGet package dependencies their project relies on. Furthermore, there is currently no way to see in the Package Manager UI what top-level dependency depends on each of the included transitive dependencies.
+
+A developer may want to install a transitive dependency as a top-level one to specify what version they want to include in their project if there is a security concern, vulnerability, or version conflict with the dependant version of a transitive dependency. The developer should have a quick and easy way to do this from the Package Manager UI in Visual Studio.
+
+## Who are the customers
+
+* All developers using Visual Studio and PackageReference projects will benefit from the additional information about the transitive dependencies and the top-level package that depend on them.
+* Developers that have installed packages that depend on transitive dependencies with security concerns, vulnerabilities, or version conflicts with other transitive dependencies will benefit from the ability to quickly install a transitive dependency as a top-level dependency and specify its version.
+
+## Requirements
+
+* UI is accessible and allows screen readers to announce pertinent information about the transitive dependencies and the top-level packages that depend on them
+* UI is localizable. All strings are in appropriate resource files and available for translation
+* Reading transitive package information is done asynchronously to minimize the impact on the responsiveness of the UI
+
+## Goals
+
+* Show transitive dependencies in the Project-level NuGet Package Manager
+
+### UI behavior goals:
+* Display a flat list of transitive depedencies along with the installed, top-level packages in the Visual Studio NuGet Package Manager UI
+* Allow the user to collapse and expand the list of transitive dependencies
+* Collapse the transitive dependencies list by default
+* For each transitive dependency, show the top-level package(s) that depend(s) on it in a tooltip, displayable on hover and readable by screen reader
+* Show the package details when a transitive dependency is selected and allow the ability to install a specific version of the transitive dependency which will promote it to a top-level package
+
+### Technical goals:
+* Enhance the InfiniteScrollList to support grouping its list content based on a field in the list item data context
+* Allow the InfiniteScrollList grouping to be enabled or disabled by the consumer so it can still be shared by the Browse, Installed, Updates tabs
+* Create an API to read transitive dependency information from the assets file, including the top-level package information for each transitive dependency
+* Cache the transitive dependency information read from the assets file so it is only read when the assets file changes
+
+## Non-Goals
+
+* Show transitive dependencies for the Solution Package Manager
+* Build transitive dependencies into the NuGet command-line interface
+* Redesign how the assets file is read
+* Design a new pattern for how the assets file information is cached
+* Add significant infrastructure for A/B experimentation
+* Build complete transitive dependency management into the Package Manager UI
+* Rebuild the PMUI or restructure to use the MVVM pattern. Attempts will be made to use MVVM where possible but it is not the goal to change the structure or patterns used in the PMUI to use the MVVM pattern where it is currently not
+* Redesign how metadata for installed packages is queried. Transitive packages will use the existing mechanism for pulling in metadata about installed packages
+
+## Solution
+
+### InfiniteScrollList
+
+**The InfiniteScrollList will be enhanced to optionally enable grouping its items by a given property and will be styled to have the group headers have an expander and a count of the elements in the group.**
+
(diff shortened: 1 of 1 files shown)


proposed/2021/TransitiveDependenciesInPMUI.md
```diff
@@ -0,0 +1,95 @@
...
+
+We will make use of existing engagement metrics to determine the feature's impact on the user's interaction with the PMUI. Namely, we will look closely at NuGet Actions (Package installs, updates, and uninstalls), as well as PMUI Refresh events on each of the tabs. We will consider adding additional metrics specific to transitive dependencies in consultation with the VS Experimentation team.
+
+
+## Considerations
+
+* The transitive packages list should be collapsed by default because the user may not need to see transitive package information every time they view the Installed tab. There is a potential opportunity to delay processing or querying transitive dependency or the top-level packages that installed the transitive dependencies until after the list is expanded. However, doing so would be premature optimization so this will only be considered if it is shown to have a significant impact on performance. Since the assets file is read from disk into memory, and processing will occur on that in-memory representation, I do not anticipate a significant impact, but this consideration can be visited as an option if we see otherwise.
```
donnie-msft, line 94: > doing so would be premature optimization so this will only be considered if it is shown to have a significant impact on performance. 

Loading all transitives up front, regardless of whether the user wants to see them, doesn't sound to me like a premature optimization. I think that should be part of this spec to consider the impact, and whether there's scenarios where we can pre-load, and some where we should not pre-load transitives.

proposed/2021/TransitiveDependenciesInPMUI.md
```diff
@@ -0,0 +1,95 @@
...
+* Transitive dependencies and the top-level packages that caused them to be installed are to be read from the assets file on disk. If that file is not present when the PMUI is loaded and installed packages are displayed, we will need to understand the impact it will have on this scenario and devise a solution. It is difficult to anticipate what the best options are without reaching that point, but some options may include doing a restore to create the file, prompting the user, or not displaying transitive dependencies and informing the user why they are not visible until after a restore is completed and the assets file is created.
```
donnie-msft, line 95: > or not displaying transitive dependencies and informing the user why they are not visible until after a restore is completed and the assets file is created.

This is my vote. The UI should immediately handle the situation where they aren't available. Not wait on anything. The UI today blocks itself on too many background tasks. Any new UI should strongly consider relaying state information to the user in a friendly way.

proposed/2021/TransitiveDependenciesInPMUI.md
```diff
@@ -0,0 +1,95 @@
...
+### UI behavior goals:
+* Display a flat list of transitive depedencies along with the installed, top-level packages in the Visual Studio NuGet Package Manager UI
+* Allow the user to collapse and expand the list of transitive dependencies
+* Collapse the transitive dependencies list by default
+* For each transitive dependency, show the top-level package(s) that depend(s) on it in a tooltip, displayable on hover and readable by screen reader
+* Show the package details when a transitive dependency is selected and allow the ability to install a specific version of the transitive dependency which will promote it to a top-level package
+
+### Technical goals:
+* Enhance the InfiniteScrollList to support grouping its list content based on a field in the list item data context
+* Allow the InfiniteScrollList grouping to be enabled or disabled by the consumer so it can still be shared by the Browse, Installed, Updates tabs
```
donnie-msft, line 38: > so it can still be shared by the Browse, Installed, Updates tabs

The first part of the bullet I thought was about giving a checkbox/etc to enable/disable grouping.
This statement is confusing. What is being shared? If you mean the user's choice to enable/disable, are we certain that's what we want? Previous discussions with PM it sounded like some settings should be per-tab, not shared.

jebriede: The "it" in this sentence refers to the InfiniteScrollList. This design adds support for the InfiniteScrollList to be grouped optionally and I'm emphasizing that it will still support displaying its contents without grouping which is important so that the same control can still be used in all tabs. The optionality here is programmatic and not driven by user action. We will not be exposing any option to allow for disabling grouping because that doesn't fit into the design.

proposed/2021/TransitiveDependenciesInPMUI.md
jebriede: It would be premature to optimize for a scenario without first measuring its impact, especially since any optimization will have tradeoffs that need to be carefully considered.

Currently, transitive dependencies are already being read in up front because of the PackageRecommender so there will be no change there in continuing to pull in transitive dependencies.

What this feature will add is an algorithm that follows the top-level dependency to the transitive dependency to build a list of top-level packages that are responsible for installing the transitive dependency. That algorithm will work against an in-memory representation of the assets file. We will need to profile how much time that algorithm takes to determine if needs to be optimized before investing in optimization that could affect both PackageRecommender and adding extra code to delay loading.

Regarding adding to the spec that we should consider the impact, I think that's what this section is doing under the title of Considerations. I call out the potential for optimization and that we will visit possibilities for improving the performance if we see the algorithm has a significant impact on the performance. Please let me know if you think there's a way I could communicate this more clearly in this section. Thanks!

proposed/2021/TransitiveDependenciesInPMUI.md
donnie-msft: > We will not be exposing any option to allow for disabling grouping because that doesn't fit into the design.

Ah ok, I wasn't aware of that decision. @JonDouglas does this sound like what you expected?

proposed/2021/TransitiveDependenciesInPMUI.md
donnie-msft: > That algorithm will work against an in-memory representation of the assets file. 

What will the UI look like when there is not an in-memory representation?

proposed/2021/TransitiveDependenciesInPMUI.md
jebriede: Thanks for your input! This sounds good to me too. Maybe we'd show a message in the place where transitive dependencies would appear or somewhere at the top of the UI. We can figure out a good place for it once we find if we need to solve this scenario and how.

proposed/2021/TransitiveDependenciesInPMUI.md
jebriede: That's covered in the Considerations section. We commented on it below.

proposed/2021/TransitiveDependenciesInPMUI.md
jebriede: The [design spec](https://github.com/NuGet/Home/blob/dev/proposed/2020/Transitive-Dependencies.md) outlines how the UI should behave. Grouping the list by top-level versus transitive packages is how the UI will make the difference between the packages (top-level vs transitive) easily distinguishable and allow for each section to be collapsed or expanded. Allowing the user to ungroup the list of installed packages has never been a consideration that I am aware of.

donnie-msft: Nevermind, since the "it" meant a specific control, I no longer have a question about a shared user-setting.

JonDouglas (approved):
Great proposal & reasonable to me! Awesome job.

proposed/2021/TransitiveDependenciesInPMUI.md
JonDouglas: > It is difficult to anticipate what the best options are without reaching that point, but some options may include doing a restore to create the file, prompting the user, or not displaying transitive dependencies and informing the user why they are not visible until after a restore is completed and the assets file is created.

Definitely agree it's hard to anticipate what the options will be here. Perhaps when we're further along we can address any concerns if they come up. My initial question & probably @nkolev92 or @zivkan could help out here is how often is the assets file not present? Perhaps just doing the right thing (restore) on demand will make the best experience.

donnie-msft (approved):

nkolev92 (commented):
Overall all of this looks pretty good! Well done!

Have we considered what happens to the labels on the installed/updates tabs wrt to deprecation/vulnerability?

Or is that future work? 

In particular, you'd get a warning on the installed tab if *any* of your top level packages are deprecated. 
I'm not arguing for either approach, mostly just bringing it up for completeness.

jebriede:
@nkolev92 thank you for your feedback! That's a great question. Both deprecation and vulnerabilities are each planned for future work and will build on the transitive dependencies being available and showing in the PMUI. For example, for vulnerabilities there is this spec:
https://github.com/NuGet/Home/blob/7f7361af32a0f0170743a752b15ced655740501b/proposed/2020/PackageVulnerability/PMUIVulnerability.md

nkolev92:
That makes sense. 

One thing that isn't immediately clear to me, would the warnings only be shown when the transitive dependencies are expanded?
I imagine that's the case, but want to clarify it.

Issue #10839, opened 2021-05-06
Simplify project file formatting for including files in NuGet packages (icon, license, README)


chrisraygill:
We should simplify the process of including files in the nupkg to ideally a single-line experience like most other package metadata.

One of the advantages of creating SDK-style projects is the ease of interpreting and editing project files. The majority of package metadata can be specified in a single line while embedded files like icons, license files, and README files will need a line in the `<PropertyGroup>` and `<ItemGroup>`. 

Converting these metadata elements into one-liners will have the following impacts:
* Make the project file look "cleaner" and more interpretable.
* Make it easier for beginners to copy and paste elements from tutorials or examples.
* Make it faster and easier to include these critical metadata elements overall.

chrisraygill:
Closing as dupe of https://github.com/dotnet/sdk/issues/2142

loic-sharma:
This feature would reduce 16 lines down to 11 lines, a 30% reduction in boilerplate in your project file. Before: 🙊


```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>

    <PackageIcon>icon.png</PackageIcon>
    <PackageReadmeFile>README.md</PackageReadmeFile>
    <PackageLicenseFile>LICENSE.txt</PackageLicenseFile>
  </PropertyGroup>

  <ItemGroup>
    <None Include="README.md" Pack="true" PackagePath="\"/>
    <None Include="icon.png" Pack="true" PackagePath="\"/>
    <None Include="LICENSE.txt" Pack="true" PackagePath="\"/>
  </ItemGroup>
</Project>
```

After: 😎

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>

    <NewPackageIcon>icon.png</NewPackageIcon>
    <NewPackageReadmeFile>README.md</NewPackageReadmeFile>
    <NewPackageLicenseFile>LICENSE.txt</NewPackageLicenseFile>
  </PropertyGroup>

</Project>
```

loic-sharma:
Reopening after offline chat with Chris. This feature request synergizes well - but is separate - from https://github.com/dotnet/sdk/issues/2142. This issue is separate as licenses, icons, and readmes must be labeled in the package's generated .nuspec manifest.

MarkPflug:
[...]

Issue #1203, opened 2015-08-17
Reduce friction for Release Notes maintenance - release notes pack support


mteper:
Too many packages do not include release notes, and I speculate that the reason lies in the friction involved in doing so, as the developer is required to actively maintain the .nuspec file. Many projects build directly from the VS project bypassing .nuspec altogether.

A few changes would improve this situation:
- add support for pulling in content from an external file
- have `nuget pack` look for releasenotes.md
- add support for a standard changelog URL
- add parameter to `nuget pack` for passing in changelog URL

*maintainer edit*

- Package Manager UI side - https://github.com/NuGet/Home/issues/1823

ashmind:
ReleaseNotes.md would be great, especially considering ASP.NET's `project.json` does not allow newlines other than `\r\n`, which is unusable for any large-ish metadata.

Alternatively -- sections from ReadMe.md for both Description and Release Notes (description has same problem with `project.json`).

leppie:
Allow release notes to be constructed from commits to git or TFS, etc.

jrossstocholm:
I strongly second this - the friction to create the release notes package configuration is too high.

We build our nuget packages from Visual Studio Online, and I would actually much prefer to simply be able to add release notes via a command line parameter. In VSO we have already set up a build parameter with version number information, so we configure the Nuget Packager build step with "Nuget arguments"

-version $(NugetVersion)

It would be great to be able to add simple text like

-releaseNotes $(&lt;insert build variable here&gt;)

yishaigalatzer:
@NickCraver has provided more insights and some good ideas. These should be addressed when we handle this suggestion. https://github.com/NuGet/Home/issues/1734

304NotModified:
In the recent updates (NLog, yeah we're listed https://github.com/NuGet/Home/issues/1734) we do add the changelog, but we found out it's pretty useless as the Visual Studio client isn't even showing the changelog... 

edit: I see that's https://github.com/NuGet/Home/issues/1823

rmarinho:
Any progress on this?

NickCraver:
My comments from #1734:

In short: I propose making the Release Notes field markdown, or at the very least have nuget.org atuomatically convert URLs into hyperlinks in the field. Right now the breakdown of the top 50 packages on nuget.org is:
- **33** have no release notes
- **12** have only a link (or a link + a few words)
- **5** have release notes in some form

[...]

PR #10867, opened 2021-05-13
Re-design of the NuGet.org Package Details page

jcjiang:
Introduction of our re-design concept for the NuGet.org package details page, which allows a developer to access at a glance the information they need to choose and use a package.

[Find the rendered spec here.](https://github.com/NuGet/Home/blob/20a62681790ca815fcbd0ee366a608c299953cec/proposed/2021/NuGetOrgDetailsPageRedesign.md)

🎉📦Please provide your comments, concerns, and feedback within this PR. 📦🎉


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=10867) <br/>All CLA requirements met.

khalidabuhakmeh:
I mentioned on Twitter, that the focus on the install bar is probably not helpful to devs and takes up a lot of visual real estate. What if we moved it to the right-hand bar and put more of an emphasis on the contents of the package?

image

NickCraver:
IMO, it's a bit weird to give the README this much presence at this point in time. While it may be at the same top place today, the current state of affairs is _most_ packages (even the top ones) seem to have 1-2 lines of text there max (am I way off here? maybe analysis has been done already). The current design pushes the other content out of view and _seems_ like it'd be less accessible. Today the #1 thing I'm looking at is dependencies, so much so that I have a user script that keeps that element expanded (it'd be nice if NuGet just remembered this - even `localStorage`), putting it on another tab is further way, especially when the result is a mostly blank landing page for most major libraries it seems like (I'm betting most in general?).

For my personal use case, I'd love to see which frameworks a package was built for, and how many dependencies it has - that can _likely_ be a sidebar item, that deep links to the tab. I'm not sure if that's a minority use case though.

It's not _directly_ related to this proposal but as a package author I really can't see release notes taking off and deserving a UI redesign around - as a part of the package that's shipped/immutable and cannot be adjusted, I'm always going to opt to keep my release notes at a URL elsewhere. Do we know how many packages might use this? I know that in https://github.com/NuGet/Home/issues/1203 there's a proposal to embed a README file in there from a path (which I agree with if used at all), but it's still immutable, and I'd definitely still just link out to readme notes.

That's a rather long way of saying I guess: can we instead let me have release notes as a URL, and put it in the sidebar if it is one - save the tab.

RichiCoder1:
Echoing twitter also, my feedback:
* Only append the version to the package command if I'm viewing a specific version
* I think the readme being at the forefront is fine (especially with the a new integrated README experience), but ya'll might want to constrain the width of the body/whole page like npmjs does. It assist reading a lot and most READMEs assume the GitHub UI. (Edit: I would in fact try to align witht the GitHub README as close as posssible when rendering it out.) The readme experience on npmjs in general is excellent.
* Personal preference, but it'd be nice if the tabs had iconagraphy to provide visual distinction
* Do people download packages and symbols often? Seems to give those a lot of real estate
* It'd be nice to see dates, specifically release date.

> For my personal use case, I'd love to see which frameworks a package was built for, and how many dependencies it has - that can likely be a sidebar item, that deep links to the tab. I'm not sure if that's a minority use case though.

Big agree. Sidebar frameworks, and keep the framework tabs for a deep dive.
Also +1 on adding a count to the dependencies tab and used by tabs.

dend:
My initial reaction is that all of the links on the right blend in - it's a sea of blue, that a user needs to parse to understand what's going on. The icons are small enough, that the differentiation is barely noticeable.

WhiteBlackGoose:
> IMO, it's a bit weird to give the README this much presence at this point in time. While it may be at the same top place today, the current state of affairs is most packages (even the top ones) seem to have 1-2 lines of text there max 

As a package maintainer, I agree. I don't want to duplicate content from a website or from github repo to a nuget page, so I simply leave a couple of references to resources there.

chrisraygill:
@WhiteBlackGoose thanks for commenting! I have a couple questions:
1. What would you prefer the default tab to be?
2. What is your concern with "duplicate content" from GitHub? With the [new embedded README feature](https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/), it should be easy to add files that already exist on GitHub like the repo README or package-specific documentation/onboarding. The benefit to the package consumer is being able to evaluate or onboard to your package more efficiently. Especially in cases where the repo README isn't package-specific, it can be a pain for consumers to find the right place to learn more about the package.

dend:
> 
> 
> > IMO, it's a bit weird to give the README this much presence at this point in time. While it may be at the same top place today, the current state of affairs is most packages (even the top ones) seem to have 1-2 lines of text there max
> 
> As a package maintainer, I agree. I don't want to duplicate content from a website or from github repo to a nuget page, so I simply leave a couple of references to resources there.

That's a fairly standard pattern, though @WhiteBlackGoose - for example, here is a package example on PyPI: https://pypi.org/project/gee2drive/

The same content is in the README: https://github.com/samapriya/gee2drive

I wouldn't expect every package user to go to GitHub to learn all the details, and speaking here as a dev myself - it's convenient to get the information where the package is.

WhiteBlackGoose:
> With the new embedded README feature, it should be easy to add files that already exist on GitHub like the repo README or package-specific documentation/onboarding

I'm taking my words back, now it sounds cool. I missed that post so assumed that there would be some limited version of readme and, as a result, I'd need to work on another version of readme specifically for nuget. If it is not so... sounds great then. With all this said, of course, readme should be the first tab.

WhiteBlackGoose:
@dend if I recall correctly, the last time I tried to upload my whole readme, it responded that it's too long 😕 . Let's see how the new version works, I'll surely try it as soon as it's released.

chrisraygill:
Hi @NickCraver ! Thanks for the thorough and thoughtful feedback 😊 I have a couple thoughts:

> IMO, it's a bit weird to give the README this much presence at this point in time. While it may be at the same top place today, the current state of affairs is _most_ packages (even the top ones) seem to have 1-2 lines of text there max (am I way off here? maybe analysis has been done already). 

You're 100% correct that most packages don't have READMEs yet. The legacy README experience had low adoption and the new [embedded README feature](https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/) is still in preview. That being said, all the mockups for the NuGet.org details re-design are thinking forward to when embedded READMEs have more adoption. Our hope is that README being the default tab will have the following effects:
1. Incentivize README adoption among package authors (which should be relatively low effort/overhead with the new README feature).
2. Make it easier for prospective package consumers to discover, evaluate, and onboard onto new packages (feedback we've heard in our surveys and interviews)

> For my personal use case, I'd love to see which frameworks a package was built for, and how many dependencies it has - that can _likely_ be a sidebar item, that deep links to the tab. I'm not sure if that's a minority use case though.

What do you see as the benefit of a sidebar item deep-linked to a tab vs. just clicking on the tab? Stands out more visually?

> It's not _directly_ related to this proposal but as a package author I really can't see release notes taking off and deserving a UI redesign around - as a part of the package that's shipped/immutable and cannot be adjusted, I'm always going to opt to keep my release notes at a URL elsewhere. Do we know how many packages might use this? I know that in #1203 there's a proposal to embed a README file in there from a path (which I agree with if used at all), but it's still immutable, and I'd definitely still just link out to readme notes.
> 
> That's a rather long way of saying I guess: can we instead let me have release notes as a URL, and put it in the sidebar if it is one - save the tab.

What makes you prefer a release notes URL over embedded markdown release notes like the new embedded README feature? It's true that it's immutable, but it should automatically pack the latest version of your changelog/release notes along with the package once you have it defined in your project file. It shouldn't be any additional overhead to update it on NuGet.org with each release beyond updating the document you maintain in your repo that you would otherwise link to.

The benefit of embedding it is that we can surface it directly on NuGet.org for much easier accessibility and a more consistent experience across packages. We may also be able to parse it to surface relevant information in client products like VS. Other ecosystems like pub.dev already use this approach: https://pub.dev/packages/http/changelog.

Thanks again for the feedback!

NickCraver:
Welcome! Answers below:
> What do you see as the benefit of a sidebar item deep-linked to a tab vs. just clicking on the tab? Stands out more visually?

To be clear: it wouldn't be just a link in the sidebar, but a section listing the frameworks, something like:

Dependencies:
- [netstandard2.0](#) (0 deps)
- [net472](#) (5 deps)

...and each could deep link to the tab for listing dependencies.

> What makes you prefer a release notes URL over embedded markdown release notes like the new embedded README feature? It's true that it's immutable, but it should automatically pack the latest version of your changelog/release notes along with the package once you have it defined in your project file. It shouldn't be any additional overhead to update it on NuGet.org with each release beyond updating the document you maintain in your repo that you would otherwise link to.

The fact that it's immutable is the main thing. You cannot amend releases notes for a release. Didn't get a link right? Didn't credit someone correctly? Didn't include a change that made it in? Didn't make a critical breaking release note that's causing a lot if issues to be filed? Things like this happen, and "just updating them next release" doesn't even work as a solution because when looking at an old version and it's release notes...you're hosed. If they're on GitHub, in the repo, GitHub pages, etc.? Easy to fix. If I fix them there, all we have is a skew between what was published and the updated version. And so, you won't find me updating this feature...I'll be updating the source-of-truth we have today, at a URL, without all of the immutability drawbacks.

Overall: it may benefit the NuGet team and users to have it right there, but it does not benefit me as a maintainer - it introduces restrictions and headaches. So as a maintainer, I wouldn't use this feature - I may be an outlier though, but want to convey that feedback so it's a known scenario.

arpadbarta:
The idea to use the README as description, would definitely be a great improvement over the current package description.
On the other hand I might belong to the minority here however, the things that I mainly look for are:
- Supported frameworks
- Dependencies 
- Last update

Would love to see the supported frameworks in a prominent area, without any navigation.

chrisraygill:
@arpadbarta what kind of information are you looking for in the "dependencies"? How many there are, whether or not your trust them, if they're mostly 1st or third party, etc.

danjagnow:
> That being said, all the mockups for the NuGet.org details re-design are thinking forward to when embedded READMEs have more adoption.

This makes a lot of sense to me, and I agree that it should help drive embedded README adoption.  When I'm evaluating a NuGet package, the first thing I want to know is what it's for.  Package authors are in the best position to communicate that.  After I know whether it might meet my needs, I start looking at dependencies, reputation, and other indicators.  Those are important, but they are secondary to understanding what the package does.

> It's not _directly_ related to this proposal but as a package author I really can't see release notes taking off and deserving a UI redesign around - as a part of the package that's shipped/immutable and cannot be adjusted, I'm always going to opt to keep my release notes at a URL elsewhere.

I understand the value of making release notes immutable, but I agree that it sets the bar pretty high for the level of maturity required to achieve that.  One thing that would help is documented examples of automated build pipelines on a variety of popular CI/CD services that demonstrate a strategy for versioning, maintaining release notes, and publishing to NuGet.org.  Alternatively, perhaps it would be possible to support publishing a NuGet package in a private draft status that allows online revisions to the release notes (and possibly embedded README) before signing (when not already signed by the author) and public release.

arpadbarta:
> @arpadbarta what kind of information are you looking for in the "dependencies"? How many there are, whether or not your trust them, if they're mostly 1st or third party, etc.

Definitely the number of them and their source. Have not thought about grouping/or differentiating somehow based on their publisher (First party and others), however that could be helpful.

jcjiang:
@khalidabuhakmeh Sweet design! Yeah, the placement of the installation bar was something our team went back and forth on a _lot._ We had heard a bunch of varying opinions on how useful that bar was and we weren't sure if the difference was because of people's experience levels. We weren't sure if the installation bar was helpful for someone just starting out, or if it would just bombard them with information they don't know how to make sense of - or don't need. 

I'm curious about your experience (and that of anyone reading!) with the installation bar, especially over time. Did you use it often at any point? When would you use it? When would you not? I'm wondering if there are some situations where the installation bar is super helpful and others where it's pretty useless.

@NickCraver REALLY appreciate the thoughts. Thanks for bringing up Release Notes - that's something I honestly always felt weird having as it is now, an immutable blob of text with no clear separation of versions. We were considering combining the Release Notes and Versions tab, with the (usually URL) notes beside the version for easy scanning. Here's what it would have looked like: 

image

I'm thinking that we could check Release Notes text input for URLs and then link directly to them, potentially right next to each version so you can select from a high-level view. What do you think?

khalidabuhakmeh:
@jcjiang I find the current bar not helpful since more users will end up using their tooling to find and install packages. Whether that be through Visual Studio, JetBrains Rider, or Visual Studio Code. The explicit version number is also more annoying. What is useful is the package name and if there is a difference in the installation process. (i.e. library dependency, templates, etc.).

Either way, I think the initial redesign is an improvement over what is there currently. Cheers!

jcjiang:
@RichiCoder1 @arpadbarta @NickCraver Thanks for bringing up frameworks! We actually have something in the works right now to show a package's target frameworks in two places on the page - badges in the header (and maybe eventually on the package search results page) and the dedicated Frameworks tab for a deep dive. This is what it might look like: 

image

and in the header:

image

As you can tell by the mock-ups, we're still making a few final decisions for the badges. One thought is to replace the colored-background .NET text with 'Target Frameworks' instead, to lessen confusion about what the badges indicate. Any parts you find confusing or would want changed?

NickCraver:
> I'm thinking that we could check Release Notes text input for URLs and then link directly to them, potentially right next to each version so you can select from a high-level view. What do you think?

I think it's a neat approach, but as an addon author I would always point it at a single URL (so I think it would just repeat here?) If that's an acceptable use case: awesome, works for me!

> As you can tell by the mock-ups, we're still making a few final decisions for the badges. One thought is to replace the colored-background .NET text with 'Target Frameworks' instead, to lessen confusion about what the badges indicate. Any parts you find confusing or would want changed?

It's nice to list the frameworks it's built for, but I wonder if the research is potentially a bit dated over time here. For a while there, it was very relevant _if_ your package was build for a particular framework (e.g. is this ready for .NET Core yet?) but that becomes less relevant and we can almost tell by the date alone. What I'm interested in is what the UI today actually does a better job of: _what does it cost me_? e.g. "what are the dependencies, on the version I'm after?" It could be a weight issue, a license issue, or something else. I always want to see the dependencies I'd be pulling in, but maybe that's a minority use case...I have no idea!

xt0rted:
As a package author I don't want to maintain a readme, changelog, or release notes outside of my repo nor do I want to deal with syncing them between environments (github, docs, wiki, nuget, etc.).

My repos all have a readme, changelog, tag, and usually a github release with a copy of the changelog and any release notes. All of this can now be automated with github actions. It's also not uncommon to have to edit these later on which is why I never include it in the nupkg or try to add it to nuget.org.

If sourcelink is used, or a repo url is provided, all of these things should automatically be pulled in and linked to. That would reduce the maintenance burden and hopefully increase the quality of information for all packages.

When looking at nuget.org the only project information I expect to see is the readme and a link to the repo. Release notes have never been relevant when searching for a package, they are relevant during updates though which dependabot handles by pulling that info from the project's repo. For anyone manually updating, having a link on the right side of the package's page to the changelog and a link to the release (if available) alongside each version feels like it should be sufficient enough.

Overall I think whatever can be done to simplify publishing and maintaining packages in the registry should be the goal. When I compare what I need to do to setup and publish to npm vs. nuget it's pretty overwhelming. I like the new readme support, but it feels like this should happen automatically and not require opting-in or configuring anything. Even simple things like a package icon feel like overkill to me.

anangaur:
> Overall I think whatever can be done to simplify publishing and maintaining packages in the registry should be the goal. When I compare what I need to do to setup and publish to npm vs. nuget it's pretty overwhelming. I like the new readme support, but it feels like this should happen automatically and not require opting-in or configuring anything. Even simple things like a package icon feel like overkill to me.

@xt0rted Thanks for your comment. Seems @loic-sharma agrees with you 😄: 
*Simplify project file formatting for including files in NuGet packages (icon, license, README) https://github.com/NuGet/Home/issues/10839*

Show ⬆️ some love by upvoting it and also providing your input. I think you go beyond the issue by asking for some auto include of artifacts. /cc: @chgill-MSFT

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+**Requirements:**
+- Download statistics 
+  - Total downloads
+  - Downloads of current version
+  - Average downloads per day
+- External links
+  - Personal site
+  - Source repository
+  - License
+- Download package
```
chrisraygill, line 98: It might be helpful if we had an info icon or similar after the "Downlod symbols" link because many people might not know what it's for.

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+
+**Requirements:**
+- Download statistics 
+  - Total downloads
+  - Downloads of current version
+  - Average downloads per day
+- External links
+  - Personal site
+  - Source repository
+  - License
```
chrisraygill, line 97: Is the typical symbol for a license the checkmark with ribbons? GitHub uses the balancing scales icon, which we may want to consider for consistency.

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+      - Of these, 23.2% are not actually owned by Microsoft. These packages have 330M total downloads. (Note that one package ID may have many package versions)
+
+
+
+### Body
+
+Tabs with titles that clearly convey what information is in them.
+
+**Requirements:**       
+- README Tab
```
chrisraygill, line 120: Consider changing the capitalization to "Readme". All caps looks a little loud and distracting compared to the other tab labels. While "README" is frequently used in documentation, both Pub.Dev and NPM use "Readme" in tab labels, probably for the reason stated above 😊

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+
+![](../../meta/resources/NuGetPackageDetails/headernew.png)
+
+Information that needs to be seen at a glance such as package ID, description, and eventually, TFM badges.
+
+**Requirements**        
+- Title   
+- Icon  
+    - Hidden for packages without icon
+- Version
```
chrisraygill, line 66: Consider showing when the version being viewed was published. Based on feedback in the issue and customer interviews, how "updated" a package is is a critical detail. It may also be worthwhile to surface the last time to package was updated as a whole in the sidebar.

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Goals and Non-Goals
+
+### Header
+
+![](../../meta/resources/NuGetPackageDetails/headernew.png)
+
+Information that needs to be seen at a glance such as package ID, description, and eventually, TFM badges.
+
+**Requirements**
```
chrisraygill, line 62: We should consider how our current warnings/notifications will look on the new design if at all different. There's a box that pops up for:
* Package is indexing
* There is a later version of this package
* The package is deprecated
* The package has a vulnerability
and maybe a few other scenarios as well.

proposed/2021/NuGetOrgDetailsPageRedesign.md
```diff
@@ -0,0 +1,140 @@
...
+
+Additional helpful information such as package installation instructions, statistics, external links. 
+
+![](meta/resources/NuGetPackageDetails/sidebar_final.png)
+
+**Requirements:**
+- Download statistics 
+  - Total downloads
+  - Downloads of current version
+  - Average downloads per day
```
chrisraygill, line 92: Are downloads per day of the current version or the overall package? We may want to consider only showing 2 of these to save on space and make it more focused. 

My vote would be keep Total and Current Version downloads and put the Per Day in the "full stats" page. It may be helpful to get customer feedback on that tho 😊

chrisraygill:
Overall the new designs look great! Most of my comments are just suggestions to consider, but I think this is already going in a good direction!

chrisraygill (approved):

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: @loic-sharma @sophiamfavro Do you know of any good resources explaining why and how one would download symbols? Maybe we can link from the page

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: Unsure, but it is the current symbol used by NuGet.org to denote license and we decided to do a direct port for this re-design (leaving additional improvements to what/how information is presented for future iterations.) Definitely something we should look at later on though

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: Great point about the difference in usage within documentation vs. tab labels! For easy reference, this is what pub.dev and NPM looks like:

image 
image

@loic-sharma  @sophiamfavro What do you two think?

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: Love it! Will make the edit

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: Great point! We talked a bit yesterday about the warnings/vulnerabilities system. The TL;DR is that there is a lot of complexity there (14 different banners with no real rhyme or reason for why some are one color and others are another) and we decided to punt the implementation outside the scope of the project.

We'll keep the current warnings/notifications, but Maryanne and I will work on re-designing them to be more clear and useful so it can be picked up by devs in the future

proposed/2021/NuGetOrgDetailsPageRedesign.md
jcjiang: Overall package!

That makes sense and I agree that we should be more intentional about what stats we show vs. hide on the "full stats" page. For this work, however, I plan on keeping these three just because they are the three displayed on the current package details page. Good to have a baseline to compare against as we start modifying how/what we show instead of just the way we show information.

jcjiang:
Thanks for all the feedback, folks! Lots of good stuff here. One bit of context here is that this re-design will be the first iteration of many. The main goal here is to do a direct port of information we currently display on the page and displaying it in a more clear way. That way we have a baseline to compare against for later iterations as we include additional data/info to display and/or modify functionality. 

Here's a summary of topics and responses:

- **Moving the installation bar to the sidebar:** Something our team went back and forth a lot, to be honest. I think the core issue here is we don't yet have a great sense of when/how people use that. Some people have told us that it is the one thing they need to see right away, others have told us that it's useless to them. With that, I plan to keep the installation bar in the header for this iteration and have it as an open question for the next.

- **Giving less emphasis to README:** Looking at the discussion, the big remaining concern is that uploading the README will add additional friction to the package publishing process. I agree that we should make this process smoother, but I feel that there is value to prominently showing the README right now for both package authors and consumers - especially in terms of driving adoption in the community. 

- **Make Release Notes more useable:** Totally agree that Release Notes is currently not very useful. Actually, the team is planning on revamping the content for most of the tabs you see, especially Frameworks and Dependencies. That's going to be in the next iteration however, just so we get a baseline for how useful people currently find these sections and get a better idea of what new functionality/info people would want to see in them. 

- **Reduce friction in package publishing/maintaining process:** Absolutely! I'm actually working on something to this effect now, with an emphasis on making package publishing/maintaining better for both new and experienced .NET developers. While this problem is out of the scope of the re-design, I will be bothering y'all for thoughts as the other proposal gets more coherent. Stay tuned!

- **Show better frameworks and dependencies info:** Talked with the team about this and while we generally agreed that the info is important for determining package quality, we were concerned that by showing certain numbers and data points without others, we remove indicators of quality from their context. In other words, a package can be very lightweight and have few dependencies but may completely bomb in other criteria. If we only show the dependency information and not the other pertinent data, we might skew the package selection process of users.
- We have some work with package quality and scoring that @JonDouglas is driving, which could be a good place to tackle this problem. The idea is we want to show better indicators of package quality, but we want to make sure we're choosing the right criteria to show. If we come up with a too simplistic definition of package quality, we would negatively impact package authors - and we really don't want that. 

- **Include release date information:** Really good point and we will include it in this re-design.

If you have any big "this re-design should not move forward without this change!" comments, now is the time! I'll close this thread end of today so the devs can start implementing this iteration ASAP, and I'll put all the other suggestions together for future iterations. I'll tag y'all in those proposals as well because it would be great to hear your thoughts.

proposed/2021/NuGetOrgDetailsPageRedesign.md
chrisraygill: Can we keep a running list of considered small design fixes that we'd like to test out after the initial iteration? My concern being that most of these are minor issues we may struggle to individually prioritize once the bulk of this feature is complete. It'll probably we overall less work to review and address them as we're refining this redesign 😊

Your approach for the initial iteration makes complete sense to me 👍

jcjiang:
Hey all! Thank you for your help here. I wanted to share that the new redesign of the package details page has gone live and you can see it by navigating to any package on NuGet.org. We have a blogpost about it here: https://devblogs.microsoft.com/nuget/introducing-the-new-nuget-org-package-details-page/

This is a first step and there is definitely a lot more work brought up in this thread that still has to be done. The first of them, displaying target framework compatibility information on NuGet.org, will be coming out sometime in October. Stay tuned!

NickCraver:
I've been playing with this most of the morning due to updating a project and its packages. I do appreciate all the effort going into this. I have hit several issues, some of which I didn't think of in mock-up phase. In hopes it'll help, summarizing feedback here and happy to expound on anything:

Pros:
- Thanks for remembering which tab I was on, if I'm commonly going for _one_ of these views like dependencies, that's handy and is appreciated. I think this is double-edged though - if I an commonly using _multiple_ tabs - which I am, it's potentially doing more harm than good.
- The dependencies are expanded when you get to them, which is something I was doing before every page.

Cons:
- Overall: everything is taking me longer. It's simply more clicks to get any information that used to be a scroll. I have to find where to navigate and then look.  I know some of this is reiterating the points I made in the mockup phase, but after actual use:
  - Most `README` sections of any package I've looked at this morning are 1 or 2 lines, but that's the landing screen...the space then seems needlessly unused.
  - README and Release Notes are often 2 lines total between the 2 (the first being a sentence or two, the second being a link) - to put these behind 2 clicks still seems not awesome, before they were much faster to consume and use. I am including Microsoft's main packages in this statement - they're often 1-liner READMEs.
- Navigation:
  - On my standard working laptop (which notably as a _taller_ 16:10 screen), the primary navigation I care about within a package is half-way down the screen. It's also the 3rd overall horizontal button group on the page.
    - It's also worth nothing rows 2 and 3 of navigation shift even further down the page when not on the latest version of a package.
  - The 2 rows of tabs (one for package manager selection and one for navigation) is really throwing me a lot more than I thought it would in these mockups. I've mis-clicked many times now, and since the first isn't of consistent height, it also moves the package nav (e.g. Cake has multiple lines and others have an optional bottom notice).
- Confusion/Simplicity:
  - I'm not sure it's a win to include the "Prefixed Reserved" notice on a package, I'd wager that confuses more users than it helps (over only a checkmark, or some other indicator): 
image
    - The intention here is to convey that this is verified in some way, but we're saying "prefix reserved" which means something to package authors and the NuGet team, but probably few consumers.
    - Tangent to design: [Clicking through the indicator](https://docs.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation), I'm having trouble parsing the description actually important here for _consumers_: "Package consumers are provided with additional information when the packages that they are consuming are not deceptive in their identifying properties." - what additional information? The best I can figure, "when" should be "that" in this description, e.g. "Package consumers are provided with additional information that the packages that they are consuming are not deceptive in their identifying properties." ...the current "when" implies this happens often and is a bit concerning to read.

Requests:
- I can't seem to link someone to a tab. If we're putting the information behind more clicks, at the least we should be able to link them directly there. IMO, something like `#versions` on the URL using HTML5 push history would do well here. It should automatically update as you navigate tabs, so that you're linking someone to what you're looking at when copying the current URL.
  - If doing this, please _replace_ history, not add to it - users almost certainly wouldn't want to navigate back through every ta when hitting the back button.
- Since Versions is behind another click now, please link me directly in places we can. For example when on an older package, I almost always want to go to the latest when there's an indicator that there is a newer version. Currently users see this: 
image
  - I propose the first line on that message be something like: "There is a newer version of this package available, the latest version is `<a>`2.0.90`</a>`".

I realized I'm probably an outlier user and my usage here may not be important or the goal, but thus far I do much prefer the previous layout that had a quick scroll to get to all the information available. If the dependencies were expanded (or remembered that you expanded them) in the last layout, it'd have been the only click saver I can think of. I am focused on productivity and in that respect, the new layout is slower to get information out of and harder to link people to (the latter of which can be fixed!). Totally get that this probably isn't designed for my use case, but I hope some of the feedback is useful in some way.

loic-sharma:
Thanks @NickCraver for taking the time to write such thoughtful and thorough feedback. I opened the following work items to track your suggestions:

* https://github.com/NuGet/NuGetGallery/issues/8822
* https://github.com/NuGet/NuGetGallery/issues/8823
* https://github.com/NuGet/docs.microsoft.com-nuget/issues/2535

> Most README sections of any package I've looked at this morning are 1 or 2 lines, but that's the landing screen...the space then seems needlessly unused.

We believe that READMEs provide an excellent "getting started" experience for users. This is aligned with .NET 6's theme of making .NET more approachable to new developers and students: https://github.com/dotnet/core/issues/5465. Unfortunately we have a chicken and egg problem. Our old design tucked away the README, so authors weren't aware they could add a README, resulting in very low README adoption. We understand that this transition is a little jarring in the short-term, but we are looking closely at how we can improve README adoption.

> The 2 rows of tabs ... is really throwing me

Do you switch often between different installation instructions? Could you explain a little more your scenario here?

> I'm not sure it's a win to include the "Prefixed Reserved" notice on a package

We received lots of feedback that folks didn't understand the checkmark by itself. We're hoping to reduce customer's confusion by adding the new "Prefix reserved" link to docs for additional context. Thanks for your feedback on the docs, these are good comments we should address!

Issue #10678, opened 2021-03-23
NuGet should not eagerly start restore when there are pending design time builds


nkolev92:
Experience tracked: [1355522](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1355522)

A couple of real time scenarios here: 

* A user edited a common file such as `Directory.Build.Props`
* Git operations such as branch switch.

NuGet has an easy tell about when to restore on solution load (wait for all the projects to be ready), during branch switching, that's not obvious because most of the projects are already ready.

This issue is blocked on [1283992](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1283992)

nkolev92:
There's a design in flight for this: 
```cs

// a new service defined inside NuGet to align with current services.

public interface IVsSolutionRestoreService4

{

                // Project system calls NuGet service (optionally) to register itself.

                // Each project can only register no more than one time

                // when project is unloaded, it need dispose the registration object.

                IDisposable RegisterRestoreInfoSource(string projectUniqueName, IVsProjectRestoreInfoSource restoreInfoSource);

}

 

 

// implemented by the project system, and is called by NuGet to know whether there is any pending restore work

// so it can hold the current batch.

public interface IVsProjectRestoreInfoSource

{

                // true means the project system plans to call NominateProjectAsync in the future.

                bool HasPendingRestore();

 

                // NuGet calls this method to wait project to nominate restoring

                // If the project has no pending restore data, it will return a completed task.

                // Otherwise a task which will be completed once the project norminate the next restore

                // the task will be cancelled, if the project system decide it no longer need restore (for example: the restore state has no change)

                // the task will be failed, if the project system runs into a problem, so it cannot get correct data to norminate a restore (DT build failed)

                Task WhenRestoreNominated();

}
```

lifengl:
After thinking about this again, I now prefer a slightly different version of it:

The difference:
 instead of each project registers itself, it can be done by providing one component for one type of project systems (depending on the implementation of the project system, one can be more efficient or easier to do that another one. Also, potentially, other features like a batch file changes might use it to block restore as well.  Basically, it decouple a source from a single project, and NuGet doesn't need manage projects by name as well.

The drawback: if NuGet wants to use the state based on a subset of projects. For example, if it decides a project which is not depending on any project touched by the restore operation & being referenced by the set, it might ignore the pending state of a specific project. Although this is quite complex, and also p2p references are dynamic, so i feel the second option sounds better.

[...]

Issue #7622, opened 2018-12-11
NominateProjectAsync should only restore the project being passed


adalon:
NuGet version: latest
VS version: dev16 (and probably dev15 too)
OS version: win10

CPS-based projects will trigger a `NominateProjectAsync` when the active configuration changes (or when the project template is unfolded too). 

Repro steps:
1- With nuget package restore settings enabled, create an empty solution 
2- Add New Project -> Class Library (.Net Standard)
3- Add New Project -> Console App (.Net Framework)
4- Change the current active configuration from Debug to Release.

At this stage the CPS-based library will trigger a `NominateProjectAsync` doing [1], [2] and [3] and the expected `restoreinformation` should only contain the `.Net Standard` project but it also contains the another `csproj` library. Resuming:

Expected: `restoreinformation` should only contain the project being nominated
Actual: `restoreinformation` is also containing other projects in the solution which were not nominated

Additional scenarios where this issue also impacts:
- `runintellisensebuild` is also triggered as the result of the `csproj` (unnecessary) nomination and might cause perf impact
- When unfolding a multi-project template which contains at least one netstandard library in it (Xamarin Cross Platform App for instance) the solution perf load might be significantly impacted if any of the other projects being unfolded contain several nuget packages to be restored (Android project for instance in the Xamarin example mentioned above). 

Finally I could not find where the project unique name is being honored in the `VsSolutionRestoreService` [4] or [5]

[1] https://github.com/dotnet/project-system/blob/ddfc471669edc4b2806de3b2a731c7c10c58f2a3/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs#L77

[2] https://github.com/dotnet/project-system/blob/ddfc471669edc4b2806de3b2a731c7c10c58f2a3/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs#L103

[3] https://github.com/dotnet/project-system/blob/ddfc471669edc4b2806de3b2a731c7c10c58f2a3/src/Microsoft.VisualStudio.ProjectSystem.Managed.VS/ProjectSystem/VS/NuGet/PackageRestoreInitiator.PackageRestoreInitiatorInstance.cs#L148

[...]

PR #10874, opened 2021-05-17
New APIs to coordinate bulk restore 

nkolev92:
Design for https://github.com/NuGet/Home/issues/10678

[Rendered](https://github.com/NuGet/Home/blob/d5c35c3c06c66bdcf23cc95062304fb55fb1e800/proposed/2021/BulkRestoreCoordination.md)

### DIFF
--- a/proposed/2021/BulkRestoreCoordination.md
+++ b/proposed/2021/BulkRestoreCoordination.md
@@ -163,3 +163,8 @@ N/A
 
 There are *many* implementation details around this feature, which can include things such as reporting that restore has been delayed etc.
 This API is designed with those opportunities in mind.
+
+## References
+
+- [Design time builds](https://github.com/dotnet/project-system/blob/main/docs/design-time-builds.md)
+- [Restore to intellisense](https://github.com/nkolev92/restore-to-intellisense/), [presentation (Microsoft only)](https://msit.microsoftstream.com/embed/channel/8855a1ff-0400-a936-24ca-f1eaa053d9d5?app=microsoftteams&sort=undefined&l=en-us#)
--- a/proposed/2021/BulkRestoreCoordination.md
+++ b/proposed/2021/BulkRestoreCoordination.md
@@ -1,0 +1,165 @@
+# Bulk restore coordination
+
+- [Nikolche Kolev](https://github.com/nkolev92)
+- Start Date 2021-05-03
+- GitHub Issue [10678](https://github.com/NuGet/Home/issues/10678)
+- GitHub PR [4058](https://github.com/NuGet/NuGet.Client/pull/4058)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a solution with SDK based projects is loaded, NuGet uses a sliding window to coordinate the number of restores being run to avoid extra restore/design time build loops.
+When a solution is loaded, any future changes will trigger a restore eagerly. Operations such as branch switch are one example.
+This proposal tackles the addition of a new API to coordinate the restore during the bulk changes operations.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+There are a few scenarios where a bulk project edit may happen that affects NuGet restore.
+
+- Branch switch
+- Directory.Build.props or other shared file edits
+- Find and replace in csproj
+
+These operations are currently not very efficient and may run extra restores and design time builds. In some cases, re-opening a solution may be more efficient than a partial branch switch.
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+There are no functional changes to any experiences.
+The user actions are all the same, the changes have a performance focus.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+#### Target scenario
+
+Ex:
+Given a solution with 3 projects.
+Proj A -> Proj B -> Proj C.
+
+Say a Directory.Build.Props is edited that affect *all* projects.
+Say the nominations come in the following order:
+
+Proj B
+Proj C
+Proj A
+
+- When the nomination for B comes in, NuGet will trigger a restore that will affect projects B & A, given that A depends on B.
+- While that first restore runs, it possible that C & A nominations have arrived. When NuGet receives these nominations and the previous restore completes, which happens last, a new restore will be kicked off, with the nomination data for C & A.
+- This second restore will affect A, B & C. Project A & B have been restored twice, thus causing unnecessary work.
+- In the ideal scenario, only 1 restore for each 3 projects is done.
+
+#### Restore batching
+
+At solution load, project-system runs DT builds and call the `IVsSolutionRestoreService` to nominate a project.
+NuGet is aware of *all* projects that are supposed to nominate, so NuGet can wait for *all* projects to be nominated.
+Currently NuGet has a sliding window to account for potentially failed projects.
+
+When the solution is loaded, NuGet doesn't have an additional heuristic to determine whether there are other projects that might need a restore done.
+The proposal is that each time NuGet determines a restore might need run, it checks whether there are *any* projects with pending *design time builds*, that *might* nominate. If yes, NuGet will wait.
+
+#### New interfaces
+
+##### IVsSolutionRestoreService4
+
+```cs
+    /// <summary>
+    /// Represents a package restore service API for integration with a project system.
+    /// Implemented by NuGet.
+    /// </summary>
+    [ComImport]
+    [Guid("72327117-6552-4685-BD7E-9C40A04B6EE5")]
+    public interface IVsSolutionRestoreService4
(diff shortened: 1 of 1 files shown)


proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,165 @@
...
+
+- [Nikolche Kolev](https://github.com/nkolev92)
+- Start Date 2021-05-03
+- GitHub Issue [10678](https://github.com/NuGet/Home/issues/10678)
+- GitHub PR [4058](https://github.com/NuGet/NuGet.Client/pull/4058)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a solution with SDK based projects is loaded, NuGet uses a sliding window to coordinate the number of restores being run to avoid extra restore/design time build loops.
```
erdembayar, line 11: Thank you creating this spec.
Could be able to include graph for possible nomination paths could happen for both before and after this change?
We don't have full context for this?

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,165 @@
...
+
+Say a Directory.Build.Props is edited that affect *all* projects.
+Say the nominations come in the following order:
+
+Proj B
+Proj C
+Proj A
+
+- When the nomination for B comes in, NuGet will trigger a restore that will affect projects B & A, given that A depends on B.
+- While that first restore runs, it possible that C & A nominations have arrived. When NuGet receives these nominations and the previous restore completes, which happens last, a new restore will be kicked off, with the nomination data for C & A.
```
erdembayar, line 54: This one doesn't have to be C & A, yeah? I believe even just nomination for C can cause this.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,165 @@
...
+
+##### IVsProjectRestoreInfoSource
+
+```cs
+    /// <summary>
+    /// Represents a service that NuGet can use to ask a project-system whether there are any pending design time builds.
+    /// Implemented by the project-system.
+    /// </summary>
+    [ComImport]
+    [Guid("35AD5FF2-6AB7-48E9-BCC0-189042410FA6")]
```
erdembayar, line 102: Did you generate this random new Guid?

proposed/2021/BulkRestoreCoordination.md
nkolev92: > Could be able to include graph for possible nomination paths could happen for both before and after this change?

I'm not sure I follow.

Currently the implementation is a sliding window. 
Effectively, NuGet waits for all projects to be nominated or no nomination for 10s before it starts a restore.

proposed/2021/BulkRestoreCoordination.md
nkolev92: Yeah, it's just an example of what could happen.

proposed/2021/BulkRestoreCoordination.md
nkolev92: Yep, we generate a new guid to associate with that interface anytime we add a new interop interface. 

If you take a look at https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Clients/NuGet.SolutionRestoreManager.Interop, all interfaces have a GUID.

proposed/2021/BulkRestoreCoordination.md
erdembayar: Here you mentioned design time build. What are the other builds? When does other builds happen in which order?
Simple graph would be enough?

nkolev92: > Here you mentioned design time build. What are the other builds? 

I [gave a presentation](https://msit.microsoftstream.com/embed/channel/8855a1ff-0400-a936-24ca-f1eaa053d9d5?app=microsoftteams&sort=undefined&l=en-us#) on that topic, please take look. The thumbnail suggests you were a part of it :) 

The design time builds are what the project-system does that leads to a nomination. 

I added a few references at the bottom of the doc.

erdembayar: Link refers all videos. What was the name of video? 
In general having simple diagram helpful for other people too, they don't have to read everything.

nkolev92: The name is NuGet and intellisense stage in Visual Studio, fixed the [link](https://msit.microsoftstream.com/video/e449a1ff-0400-9887-1aca-f1eb55f40c4b?channelId=8855a1ff-0400-a936-24ca-f1eaa053d9d5). 

> In general having simple diagram helpful for other people too, they don't have to read everything.

https://github.com/nkolev92/restore-to-intellisense/ is closest thing to a write-up, and I added it as a reference.
It is a very broad topic, which is why I have to refer you to the video again. 

This is design that's been discussed with various SMEs on the perf, CPS and project-system teams.

erdembayar: Ok. Pardon me for my short memory. 😭 
I believe your referenced links are enough.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,170 @@
...
+
+- [Nikolche Kolev](https://github.com/nkolev92)
+- Start Date 2021-05-03
+- GitHub Issue [10678](https://github.com/NuGet/Home/issues/10678)
+- GitHub PR [4058](https://github.com/NuGet/NuGet.Client/pull/4058)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a solution with SDK based projects is loaded, NuGet uses a sliding window to coordinate the number of restores being run to avoid extra restore/design time build loops.
```
erdembayar, line 11: By `sliding window` means `sliding window of time frame` in below?

https://github.com/NuGet/NuGet.Client/blob/1ef0444b1ff1ca79ebde7a6e3a717a4ff74b1832/src/NuGet.Clients/NuGet.SolutionRestoreManager/SolutionRestoreWorker.cs#L505

nkolev92: yes, that line + the 3 lines before.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,170 @@
...
+Say the nominations come in the following order:
+
+Proj B
+Proj C
+Proj A
+
+- When the nomination for B comes in, NuGet will trigger a restore that will affect projects B & A, given that A depends on B.
+- While that first restore runs, it possible that C & A nominations have arrived. When NuGet receives these nominations and the previous restore completes, which happens last, a new restore will be kicked off, with the nomination data for C & A.
+- This second restore will affect A, B & C. Project A & B have been restored twice, thus causing unnecessary work.
+- In the ideal scenario, only 1 restore for each 3 projects is done.
```
erdembayar, line 56: Should be strive for restore lowest level first and higher ones later? So all projects only restore once, otherwise we always restore higher ones if any lower project nominated. Of course it can introduce other delays, in case of `A` -> `B`->`C` .. -> `Z` solution we keep waiting until  `Z` to nominate, while others are asking for nomination.

nkolev92: That's a possibility, but there are some caveats.

When actual restore runs there's a lot of overlap among different projects, so restoring them at the same time allows us to minimize duplication, and to cache certain results in memory. For example many of the SourceRepository resources have lifetime caches, such as HttpFileSystemBasedFindPackageByIdResource. 

Just because NuGet is not running is a restore that doesn't mean that there's nothing going on either. There's other components utilizing the resources available. 

That change wouldn't be trivial, so one important consideration the project reference depth and that's not something we've studied yet to decide whether that effort would be worth it.

At this point we're interested in the total time it takes to complete the intellisense stage, as that's when the user can start doing work.

In a different world, maybe VS would have an idea which projects the customer is interested in working on (or had last open) and we'd priority getting that project ready for work sooner, but that's not what's happening today.

Related https://github.com/NuGet/Home/issues/7622.

We may indeed change it some day, but for now the focus is on minimizing the restore count.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -0,0 +1,170 @@
...
+Proj A -> Proj B -> Proj C.
+
+Say a Directory.Build.Props is edited that affect *all* projects.
+Say the nominations come in the following order:
+
+Proj B
+Proj C
+Proj A
+
+- When the nomination for B comes in, NuGet will trigger a restore that will affect projects B & A, given that A depends on B.
```
erdembayar, line 53: Last question. Here nomination means something changed so CPS is telling VS this project is ready for restore. Is it right?

nkolev92: Yes, mostly.

I am going to nitpick your statement an attempt to clearly communicate the separation. 

> CPS is telling VS this project is ready for restore

Firstly:

> CPS is telling VS

CPS is really the project-system. In theory it should be CPS that's calling NuGet, but it's actually managed languages project system right now.

> this project is ready for restore

There's a model at project system side that's generated from design time builds. When a design time build is run, the data is collected and compared against the last data that project-system sent to NuGet. 
If it is different, a nomination call happens. 

Each nomination call tells NuGet that a restore *may* be needed. It's NuGet that decides whether a restore runs and if it does whether any changes occur. 

Hope that clarifies it.

Issue #10793, opened 2021-04-22
Full development cycle with Symbols


BlackGad:
# Details about Problem

NuGet system implementation currently has almost all parts for full development cycle with symbols.

This is basically first stop on the long road to resolving target issue https://github.com/NuGet/Home/issues/6579.

So lets start with stages of full development cycle:
1. Create and publish - Includes build project, pack binary package, pack symbols package and publish it to the registries.
2. Restore and build - Includes binary package restore and dependent project build.
3. Launch and debug - Execute application from binaries, debugger attachment etc.

# Current implementation state

## Create and publish
![pack](https://user-images.githubusercontent.com/8100521/116000329-6632a680-a5f8-11eb-9701-35544840a48a.png)

This stage fully implemented. Tools and applications which uses NuGet can pack binaries into packages accordingly to nuspec configuration. With additional options they can automatically create snupkg symbols packages. Also v3 web registries can consume binary AND symbols packages for future usage.

From technical perspective both .nupkg and .snupkg similar because same code used for their creation (just minor nuspec changes and exclude filters).

**Update:** Official NuGet registry reformat snupkg package internally to Symbols artifact for built-in Symbols server

## Restore and build

![Build_current](https://user-images.githubusercontent.com/8100521/115671942-b50be200-a353-11eb-99ef-763102438494.png)

Stage begins from restore step. Tools and applications restores existing (install new ones) packages into specific folder to allow builder (Visual Studio on a picture) compose real application. NuGet protocol(framework) has built in discovery and cache mechanisms that why actual tools do not care about implementation. Everything what they must to do is call proper API methods.

## Launch and debug

![Debug_current_1](https://user-images.githubusercontent.com/8100521/116000482-16081400-a5f9-11eb-89f0-471b809bbb5c.png)

Executing our binaries and attach debugger (Visual Studio here). In order to debug external binaries it is required to load their symbols. 
Generic approach for Visual Studio is to use symbols from only one of [three places](https://docs.microsoft.com/en-us/visualstudio/debugger/specify-symbol-dot-pdb-and-source-files-in-the-visual-studio-debugger?view=vs-2019#symbol-file-locations-and-loading-behavior):
- The location that is specified inside the DLL or the executable (.exe) file.
- The same folder as the DLL or .exe file.
- Debugger options for symbol files.

Picture above describes current Visual Studio integration for .snupkg (Option 3). Visual Studio downloads from existing NuGet symbol server Symbols and map them to actual runtime.

### Issues
[...]

PR #10899, opened 2021-05-27
Proposal of snupkg restore functionality and debug improvement

BlackGad:
Design for https://github.com/NuGet/Home/issues/10793

### DIFF
--- a/proposed/2021/NuspecRestore.md
+++ b/proposed/2021/NuspecRestore.md
@@ -25,49 +25,90 @@ For proper debugging you need to provide [debug symbols](https://en.wikipedia.or
 
 With general `nupkg` and symbols `snupkg` packages end user is able to fulfil debugger with all required information.
 
-Detailed instruction for the framework owner:
-- Place general `nupkg` and symbols `snupkg` packages into discovarable feed ([local](https://docs.microsoft.com/en-us/nuget/hosting-packages/local-feeds) or any remote one). Note: `nupkg` can be located in one feed and `snupkg` in another but their id and version must be identical.
-
-Detailed instruction for the end user:
+#### Detailed instruction for the end user:
+- Place general `nupkg` and symbols `snupkg` packages into discovarable feed ([local](https://docs.microsoft.com/en-us/nuget/hosting-packages/local-feeds) or any remote one). **Note**: `nupkg` can be located in one feed and `snupkg` in another but their id and version must be identical.
 - Install ([nuget.exe install -Symbols](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-install), [dotnet add package --symbols](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-package?tabs=netcore2x)) or restore ([nuget.exe restore -Symbols](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-restore), [dotnet restore --symbols](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore)) NuGet package.
-- Make sure your debugger consumed all required data. 
-    Generic approach for Visual Studio is to use symbols from only one of [three places](https://docs.microsoft.com/en-us/visualstudio/debugger/specify-symbol-dot-pdb-and-source-files-in-the-visual-studio-debugger?view=vs-2019#symbol-file-locations-and-loading-behavior):
-    - The location that is specified inside the DLL or the executable (.exe) file.
-    - The same folder as the DLL or .exe file.
-    - Debugger options for symbol files.
+- Make sure your debugger consumed all required data.
 
-<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
-<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+Now end user can debug 3th party code line by line and finally close tone of its own issues :)
 
 ### Technical explanation
 
-<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+#### Things to add
+Options that will **force** to restore `snupkg` as well as `nupkg`.
+1. New `-Symbols` option to [nuget.exe restore](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-restore) command that will force to restore all available symbols for all installed packages.
+2. New `-Symbols` option to [nuget.exe install](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-install) command will force to restore specific package symbols.
+3. New `--symbols` option to [dotnet restore](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore) command that will force to restore all available symbols for all installed packages.
+4. New `--symbols` option to [dotnet add package](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-package?tabs=netcore2x) command will force to restore specific package symbols.
+
+**Optionally**: For proper integration into real project - symbols restore configuration can be placed in new file (similar to *nuget.config* for example *nuget.config.user*) on solution level. It is crucial to have ability to exclude it from subversion.
+    
+Inside Visual Studio in the **NuGet Package Manager** on **Installed** tab for every package add option for **symbols** install as well (configuration placed in *nuget.config.user*). Built-in **Restore** mechanism will respect this option in same way as common package miss situation.
+
+#### As a result
+[nuget.exe restore](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-restore) and [dotnet restore](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore) commands will do their job with symbols restore option only for configured packages.
+
+Based on source project version above commands will:
+1. For any project structure will download and place **package** and **symbols** content into the cache folders (if cache allowed). Default place for package content is `c:\Users\<user>\.nuget\packages\`. Default place for symbols content may be `c:\Users\<user>\.nuget\symbols\`.
+2. In old project structure will merge **package** and **symbols** content into destination **packages** folder (in reverse to split pack operation).
+3. In new project structure will do same the operation if **packages** folder set in `nuget.config` file.
+
+Further symbols transport will be handled by **MSBuild**. In old project structure symbols will be copied as assembly satellites to output folder and in new project structure this action will be handled by `PackageReference` (copy satellites from external **packages** folder or gather assemblies and symbols from the cache).
+
+Also it is important to implement symbols discovery operation in registries independent from package registries. This will allow to place `snupkg` files in local NuGet folder (local file based registry). 
+
+**Hint**: Generic approach for **Visual Studio** is to use symbols from only one of [three places](https://docs.microsoft.com/en-us/visualstudio/debugger/specify-symbol-dot-pdb-and-source-files-in-the-visual-studio-debugger?view=vs-2019#symbol-file-locations-and-loading-behavior):
+- The location that is specified inside the DLL or the executable (.exe) file.
+- The same folder as the DLL or .exe file. (**We want to reuse this ability**)
+- Debugger options for symbol files. (Implemented by official NuGet.org after `snupkg` file upload)
+
+Because `snupkg` technically is the same `nupkg` but with limited content and limited metadata all internal code for the restoration and security can be reused. 
 
 ## Drawbacks
 
-<!-- Why should we not do this? -->
+From my point of view this feature is missed part from the very beginning. Split package into assemblies and symbols and merge them back is atomic functionality. Split was implemented merge was not.
 
 ## Rationale and alternatives
 
-This is basically first stop on the long road to resolving target issue https://github.com/NuGet/Home/issues/6579.
+### Pack symbols into generic package
 
-<!-- Why is this the best design compared to other designs? -->
-<!-- What other designs have been considered and why weren't they chosen? -->
-<!-- What is the impact of not doing this? -->
+For older projects you can pack `pdb` files into generic package. As a result your package size will be huge.
+
+### Symbols server
+
+Public feeds can provide [symbols server](https://docs.microsoft.com/en-us/windows/win32/dxtecharts/debugging-with-symbols) based on `snupkg` package on upload via well known entry
+```json
+{
+  "version": "3.0.0",
+  "resources": [
+   {
+      "@id": "https://www.nuget.org/api/v2/symbolpackage",
+      "@type": "SymbolPackagePublish/4.9.0",
+      "comment": "The gallery symbol publish endpoint."
+    }
+  ]
+}
+```
+
+It is require additional feeds backend implementation.
+
+### Manual
+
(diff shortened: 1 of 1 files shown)


JonDouglas:
Thank you so much for the contribution @BlackGad! We will keep this proposal open for the next couple of weeks to gather upvotes and feedback from the community. We'll then review it as a team and get back to you shortly. Thanks again in the meantime!

If you're seeing this message, please 👍 or provide your feedback on this proposal in this PR as to why we should or shouldn't do this.

Thank you everyone!

BlackGad:
Is it reasonable to advertise this PR for community? Maybe you can invite some via the twitter?

JonDouglas:
@BlackGad I literally just tweeted a minute before your comment :) https://twitter.com/_JonDouglas/status/1399788019851407374

BlackGad:
Well :) Is it time to continue or no?

JonDouglas:
@BlackGad I have scheduled a review for this week & we will update this post shortly! Sorry for the delay as we're slightly behind this time of year.

BlackGad:
> @BlackGad I have scheduled a review for this week & we will update this post shortly! Sorry for the delay as we're slightly behind this time of year.

Nice to hear this. If you need any explanations or details fill free to ask in any form

BlackGad:
Did you manage to do a review?

BlackGad:
@JonDouglas congratulations once again! :) When you will be able to review this proposal?

JonDouglas:
@loic-sharma will you be able to help drive this forward and share thoughts about the review with @BlackGad while I'm out?

BlackGad:
@JonDouglas , @loic-sharma guys?

BlackGad:
@JonDouglas traditional Monday ping.

loic-sharma:
Hello, apologies for the late response. Thank you for creating this proposal! I definitely agree that our .NET debugging story is incomplete and needs work 😅

I'm not sure I fully understand your scenario just yet. Could you explain why you need to manually fetch your symbols instead of letting the debugger get them for you? Is it for improved exceptions at runtime?

I'm also curious as to the benefits of adding "symbol gathering" to NuGet restore. If we provided a separate standalone tool, would that work for you?

BlackGad:
Details described in [issue](https://github.com/NuGet/Home/issues/10793) but there is a brief overview

# Debugging
1) Debugging is not easy thing :)
2) Currently to manage debugging - build tools generates additional symbols near the generic output for future hints for the debug tool
3) Common symbols discovery mechanism for the MS libraries(applications) described [here](https://docs.microsoft.com/en-us/visualstudio/debugger/specify-symbol-dot-pdb-and-source-files-in-the-visual-studio-debugger?view=vs-2019#symbol-file-locations-and-loading-behavior)

# Abstract example
As far as I understand you are Web developer so will try to explain in this domain (not accurate example I must say). 
To present STATIC web page for the user developer must:

## Web example diagram

web_example

1) Create web page
    - Html with DOM structure
    - Appropriate css styling
2) Deliver to client side
    - Pack html to some kind of blob (1 on diagram)
    - Pack css to some kind of blob (2 on diagram)
    - Transfer to the client side
    - Unpack html blob (3 on diagram)
    - Unpack css blob (4 on diagram)
3) Display web page (5 on diagram)
    - Make sure client application understand how to discover html
    - Make sure client application understand how to discover css
    - Make sure client application understand how to link them and display

**Blobs** - byte array
**Client application** - some kind of web engine (Browser magic box)

## Generic application development

Now lets change labels on the diagram

dev_example

1) Create application
    - Application binaries
    - Debug symbols binaries
2) Deliver to client side
    - Pack application binaries to nupkg (1 on diagram)
    - Pack symbols to snupkg (2 on diagram)
    - Publish to NuGet feed
    - Install on the client side
    - Unpack application binaries (3 on diagram)
    - Unpack symbols (4 on diagram)
3) Run application in debug mode (5 on diagram)
    - Make sure runtime understand how to discover application binaries
    - Make sure runtime understand how to discover symbols binaries
    - Make sure debug runtime understand how to link them and display appropriate information

# So what NuGet currently can do:
1) NuGet package framework able to pack generic output (nupkg) and debug symbols (snupkg) into separated packages
2) NuGet framework able to publish generic output and the debug symbols to remote NuGet feeds
3) NuGet framework able to restore generic output packages to target system from the NuGet feed (remote AND local)

Basically it is artifacts split and publish mechanisms. Also there is only half of delivery mechanism implementation right now (nupkg only).

# What NuGet missing
1) Ability to restore snupkg in same way how it do with the generic ones
2) snupkg delivery mechanism for automatic artifacts merge (place symbols near generic output) to enable one of the options of generic symbols discovery mechanism. 

# Important
1) NuGet framework symbols restore procedure must be flexible and not attached to generic binaries source NuGet feed. (Then you can use local file system repository as a source for the symbols for binaries from remote NuGet feed)
2) Existing toolchains must be extended for symbols restore (separate tool will add additional complexity)
3) NuGet protocol API must be extended for symbols restore (all 3-th party application will have easy way to support standard restore symbols mechanism)
4) NuGet protocol must be end to end tool for pack, publish, delivery and unpack stages.

# Notes
It is a brief overview without technical details. This PR and initial issue have additional information about technical implementation, API definitions, problem areas so must refer you to them :)

This PR: https://github.com/NuGet/Home/pull/10899
Initial issue: https://github.com/NuGet/Home/issues/10793
PR text itself: https://github.com/BlackGad/Home/blob/dev/proposed/2021/NuspecRestore.md

# And the answer

> I'm not sure I fully understand your scenario just yet. Could you explain why you need to manually fetch your symbols instead of letting the debugger get them for you? Is it for improved exceptions at runtime?

Current debugging mechanism assumes that you MUST have remote NuGet feed which also hosting Symbol server. Local file system feeds are not supported. There is no easy way to fill symbols manually [explanation on real case](https://github.com/NuGet/Home/issues/10793#issuecomment-853708939) (We are the legion that requires this %))

> I'm also curious as to the benefits of adding "symbol gathering" to NuGet restore. If we provided a separate standalone tool, would that work for you?

One more spike?) snupkg is the same nupkg. All procedures that can be applied to nupkg can be applied to snupkg (discovery, security, install etc). Also NuGet already have API to create and publish snupkg so will be at least strange decision to use external tool for install and use.

BlackGad:
@loic-sharma 
Real world is cruel :) Often frameworks used in different from authors idea way.

Simplest solution - application have no additional component references so it uses required nuget packages directly.

Complex solutions requires additional packed modules and this modules require another internal modules so on. Real build often 1 hour CI pipeline build. End product bug means bug in some separate component and to debug and fix you need to have ability rapidly change and merge components locally.

I already said at the very beginning that this functionality is first step to manage above situation %)

loic-sharma:
This seems like an awesome feature to improve user's debugging/error experiences. I created a prototype for this proposal: https://github.com/loic-sharma/symbols

How this works:

1. Listen to the build for assemblies that are written to the output directory without corresponding symbols ([LINK](https://github.com/loic-sharma/symbols/blob/9cdee5246ba114301be5921d94c546cebc3ef49a/src/Cobalt.Symbols.MSBuild/build/Cobalt.Symbols.MSBuild.targets#L5-L10))
2. Pass the paths to these assemblies to a tool that download symbols ([LINK](https://github.com/loic-sharma/symbols/blob/9cdee5246ba114301be5921d94c546cebc3ef49a/src/Cobalt.Symbols.MSBuild/build/Cobalt.Symbols.MSBuild.targets#L12-L14))
3. For each assembly, try and download its corresponding symbols from nuget.org's symbol server ([LINK](https://github.com/loic-sharma/symbols/blob/9cdee5246ba114301be5921d94c546cebc3ef49a/src/Cobalt.Symbols.Downloader/Program.cs#L24-L36)) using the [symbol query protocol](https://github.com/dotnet/symstore/blob/main/docs/specs/Simple_Symbol_Query_Protocol.md).

My findings:

1. Instead of restoring `.snupkg`s, it is simpler to download individual symbols
2. This doesn't require any special NuGet integration. It can be done entirely by the .NET SDK

 Some open questions:

1. How do we configure what symbol servers to download symbols from?
2. Should this feature be opt-in? Or should it be on by default?

I'll ping some folks at Microsoft to get their thoughts on this idea.

BlackGad:
> Instead of restoring .snupkgs, it is simpler to download individual symbols

This will brake security for signed snupkg's :( https://github.com/NuGet/Home/issues/10793#issuecomment-838644605

> Should this feature be opt-in? Or should it be on by default?

Symbols restore option must be opt-in only. Symbols restore process must not impact nupkg restore option and also must be controllable.

Also you faced with one of issues that I collect earlier in source issue. https://github.com/NuGet/Home/issues/10793#issuecomment-826388332

I insist on same  behavior for snupkg as for the nupkg. Current internal protocol has implementation for web and local resources (feeds) discovery with solved transfer/auth/security features. If snupkg will behave in different way - you will be forced to solve same features separately. snupkg is the same nupkg with some content and metadata filters. Also there are already implemented  caching and restore mechanisms (for VS old and new project structures).

There is no easy road for this feature because different parts of it impact NuGet, VS, NuGet registry teams  :) 
I introduce a [plan (see v2 - 1)](https://github.com/NuGet/Home/issues/10793#issuecomment-827496500) that will impact ONLY NuGet team at the very beginning (stage 1). Then it will require other cases implementation.

loic-sharma:
> This will brake security for signed snupkg's :( #10793 (comment)

Why would it break security? This is how your debugger works today. The assembly contains a checksum for its corresponding symbols ([LINK](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#pdb-checksum-debug-directory-entry-type-19)). So once you download symbols, you can check it hasn't been tampered by recalculating its checksum.

FYI, nuget.org does not verify .snupkg's signatures today. We do, however, check that the uploaded symbols match assemblies previously uploaded to nuget.org.

> Symbols restore option must be opt-in only. Symbols restore process must not impact nupkg restore option and also must be controllable.

Yup, that's my thinking as well. 

> I insist on same behavior for snupkg as for the nupkg. Current internal protocol has implementation for web and local resources (feeds) discovery with solved transfer/auth/security features. If snupkg will behave in different way - you will be forced to solve same features separately. snupkg is the same nupkg with some content and metadata filters. Also there are already implemented caching and restore mechanisms (for VS old and new project structures).

I see where you're coming from. That said, symbols have been around for decades and changing protocol details would mean losing decades worth of tooling and integrations. This is not a decision we should make lightly.

The following problems are already solved for symbols:

* Transfer/restore - The Simple Symbol Query Protocol (SSQP) is to download symbols
* Security - The assembly contains a checksum for its corresponding symbols. In other words, you can trust a symbol if you already trust its corresponding assembly.
* Caching - Your machine has a symbol store cache.

That only leaves authentication. NuGet doesn't bring much to the table here as out of the box it only supports HTTP basic authentication and client certificate authentication. Hence, I don't think there's a strong case yet to move away from the existing symbol server protocol.

BlackGad:
Ok, I already mention earlier that it is first step for local debugging. If we add support for snupkg into the protocol - than we can manage main thing that I need for proper local debugging of composite applications. All features above cover more than I need, but purpose is organic for current implementation state of protocol.

Idea behind all of this is to call command line to end product for proper assemblies and symbols substitution. Because nupkg become not only assemblies blob. 

Can't find my initial case, it is possible that it was missed somewhere in other messages.

# Real case

Lets imagine we have multilayered product. Every layer it is some kind of nupkg. 
1. Package that contains basic logic - Base.dll
2. Package that contains some device driver specific logic (which uses Base.dll as well)
3. Package that contains web UI resources and msbuild scripts for proper output folder delivery
4. Final product that hosts web UI with katana, implements service, WPF ui etc.

Now client report a bug. After the research we understand that bug can be in Base.dll. To determine this we need restore symbols and debug in code issue line by line. Our nuget feed do not hosts symbol server. Ok we download symbols as an artifact and copy pdb files into proper location near the assemblies (file by file for every platform) then determine that bug there, so we need to fix it.

We clone repo for base.dll and trying to fix an issue. To test, that issue was fixed we compile new base nuget package. 

## Current behavior 

1. If we are lucky this is legacy project with v2 project structure. Why lucky? Because that protocol version can work with local packages folder - so we can substitute inside it content directly. Just unpacking nupkg and snupkg folder to currently used by product base nuget version. Manually or with very custom scripts, CLI utils.
2. If we are not lucky we go to the nuget global cache folder find for currently used version. Backup it and do the same thing from 1. Then restore cached content. But wait it is only nupkg content, because in new structure satellite files will not be copied automatically. So it must be trick inside product msbuild project or manual operation for all platforms.
3. Launch product and ensure bug fixed. 
4. Commit base repo code and wait for ci build. 
5. Update base NuGet package version and commit to product repo

Note: Package version will be available only after CI build. Base NuGet package content cannot be copied to output folder because of automatic build output files replacement. We are using 3.5, 4 and 5.0 net frameworks in same time.

Now imagine that bug was inside package with device driver logic because some issues inside base.dll. We will receive dependency tree. And now we need to work with 3 repos waterfall substitution.

# Wished behavior 
1. Run nuget.exe with limited configuration where the only feeds will be output folders of required nuget packages with no cache option and force reinstall
2. Launch product and ensure bug fixed. 
3. Commit base repo code and wait for ci build. 
4. Update base NuGet package version and commit to product repo

This is simplified developer flow. Ideally it would be 2 command lines. But, hope, it is a future. Not relevant to this purpose.

BlackGad:
And again. You insist that symbols restore handled by SSQP. What to do when you have no symbol server? 

Current builds already have assemblies and symbols in right place for debug. NuGet selects binaries and pack them to package, NuGet selects symbols and pack them to package as well, NuGet can install packed nupkg, but Nuget cannot install packed symbols! From my prospective it is the same situation like brief scheme above for web but at stage css unpack we tell browser - hey bro your css in another castle. Decide yourself how to fetch and link it yourself ;) and it is not my business that technically it similar to already delivered HTML

papeh:
My company manages shared dependencies for our desktop apps using NuGet. We like to ship symbols to our users so that automated error reports contain line numbers in crash stacks. This requires downloading symbols at build time. So we would find this feature useful, although most of our products have migrated to the new <PackageReference>.

We're using this hack when we collect the files to copy to our output directory:
```
<Target Name="CollectAssemblyAndPdbPaths">
	<ItemGroup>
		<NugetPacks Include="Package1"><Version>1.0</Version><Path>lib/net461/*.*</Path></NugetPacks>
	</ItemGroup>
	<DownloadFile Address="https://www.nuget.org/api/v2/symbolpackage/%(NugetPacks.Identity)/%(NugetPacks.Version)"
		DownloadsDir="$(SymbolsDir)" LocalFilename="%(NugetPacks.Identity).snupkg" />
	<Unzip ZipFilename="$(SymbolsDir)/%(NugetPacks.Identity).snupkg" ToDir="$(SymbolsDir)/%(NugetPacks.Identity).%(NugetPacks.Version)" />
	<ItemGroup>
		<NugetFiles Include="$(PackagesDir)/%(NugetPacks.Identity).%(NugetPacks.Version)/%(NugetPacks.Path)" />
		<NugetFiles Include="$(SymbolsDir)/%(NugetPacks.Identity).%(NugetPacks.Version)/%(NugetPacks.Path)" />
	</ItemGroup>
</Target>
```

donnie-msft:
Hi, we have removed our "proposals" folder, so please move this proposal to the "accepted" folder. 
Thanks!

PR #10967, opened 2021-06-24
Update the bulk restore coordination spec

nkolev92:
Per a bunch of recent discussion, updating the spec with all the relevant information from those meetings and discussions.

cc @lifengl @dtivel @aortiz-msft

### DIFF
--- a/proposed/2021/BulkRestoreCoordination.md
+++ b/proposed/2021/BulkRestoreCoordination.md
@@ -38,7 +38,8 @@ The primary goal is improving the branch switching scenario through Visual Studi
 There are no functional changes to any experiences.
 The user actions are all the same, the changes have a performance focus.
 
-The user *may* should be able to observed differences in any of the above mentioned scenarios, bulk edit, branch switch. 
+The user *may* be able to observe differences in any of the above mentioned scenarios, bulk edit, branch switch. 
+
 The time it takes for the user to be able to do work should be shorter than before.
 
 ### Technical explanation
--- a/proposed/2021/BulkRestoreCoordination.md
+++ b/proposed/2021/BulkRestoreCoordination.md
@@ -141,9 +141,9 @@ This will allow NuGet to remove the sliding window as NuGet can report progress 
 #### Implementation details
 
 - Given that there are risks with this proposal, the NuGet implementation will have a means of disabling this feature.
-- NuGet will only account for projects that have registered an IVsProjectRestoreInfoSource. A PackageReference based project does not have to register an IVsProjectRestoreInfoSource.
+- NuGet will only account for projects that have registered an IVsProjectRestoreInfoSource. A PackageReference based project does not have to register an IVsProjectRestoreInfoSource; however, not registering means that these projects will not participate in the bulk coordination and add a performance hit.
 - At solution load time, NuGet should be able to remove the sliding window, and rely on reporting information to the customer about potentially delayed restores. The frequency and the means of reporting can be determined at a later point.
-- In any non solution load scenario, NuGet can loop through all IVsProjectRestoreInfoSource objects, calling `HasPendingNomination`. When the first project reports `true` NuGet will call `WhenNominated`. There is no need for NuGet to call `WhenNominated` on all pending projects.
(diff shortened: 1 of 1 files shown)


nkolev92:
Just wanted to make sure I get a pair of eyes on this. 

cc @NuGet/nuget-client @dtivel @zivkan @aortiz-msft

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -34,6 +38,9 @@ These operations are currently not very efficient and may run extra restores and
 There are no functional changes to any experiences.
 The user actions are all the same, the changes have a performance focus.
 
+The user *may* should be able to observed differences in any of the above mentioned scenarios, bulk edit, branch switch.
```
zivkan:
```suggestion
The user *may* be able to observe differences in any of the above mentioned scenarios, bulk edit, branch switch. 

```

zivkan (approved):

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -127,11 +138,33 @@ The proposal is that each time NuGet determines a restore might need run, it che
     }
 ```
 
+#### Implementation details
+
+- Given that there are risks with this proposal, the NuGet implementation will have a means of disabling this feature.
+- NuGet will only account for projects that have registered an IVsProjectRestoreInfoSource. A PackageReference based project does not have to register an IVsProjectRestoreInfoSource.
```
dtivel: A PackageReference-based project is not required to register an IVsProjectRestoreInfoSource; however, not registering means that the project is responsible for restoring itself.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -127,11 +138,33 @@ The proposal is that each time NuGet determines a restore might need run, it che
     }
 ```
 
+#### Implementation details
+
+- Given that there are risks with this proposal, the NuGet implementation will have a means of disabling this feature.
```
dtivel, line 143: To be clear, disabling this feature means returning to current (released) behavior?  That means the 10-second sliding window can't be removed; it must be disabled when this feature is enabled and enabled otherwise.

proposed/2021/BulkRestoreCoordination.md
```diff
@@ -127,11 +138,33 @@ The proposal is that each time NuGet determines a restore might need run, it che
     }
 ```
 
+#### Implementation details
+
+- Given that there are risks with this proposal, the NuGet implementation will have a means of disabling this feature.
+- NuGet will only account for projects that have registered an IVsProjectRestoreInfoSource. A PackageReference based project does not have to register an IVsProjectRestoreInfoSource.
+- At solution load time, NuGet should be able to remove the sliding window, and rely on reporting information to the customer about potentially delayed restores. The frequency and the means of reporting can be determined at a later point.
+- In any non solution load scenario, NuGet can loop through all IVsProjectRestoreInfoSource objects, calling `HasPendingNomination`. When the first project reports `true` NuGet will call `WhenNominated`. There is no need for NuGet to call `WhenNominated` on all pending projects.
```
dtivel: When the first project reports `true` NuGet will call `WhenNominated` _on that project and any other project_?

dtivel (approved):

proposed/2021/BulkRestoreCoordination.md
nkolev92: We'd go back to the previous approach that will likely have a longer sliding window at solution load, yes.

proposed/2021/BulkRestoreCoordination.md
nkolev92: I made it more explicit. NuGet will call `WhenNominated` on that project only.

aortiz-msft:
@nkolev92 - Sorry about the post-merge feedback, but could that HasPendingNomination from Project-System is extremely quick (version comparison) under "Rationale and alternatives"?

nkolev92:
Sure, adding it in https://github.com/NuGet/Home/pull/11008.

PR #10992, opened 2021-07-01
New issue templates

zkat:
This adds templates for:

* [Bug Reports](https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/.github/ISSUE_TEMPLATE/BUG.yml)
* [Feature Requests](https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/.github/ISSUE_TEMPLATE/FEATURE.yml)
* [DCRs](https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/.github/ISSUE_TEMPLATE/DCR.yml)

(click links to preview)

Also, it adds a [`config.yml`](https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/.github/ISSUE_TEMPLATE/config.yml) that adds an option to open a NuGet Gallery issue, redirecting folks to the Gallery repo.

Finally, this adds a [`SECURITY.md`](https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/SECURITY.md), taken from https://github.com/dotnet/runtime/blob/main/SECURITY.md, which will show another option to "View Policy" for security issues.


zkat:
lol don't merge this yet, it's gonna 500

zkat:
alright, it works now. You can preview the form here: https://github.com/NuGet/Home/blob/dev-zkat-new-issue-template/.github/ISSUE_TEMPLATE/BUG.yml /cc @nkolev92 @zivkan

zkat:
https://github.com/zkat/test/issues/1 Here's what a resulting issue looks like.

.github/ISSUE_TEMPLATE/BUG.yml
```diff
@@ -0,0 +1,89 @@
...
+      label: Worked Before?
+      description: Is this a regression? Did it work in a previous version?
+      options:
+        - label: This used to work.
+    validations:
+      required: false
+  - type: input
+    id: lastworking
+    attributes:
+      label: Which Version?
```
donnie-msft: Consider just having this in the "Worked Before?" section to reduce headings/bulk.

donnie-msft (dismissed):
I really like it, and didn't know that GH allowed dropdowns/etc now!

.github/ISSUE_TEMPLATE/BUG.yml
zkat: You can't omit labels, unfortunately. I tried to do this because yeah, that's a lot of headings :(

.github/ISSUE_TEMPLATE/BUG.yml
```diff
@@ -0,0 +1,89 @@
...
+      value: "# Details About the Problem"
+  - type: dropdown
+    id: product
+    attributes:
+      label: NuGet Product Used
+      multiple: true
+      description: Which of our NuGet clients does this apply to?
+      options:
+        - NuGet.exe
+        - Visual Studio Package Management UI
```
dominoFire, line 22: Please add PMC:

```suggestion
        - Visual Studio Package Management UI
        - Visual Studio Package Manager Console
```

zkat:
@dominoFire @donnie-msft I've added FEATURE and DCR templates, btw.

.github/ISSUE_TEMPLATE/QUESTION.yml
```diff
@@ -0,0 +1,48 @@
...
+description: "Just asking a question (we don't necessarily always answer these)."
+title: "[Question]: "
+labels: ["Resolution:Question", "Triage:Untriaged"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+       Thank you for taking the time to file this issue. In order to best understand what you're asking about and what your needs are, we just need a few details from you.
+       
+       Please note that the NuGet issue tracker is not a support forum, and it's possible your question might not be answered at all and automatically closed after some time. Please consider asking your question on Stack Overflow or the VS Developer Community.
```
dominoFire: Is NuGet/Home GitHub discussion encouraged? Should we mention it here? https://github.com/NuGet/Home/discussions

zkat: oh yeah. I forgot we'd enabled that. I've added it to the config.yml and removed the QUESTION.yml template.

dominoFire (approved):

.github/ISSUE_TEMPLATE/BUG.yml
```diff
@@ -0,0 +1,85 @@
...
+       
+       The more detail you provide, the more likely it will be for us to be able to identify what is going on and how to solve it!
+  - type: dropdown
+    id: product
+    attributes:
+      label: NuGet Product Used
+      multiple: true
+      description: Which of our NuGet clients does this apply to?
+      options:
+        - NuGet.exe
```
nkolev92, line 21: What's the motivation behind the ordering?

nkolev92 (commented):
The love the amount of context this provides!
Hopefully this help with triaging overall. 

The one concern I have is the sheer verbosity of the template. 

We went away from the Repro/expected/actual in the template a while ago as an attempt to reduce verbosity and make the template less intimidating. 

I'm afraid this may have a similar effect :/

nkolev92:
Another thing that I noticed when I looked at https://github.com/zkat/test/issues/new?assignees=&labels=&template=BUGS.yml&title=%5BBug%5D%3A+ is the amount of vertical space this takes. 
It's too long imo.

.github/ISSUE_TEMPLATE/BUG.yml
zkat: there is no motivation behind the ordering.

nkolev92: Should we do alphabetical instead?

zkat:
tbh I think it's fine to make the template a bit more verbose if it's a form: We _do_ want this data from people, and it _should_ be required. I don't think issues where we have to constantly ask users for more context or details etc are actually that useful to us.

I think shortening the freeform template made more sense because what people would otherwise do was just ignore certain sections (because they could), or they would just get frustrated and wipe out the whole template.

Since we can actually enforce quality now, I think we _should_ expect customers to take a little more time to tell us about their issue. And if they get _very_ frustrated, they can always file a blank issue and avoid the template altogether, and we can decide what to do with the issue from there. But we do spend a lot of time asking customers for more details and waiting for answers that could, and should, have been given from the get-go.

.github/ISSUE_TEMPLATE/BUG.yml
zkat: sure. Pushed.

nkolev92:
I think a lot of the expected/actual assumes most bugs are appropriate for that pattern. I don't believe that would lead to better issues (we've had this exact template in the past)

If you look at https://github.com/nuget/home/issues?q=is%3Aopen+is%3Aissue+label%3AType%3ABug and click through the issues that are not created by internal customers you'll notice only a few are appropriate for that. 

Most of the time it's `actual: error with logs`, `expected: not getting an error`, which is not necessarily super helpful. 

The developer community template for Visual studio for example, only has an impact dropdown and an open ended section for expected/actual, repro.

zkat:
@nkolev92 so how about I:
1. Remove expected/actual sections
2. Add some text to the repro steps placeholder to imply that you should add expected/actual there ,but make it optional. 
3. Add an Impact dropdown

nkolev92:
I am fine with that. 

Please don't consider my comments as ultimatums or anything like that. 
I just want to start the conversation about it, as receiving feedback on the quality of our issue templates is something we haven't done much of :)

zkat:
we're working together, and your input makes things even better. I'll go do this stuff and merge, then :)

PR #11024, opened 2021-07-13
oops! this isn't a dcr

zkat:


nkolev92 (approved):

nkolev92:
@zkat maybe you can add nuget-client as codeowners of the .github folder https://github.com/NuGet/Home/blob/dev/.github/CODEOWNERS so that you don't have to manually request it again.

Issue #10865, opened 2021-05-13
Don't hit registration to gather deprecation/vulnerability metadata when on the browse tab


nkolev92:
Blocked on https://github.com/NuGet/NuGetGallery/issues/7297. 

A lot of the feedback centered around the PM UI has to do with how sluggish and heavy the UI looks. 
Some of the problems are technical decisions in the PM UI, some are UX driven, others are a little bit of both. 
 
Say you create a new project and you open the PM UI. 
0 remote calls happen because NuGet will default to the installed tab. 
Switch to the browse tab. 
You get 1 search call + N registration calls, where N is the number of loaded packages. 
The logic behind this registration call is fetching vulnerability and deprecation information. 
 
Registration pages can be just as large and as heavy to parse as the search ones. 
 
This "trick" is there due to https://github.com/NuGet/NuGetGallery/issues/7297. 
 
Given the repeated feedback about the clunkiness of the PM UI, I think we should ask the server team to prioritize this ask, and we should follow up on the equivalent client side.

Alternatively, we can hack it and *only* fetch metadata whne nuget.org is the source. This idea is bad. Current perf is worse.

nkolev92:
cc @donnie-msft @jebriede @aortiz-msft

nkolev92:
Triage: This change is blocked on a server feature. @JonDouglas to make sure it's on the unified backlog.

nkolev92:
Prototype in https://github.com/NuGet/Home/issues/11483. 

Already on the backlog, so need to keep it assigned to Jon.

loic-sharma:
FYI, a simpler solution may be to reconsider the registration API's paging strategy. Currently, the registration API splits package's information into groups of 64 package versions. This page size appears to be much too low. Making registration pages include more package versions would reduce the number network requests the client needs to do without needing to update the protocol or clients.

See: https://github.com/NuGet/Home/issues/8058#issuecomment-626465454

> I made [a tool](https://github.com/loic-sharma/nuget-mirror) to help estimate response sizes if we removed paging from the registration API altogether. The resulting registration indexes had the following size on disk after gzip compression:
> 
> * Average size: 1.15 KB
> * 90th percentile size: 1.82 KB
> * 99th percentile size: 8.60 KB
> * 99.9th percentile size: 25.1 KB
> * 99.99th percentile: 60.76 KB
> * Max size size: 223 KB
> 
[...]

PR #11033, opened 2021-07-15
Technical spec for the Vulnerabilities feature in the PMUI

### LINKED ISSUE #10987: Write tech spec for vulnerabilities MVP in PMUI

jebriede:
Fixes: https://github.com/nuget/home/issues/10987

Design spec PR by @chgill-MSFT: https://github.com/NuGet/Home/pull/11054


proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+
+The Details Pane will have a new section in the view called "Vulnerabilities" that will be visible when a package has one or more vulnerabilities. The Vulnerabilities section will display the highest severity of the vulnerabilities and show a list of vulnerabilities, their severity, and a link to the advisory. To stay consistent with the current design of the details pane, the vulnerabilities section will be created as its own UserControl that will be used by the DetailControl.
+
+The PackageManagerTopPanel will be modified to handle showing a warning icon when there are either deprecated or installed packages, or both. A new method will be added to this class to handle getting the count of the vulnerable packages installed which will be used in the tooltip and also to determine whether or not to show the warning icon for vulnerabilities.
```
drewgillies: Typo? I expect this should be deprecated or _vulnerable_...

drewgillies:
Looks good, @jebriede --checking on something: seeing the vulnerability information when clicking a package in the browse pane is to cover the fact that a registration blob needs to be loaded explicitly for this detail?

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+
+The PackageItemViewModel will have a new method to query the Vulnerability status of a given package asynchronously, much like it currently reads package deprecation status. This information will be used, along with the deprecation status, to determine if the warning icon should appear by the package in the package list. The tooltip for the icon will be updated to display information about the vulnerabilities and/or deprecated packages.
+
+The NuGetPackageSearchService will be updated to query the Vulnerability information for a given package so it can be used to by the PackageItemViewModel to know about vulnerabilities. Some code may need to be refactored here so the deprecation data and the vulnerability data is pulled in by the same query as package metadata and then separated into deprecation and vulnerability data by the caller rather than duplicating the call to the service for each chunk of data.
+
```
zivkan, line 51: I don't know if deprecation already handles this, but when the sources list is set to "all", this is a special case in PM UI, where when multiple sources contain the same package ID, we join them into a single entry. NuGet is built on the assumption that all packages are immutable and globally unique, so when this is violated and package description or package icon is different on different sources, it's effectively random which package metadata you'll actually get.

I really hope that package deprecation doesn't do this, but if it doesn't, I think it's really important for vulnerabilities to ensure that they're all listed, since security is so important.

In the case where multiple package sources list vulnerabilities, how will the list be merged?  If two sources list the same CVE/AdvisotyUrl, will the CVE be listed once or twice? If they list the same CVE, but for some reason have different severities, how will this be handled?

zivkan:
The title of this PR says "initial draft".  Does that mean it's still not going to be a completed spec after this PR is merged?

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: Don't forget to verify the screen-reader output of the new text.

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+The PackageManagerTopPanel will be modified to handle showing a warning icon when there are either deprecated or installed packages, or both. A new method will be added to this class to handle getting the count of the vulnerable packages installed which will be used in the tooltip and also to determine whether or not to show the warning icon for vulnerabilities.
+
+The PackageItemViewModel will have a new method to query the Vulnerability status of a given package asynchronously, much like it currently reads package deprecation status. This information will be used, along with the deprecation status, to determine if the warning icon should appear by the package in the package list. The tooltip for the icon will be updated to display information about the vulnerabilities and/or deprecated packages.
```
donnie-msft: > The tooltip for the icon will be updated to display information about the vulnerabilities and/or deprecated packages.

How will this look? Is there a sample from a PM or your own mockup? I'd be interested to see the layout and understand what we are deeming as worthy of tooltip text. Eg, is the max severity only shown? Are all severities listed? How big can this tooltip become in the average/worst cases?

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+## Goals
+
+* Show a warning icon in the Installed tab with a tooltip indicates how many vulnerable packages are installed.
+* Show a warning icon next to the package version of a vulnerable package in the package list. This icon will show for packages in the Browse and Installed tabs.
+* Show detailed information about the package vulnerabilities in the package details pane when the user clicks on a package.
+* Track via telemetry when an action (installed, update, uninstall) is performed on a package with a vulnerability from within the Package Manager UI.
+
+## Non-Goals
+
+* Improve the performance of querying package vulnerability information.
```
donnie-msft: I don't understand what this means given that we don't show vulnerability info today, right?

Secondly, we really shouldn't defer performance considerations given we already have a lot of debt in this space. Can you expand on what you mean here just so it's more clear what impact the new querying will have on perf?

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+* Make any improvements to the package deprecation experience.
+* Change the registration API to pull in more details about a package vulnerability. This will be considered in a future iteration.
+* Show a warning icon in the Solution Explorer when packages with vulnerabilities are installed. This will be considered in a future iteration.
+
+## Solution
+
+A list of vulnerabilities for a package is read from the server, along with other data, whenever a package version is selected (DetailControlModel.SetCurrentPackageAsync) or when the selection version changes (DetailControlModel.SelectedVersionChangedAsync). This data is stored in the DetailedPackageMetadata and is used by the Details Pane. The vulnerabilities information exposed in the PackageSolutionDetailControlModel which implements DetailedPackageMetadata is currently limited to whether a vulnerability exists and what the maximum severity of the vulnerability is. This will be changed to expose instead a list of vulnerabilities, their severity, and their advisory links.
```
donnie-msft: > DetailedPackageMetadata is currently limited to whether a vulnerability exists and what the maximum severity of the vulnerability is. 

I wasn't aware that Vulnerability info was part of this class at all. Is that true or can this be made more clear?

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+
+The Details Pane will have a new section in the view called "Vulnerabilities" that will be visible when a package has one or more vulnerabilities. The Vulnerabilities section will display the highest severity of the vulnerabilities and show a list of vulnerabilities, their severity, and a link to the advisory. To stay consistent with the current design of the details pane, the vulnerabilities section will be created as its own UserControl that will be used by the DetailControl.
```
donnie-msft: Since it would be relatively rare that this section be shown, a perf consideration would be to make sure it doesn't try to render/bind or fire propertyChanged notifications for these new controls unnecessarily. 

We do have some perf issues where package `SelectionChanged` causes so many bindings in the details pane, we overwhelm some screen readers (and likely contributes negatively to perf in general). Something to be mindful of.

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Requirements
+
+* A warning icon will be shown in the Installed tab and in the package list if the project has any vulnerable packages installed. These icons will be the same ones that are used when deprecated packages are installed. In the case where a package version is both vulnerable and deprecated, only one icon will still be shown, but the tooltip on the icon will reflect the vulnerability and deprecation.
+* Maintain the functionality of the deprecated warning icon in the package list as well as the installed tab. The icon will be used to indicate vulnerabilities, but will need to still indicate deprecated packages.
+* There can be one or more vulnerabilities for a package version so the details pane will need to be able to display a list of vulnerabilities, the severity levels, and the advisory links for each.
+
+## Goals
+
+* Show a warning icon in the Installed tab with a tooltip indicates how many vulnerable packages are installed.
```
donnie-msft: To be more clear that no new visual is being added:

```suggestion
* Show the existing warning icon (used for deprecation) in the Installed tab with a tooltip indicates how many vulnerable packages are installed.
```

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+* Show a warning icon next to the package version of a vulnerable package in the package list. This icon will show for packages in the Browse and Installed tabs.
```
donnie-msft:
```suggestion
* Show the existing warning icon (used for deprecation) next to the package version of a vulnerable package in the package list. This icon will show for packages in the Browse and Installed tabs.
```

donnie-msft: > This icon will show for packages in the Browse and Installed tabs.

Why not Updates/Consolidate? I believe the XAML is shared, so will this be explicitly coded as a condition?

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: While we don't show vulnerability info yet, we are already querying for vulnerability data in the PMUI today whenever the package is clicked or a version is changed in the details pane.

The non-goal of this MVP feature is to change the way the vulnerability information is obtained that may improve performance, which will require some changes on the server side as well as the client. That work is being tracked here: https://github.com/NuGet/Home/issues/10865

This doesn't imply performance considerations will be deferred for newly written code, it only calls out that it is not a goal of this feature to improve the existing querying of vulnerability information.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: Vulnerability data is part of DetailedPackageMetadata and it is exposed as a list of vulnerabilities. However, I meant to state that DetailControlModel is the one that is limited to whether a vulnerability exists and what the max severity of the vulnerability is. I'll fix the class name in my text. But yes, vulnerability data is present in both those classes already today.

jebriede: https://github.com/NuGet/NuGet.Client/blob/467c6eb5772e9506647986df8b675b6bedb5d02e/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailedPackageMetadata.cs#L115

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.UI/Models/DetailControlModel.cs#L411

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: I see, so it's already in the registration payload, and you're saying we won't be changing that yet.

I'd suggest the bullet be changed to something like:

```suggestion
* Modifying the registration query for vulnerability data will happen in a separate effort: https://github.com/NuGet/Home/issues/10865
```

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: I would imagine any issues with the bindings are more likely a problem with the property being bound to than the binding itself. For example, a binding should not be made to a Lazy property because it forces the evaluation of the Lazy to happen on the UI thread and I know we have a few cases where this is being done. If such issues are present, it would likely cause performance issues. I will be avoiding any such pitfalls. If you mean the bindings themselves of collapsed controls are causing performance issues, have we pulled a perf trace that shows this? Do you have that available?

Collapsed controls will not be picked up by screen readers. Can you expand on what you mean that bindings overwhelm screen readers?

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: @chgill-MSFT has a spec that he'll be adding to source control that will act as the source of truth for the look and design of the feature. If you take a look at the screenshot in the e-mail spec, it shows a good picture of what the tooltip will look like. I would rather not include that image in the technical spec as well, because if it changes in the design spec, we'll need to update it in two places. As soon as the design spec is up in the repo, I'll add a reference to it in this spec.

jebriede:
> The title of this PR says "initial draft". Does that mean it's still not going to be a completed spec after this PR is merged?

No, I intend this to be the spec once committed. I updated the title and removed "initial draft"

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: > Looks good, @jebriede --checking on something: seeing the vulnerability information when clicking a package in the browse pane is to cover the fact that a registration blob needs to be loaded explicitly for this detail?

Thanks Drew! It's partially because of where the data is queried from but also because the details pane is where more information is available about a package so the added details about the vulnerability will appear there as well.

jebriede: > > This icon will show for packages in the Browse and Installed tabs.
> 
> Why not Updates/Consolidate? I believe the XAML is shared, so will this be explicitly coded as a condition?

@donnie-msft: Good catch! It should show up there too. I'll remove this sentence altogether.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: @chgill-MSFT is vulnerability information available in any feed other than nuget.org?

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: > Collapsed controls will not be picked up by screen readers. Can you expand on what you mean that bindings overwhelm screen readers?

I don't see where the spec indicates they will be collapsed. Either way, a collapsed control like a TreeView, for example, still fires UIA events when the binding changes (eg, our Dependencies TreeView). This is to notify any clients of the new data in that control.  

Simply put: please just make sure controls aren't being unnecessarily made visible and rebinded. Preferably, make them non-visible (or whatever the control specifies) to prevent unnecessary rebinding.

> Can you expand on what you mean that bindings overwhelm screen readers?

[NVDA becomes sluggish with WPF TreeView UIAutomation Events #11109](https://github.com/nvaccess/nvda/issues/11109)

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: > is vulnerability information available in any feed other than nuget.org?

I don't think that's the point. The point is if 2 sources have *the same package*, we need to ensure the one with vulnerability flags (even if it's only nuget.org), is prioritized/loaded and not the one that doesn't contain vulnerability data at all.

chrisraygill: @jebriede can you check the code to see how/if this is handled for deprecation?

@joelverhagen do/can other feeds have their own vulnerability information? My guess is no since I believe the GitHub data is NuGet.org-specific.

@zivkan my understanding is that the PMUI will show the package details from whichever feed responds first. Is that correct? If so, do we continue to listening for a response from the other feeds after one has responded?

I'm wondering if we can show the vulnerability/deprecation information from NuGet.org even if it's not the first feed to respond.

joelverhagen: > do/can other feeds have their own vulnerability information? My guess is no since I believe the GitHub data is NuGet.org-specific.

It is technically possible. NuGet.org simply fills out some documented properties in JSON API responses with vulnerability information. Any V3 feed could do this, sourcing info however they choose. AFAIK there is no logic in NuGet client to limit vulnerability information (or deprecation, or prefix reservation, etc)  from coming from NuGet.org -- which is good.

zivkan: > my understanding is that the PMUI will show the package details from whichever feed responds first. Is that correct? If so, do we continue to listening for a response from the other feeds after one has responded?

I don't know. Someone who knows more can reply, otherwise someone working on the feature needs to investigate. I'm just saying I think this is a scenario that this spec should address.

> I'm wondering if we can show the vulnerability/deprecation information from NuGet.org even if it's not the first feed to respond.

With enough code changes, even if NuGet today is ignoring responses after the fastest server responds first, we can change it. Might be costly if the software design doesn't make this easy, in which case it depends on valuable it is to the PMs & division.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: Rest assured I will not be making the vulnerabilities section visible when there are no vulnerabilities to display as that would be quite undesirable. :-)

>I don't see where the spec indicates they will be collapsed.

The spec says the control will be visible when there are vulnerabilities, implying it will not be visible otherwise. In WPF when a control is not visible and you want it to not take up space, you make it collapsed. I feel adding that level of detail would be too much about how WPF works but if you think it isn't clear without that detail, I can add it to the spec.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: For the sake of the MVP I think it is out of scope to try and resolve an existing issue with how the "All" package sources code decides to take metadata from only one package source and ignore the rest. The scenario is a bit of an edge case but if displaying vulnerabilities when the All package source is selected in the PMUI could potentially not show metadata from nuget.org, we can make note of that in our release notes and fix that in the future. I'd like to get an opinion from @aortiz-msft on the matter.

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: > The spec says the control will be visible when there are vulnerabilities,

Ah I see the confusion. Visible isn't the same thing as collapsed. If you mean `Visibility.Collapsed`, you could put that in the spec, because others like myself may think you're talking about a collapsed state of a control, such as a TreeView or the like.

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: I've replied to your offline thread. 
I don't believe it's an edge case or out of scope to consider whether multiple sources could present a scenario where the "win" misses vulnerability data. An MVP for vulnerabilities should present consistent/reliable warnings to the user. Otherwise to me, it really takes the "viable" out of MVP.

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.PackageManagement.VisualStudio/PackageFeeds/MultiSourcePackageMetadataProvider.cs#L206

donnie-msft: Eyeballing the code, it seems possible that these secondary queries are all-or-none, so there may not be a problem. Something you can (should, IMO) do is write a test that shows how, given multiple sources with and without vulnerability data, the UI is always saturated with the vulnerability data.

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
+# Expose package vulnerability information in the Package Manager UI - Technical Spec
+
+* Status: **In Review**
+* Author(s): [Jean-Pierre Briedé](https://github.com/jebriede)
+* Issue: [10981](https://github.com/NuGet/Home/issues/10981) Expose package vulnerabilities information in the PMUI
+* Type: Feature
+
+## Problem Background
+
+Developers that install NuGet packages through Visual Studio may be installing packages that have known vulnerabilities or they may have a project that is depending on a vulnerable version of a NuGet package. There is currently no information in the Package Manager UI that indicates that a package version is vulnerable so developers may continue to use a vulnerable package in their project. Users should be able to see a warning about a package version they are browsing or one they have installed so they can be informed of the vulnerability and hopefully update to a different version of the package that is not vulnerable or use a different package.
```
kartheekp-ms, line 10: > Users should be able to see a warning about a package version they are browsing or one they have installed so they can be informed of the vulnerability.

Is this going to the case even for vulnerable transitive dependent packages also? Let's say I am searching for `abcd` package from a source. This package and direct dependencies don't have any known vulnerability, but a transitive dependency has a vulnerability.  Curious about what the user experience is going to be in this case.

Looking at this https://github.com/NuGet/Home/blob/dev/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md#display-vulnerabilities-in-visual-studio-pmui doc, I think it covers transitive dependencies also but not referenced in this spec. Hence not sure.

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Who are the customers
+
+* All users that install a NuGet packages using the Package Manager in Visual Studio.
+* All users that manage their installed NuGet packages using the Package Manager in Visual Studio.
+
+## Requirements
+
+* A warning icon will be shown in the Installed tab and in the package list if the project has any vulnerable packages installed. These icons will be the same ones that are used when deprecated packages are installed. In the case where a package version is both vulnerable and deprecated, only one icon will still be shown, but the tooltip on the icon will reflect the vulnerability and deprecation.
```
kartheekp-ms, line 19: Do we have any plan to display vulnerabilities differently based on the severity of it?

https://github.com/NuGet/Home/blob/dev/proposed/2020/PackageVulnerability/FlagVulnerablePackages.md#display-vulnerabilities-in-visual-studio-pmui doc says, `Installed tab shows a warning icon that leads to vulnerable package, when the package has High or Critical vulnerabilities`. Reading from this spec it looks like a warning icon will be displayed irrespective of severity of the vulnerability. It is possible that I may be missing some context here.

kartheekp-ms (commented):
Left few comments to start with..

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: @donnie-msft thanks for pointing out the code that deprecation uses to handle merging the deprecation status by always ensuring the non-null data is taken. I misunderstood some of Andy's comment to mean that we don't have a mechanism for merging the contents but I see now that's not what he was saying.

In the case where there are two sources that both provide vulnerability data (so they're both non-null), we'll want to take the one that exposes the highest severity of vulnerability and not just take the first vulnerability metadata that is not null like deprecation is doing. That would at least expose the vulnerabilities with the highest severity.

I'll add a line to the spec about merging vulnerability data for the All sources case.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: @kartheekp-ms The tooltip will reflect the maximum severity of the vulnerabilities on the package. The icon will be the same regardless of severity.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: Transitive dependencies is a separate feature and exposing vulnerabilities on them will be another iteration on vulnerabilities when transitive dependencies is built.

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,54 @@
...
+* There can be one or more vulnerabilities for a package version so the details pane will need to be able to display a list of vulnerabilities, the severity levels, and the advisory links for each.
+* When multiple package sources have the same package and All package sources is selected, the vulnerability data will need to be merged to ensure vulnerability information is shown for a package if any of the packages sources provides vulnerability information for the given package.
+
+## Goals
+
+* Show the existing warning icon, used for deprecation, in the Installed tab with a tooltip that indicates how many vulnerable packages are installed.
+* Show the existing warning icon, used for deprecation, next to the package version of a vulnerable package in the package list.
+* Show detailed information about the package vulnerabilities in the package details pane when the user clicks on a package.
+* Track via telemetry when an action (installed, update, uninstall) is performed on a package with a vulnerability from within the Package Manager UI.
```
donnie-msft, line 29: In the telemetry PR, you requested changes such that an array of severities is included along with a count of vulnerable packages per action. Can whatever goal prompted that change be made more explicit here?

proposed/2021/VulnerabilityInfoInPMUI.md
```diff
@@ -0,0 +1,54 @@
...
+
+The Details Pane will have a new section in the view called "Vulnerabilities" that will be visible when a package has one or more vulnerabilities, otherwise will have a visibility value of `Collapsed`. The Vulnerabilities section will display the highest severity of the vulnerabilities and show a list of vulnerabilities, their severity, and a link to the advisory. To stay consistent with the current design of the details pane, the vulnerabilities section will be created as its own UserControl that will be used by the DetailControl.
+
+The `PackageManagerTopPanel` will be modified to handle showing a warning icon when there are either deprecated or vulnerable packages, or both. A new method will be added to this class to handle getting the count of the vulnerable packages installed which will be used in the tooltip and also to determine whether or not to show the warning icon for vulnerabilities.
```
donnie-msft, line 46: The existing pattern is to pass in counts of packages or deprecation to an existing method. Are you sure a new method is needed in this view?

donnie-msft (commented):
A comment of mine was marked resolved about mentioning screen-readers, but I didn't see anything in the spec.

EDIT: readme spec was removed 👍

jebriede:
> A comment of mine was marked resolved about mentioning screen-readers, but I didn't see anything in the spec.
> 
> EDIT: readme spec was removed 👍

@donnie-msft: Can you please rephrase this in a way that is actionable? I'm not quite sure if there's anything pending from this comment or not and if so, what that is. Thanks!

donnie-msft:
> A comment of mine was marked resolved about mentioning screen-readers, but I didn't see anything in the spec.
> > @donnie-msft: Can you please rephrase this in a way that is actionable? 

@jebriede  The action would be adding to your spec how you expect screen-readers to announce Vulnerability information, and that you'll test that.

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: Yes. See: https://github.com/NuGet/NuGet.Client/blob/3ec538cb909cf5b9b0f663ef6f7a6891406d21aa/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerControl.xaml.cs#L936-L958

proposed/2021/VulnerabilityInfoInPMUI.md
jebriede: Let's chat offline about this.

proposed/2021/VulnerabilityInfoInPMUI.md
donnie-msft: @jebriede thanks for the link, but I think that reveals some confusion as your link points to a different file. The line in your spec actually states `PackageManagerTopPanel`, which was my question. So was that just a typo then?

jebriede: @donnie-msft No, it wasn't a typo. I linked the method added to the PackageManagerControl, but if you follow where that method is used, you can see where the counts are passed to the PackageManagerTopPanel.

jebriede: Looking over it in further detail, what I wrote is accurate. In my implementation, I created a new method for the PackageManagerTopPanel to handle getting (receiving) the count of vulnerable packages installed that is used in the tooltip and used to determine whether to show the warning icon for vulnerabilities. The level of specificity here describes what will get and use the vulnerability count to determine the tooltip and the whether the warning should appear.
https://github.com/NuGet/NuGet.Client/blob/3ec538cb909cf5b9b0f663ef6f7a6891406d21aa/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs#L110-L149

donnie-msft: [That method already existed](https://github.com/NuGet/NuGet.Client/blob/release-5.10.x/src/NuGet.Clients/NuGet.PackageManagement.UI/Xamls/PackageManagerTopPanel.xaml.cs#L110), which was what I thought you were adding a parameter into, but was confused by "new method". A consideration would be,, if you're going to write in specs very minute details, such as creating a new method, then make sure it's clear whether you mean refactoring a method, or actually introducing a whole new paradigm to the code.

donnie-msft (approved):
Going to go ahead and approve, considering some of this work is already merged and remaining comments are drying up.

PR #11047, opened 2021-07-19
Visual Studio Package Install Deep Link Proposal

chrisraygill:
Hi everyone! 

Here is my proposal for supporting Visual Studio NuGet Package Manager UI deep links. We have significant evidence that users vastly prefer installing NuGet packages via the PMUI in Visual Studio.

This proposal is meant enable users to install packages through the VS PMUI even when their decision to install occurs outside of VS, such as on NuGet.org, from a coworkers suggestions, or a documentation page.


nkolev92 (commented):
Looks pretty good. 

Another `edge case` is when multiple PM UIs are open, either multiple projects, or a project and solution level pm ui.

chrisraygill:
> Looks pretty good.
> 
> Another `edge case` is when multiple PM UIs are open, either multiple projects, or a project and solution level pm ui.

@nkolev92 - For consistency sake, this is a good argument for opening the solution level PMUI no matter what. That being said, do you know if we can tell which PMUI window is "in focus"? Does such a concept exist in VS?

nkolev92:
I think there are ways to figure that out.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+
+We will use telemetry the measure % of VS package installs that come from a deep link and surveys to understand satisfaction with available installation options for NuGet packages.
+
+## Scope
+
+* Generate a deep link URL that opens the NuGet package in the Visual Studio PMUI
+* Handler is just for Visual Studio
+* Only NuGet.org packages are supported
+
+## Scenarios
```
timheuer, line 37: 0. Visual Studio isn't installed
I recommend spec'ing this out to be clear.  I would assume (see later comments of mine) that you'd probably NOT show the option if VS isn't installed, but call this scenario out.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+* Can specify version
+* No version specified defaults to latest version available
+* Structure: TBD
+
+* Example from VS Code extension deep link: [vscode:extension/dracula-theme.theme-dracula](vscode:extension/dracula-theme.theme-dracula)
+
+### Scenario 1 - Visual Studio isn’t open
+
+1. Deep link URL is clicked.
+2. Open Visual Studio to solution selector window and wait for solution selection
```
timheuer, line 59: I wonder how frequent the solution selector window is used/known to most VS users...this *may* be a jarring experience.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+- GitHub Issue: https://github.com/NuGet/Home/issues/11032
+- GitHub PR: N/A
+
+## Summary
+
+Support deep links for NuGet.org packages that, when clicked, open the package in the Visual Studio NuGet Package Manager UI (PMUI) for a 1 click install experience.
+
+Applications:
+
+* Add “Install with Visual Studio” link to package README, documentation, or tutorials
```
timheuer, line 14: UX nit: The suggestion here is 'Install with Visual Studio' however looking at the scenarios the truth is 'launch PMUI with pre-selected search/details' and the user still has to click 'install' at some point.  Just making sure we have truth in advertising.  Consider testing with "Open in Visual Studio"?  You provide an example of VS Code/Dracula -- I can't recall if they use similar language, but the pattern is similar...the vscode:// protocol opens the detail page and still requires the user to click Install -- so follow similar language there whatever the original entry point is for that deep link.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+* Option 1: Open a dialogue explaining to the customer “You are attempting to install a package from NuGet.org but you do not have it enabled as a source. Would you like to enable it or cancel?”
+* Option 2: Switch feed setting to “All” automatically
+
+**Q: What package types should support downloading in Visual Studio (library, template, etc.)?**
+
+Currently, anything but library packages fail to install through the PMUI and don’t make sense to install through the PMUI. 
+
+**Q: What Visual Studio operation can we use as prior art?**
+
+•	Open repo in Visual Studio on GitHub.
```
timheuer, line 142: This was an experiment for Visual Studio Codespaces...I think this actually did have some detection of VS installed -- should check with the team there and if any header information was used, etc. That would be good to 'light up' the option knowing you have some rate of success versus offering an option where someone may not have VS (or the supported version of VS) installed and send them on a wild chase.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Motivation 
+
+Today, customers must copy a CLI command to install a NuGet package to their project from NuGet.org, or search for the package in Visual Studio. We should enable customers to directly install the package from the Visual Studio package manager UI so they can continue to use their preferred package management tool.
+
+From our NuGet.org survey, 70% of our respondents reported using the Visual Studio NuGet UI to install packages. Additionally, 87% of our respondents use Visual Studio for developing .NET apps. 
+
+Since users prefer installing packages via the VS package manager UI by a large margin relative to CLI methods, we should support that as a first-class experience.
+
+## Success metric
```
timheuer, line 27: Please also think about 'health' -- it isn't just the % of installs coming from the deep link, but we should understand the health...% of actual success from NuGet.org to full install, but also what is the bail-out rate...that would be interesting -- those who clicked install from the deep link, but never actually completed the task.

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+
+We will use telemetry the measure % of VS package installs that come from a deep link and surveys to understand satisfaction with available installation options for NuGet packages.
+
+## Scope
+
+* Generate a deep link URL that opens the NuGet package in the Visual Studio PMUI
+* Handler is just for Visual Studio
```
timheuer, line 34: VSMac as well?

timheuer (changes requested):
Some things to think about.

proposed/2021/VisualStudioPackageInstallDeepLink.md
chrisraygill: Don't customers see it every time they launch VS to get to their solution? At least I don't typically click on the .sln file directly since it's dug deep in some folder somewhere. I usually launch VS from my taskbar and choose the solution from the selection menu.

proposed/2021/VisualStudioPackageInstallDeepLink.md
chrisraygill: This is a good point!

proposed/2021/VisualStudioPackageInstallDeepLink.md
timheuer: Ahhhh. I thought you were referring to an older dialog.  If you are referring to the dialog that is this:
image
That makes sense. I've never heard anyone call this that but makes more sense now :-)

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+   ![Project level PMUI Browse tab filtered to target package](./meta/resources/VisualSutdioPackageInstallDeepLink/Project%20level%20filtered%20package%20results.png)
+5. User chooses relevant projects and installs the package as desired.
+
+## Drawbacks
+
+* Older versions of Visual Studio won't support this.
+* Current proposal is limited to NuGet.org packages - generalizing to packages from any feed will increase engineering complexity significantly as we would need a custom way to isolate the target package in the Browse tab.
+
+## Prior Art
+
```
dominoFire, line 112: I would add [Teams deep links](https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/build-and-test/deep-links) as prior art. Those use teams.microsoft.com site to generate links accordingly, either for web of desktop. 

https://docs.microsoft.com/en-us/microsoftteams/platform/concepts/build-and-test/deep-links

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+2.	Visual Studio is open without a solution selected
+3.	A solution is open, but the PMUI isn’t
+4.	Project level PMUI is open
+5.	Solution level PMUI is open
+
+## Design
+
+### Deep link design
+
+* Must specify package ID
```
dominoFire, line 49: Since this link will be made for opening VS, can we include extension Id and tool window ID so we can target any VS extension / window, not just PM UI ?

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+**Q: What happens if the package exists on multiple feeds?**
+
+We follow today’s normal installation behavior, though imperfect, and install the package from whichever feed responds. This is an inherent risk of package version duplication across multiple feeds and addressing the issue is not within the scope of this feature.
+
+We plan to address the dependency confusion issue with the upcoming [package namespaces feature](https://github.com/NuGet/Home/blob/dev/proposed/2021/PackageNamespaces.md).
+
+**Q: What happens if the package is already installed?**
+
+The user flow should proceed exactly as it would if the package were not already installed. This is consistent with today’s behavior where we show installed packages in the Browse tab.
+
```
dominoFire, line 159: How to pick which NuGet config file should be modified when adding a new source?

proposed/2021/VisualStudioPackageInstallDeepLink.md
```diff
@@ -0,0 +1,163 @@
...
+### Scenario 1 - Visual Studio isn’t open
+
+1. Deep link URL is clicked.
+2. Open Visual Studio to solution selector window and wait for solution selection
+   ![Visual Studio solution selection menu](./meta/resources/VisualSutdioPackageInstallDeepLink/Solution%20selection%20menu.png)
+3. When a solution is selected, open the solution level PMUI to the Browse tab.
+    1. Search query for [PackageId: “PackageName”] is prepopulated – yielding only the target package as a result.
+    2. Target package should be pre-selected with its package details open to the target version from NuGet.org.
+    ![Solution level PMUI Browse tab filtered to target package](.\meta/resources/VisualSutdioPackageInstallDeepLink/Filtered%20Package%20Search%20Results.jpg)
```
dominoFire, line 64: Image is broken

```suggestion
    ![Solution level PMUI Browse tab filtered to target package](./meta/resources/VisualSutdioPackageInstallDeepLink/Filtered%20Package%20Search%20Results.jpg)
```

proposed/2021/VisualStudioPackageInstallDeepLink.md
zivkan: Out of curiosity, how would a website (such as nuget.org, azure devops, artifactory, etc) detect whether or not VS is installed?  Does the browser have APIs to determine whether or not a specific protocol handler is installed?

kartheekp-ms: It looks `github.com` has that integration already. If you click on `Open with Visual Studio` on `NuGet.Client` repo for example it opens a `Microsoft Visual Studio Web Protocol Handler` and user must select the VS version, then the repo is cloned locally.

image

zivkan: The "choose which version of VS" looks like the standard Windows prompt, not github.  If you have multiple monitors, move your browser to the non-primary monitor, and see that the dialog appears on the primary screen.

I don't have a WIndows machine available right now to validate that the "Open with Visual Studio" is not displayed/hidden when VS is not installed, which is what the suggestion is.

jeffkl:

PR #11185, opened 2021-08-26
Remove Package Namespace modes for 17.0

nkolev92:
Fixes https://github.com/NuGet/Client.Engineering/issues/1106

### DIFF
--- a/proposed/2021/PackageNamespaces.md
+++ b/proposed/2021/PackageNamespaces.md
@@ -82,7 +82,8 @@ Next, add `<namespace>` elements under the `<packageSource>` with an `id` to spe
 </packageNamespaces>
 ```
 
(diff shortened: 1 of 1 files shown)


erdembayar (approved):

kartheekp-ms:
I have not reviewed the changes yet. Maybe we should add your response https://github.com/NuGet/NuGetGallery/pull/8754#discussion_r695967127 for `Would it be a good idea to use a "catch all" namespace (<namespace id="*" />) for the nuget package source? ` question in this spec.

nkolev92:
@kartheekp-ms 

Do you think that'd be something that's better suitable for the blog/actual feature docs?

aortiz-msft (approved):

clairernovotny (approved):

PR #11250, opened 2021-09-20
Chgill-update-namespaces-design-doc

chrisraygill:
Fixes: https://github.com/NuGet/Client.Engineering/issues/1161

Update the Package Namespaces design doc with the new Package Source Mapping syntax and preview 4 rules.


nkolev92:
Given that there was filename changes involved, it was hard to actually look at the actual changes. 

So I created a bbranch that kept the changes in the same doc to make it easier for review https://github.com/NuGet/Home/tree/dev-nkolev92-cgill-namespaces-doc. 

@chgill-MSFT You could consider squashing the commits into 2 to make it easier to review ;)

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+
+ compatible with the following tools:
+
+* [Visual Studio 2022 preview 4](https://visualstudio.microsoft.com/vs/preview/) and later
+* [nuget.exe 6.0.0-preview.4](https://www.nuget.org/downloads) and later
+* [.NET SDK 6.0.100-rc.1](https://devblogs.microsoft.com/nuget/introducing-package-source-mapping/#package-source-mapping-rules) and later
+
+Older tooling will ignore the Package Source Mapping configuration. To use this feature, ensure all your build environments use compatible tooling versions.
+
+Package Source Mappings will apply to all project types – including .NET Framework – as long as compatible tooling is used for build and restore.
```
nkolev92: nit: `including .NET Framework` seems unnecessary. Is there a guideline?

I usually try not to associate tooling with the framework explicitly, even if customers do it themselves.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
+# Package Source Mapping
+
+- [Jon Douglas](https://github.com/JonDouglas) & [Nikolche Kolev](https://github.com/nkolev92) & [Chris Gill](https://github.com/chgill-MSFT)
+- Start Date: (2021-02-19)
+- GitHub Issue ([6867](https://github.com/NuGet/Home/issues/6867))
+
+## Status
```
nkolev92, line 7: Love this update :)

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+* [.NET SDK 6.0.100-rc.1](https://devblogs.microsoft.com/nuget/introducing-package-source-mapping/#package-source-mapping-rules) and later
+
+Older tooling will ignore the Package Source Mapping configuration. To use this feature, ensure all your build environments use compatible tooling versions.
+
+Package Source Mappings will apply to all project types – including .NET Framework – as long as compatible tooling is used for build and restore.
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+NuGet allows access to all package IDs on a package source. However, many .NET developers have needs to filter & consume a subset of the package IDs available on public, private, and local sources. Although NuGet supports a concept of [trusted-signers](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers) to trust packages signed by specific authors, users would like to optionally specify which package IDs shoulder be allowed from which sources.
```
nkolev92:
```suggestion
NuGet allows access to all package IDs on a package source. However, many .NET developers have needs to filter & consume a subset of the package IDs available on public, private, and local sources. Although NuGet supports a concept of [trusted-signers](https://docs.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-trusted-signers) to trust packages signed by specific authors, users would like to optionally specify which package IDs should be allowed from which sources.
```

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+  <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  <add key="contoso" value="https://contoso.com/packages/" />
+</packageSources>
+```
+
+**Definition:**
+
+Add a new `<packageSourceMapping>` element within the `NuGet.config` using the following syntax:
+
+```xml
```
nkolev92, line 71: Seems like there's some duplication here with the global packages folder.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+        <package pattern="NuGet.*" />
+    </packageSource>
+ 
+</packageSourceMapping>
+```
+
+Repeat this process for all sources.
+When using package source mapping, every source should have a list of allowed package patterns.
+A user may also pin a specific package id instead of broader package ID prefix.
+The most specific pattern that matches a package ID will have the highest precedence. Exact package IDs will always have the highest precedence over any prefix.
```
nkolev92: Maybe change: Exact package IDs will always have the highest precedence over any prefix.
 
to 

Package id patterns will always have the highest precedence over any prefix.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+When using a combination of public, private, and local sources defined in `NuGet.config` file(s), a user can add a new `<packageSourceMapping>` element to opt-in to the feature. Lastly by adding individual `<package pattern="">` elements in the `<packageSource>` node, the source will only allow the matching package IDs from the respective package source.
+
+**Prerequisites**
+
+1. Create a repo-level `nuget.config` file by executing `dotnet new nugetconfig` at the repo root.
+2. Define a [repo-specific global packages folder](https://docs.microsoft.com/nuget/reference/nuget-config-file#config-section) in your `nuget.config`.
```
nkolev92: This is probably merely a recommendation, rather than a requirement. 

I think calling out the global packages folder that frequently is unnecessary.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+When using a combination of public, private, and local sources defined in `NuGet.config` file(s), a user can add a new `<packageSourceMapping>` element to opt-in to the feature. Lastly by adding individual `<package pattern="">` elements in the `<packageSource>` node, the source will only allow the matching package IDs from the respective package source.
+
+**Prerequisites**
+
+1. Create a repo-level `nuget.config` file by executing `dotnet new nugetconfig` at the repo root.
```
nkolev92: Have or create one is probably the better language here.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+When using a combination of public, private, and local sources defined in `NuGet.config` file(s), a user can add a new `<packageSourceMapping>` element to opt-in to the feature. Lastly by adding individual `<package pattern="">` elements in the `<packageSource>` node, the source will only allow the matching package IDs from the respective package source.
+
+**Prerequisites**
```
nkolev92, line 47: Sounds like most of these are recommendations rather than prereqs.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+The global packages folder is an *append only* resource. This means NuGet only ever installs a new package. There is **no** refreshing, or overriding, or `re-installing` packages of any kind.
+- When requesting a particular version (not floating), local, file sources are currently preferred - When installing a package, NuGet checks local sources independently before checking http sources.
+- Package installation is operation based - If 3 projects are being restored during that operation, and all those projects have a dependency to `Newtonsoft.Json`, version `9.0.1`, in regular scenarios, only 1 project will *actually* download and install the package. The other projects will use the already installed version.
+- `.nupkg.metadata` - Each package installation directory contains a `.nupkg.metadata` file which signify that a package installation is complete. This is expected to be the last file written in the package directory. This file is *not* use during build. NuGet.Client writes the package source information inside this file.
+
+#### Package Source Mapping rules
```
nkolev92, line 171: They were deleted, but we shoiuld keep the package installation rules. 

The doc contains both feature and technical aspects.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+
+1. Two types of package ID patterns are supported:
```
nkolev92:
```suggestion
1. Two types of package patterns are supported:
```

nkolev92: I see this matches the blog. 

I have some feedback on the particular parts, but I think we can do that in the docs instead of the spec. 
Up to you. Happy to provide the feedback.

implemented/2021/PackageSourceMapping.md
```diff
@@ -0,0 +1,571 @@
...
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+- Different modes or strategies can be considered in future iterations of the feature.
+- The current proposal focuses on the nuget.config experience only. Adding a capability to source pin through RestoreSources is a future task.
+- NuGet can allow users to filter their Package Source Mapping per source within CLI & IDE experiences.
+- NuGet can allow a user to add a full or glob package ID namespace at install time with an additional click/parameter in Visual Studio or CLI.
+- NuGet can combine this feature with `Package Lock Files` to allow a user to ensure the lock file is generated under the allowlist of package ID patterns.
+- Create a better experience for this feature interacting with the global packages folder so that packages in the GPF are validated against the source mapping configuration.
+- Enable users to validate the outcome of their configuration with a CLI command, maybe apart of `dotnet list package`
```
nkolev92:
```suggestion
- Enable users to validate the outcome of their configuration with a CLI command, maybe a part of `dotnet list package`
```

implemented/2021/PackageSourceMapping.md
chrisraygill: Removed the specific callout here as unnecessary🙂

implemented/2021/PackageSourceMapping.md
chrisraygill: Agree 👍

implemented/2021/PackageSourceMapping.md
chrisraygill: Agree 👍

implemented/2021/PackageSourceMapping.md
chrisraygill:
```suggestion
The most specific pattern that matches a package ID will have the highest precedence. Patterns with exact IDs will always have the highest precedence over a prefix pattern.
```

implemented/2021/PackageSourceMapping.md
chrisraygill: AFAICT the rules I included kept all the relevant technical details from the previous rules. Here are the old ones:

> #### Package installation rules
> 
> - When the requested package is already installed in the global packages folder, no source look-up will happen. The namespaces are irrelevant.
> - Only sources with namespace metadata will be used.
> - When a source has no metadata, the default is *none*.
> - When a package id needs to be downloaded, it is first compared against sources with namespaces. If any match, only those sources are used. If none of the namespaces match, then the sources without namespaces are considered, if any.
> - A namespace with a specific package id is *always* preferred over a namespace with a prefix.
> - When multiple different namespaces match the package id, the most specific one will be considered.

Anything you see missing?

implemented/2021/PackageSourceMapping.md
nkolev92: That's even better. 

Here's how I attempted to solve the problem in the feature docs: 

https://review.docs.microsoft.com/en-us/nuget/consume-packages/package-source-mapping?branch=dev-nkolev92-addPackageSourceMappingDocumentation

Figured a visual solution it's easier than any text :D

nkolev92 (approved):
Me & @chgill-MSFT synced offline. 

Thanks for handling this @chgill-MSFT !

implemented/2021/PackageSourceMapping.md
nkolev92: My focus was primarily on how the global packages folder is used as the source of truth. 
Should be better now.

PR #11274, opened 2021-09-27
add new internal issue template and disable blank issues

zkat:
Template preview: https://github.com/NuGet/Home/blob/zkat/internal-template/.github/ISSUE_TEMPLATE/INTERNAL.yml

### DIFF
--- a/.github/ISSUE_TEMPLATE/INTERNAL.yml
+++ b/.github/ISSUE_TEMPLATE/INTERNAL.yml
@@ -1,7 +1,5 @@
 name: Team Internal
 description: Issue template used for tasks created by the NuGet team for themselves. Internal Use Only.
-title: ""
-labels: []
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/INTERNAL.yml
+++ b/.github/ISSUE_TEMPLATE/INTERNAL.yml
@@ -1,0 +1,41 @@
+name: Team Internal
+description: Issue template used for tasks created by the NuGet team for themselves. Internal Use Only.
+title: ""
+labels: []
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ## Preflight Checklist
+
+        **Do not file this issue until you've gone through this checklist.**
+
+        ### Basic Labels (required)
+
+        * Add **Regression** labels if this is a regression.
+        * Add a **Type:** label
+        * Add **Area:** and/or **Functionality:** labels
+        * Add a **Tenet:** if applicable
+
+        ### Triage Labels (required)
+
+        * Add a **Priority:** label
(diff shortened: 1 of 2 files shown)


nkolev92 (commented):
Any way I can see a sample issue created through this template?

I'm not sure what problem we're solving here. 

We have a lot of duplication about labeling etc. 
I would also prefer *not* to have to do a click through to create an issue. I think this is an overkill right now and I think we should keep the blank issues.

zkat:
None of the template text will show up in the issue. This is basically equivalent to a standard issue, but with a block of reminder text preceding it.

Note that if you're clicking on `new issue` to create an issue, you still have to do the same number of click throughs: one.

.github/ISSUE_TEMPLATE/INTERNAL.yml
```diff
@@ -0,0 +1,39 @@
...
+      value: |
+        ## Preflight Checklist
+
+        **Do not file this issue until you've gone through this checklist.**
+
+        ### Basic Labels (required)
+
+        * Add **Regression** labels if this is a regression.
+        * Add a **Type:** label
+        * Add **Area:** and/or **Functionality:** labels
```
nkolev92, line 15: Product probably deserves a mention here. 

Area is usually less important than Functionality and Product.

.github/ISSUE_TEMPLATE/INTERNAL.yml
```diff
@@ -0,0 +1,39 @@
...
+    attributes:
+      value: |
+        ## Preflight Checklist
+
+        **Do not file this issue until you've gone through this checklist.**
+
+        ### Basic Labels (required)
+
+        * Add **Regression** labels if this is a regression.
+        * Add a **Type:** label
```
nkolev92, line 14: I think right now, it's either `Type` or `Category:Engineering`. 

I'd prefer to have had Type:Engineering, but I can't make that call :)

nkolev92 (commented):
Minor feedback on the template. 

My primary concern is velocity, so as long as that's not affected, I'm fine with whatever the team finds comfortable.

.github/ISSUE_TEMPLATE/INTERNAL.yml
nkolev92: We're having the `Type:Engineering` discussion in https://teams.microsoft.com/l/message/19:00d019d741b44e54baadd67ef7aeb41d@thread.skype/1634250158470?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=ad34ac55-6785-4e7c-82c3-7ba405d30c3c&parentMessageId=1634250158470&teamName=NuGet&channelName=Client&createdTime=1634250158470

zivkan:
Closing as stale. If someone else feels passionate about this, please re-open or create a new PR.

PR #11324, opened 2021-10-20
Fix image links

Moneypulation:
Mockup images seemed to be specified through the wrong path. Fixed the directory name


erdembayar (commented):
Thank you for your contribution.
This new feature was renamed to `packageSourceMapping` from `PackageNamespaces`.
So proper fix would be just move same images to `packageSourceMapping` path and leave md as it's now.

nkolev92 (approved):
Approving so we can merge.

nkolev92:
@erdembayar I think your feedback makes sense.

I'll create a follow up PR fixing it.

Issue #9043, opened 2020-01-10
Search query isn't properly URL encoded


loic-sharma:
## Details about Problem

The search query is not properly URL encoded:

1. This breaks NuGet.org's ["plus" operator](https://docs.microsoft.com/en-us/nuget/consume-packages/finding-and-choosing-packages#search-syntax) for queries like `foo+bar`
1. This allows a user to craft a search query that appends arbitrary query strings.

It looks like this line is responsible: https://github.com/NuGet/NuGet.Client/blob/74d1935e4e25a90df536ae7cac6cf7c73c6ea678/src/NuGet.Core/NuGet.Protocol/Resources/RawSearchResourceV3.cs#L48

NuGet product used: VS UI
VS version: 16.5.0 Preview 2.0 

## Detailed repro steps so we can see the same problem

1. Open Fiddler
1. Open the package manager UI
1. Select the "Browse" tab
1. Search for `foo&FOO=BAR`
1. Check the request on Fiddler. You should see: `GET https://azuresearch-usnc.nuget.org/query?q=foo&FOO=BAR&skip=0&take=26&prerelease=true&semVerLevel=2.0.0 HTTP/1.1`

![image](https://user-images.githubusercontent.com/737941/72115851-40976600-32fd-11ea-8266-eff0639a0909.png)

nkolev92:
@loic-sharma 
Can you help us understand the impact here?
When was the plus operator added? 
Do we know how frequently it's used in the search queries?

loic-sharma:
> Can you help us understand the impact here?

This is potentially a security issue. We should review this with security experts as they may want to increase the priority.

It's difficult for us to analyze the full customer impact as the requests themselves are malformed. For example, if a customer searches for `foo&X=Y` in Visual Studio's search bar, our telemetry will think that the user searched for just `foo`. That said, we do receive hundreds of searches each day that do not properly encode `+`:

![MicrosoftTeams-image](https://user-images.githubusercontent.com/737941/84538134-ffd5bb80-aca5-11ea-9c9a-45132c87d49b.png)

> When was the plus operator added?

A long, long time ago. Easily 5+ years ago :)

PR #11374, opened 2021-11-11
Search by Target Framework & Supported Platforms on NuGet.org

JonDouglas:
Introduction to a filtering concept on NuGet.org to allow filtering for target framework monikers and supported platforms of various generations of .NET.

[Rendered Spec](https://github.com/NuGet/Home/blob/c07fe38f029f5dd78c4ae264fc28e954b398f93b/proposed/2021/TargetFrameworkAndPlatformsFilters.md)

Thank You 🎉


proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a developer is browsing for a NuGet package to include in their project, they can filter packages based on generations of .NET including the target framework & supported platforms that package targets.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. Ontop of it being one of the biggest error codes we see, it is also the number #1 request on the `NuGetGallery` repository for good reason. It should be straight-forward for a developer to separate the generation of .NET they are currently targeting and find packages that are supported on the platforms that the respective generation supports.
```
loic-sharma: Consider linking to the NuGetGallery issue since that's an excellent motivation. This helps interested folks dig deeper into background

```suggestion
Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. Ontop of it being one of the biggest error codes we see, it is also [the number #1 request on the `NuGetGallery` repository](https://github.com/NuGet/NuGetGallery/issues/3098) for good reason. It should be straight-forward for a developer to separate the generation of .NET they are currently targeting and find packages that are supported on the platforms that the respective generation supports.
```

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+When browsing for packages on NuGet.org, a developer will be able to filter packages based on compatibility filters. These filters will include:
+
+1. The **generation of .NET** such as .NET Framework, .NET Core, .NET (5, 6, and later), and Any.
+2. The **supported platforms** for the generation such as .NET Standard, Windows, macOS, Linux, Android, iOS, and more.
+3. The **version number** of the generation being used such as .NET 5, .NET Core 3.1, .NET Framework 4.8, and more.
+
+This proposal only accounts for the **majority of packages on NuGet.org**. There are many unique TFMs in the .NET ecosystem, but for the sake of creating a concept that will scale with the future of .NET, many unsupported & low usage TFMs are not considered. Thus this proposal only focuses on the [most common target frameworks](https://docs.microsoft.com/dotnet/standard/frameworks#latest-versions) & omits considering TFMs with less than 5,000 packages.
```
loic-sharma:
```suggestion
This proposal only accounts for the **majority of packages on NuGet.org**. There are many unique TFMs in the .NET ecosystem, but for the sake of creating a concept that will scale with the future of .NET, many unsupported & low usage TFMs are not considered. Thus this proposal only focuses on the [most common target frameworks](https://docs.microsoft.com/dotnet/standard/frameworks#latest-versions) & omits considering TFMs with fewer than 5,000 packages.
```

Sorry for being annoying 😂

loic-sharma: Also consider clarifying if the 5,000 bar is on package IDs or package versions:

```suggestion
This proposal only accounts for the **majority of packages on NuGet.org**. There are many unique TFMs in the .NET ecosystem, but for the sake of creating a concept that will scale with the future of .NET, many unsupported & low usage TFMs are not considered. Thus this proposal only focuses on the [most common target frameworks](https://docs.microsoft.com/dotnet/standard/frameworks#latest-versions) & omits considering TFMs with fewer than 5,000 package IDs.
```

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+  - Version filter populates all known version for .NET Framework and .NET Standard.
+    - 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 2.0, 2.1, 2.2, 3.0, 3.1
+
+For further scenarios, see the [framework precedence](https://nugettools.azurewebsites.net/6.0.0-preview.3/framework-precedence) tool to understand how the version matrix would be calculated.
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/VersionNumbers.png)
+
+#### Complete Experience
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/CompleteExperience.png)
```
loic-sharma, line 123: The `Versions` section in the `Filter` pane may be confusing to customers: they may think this is referring to package versions instead of target framework versions

loic-sharma: Should we move the supported platform fields to the `Filter` pane? Benefits:

1. My gut feeling is that most users will filter on generations, but only a few users will filter on supported platforms and versions. Moving to the `Filter` pane reduces the number of controls visible by default
2. This may be better for accessibility. It's clear an `Apply` button will refresh the page to apply the filters. It's less clear that a `.NET Standard` button will refresh the page. I would ask @ryuyu if they have any accessibility concerns with the current controls.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+    - `net6.0-tvos`
+  - .NET Standard
+- Tizen:
+  - .NET 6+
+    - `net6.0-tizen`
+  - .NET Standard
+
+**Note:** .NET Standard is considered a special case that supports majority of these platforms depending on the [version of .NET standard](https://docs.microsoft.com/dotnet/standard/net-standard#net-implementation-support). If the framework precedence of a target framework supports .NET Standard, it's respective versions supported will be implicitly included.
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/SupportedPlatforms.png)
```
loic-sharma, line 90: We use slightly different UI for generations versus supported platforms:

1. You can select only a single generation but multiple support platforms
2. There's an "Any" option for generations, but none for supported platforms (I imagine that either all supported platforms would be selected by default, or, no filter is applied if no supported platform is selected)

Would it make sense to unify the UI for generations and supported frameworks to use the similar controls?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
loic-sharma: I imagine that the vast majority of users are confused by `.NET` vs `.NET Core` vs `.NET Framework` and don't understand target frameworks. Could we add tooltips or links to docs to help our customers understand all this jargon?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+```
+framework:net472 framework:net471 framework:net47   # Match the target framework monikers
+
+platform:windows platform:macos platform:linux    # Match the platforms
+
+implementation:netframework implementation:netcore implementation:net  # Match the generation of .NET including both supported frameworks & platforms.
+```
+
+- `framework:` - Allows a short-form TFM value to serach on.
```
loic-sharma:
```suggestion
- `framework:` - Allows a short-form TFM value to search on.
```

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+- `platform:` - Allows a platform value to search on.
+- `implementation` - Allows a generation of .NET to search on (includes all frameworks of the generaiton).
```
loic-sharma:
```suggestion
- `implementation:` - Allows a generation of .NET to search on (includes all frameworks of the generation).
```

loic-sharma: Are the `platform:` and `implementation:` filters necessary? Can we power the MVP using only the `framework:` filter?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+For further scenarios, see the [framework precedence](https://nugettools.azurewebsites.net/6.0.0-preview.3/framework-precedence) tool to understand how the version matrix would be calculated.
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/VersionNumbers.png)
+
+#### Complete Experience
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/CompleteExperience.png)
+
+## Search Syntax
```
loic-sharma: I would strongly recommend using query parameters instead of nuget.org's search syntax:

* Other search filters (semver level, prerelease, package type) are implemented using query parameters: https://docs.microsoft.com/nuget/api/search-query-service-resource#request-parameters
* nuget.org's search syntax is not part of the official NuGet protocol. Ideally Visual Studio uses the NuGet protocol and does not rely on APIs that are specific to nuget.org
* Visual Studio does not properly encode nuget.org's search syntax: https://github.com/NuGet/Home/issues/9043
* nuget.org's search syntax has a strange design that is highly inconsistent with other search syntaxes
*  It is easier for other server implementations to accept new query parameters then

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Let's keep these comments up to get more feedback from the community. They are great suggestions.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+When browsing for packages on NuGet.org, a developer will be able to filter packages based on compatibility filters. These filters will include:
+
+1. The **generation of .NET** such as .NET Framework, .NET Core, .NET (5, 6, and later), and Any.
```
loic-sharma, line 25: Customers may be really confused by `.NET` and `.NET Core`. Support for the last .NET Core TFM will be dropped in December 2022: https://dotnet.microsoft.com/platform/support/policy/dotnet-core#lifecycle

Should we avoid the confusion and "merge" `.NET` and `.NET Core`?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: I would say `platform:` is a necessary filter. The other not so much other than being a helpful shorthand.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Should we support both? The search syntax NuGet.org provides is quite helpful & this would be a great addition.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Will need to get more feedback from others regarding how .NET Core & .NET 5+ are perceived. .NET 5.0 is still technically `.NETCoreApp,Version=v5.0` for example.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: I can add some more words around the tabs & pill controls being used for these concepts as shown in the mockups.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
loic-sharma: The search request would look something like this using query parameters: `?q=hello world&framework=net472&framework=net461`

loic-sharma: > Should we support both?

I feel mixed about this. Personally I would vote 'no' because the syntax has so many problems, but, this would have the benefit of allowing VS users to filter packages before we update its UI.

lukemcdo:
This is a cool idea, but if NuGet is really going to add this feature, I'd like to see something added for Fable if possible.

JonDouglas:
> This is a cool idea, but if NuGet is really going to add this feature, I'd like to see something added for Fable if possible.

Can you help elaborate on the Fable scenario? I read through the documentation about authoring Fable libraries. You would be able to filter for .NET Standard libraries with a search term like "Fable" for your use-case. You wouldn't be able to find fable folder assets with this proposal however. Maybe in another proposal you could search for specific top-level folders like fable to help even further.

lukemcdo:
> 
> 
> > This is a cool idea, but if NuGet is really going to add this feature, I'd like to see something added for Fable if possible.
> 
> Can you help elaborate on the Fable scenario? I read through the documentation about authoring Fable libraries. You would be able to filter for .NET Standard libraries with a search term like "Fable" for your use-case. You wouldn't be able to find fable folder assets with this proposal however. Maybe in another proposal you could search for specific top-level folders like fable to help even further.

I would like to see general search "filters," where TFMs are filters, and a theoretical folder assets proposal can also be a filter.

Then, within the front-end GUI, a combination filter such as .NET Standard + contains a fable top-level folder could appear alongside TFMs.

lukemcdo:
I'm also concerned because in general, this looks somewhat incomplete. For example, UWP theoretically supports .NET Standard 2.0, however, there are many libraries that are surprisingly incompatible and/or surprisingly compatible, and can only really be tested for support and self-reported by the vendor.

Thinking about it as a general problem, .NET Standard isn't/wasn't as standard as intended, and there should be a way for packages to manually mark which "sub-TFMs" a package supports. Maybe some sort of cosmetic-only marker in the package could be used for both the Fable and the UWP scenario? (As well as any other legacy scenarios I'm not thinking about.)

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+![](meta/resources/TargetFrameworkAndPlatformFilters/CompleteExperience.png)
+
+## Search Syntax
+
+Lastly, NuGet should add a protocol in which supports [search syntax](https://docs.microsoft.com/nuget/consume-packages/finding-and-choosing-packages#search-syntax) for frameworks, platforms, and potentially *implementations**.
+
+Here's an ideal `<property>:<term>` for each use-case:
+
+```
+framework:net472 framework:net471 framework:net47   # Match the target framework monikers
```
skofman1, line 134: What would be the behavior when a user selects several platforms/frameworks in the UX, will it be 'AND' or 'OR' between the selections? (this might be out of scope for this spec)

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+    - `net6.0-tizen`
+  - .NET Standard
+
+**Note:** .NET Standard is considered a special case that supports majority of these platforms depending on the [version of .NET standard](https://docs.microsoft.com/dotnet/standard/net-standard#net-implementation-support). If the framework precedence of a target framework supports .NET Standard, it's respective versions supported will be implicitly included.
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/SupportedPlatforms.png)
+
+#### Version Numbers
+
+There are many unique version numbers for all generations of .NET & supported platforms. The filtering experience should be smart enough to create a version list based on the current selection. Here are a few scenarios:
```
jnyrup, line 94: When I saw the screenshot my first reaction was "why would anyone want to list all packages that has a version 2.0".

I now see it's about TFMs, but it still raises some questions for me:
1) Are the selected TFMs what a _package_ targets, or what TFMs _I_ target and hence a nuget must support.
E.g. if my app targets net462 and netcoreapp2.0, a nuget targeting netstandard2.0 would be compatible.

2) Regarding the "The filtering experience should be smart enough to create a version list based on the current selection". I'm wondering about the "smart" part.
Say my app multi targets .net framework 4.7.2 and .net core 2.0.
Would selecting `.NET Framework` and `.NET Core` and then selecting `4.7.2` and `2.0` from the version selector imply that the combined TFMs selected are `.NET Framework 2.0`, `.NET Framework 4.7.2` and `.NET Core 2.0`?

JonDouglas: 1. It would be what a package provides assets for. At least for now. So in your case, if you filtered for `.NET Framework` -> `Windows` -> `4.6.2`, you would see packages providing assets for `net462` or are compatible in precedence with `net462`. https://nugettools.azurewebsites.net/6.0.0-preview.3/framework-precedence?framework=net462&exludedIdentifiers= (I'll add a note to make this more clear in the proposal)
2. Yes, that is right. You would get packages matching `.NET Standard 2.0` as well in this case as its implicitly included. One note is that the current proposal uses `Any` for all generations(.NET Framework and .NET Core and .NET)  & only allows individual selection outside of that(.NET Framework or .NET Core or .NET).

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: I believe current user-facing search syntax only supports `OR`. The examples here match what is currently supported.

Supporting `AND` and other advanced search syntax qualifiers would be nice & is likely supported:

https://docs.microsoft.com/en-us/azure/search/query-simple-syntax

For example, the `+` operator on a NuGet Azure Search query seems to work and support `AND`.

EX: `framework:net472+net471+net47`

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+```
+framework:net472 framework:net471 framework:net47   # Match the target framework monikers
+
+platform:windows platform:macos platform:linux    # Match the platforms
+
+implementation:netframework implementation:netcore implementation:net  # Match the generation of .NET including both supported frameworks & platforms.
+```
+
+- `framework:` - Allows a short-form TFM value to search on.
```
zivkan, line 141: FYI, NuGet client has supported sending `&supportedFramework={tfm}` since 2014: https://github.com/NuGet/NuGet.Client/commit/3833a2017a2e167795a45b4ebfce9d48d9809964#diff-734b4bcd455eef711c1fccb3e1488d8fe8dac8dee9f9f6e0fd937c97e3e328f6R110. 

Here's the equivalent code from the current dev branch, in case you want to see what other query parameters client already supports: https://github.com/NuGet/NuGet.Client/blob/927e225f1bbefd3e4df3b5cc3c9371d6f65b723c/src/NuGet.Core/NuGet.Protocol/Resources/PackageSearchResourceV3.cs#L136

I'm not 100% sure, but I think client may still be sending this today for packages.config projects, and possibly "legacy" PackageReference projects. I know, however, that in Visual Studio, SDK style projects do not pass any target frameworks to the search service.

I'm confused over this comment: https://github.com/NuGet/Home/pull/11374/files#r748213065

Depending on the intent, it's either critical to avoid re-using the same parameter, or it might be acceptable to use.

However, if the server implementation is not perfect, old versions of VS do not have a checkbox, or any other way for customers to turn off sending `supportedFramework` in their search HTTP request. So, maybe we want to intentionally avoid `supportedFramework`, to ensure that false negatives do not make packages inaccessible to customers on clients that do not have the flexibility to enable/disable this filtering parameter.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+- `platform:` - Allows a platform value to search on.
+- `implementation:` - Allows a generation of .NET to search on (includes all frameworks of the generation).
```
zivkan, line 143: I feel like `platform` and `implementation` are largely there because of the decision to make `framework` return packages with assets of that framework, rather than packages compatible with that framework.

If, in the future, a search parameter to search by compatibility is added, would `platform` and `implementation` still be useful?  I get that some, particularly new, developers don't know what version of .NET their project is targeting, so they might not know to choose/type `net6.0-android` vs `net7.0-android`, but I feel like Visual Studio's project system makes not unreasonably difficult to find, especially since single clicking a project file opens the xml, or you can right click the project and select edit, where `<TargetFramework>net6.0-android</TargetFramework>` is fairly clear to see.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
zivkan: I'm still/more confused about what your response 1 is saying.

> what a package provides assets for

My understanding is that this means if a package only has `lib/netstandard2.0/my.dll`, and I search `framework:net462`, this package will **not** be in the search results. It will only return packages that have `lib/net462/*` (or from the other assets types, not just lib).

> or are compatible in precedence with net462

So it is returning all packages that are compatible with projects that are targeting net462, directly contradicting my understanding of the above?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Thanks Andy. `supportedFramework` will help with the technical explanation of the spec if we choose to re-use it. It does require a fully qualified name such as `.NETFramework,Version=v4.5` vs. the shorthand `<TargetFramework>net45</TargetFramework>` that you might include in your project file. We'll have to see how easy it is to support common names & whether the existing functionality will suffice.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Definitely. I think an ideal experience would allow a developer to fully qualify their scenario for searching packages on NuGet.org or their context of their project in their favorite client experience (VS/CLI/etc) to fully qualify it for them & be provided compatible assets entirely.

For now since no such thing exists, these additions just help narrow down results as helpful additions. The useful of them may diminish as tooling gets smarter for compatibility scenarios (i.e. matching TFMs in a project or solution directly with round-trip compatible TFMs)

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: For a `net462` scenario, the precedence includes `netstandard1.3 -> netstandard2.0`. .NET Standard is implicitly included in search results & thus would be included.

If we allow the ability to disable the implicit `.NET Standard` functionality, then only packages that have `net462` and prior `net` version assets would show up. 

There's probably a further discussion on where do we draw the line on precedence for older/out of support TFMs that are still "compatible". I think maybe that line to draw is on the framework itself & netstandard to start.

JonDouglas:
> I'm also concerned because in general, this looks somewhat incomplete. For example, UWP theoretically supports .NET Standard 2.0, however, there are many libraries that are surprisingly incompatible and/or surprisingly compatible, and can only really be tested for support and self-reported by the vendor.
> 
> Thinking about it as a general problem, .NET Standard isn't/wasn't as standard as intended, and there should be a way for packages to manually mark which "sub-TFMs" a package supports. Maybe some sort of cosmetic-only marker in the package could be used for both the Fable and the UWP scenario? (As well as any other legacy scenarios I'm not thinking about.)

Thanks for the feedback. This is a work in progress & will continue to evolve with the feedback we get.

There is no silver-bullet solution to provide complete compatibility for the extensive matrix of past/present/future frameworks, platforms, etc of .NET. The hope of this proposal is to capture the majority & build on-top of it afterwards.

The framework precedence tool mentioned in this proposal may help provide a better idea of how compatibility is defined for best matches & we'd be using that to help developers find compatible packages with these filters.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
loic-sharma: > I believe current user-facing search syntax only supports `OR`.

We `OR` terms for the same field but you must match at least one term per field. For example imagine my query is `color:blue color:red make:Toyota make:Mazda`: this would match “blue Toyota”, or “blue Mazda”, but not “blue Hummer”.

> Supporting `AND` and other advanced search syntax qualifiers would be nice & is likely supported

Yup this could be supported, but we should be thoughtful here as the search syntax is quite confusing. Making the same syntax have different behaviors on different fields may complicate things further. 

Alternatively, we could instead revamp the current search syntax to be more sensible. We would need to understand how the current search syntax is used to make sure such a revamp is successful.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
jcjiang: The unclear relationship between certain TFMs and their proper names came up for the displaying on package details feature as well - the question was, will less experienced .NET developers know what frameworks are compatible for their projects by their proper name or the TFMs they see in their package assets? For that we decided to stick to proper names when we could and show the TFM names when the relationship was very unclear. 

For filters it will be way less of a problem because you are defining three major generations as opposed to dealing with more obscure TFMs, but I wonder if there are any cases where we might need to show info beyond the proper names.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a developer is browsing for a NuGet package to include in their project, they can filter packages based on generations of .NET including the target framework & supported platforms that package targets.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. Ontop of it being one of the biggest error codes we see, it is also [the number #1 request on the `NuGetGallery` repository](https://github.com/NuGet/NuGetGallery/issues/3098) for good reason. It should be straight-forward for a developer to separate the generation of .NET they are currently targeting and find packages that are supported on the platforms that the respective generation supports.
```
nkolev92:
```suggestion
Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. On top of it being one of the biggest error codes we see, it is also [the number #1 request on the `NuGetGallery` repository](https://github.com/NuGet/NuGetGallery/issues/3098) for good reason. It should be straight-forward for a developer to separate the generation of .NET they are currently targeting and find packages that are supported on the platforms that the respective generation supports.
```

proposed/2021/TargetFrameworkAndPlatformsFilters.md
nkolev92: This is more about the naming convention, but `generations` as I understand is usually referred to the different versions of the frameworks. 

Things like `.NET`, `.NET Core`, `.NET Framework` are really just called frameworks. 

Is this terminology used anywhere else?

I just want to make sure we're consistent.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/SupportedPlatforms.png)
+
+#### Version Numbers
+
+There are many unique version numbers for all generations of .NET & supported platforms. The filtering experience should be smart enough to create a version list based on the current selection. Here are a few scenarios:
+
+- User selects `Any` for the .NET generation.
+  - Version filter populates all known versions for .NET Framework, .NET Core, .NET, and .NET Standard.
+    - 1.0, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, 2.0, 2.1, 2.2, 3.0, 3.1, 3.5, 4.0, 4.5, 4.5.1, 4.5.2, 4.6, 4.6.1, 4.6.2, 4.7, 4.7.1, 4.7.2, 4.8, 5.0, 6.0
```
nkolev92, line 98: I have concerns about showing .NET Framework packages as compatible here, especially later versions (such as .NET 4.8).

These are not fully compatible. Should we differentiate between the 2?

If we won't we should call it out explicitly.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
nkolev92: One special consider woudl be how we handle things such as `net5.0-windows`. 

There is an implicit platform version set by the SDK, that NuGet does not understand. 

This would be a larger concern with net5.0-android. 

Alternatively, we could not respect combinations such as this.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Yeah we can use `Framework`, but I used generations in this spec to avoid confusion around `.NET Framework`. Perhaps we define a generation as a `framework` and a TFM as a `target framework`. :)

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,196 @@
...
+- [debian](https://packages.debian.org/) has filters for release suites and supported architectures.
+- [winget](https://docs.microsoft.com/windows/package-manager/winget/search) allows you to filter by a supported app moniker.
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- Should there be an advanced option to disable .NET Standard from being implicit/on-by-default for more exact matches?
```
zivkan, line 193: Why? What benefit to customers would there be to exclude .NET Standard-only libraries from search results?

proposed/2021/TargetFrameworkAndPlatformsFilters.md
anangaur: Can we support the same mechanism as GitHub for filters?

anangaur: Will selecting the UI filters also put the corresponding text filters in the search box? (same as GitHub)

jcjiang (approved):
Really great work here, Jon!

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Here is the official terminology. `Framework` seems a bit more accurate. 

> implementation for .NET “flavors” (.NET
Framework, .NET Core, Mono). This describes
the runtime, tools, libraries/frameworks for
building different groups of applications.
There are shared components across
implementations (i.e. compilers, services, .NET
Standard).

> framework for APIs (.NET Framework base
class library, .NET Core framework, Mono
base class library). Libraries is OK here too
depending on context. 

https://github.com/dotnet/brand/blob/main/dotnet-brand-guidelines.pdf

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: The original thinking was to be able to search for platform specific assets rather than .NET Standard as a fallback.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: In this scenario, it's just a filter for "Any". 

If a user selected `.NET Framework` & a specific version of framework like 4.5, they should only see `.NET Standard 1.0-1.1` compatible assets.

i.e. net45+ https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: What is the concern? 

`net6.0-androidXX.X` would represent `.NET 6` + `Android`. The platform version is not in scope of this spec today if it is defined.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Agreed. The proper name is ideal for this scenario. We will stay with them to keep them consistent with other badges and similar.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: @anangaur That was an original hope. The problem is that NuGet.org does not support simple search syntax like `AND` in the frontend experience (although the backend does support it). If we can remove that barrier, we could do a search syntax autocompletion in the textbox like GitHub.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
nkolev92: makes sense. Resolving.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
nkolev92: Can you mix framework and platform in the queries? 

For example, framework=net5.0-android and platform=macos. 

It's  easy to get into a situation where the queries contradict.

JonDouglas: Ah gotcha. No you wouldn't be able to do this with modern TFM. It's one framework and multiple platforms.

.NET 6 + Android + macOS = net6.0-android && net6.0-macos

nkolev92: So is `net6.0-android` allowed in the framework part?

JonDouglas: I don't know right now. I would say that if it's a valid short form TFM, then yes. But it should be smart enough to treat the `-<platform>` as a platform filter in the search call. We will see what's possible shortly. If you did support it, it may be a similar case above where:

`net6.0-android + macOS = net6.0-android && net6.0-macos`

The goal in my eyes is to separate framework completely from platform for modern TFMs so the user experience is straight-forward of "you want to target this version of .NET on this platform"! Will learn more with more feedback & interviews.

nkolev92: > But it should be smart enough to treat the -<platform> as a platform filter in the search call

Yeah, I'd guess that's 100% doable. 
Mixing and matching specific frameworks with platforms and the platform filter would be weird.

Let's see what the customer dev says. 

We can always say you can't mix them early on, but then allow it later.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Added a couple notes to address these. For now, we will keep platforms near these main filters similar to other ecosystems like pub.dev.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Resolving this feedback for now as these questions are more suited for the implementation. We definitely want ways to search by TFM in both the API & existing search syntax. The former being more important than the latter from the SDK perspective.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Resolving as this has a path forward.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Resolving as this has a path forward. Added a note to `unresolved questions` for now.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Resolving as this has a path forward, adding it under `unresolved questions` with the current answer.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Added as an unresolved question for now.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: Adding an unresolved question regarding the naming of `implementation` search parameter. Other comments stand as "as tooling gets smarter, the usefulness will diminish".

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,205 @@
...
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/CompleteExperience.png)
+
+**Note:** There's opportunity to put a tooltip in the complete experience to help explain what the difference of .NET, .NET Core, .NET Framework is & link to a documentaiton page.
+
+## Search Syntax
+
+Lastly, NuGet should add a protocol in which supports [search syntax](https://docs.microsoft.com/nuget/consume-packages/finding-and-choosing-packages#search-syntax) for frameworks, platforms, and potentially *implementations**.
+
+Here's an ideal `<property>:<term>` for each use-case:
```
joelverhagen, line 131: I am concerned about making the filtering syntax based on the `q=` (query term) parameter to search instead of a different query parameter (similar to what @zivkan is alluding to in https://github.com/NuGet/Home/pull/11374#discussion_r748346221). This search term will not be recognized by other search implementations (e.g. Azure DevOps) and may lead to bad results on those feeds. This is mainly relevant with the "All" source selected in VS PM UI, browse tab.

If we make this filter a query string (like `framework=X&q=json`) it will be ignored by other search services. Ignoring unknown query parameters is very standard in REST/HTTP whereas the parsing of `q=` (the search term) is much less forgiving or standardized.

Take for example, this URL:
https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/9d15d80a-6afc-4f7e-901b-9378146a4b8b/nuget/v3/query2/?q=foo%20framework:net6.0
This is search "foo framework:net6.0" on an a public Azure DevOps feed. **Zero results.**

Compare this to:
https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/9d15d80a-6afc-4f7e-901b-9378146a4b8b/nuget/v3/query2/?q=foo&0framework=net6.0
This is sending a `framework=net6.0` query parameter to the same feed. **One result.**

There is a whole technical discussion needed on what to do with feeds of mixed capabilities (nuget.org having framework filtering and Azure DevOps not yet having it) but I fear that this approach of hijacking the search text for this will limit our solution space or force us into nasty hacks.

We _could_ say that this syntax only works on NuGet.org UI but this would be a change in patterns. Today, any search you can do on nuget.org can be done similarly in PM UI to get the same results.

As an aside, this approach also diverges from the implementation of other nuget.org advanced filters, e.g. sorting and package type filtering. These are both handled with query parameters rather than adding `<property>:<term>` to the search text.

I see a direction here that leads to an advanced search query building UX on NuGet.org (similar to GitHub's or Google's old page) that builds up the search text with `<property>:<term>` specifiers, and I like that idea and direction, but it's hard for me to see how this would work well in the near tear where there is little consistency in search syntax across feeds.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,205 @@
...
+- .NET Core:
+  - `netcoreapp1.0, netcoreapp1.1, netcoreapp2.0, netcoreapp2.1, netcoreapp2.2, netcoreapp3.0, netcoreapp3.1`
+- .NET:
+  - `net5.0, net6.0`
+- Any:
+  - .NET Framework, .NET Core, .NET
+
+![](meta/resources/TargetFrameworkAndPlatformFilters/GenerationsOfNET.png)
+
+#### Supported Platforms
```
joelverhagen, line 46: @advay26 and I talked a bit about maybe phasing out platforms and frameworks. Today, the proposed UI for the nuget.org package details page only refers to frameworks. There is no platform notion for display yet. We could consider implementing first the framework filtering to align with the upcoming package details page changes and only add the platform filtering when it's materialized in a well understood way on the details page.

As an aside, the platform concept is a bit more opaque to me (perhaps due to my high exposure to NuGet-only concepts) so I'm wondering if we actually have consensus in the .NET ecosystem on what "platform" means.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
joelverhagen: I think this is a key point to clarify in the spec. Could we provide some example search queries and matching packages? For example, what are 1-3 packages that `json framework:net6.0` would return? Would it include Newtonsoft.Json?

Regarding:
> I feel like `platform` and `implementation` are largely there because of the decision to make `framework` return packages with assets of that framework, rather than packages compatible with that framework.
Where is this noted and/or decided?

From a phasing/scoping perspective, it totally makes sense to start with searching first on package assets then later on package compatibility, but that sounds like just phases not something we should bake into our search syntax design unless we really want to support both modes in parallel. Sorry if I am missing context.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: This spec leapfrogs a bit of package details as it was written before it. I believe we can revisit package details with inclusion of platforms or have a feature flag for platforms until package details has parity.

**I've added that as an opportunity at the bottom of the spec just now.**

We definitely want a consistent experience & I'll visit that again shortly.

How .NET defines a platform is the following:

```
platform for operating systems and chipsets
(Windows, macOS, Linux) with distros for Linux
flavors. Unless prefixed by “developer” which
can mean a programming platform
```
from https://github.com/dotnet/brand/blob/main/dotnet-brand-guidelines.pdf

image

Really what we're trying to do here is break up TFM strings from being opaque to having individual meaning in parts. Such as this scenario from .NET 5:

https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#valid-platforms

> So with a TargetFramework of net5.0-android, NuGet should parse out the platform as Android. The core .NET SDK will then set the TargetPlatformIdentifier to android. The Android workload should then normalize the casing of the TargetPlatformIdentifier property to Android. Similarly, the iOS Workload and Windows targets should normalize the TargetPlatformIdentifier to iOS and Windows.

For filtering purposes, we want to show packages that have support for that platform with platform-specific assets and/or .NET Standard(portable) assets if selected.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: I don't have strong opinions because I'm quite ignorant about search. I'm hoping we can first define the frontend experience for the user via search syntax that they can use alternative to clicking through filters & we do the right thing in the backend to generate the proper search result.

I used the current experience of typing in something like `authors:microsoft` into NuGet.org as a reference & that in turn would generate the respective query(q=?) parameters. If it's more ideal to follow the advanced filters paradigm of `framework=` then by all means! 

The goal at the end of the day would be to allow someone to apply filters, possibly generate those in the search box itself as hint text, and give the user the respective results with those filters applied.

joelverhagen: If the user experience requires typing into the search box, we may need to do weird hacks like parsing out framework/platform/implementation search tokens and converting them behind the scenes to query parameters -- at least on the VS side as to not mess with other feeds that don't yet support compatibility filtering. Hopefully we don't have to do that.

I think it's worth adding a note to this section or change emphasis such that the filter experience is the priority and the search text is optional pending feasibility investigation. That's my opinion at least 😄.

On the VS side, I am imagining this filtering is done implicitly based on project context or with a similar filtering UI that's on NuGet.org. The reason I bring this up is that we've attempted to keep reasonable parity between the two search experiences (nuget.org + VS PM UI).

advay26 (approved):

proposed/2021/TargetFrameworkAndPlatformsFilters.md
JonDouglas: > I think it's worth adding a note to this section or change emphasis such that the filter experience is the priority and the search text is optional pending feasibility investigation. That's my opinion at least 😄.

Yep let's do that.

> On the VS side, I am imagining this filtering is done implicitly based on project context or with a similar filtering UI that's on NuGet.org. The reason I bring this up is that we've attempted to keep reasonable parity between the two search experiences (nuget.org + VS PM UI).

Yep, I think right now even with `supportedFramework`, I don't think we can capture a similar browse/search experience in the client until we do the nuget.org side. Even if we matched a list of project frameworks, we may want users to look for platform specific vs. portable assets which isn't possible today I don't believe.

proposed/2021/TargetFrameworkAndPlatformsFilters.md
```diff
@@ -0,0 +1,210 @@
...
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When a developer is browsing for a NuGet package to include in their project, they can filter packages based on generations of .NET including the target framework & supported platforms that package targets.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. On top of it being one of the biggest error codes we see, it is also [the number #1 request on the `NuGetGallery` repository](https://github.com/NuGet/NuGetGallery/issues/3098) for good reason. It should be straight-forward for a developer to separate the generation of .NET they are currently targeting and find packages that are supported on the platforms that the respective generation supports.
+
```
anangaur, line 16: Given the problem statement in this section: 

> Finding a package that is compatible with your project is a daily challenge for the average developer in the .NET ecosystem. Over 21% of developers fail to install a package today because of an incompatibility; it's one of NuGet's top errors known as `NU1202`. 

Will showing the TFM badges in search results go a good distance towards solving the finding the "right" package?

joelverhagen (approved):

zivkan:
@JonDouglas there hasn't been activity for a few months now. We're trying to progress/close open pull requests.  Do we need more reviews?

JonDouglas:
@zivkan I will merge this as I was supposed to do that start of January. Thanks for the reminder.

PR #11391, opened 2021-11-16
Central Package Management - Transitive Pinning

JonDouglas:
This document contains the behaviors known today of the central package management's transitive pinning feature & has a few questions/proposal that we would love to get your feedback on. Please feel free to drop a comment on this PR, or add a review yourself to help us.

[Rendered Document](https://github.com/NuGet/Home/blob/75c2b5f5dd6e7ddee3e3e92ad00cf4e878cfcf87/proposed/2021/CentralPackageManagementTransitive.md)

Thank you! 😄

### DIFF
--- a/proposed/2021/CentralPackageManagementTransitive.md
+++ b/proposed/2021/CentralPackageManagementTransitive.md
@@ -1,0 +1,201 @@
+# Specifying Versions for Transitive Dependencies
+Central Package Version Management (CPVM) is a feature set in NuGet that allows you to manage package versions in one place.  Consider the following packages are available:
+
+| Id | Version | Dependencies |
+|-----:|:---------|-----|
+| PackageA | 1.0.0 | PackageC >= 3.0.0 |
+| PackageB | 2.0.0 | PackageC >= 4.0.0 |
+| PackageC | 3.0.0 |  |
+| PackageC | 4.0.0 |  |
+| PackageC | 5.0.0 |  |
+
+## Referencing packages without CPVM
+This section explains how to specify a different version of a transitive dependency when using the standard mechanism for referencing packages and specifying versions.
+
+**`ClassLibrary1.csproj`**
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="PackageA" Version="1.0.0" />
+    <PackageReference Include="PackageB" Version="2.0.0" />
+  </ItemGroup>
+</Project>
+```
+
+`PackageA` and `PackageB` both transitively depend on `PackageC` but different versions, `3.0.0` and `4.0.0` respectively.  NuGet will unify this to the minimum version that satisifies the requirements which is **`4.0.0`**.
+
+During Pack, the resulting package will have the top-level dependencies on `PackageA` and `PackageB` and the transitive dependencies of `PackageC` will be resolved by the consumer:
+
+**`ClassLibrary1.nuspec`**
+```xml
+<dependencies>
+  <group targetFramework="net5.0">
+    <dependency id="PackageA" version="1.0.0" exclude="Build,Analyzers" />
+    <dependency id="PackageB" version="2.0.0" exclude="Build,Analyzers" />
+  </group>
+</dependencies>
+```
+Since version `4.0.0` of `PackageC` was used in this scenario, it is okay for the generated package to transitively depend on `4.0.0` since consumers of this library will end up using hte same version of `PackageC` that was used by this package.
+
+### Overriding transitive versions
+If a user wants to override the transitive version of `PackageC`, they must add an explicit `PackageReference` item:
+```diff
+<ItemGroup>
+  <PackageReference Include="PackageA" Version="1.0.0" />
+  <PackageReference Include="PackageB" Version="2.0.0" />
++ <PackageReference Include="PackageC" Version="5.0.0" />
+</ItemGroup>
+```
+
+This eclipses the transitive version of `PackageC` since it is newer which results in referencing `PackageC` version **`5.0.0`**.  This affects the restored packages, reference assemblies passed to the compiler, files copied to the output directory, and dependencies in used during package creation.
+
+```diff
+<dependencies>
+  <group targetFramework="net5.0">
+    <dependency id="PackageA" version="1.0.0" exclude="Build,Analyzers" />
+    <dependency id="PackageB" version="2.0.0" exclude="Build,Analyzers" />
++   <dependency id="PackageC" version="5.0.0" exclude="Build,Analyzers" />
+  </group>
+</dependencies>
+```
+Note that since **`PackageC` is no longer considered transitive, it must now be elevated to be a top-level dependency of my package**.  This is because `ClassLibrary1` was built against version `5.0.0` of `PackageC` so the transitive dependencies of `PackageA` and `PackageB` are no longer accurate.  The act of "pinning" a version of a transitive dependency effectively transformed it into an explicit dependency.
+
+## Referencing packages with CPVM
+
+To manage the package versions in one location, the `Version` metadata is removed from `PackageReference` items in the project:
+
+**`ClassLibrary1.csproj`**
+```diff
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+  <ItemGroup>
+-   <PackageReference Include="PackageA" Version="1.0.0" />
++   <PackageReference Include="PackageA" />
+-   <PackageReference Include="PackageB" Version="2.0.0" />
++   <PackageReference Include="PackageB" />
+  </ItemGroup>
+</Project>
+```
+
+And a `Directory.Packages.props` file is created with the following content:
+
+**`Directory.Packages.props`**
+```xml
+<Project>
+  <ItemGroup>
+    <PackageVersion Include="PackageA" Version="1.0.0" />
+    <PackageVersion Include="PackageB" Version="2.0.0" />
+  </ItemGroup>
+</Project>
+```
+
+All projects in the repository that reference `PackageA` or `PackageB` explicitly will use the specified value in a corresponding `PackageVersion` item.
+
+If a user wants to override the transitive version of `PackageC`, they must add an explicit `PackageReference` item:
+```diff
+<ItemGroup>
+  <PackageReference Include="PackageA" />
+  <PackageReference Include="PackageB" />
++ <PackageReference Include="PackageC" />
+</ItemGroup>
+```
+
+And add a corresponding `PackageVersion` item to Directory.Packages.props:
+```diff
+<ItemGroup>
+  <PackageVersion Include="PackageA" Version="1.0.0" />
+  <PackageVersion Include="PackageB" Version="2.0.0" />
++ <PackageVersion Include="PackageC" Version="5.0.0" />
+</ItemGroup>
+```
+
+The result is the same as before where the resolved `PackageC` is version **`5.0.0`** and affects the restored packages, reference assemblies passed to the compiler, files copied to the output directory, and dependencies in used during package creation.  The transitive dependency is no longer considered transitive and instead is an explicit dependency.
+
+### Proposal to allow transitive version selection based on specified PackageVersion items
+In the previous example, a user must specify an explicit `PackageReference` to a package to override the transitive version.  This elevates the reference from transitive to explicit and behaves as a user expects.  However, some users want a `PackageVersion` to automatically override a transitive version even without an explicit top-level `PackageReference`.  The reason for this feature request is for better **scale** and **performance**.
+
+#### Addressing Scale
+For larger repositories, requiring users to specify explicit `PackageReference` items in each project can become difficult.  This is because you may need to add it to dozens of projects in order to unify the version used across the repo.  Since CPVM is used to centralize package versions, it makes sense to have transitive reference version overrides to be implied when specifying a `PackageVersion`.  There are however at least two drawbacks of this approach:
+
+1. **Confusion** about "pinning" a transitive version - Many users consider overriding a transitive version to be pinning that version in the graph.  In some cases they consider this gesture only part of Restore and Build.  But since they've overridden a transitive version, this also flows to Pack which affects the versions used as dependencies.  When transitive versions are are "pinned", they are really just elevated to top-level dependencies which causes this confusion.  We will need to educate users on how this feature works so they understand that it is not really just pinning a version but instead is actually making it an implicit reference.
+
+3. How to opt out - If transitive version overrides are automatic when a `PackageVersion` is specified, there will need to be a way for a user to opt out of this behavior for one or more `PackageReference` items.  In previous versions of NuGet, a user could only opt into transitive version overrides by _adding_ a `PackageReference` item.  But to opt out of newer behavior, they will not have a `PackageReference` item to remove.  An MSBuild property could be used to opt an entire project out of transitive version overrides but that would not address a single `PackageReference` item.
+
+Some ideas on this are the following:
+
+**Metadata on a PackageReference to opt out**
+```xml
+<ItemGroup>
+   <PackageReference Include="PackageC" Pin="false" />
+</ItemGroup>
+```
+**Drawback**: A user has to _add_ a `PackageReference` to opt it out of behavior which is not a great user experience.
+
+**Users follow a pattern to disable a PackageVersion item**
+
+```diff
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    ...
++   <PinPackageC>false</PinPackageC>
+  </PropertyGroup>
+</Project>
+```
+
+**Directory.Packages.props**
+```diff
+<ItemGroup>
+- <PackageVersion Include="PackageC" Version="5.0.0" />
++ <PackageVersion Include="PackageC" Version="5.0.0" Condition="'$(PinPackageC)' != 'false'" />
+</ItemGroup>
+```
+
+**Drawbacks**: Requires users to follow a well documented convention which leaves a lot of room for error.
+
+#### Addressing Performance
+Users have provided feedback that overriding transitive versions improves the overall time of restore because it simplifies the graph by reducing the amount of walking performed by the resolution algorithm.  Obviously it would not scale for each and every project in a tree to specify all `PackageReference` items so it only makes sense for the central version file to contain all of the versions.  Large repositories would experience more benefit from this since smaller repositories generally have fast restore times.  One drawback of relying on transtiive package version overrides is that restore could get slow again as users add top-level packages but don't add new transitive dependencies or update the versions.  There would be nothing built-in to remind them that the restore graph has gotten more complex so they would be in a constant battle of updating all their versions.  This could be handled by tooling but it would be a better user experience if NuGet restore performance was made better across the board and the act of walking large graphs be made faster and more optimized.  This could be done by caching graph walks for subsequent restores or profiling and optimizing the algorithm.  If we improved the performance of restore, users would not need to rely on transitive package version overrides to get fast restore times.
+
+### Other considerations
+
+#### Respecting assets of transitive dependencies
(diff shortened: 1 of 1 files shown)


proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+  <PackageVersion Include="PackageA" Version="1.0.0" />
+  <PackageVersion Include="PackageB" Version="2.0.0" />
++ <PackageVersion Include="PackageC" Version="5.0.0" />
+</ItemGroup>
+```
+
+The result is the same as before where the resolved `PackageC` is version **`5.0.0`** and affects the restored packages, reference assemblies passed to the compiler, files copied to the output directory, and dependencies in used during package creation.  The transitive dependency is no longer considered transitive and instead is an explicit dependency.
+
+### Proposal to allow transitive version selection based on specified PackageVersion items
+In the previous example, a user must specify an explicit `PackageReference` to a package to override the transitive version.  This elevates the reference from transitive to explicit and behaves as a user expects.  However, some users want a `PackageVersion` to automatically override a transitive version even without an explicit top-level `PackageReference`.  The reason for this feature request is for better **scale** and **performance**.
```
jjmanton, line 122: I disagree that the only reason for this request is scale and performance. 

We have multiple top level projects in a solution, and due to some deployment complexity, install them side by side in the same folder. We want the dependencies to all be on the same version, even when Nuget is not able to resolve across dependency trees. Otherwise, we might end up with a dll of unknown version depending on which build finished first.

jjmanton:
Here are my answers, if it is helpful. I work on a large solution, with around 270 projects, ~3.4m lines of C#, mix of Net Framework and Netstandard projects for background.

**- Are you confused about the transitive behavior of what "pinning" a reference does today? (i.e. promotes a transitive dependency to a top-level dependency implicitly).**

Nope, we do this on occasion. It sometimes creates maintenance issues, but we are usually able to spot them. For example, Reference Package A, Dependency on Package B. Add Top Level for Package B to pin version. On upgrading Package A, it drops dependency on B. My understanding is that CPVM might alleviate this.

**- Is it important to you to be able to optionally remove a single PackageReference from the same "pinning" behavior?**

Not for our organization. Our primary goal of using CPVM is to ensure consistent versions across all projects. We want this for not only technical, but operational constraints as well.

**- Would you expect IncludeAssets, ExcludeAssets, or PrivateAssets to be respected when overriding a dependency?**

Can this question be clarified a bit? If I Pin a version centrally of a Roslyn analyzer, would that change how it is consumed? I expect that the way in which the dependency was defined transitively would remain constant. However, the version might change. I would not expect transitive pinning to change the behavior or promote to a top level dependency. If I explicitly override, adding the package as a top level dependency, I DO expect to have to manually define the flow of assets.

**- Do you understand that this feature is primarily designed for usability & maintainability of your packages in a central location? It does not guarantee any performance benefits with as your projects scale.**

Absolutely! And we are very much looking forward to the feature.

proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Addressing Scale
+For larger repositories, requiring users to specify explicit `PackageReference` items in each project can become difficult.  This is because you may need to add it to dozens of projects in order to unify the version used across the repo.  Since CPVM is used to centralize package versions, it makes sense to have transitive reference version overrides to be implied when specifying a `PackageVersion`.  There are however at least two drawbacks of this approach:
+
+1. **Confusion** about "pinning" a transitive version - Many users consider overriding a transitive version to be pinning that version in the graph.  In some cases they consider this gesture only part of Restore and Build.  But since they've overridden a transitive version, this also flows to Pack which affects the versions used as dependencies.  When transitive versions are are "pinned", they are really just elevated to top-level dependencies which causes this confusion.  We will need to educate users on how this feature works so they understand that it is not really just pinning a version but instead is actually making it an implicit reference.
```
zivkan, line 127: Should we explain a bit **why** a pinned package is elevated to a direct dependency?  A brief example of how if this was not the case, the package author might accidentally use an API only available in the higher version of the transitive package, and then when another developer uses the package author's package in their own app, the library crashes due to missing method exceptions, or similar.

proposed/2021/CentralPackageManagementTransitive.md
swythan: Going by the subsequent paragraph, I think @JonDouglas is using the term "scale" in the sense of "managing the versions in every .csproj individually doesn't _scale_", That seems to largely align with your use-case, in that you _can_ do what you need now, but it's an unreasonable amount of work. 

I'd certainly agree that all of that meaning isn't immediately clear from the phrase "better **scale** and **performance**" (although I can't immediately think of better wording). It made me think of the scale of the dependency graph, or something like that.

JonDouglas: Thank you both. Yes the wording can use some word smithing. It was a hidden benefit that users use this feature today for. We just want to express that perf isn't any guarantee here and especially as projects scale(become larger), it is even less of a guarantee. Rather the focus is on UX, convenience, and more.

tompazourek:
Regarding your ideas on **how to opt-out of pinning for a specific package**, I have never actually needed to do this yet. But the MSBuild property and `Condition` pattern looks most natural to me. I understand your comment about coming out with some convention for naming the MSBuild properties, but on the other hand, it gives you more power, e.g.:

- you can have a single property for a group of packages,
- or you can alter it based on which type of build you currently do.

So the **Users follow a pattern to disable a PackageVersion item** seems like the ideal solution to me.

---

And for the **questions**:

> Are you confused about the transitive behavior of what "pinning" a reference does today? (i.e. promotes a transitive dependency to a top-level dependency implicitly).

No, actually it's very simple to think about. When transitive pinning still used to work, I always explained it to developers as follows: If you add a version into Directory.Packages.props, it will make sure that the package is always that version, regardless of where in the dependency graph it is (both direct and transitive). And that is a very simple idea to grasp, in my opinion. Specify a version for PackageA => everywhere where PackageA is used even transitively, it will have that version.

However, I have never yet used CPVM for projects where I would Pack any packages, it never seemed like a good fit yet. I don't often encounter big complex projects that produce library packages, and I think that big complex projects benefit from CPVM (and transitive pinning) the most.

So I'm coming from a perspective where I don't have experience using CPVM together with Pack... To be honest, I don't think transitive pinning is great for building libraries at all. When building libraries, I prefer everything to be more explicit.


> Is it important to you to be able to optionally remove a single PackageReference from the same "pinning" behavior?

No, I have never needed to do this yet. It feels nice to have the option, but I never had a use-case for it yet.


> Would you expect IncludeAssets, ExcludeAssets, or PrivateAssets to be respected when overriding a dependency?

I am not sure if I understand this correctly. But for me, all this CPVM stuff should ideally be only about deciding which version is chosen. So for me, ideally, everything would work the same whether I pick specific version or I don't. But if it makes things too complicated to achieve that, I don't have a problem with it working differently. If the underlying mechanism is that it "automatically" adds the direct PackageReference, and that does have its implications, that's fine with me, even though not ideal.


> Do you understand that this feature is primarily designed for usability & maintainability of your packages in a central location? It does not guarantee any performance benefits with as your projects scale.

Yes, I don't care too much about performance as long as it's not like 10x slower. I want to be able to reason about which versions are used in which projects, and the **great simplification of dependency management** that I get is worth it regardless of the performance.

The idea that I mentioned: *"Specify a version for PackageA => everywhere where PackageA is used even transitively, it will have that version."* is such a great tool for complex solutions with 100-200 projects in it, that it's very worth it in my eyes.  Without it, whenever a new dependency is added somewhere, I would need to manually go through the entire dependency graph, and figure out where I need to add explicit PackageReference, and where I need to remove them, just to get the transitive dependencies under control. And that is very difficult. Any change of dependencies between projects could have unforeseen consequences. But with CPVM and transitive pinning, it's all much simpler, I just write down the version, and it's that version everywhere, even transitively. No need to figure out which of the 100-200 projects need some new explicit dependency just to make the versions aligned. When it's automatic, all of this is hard work is gone, and it's simple to think about. I have a sort of a "guarantee" that the version is the same everywhere if that makes sense. And this "guarantee" is worth a lot to me.

When I adopted CPVM for use in a big solution like that, I **didn't actually see a performance increase, but instead a slight performance decrease**. From my measurements, a clean restore went from ~30 seconds to ~34 seconds with CPVM, and an incremental restore (when nothing was changed) from ~4 seconds to ~9 seconds with CPVM. This was in times of MSBuild 16.7. So I already experienced a performance penalty, and still adopted it.

Hope this feedback helps 🙂

proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+
+3. How to opt out - If transitive version overrides are automatic when a `PackageVersion` is specified, there will need to be a way for a user to opt out of this behavior for one or more `PackageReference` items.  In previous versions of NuGet, a user could only opt into transitive version overrides by _adding_ a `PackageReference` item.  But to opt out of newer behavior, they will not have a `PackageReference` item to remove.  An MSBuild property could be used to opt an entire project out of transitive version overrides but that would not address a single `PackageReference` item.
+
+Some ideas on this are the following:
```
marcin-krystianc, line 131: There is also a syntax which leverages already existing `Remove` attribute (https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2022#BKMK_RemoveAttribute):
```xml
<ItemGroup>
  <PackageVersion Remove="PackageC" />
</ItemGroup>
```
Such entry removes the globally defined `PackageVersion/PackageC` item so the result is same as if it was not defined at all (thus pinning will not happen for `PackageC`)

JonDouglas:
Thank you all so much for your feedback. We'd like to keep this open for a little bit to collect more feedback. This is very helpful!

proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+**`ClassLibrary1.nuspec`**
+```xml
+<dependencies>
+  <group targetFramework="net5.0">
+    <dependency id="PackageA" version="1.0.0" exclude="Build,Analyzers" />
+    <dependency id="PackageB" version="2.0.0" exclude="Build,Analyzers" />
+  </group>
+</dependencies>
+```
+Since version `4.0.0` of `PackageC` was used in this scenario, it is okay for the generated package to transitively depend on `4.0.0` since consumers of this library will end up using hte same version of `PackageC` that was used by this package.
```
nkolev92, line 42:
```suggestion
Since version `4.0.0` of `PackageC` was used in this scenario, it is okay for the generated package to transitively depend on `4.0.0` since consumers of this library will end up using the same version of `PackageC` that was used by this package.
```

proposed/2021/CentralPackageManagementTransitive.md
nkolev92: Does line 65 cover it?

> Note that since **`PackageC` is no longer considered transitive, it must now be elevated to be a top-level dependency of my package**.  This is because `ClassLibrary1` was built against version `5.0.0` of `PackageC` so the transitive dependencies of `PackageA` and `PackageB` are no longer accurate.  The act of "pinning" a version of a transitive dependency effectively transformed it into an explicit dependency.

proposed/2021/CentralPackageManagementTransitive.md
nkolev92: I'd definitely add `unification` as the primary purpose or something along the lines that @jjmanton is suggesting.

proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+```diff
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    ...
++   <PinPackageC>false</PinPackageC>
+  </PropertyGroup>
+</Project>
+```
+
+**Directory.Packages.props**
```
nkolev92, line 152: Are we missing out on the whole `disable package pinning for the project` option?

proposed/2021/CentralPackageManagementTransitive.md
lazy8: Thanks again everybody for tackling this issue! It's really not trivial to get this right.

To give a (hopefully) easy-to-understand example of what `unification` could mean:

- I have several plug-ins projects meant for a common main App.
- They are built separately, but need to run together in the same app domain.
- Plug-Ins will depend on App, but they are not allowed to depend on each other.
- Each plug-in might use any external nuget package reference it likes.

Transitive pinning would give me a central file where to synchronize the (transitive) dependencies of all plug-ins to common versions. This can then be collected into the Main App's `assemblyBinding` version override list, and I am done.

To find conflicting version issues before they hit at runtime, I can do a simple run of `dotnet list package <PROJ> --include-transitive` over all plug-ins and the main App, and compare the resolved versions. If there is a conflict, I can resolve it centrally to a common version.

### Another advantage of TP against elevation: Detect and remove no-longer-needed dependencies 
Another (I think) important point where elevation of transitive packages becomes unmanageable, is when you think about how your App evolves over time:

If I have to elevate conflicting packages to top-level dependencies, I wouldn't know how to detect when it becomes unneeded in a later version of that 3rd-party-package. Thus, I would *never be able to get rid of it again*. With transitive pinning, it would disappear automatically from my dependency graph and the compile folder, and I can check for that and remove the pinned version specification at my convenience.

### Minor point: Managing the stability of your app's full DLL set
When transitive pinning is used together with lock-files and a transitive company nuget mirror, I can conveniently manage and _guarantee_ DLL stability between re-builds (e.g. on CI infrastructure), which I think is vital against accidental or malicious changes (= new nuget version releases) in the supply-chain-graph.

lazy8:
Thanks again everybody for tackling this issue! It's really not trivial to get this right.

To give a (hopefully) easy-to-understand example of what `unification` could mean:

- I have several plug-ins projects meant for a common main App.
- They are built separately, but need to run together in the same app domain.
- Plug-Ins will depend on App, but they are not allowed to depend on each other.
- Each plug-in might use any external nuget package reference it likes.

Transitive pinning would give me a central file where to synchronize the (transitive) dependencies of all plug-ins to common versions. This can then be collected into the Main App's `assemblyBinding` version override list, and I am done.

To find conflicting version issues before they hit at runtime, I can do a simple run of `dotnet list package <PROJ> --include-transitive` over all plug-ins and the main App, and compare the resolved versions. If there is a conflict, I can resolve it centrally to a common version.

### Another advantage of TP against elevation: Detect and remove no-longer-needed dependencies 
Another (I think) important point where elevation of transitive packages becomes unmanageable, is when you think about how your App evolves over time:

If I have to elevate conflicting packages to top-level dependencies, I wouldn't know how to detect when it becomes unneeded in a later version of that 3rd-party-package. Thus, I would *never be able to get rid of it again*. With transitive pinning, it would disappear automatically from my dependency graph and the compile folder, and I can check for that and remove the pinned version specification at my convenience.

### Minor point: Managing the stability of your app's full DLL set
When transitive pinning is used together with lock-files and a transitive company nuget mirror, I can conveniently manage and _guarantee_ DLL stability between re-builds (e.g. on CI infrastructure), which I think is vital against accidental or malicious changes (= new nuget version releases) in the supply-chain-graph.

HugoLemieux:
Here are my answers:
1. Are you confused about the transitive behavior of what "pinning" a reference does today? (i.e. promotes a transitive dependency to a top-level dependency implicitly).
No and I think it is good that this also flows to Pack.
2. Is it important to you to be able to optionally remove a single PackageReference from the same "pinning" behavior?
No but it could be nice to have the option.
3. Would you expect IncludeAssets, ExcludeAssets, or PrivateAssets to be respected when overriding a dependency?
Yes and I think it is good to respect them.
4. Do you understand that this feature is primarily designed for usability & maintainability of your packages in a central location? It does not guarantee any performance benefits with as your projects scale.
Yes.

Great work @JonDouglas for having analyzed and documented these impacts.

nicolascotton:
1.	Are you confused about the transitive behavior of what "pinning" a reference does today? (i.e. promotes a transitive dependency to a top-level dependency implicitly).
**No, I’m not.**
2.	Is it important to you to be able to optionally remove a single PackageReference from the same "pinning" behavior?
**Not that much.**
3.	Would you expect IncludeAssets, ExcludeAssets, or PrivateAssets to be respected when overriding a dependency?
**I would.**
4.	Do you understand that this feature is primarily designed for usability & maintainability of your packages in a central location? It does not guarantee any performance benefits with as your projects scale.
**Yes.**

jimmylewis:
1. Are you confused about the transitive behavior of what "pinning" a reference does today? (i.e. promotes a transitive dependency to a top-level dependency implicitly).
**No.**
2. Is it important to you to be able to optionally remove a single PackageReference from the same "pinning" behavior?
**No, I have not yet come up with a case where I would want a specific PackageReference to behave differently from other PackageReferences in the same project (this just seems to create extra confusion).  Currently the proposals for CPVM allow me to:**
    * Toggle CPVM on and off for the project 
    * Toggle (opt-out) transitive pinning for the project
    * Use `<PackageVersion Update=...>` in the project to override the value in Directory.Package.props if I need a different version pinned on a one-off basis.
**This is sufficient flexibility for any scenario I've encountered thus far.**
3. Would you expect IncludeAssets, ExcludeAssets, or PrivateAssets to be respected when overriding a dependency?
**Yes, if the transitive dependency has any of these set, I would expect the promoted dependency to retain those settings.  If I manually promote, then my new entry would specify how those assets should be handled.**
4. Do you understand that this feature is primarily designed for usability & maintainability of your packages in a central location? It does not guarantee any performance benefits with as your projects scale.
**Yes, I desire this feature for the maintainability aspects.  I would even trade off some build performance for improved maintainability of large dependency graphs.**

czdietrich:
Does this mean that CPVM is then no longer a preview feature and becomes an official feature?

czdietrich:
And another question I'd to ask. One of the main reasons to use CPVM is to ensure that all (top level) projects use the same version of a DLL, which on the other hand is often needed for installers, since all binaries usually end up in the same installation / program files directory.

CPVM currently only handles `PackageReferences` but does not care about `FrameworkReferences`. So if for example you use `Microsoft.Extensions.Logging` as `PackageReference` in one project and `Microsoft.AspNetCore.App` as `FrameworkReference` in another project which brings `Microsoft.Extensions.Logging.dll` as direct dependency, you can still end up with 2 different versions of the same DLL for two projects. 
And even if you synchronize the version of the package and of the framework reference, you will still end up with two different DLLs but with the same version, since the DLL of the framework reference is compiled `Ready2Run` while the NuGet reference is not.

Currently the only workaround I am aware of is to use the same framework references for all top level projects.

I'm not sure whether this issue belongs to CPVM, but maybe it should be considered to belong there. Is there any idea or plan to somehow add a clean solution for this issue?

Thanks in advance.

proposed/2021/CentralPackageManagementTransitive.md
```diff
@@ -0,0 +1,201 @@
...
+
+3. How to opt out - If transitive version overrides are automatic when a `PackageVersion` is specified, there will need to be a way for a user to opt out of this behavior for one or more `PackageReference` items.  In previous versions of NuGet, a user could only opt into transitive version overrides by _adding_ a `PackageReference` item.  But to opt out of newer behavior, they will not have a `PackageReference` item to remove.  An MSBuild property could be used to opt an entire project out of transitive version overrides but that would not address a single `PackageReference` item.
```
caghand, line 129: I don't think that a project-wide MSBuild property is needed...  And it can always be added later, if there is demand.  I believe that the overwhelming majority of users will want to keep transitive version pinning, with some rare exceptions for some individual packages.

proposed/2021/CentralPackageManagementTransitive.md
caghand: I don't think that a project-wide MSBuild property is needed... And it can always be added later, if there is demand. I believe that the overwhelming majority of users will want to keep transitive version pinning, with rare exceptions for some individual packages in individual projects.
Note that I am not an expert on this, so this is a subjective opinion.

proposed/2021/CentralPackageManagementTransitive.md
caghand: Very interesting suggestion!  Yes, if you are intent on using the `Update` attribute to override the package version, then it makes sense to put the `Remove` attribute to good use, in a similar way.

proposed/2021/CentralPackageManagementTransitive.md
nkolev92: When it comes to features that can be used to tweak things, you'd normally go "outwards". 
Meaning you start with the least specific version and only go to the more specific ones when necessary. 

Regardless, my comment is primarily about calling out the reasonable options, and I think the project version one that has merit.

jeffkl (approved):

PR #11492, opened 2022-01-08
Update package events spec

nkolev92:
Updated based on discussions in https://github.com/NuGet/NuGet.Client/pull/4359. 

cc @zivkan

### DIFF
--- a/proposed/2021/ProjectUpdateEventsInVisualStudio.md
+++ b/proposed/2021/ProjectUpdateEventsInVisualStudio.md
@@ -46,7 +46,7 @@ The API and the shape is fairly self explanatory.
 
 using System.Collections.Generic;
 
-namespace NuGet.SolutionRestoreManager
+namespace NuGet.VisualStudio
 {
     /// <summary>
     /// NuGet project update events.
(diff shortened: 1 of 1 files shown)


nkolev92:
Can I please get a review @zivkan 
The spec matches what was implemented.

zivkan (approved):

proposed/2021/ProjectUpdateEventsInVisualStudio.md
```diff
@@ -42,28 +44,25 @@ The API and the shape is fairly self explanatory.
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
 
 namespace NuGet.SolutionRestoreManager
```
zivkan: It's probably too late now since the implementation is already merged and this is just updating the spec,  but I'm not sure that I feel `NuGet.SolutionRestoreManager` is the right namespace. I think it belongs in `NuGet.VisualStudio` with all the other "useful for everyone, not just project systems" APIs.  But as I said, it's probably too late now.

nkolev92: There's no adoption yet and it can be tweaked. 

I can change it.

zivkan: I wasn't aware it hadn't been adopted yet. I thought since it was created for a partner team, they might have adopted it as soon as we inserted it into an internal preview of VS.

Anyway, as stated, I feel it's useful for many extensions, not just project systems, so as part of API design/review, my feedback is that SolutionRestoreManager doesn't feel like the ideal place. But since all our MEF APIs are in one assembly now, it doesn't make a huge difference, really.

Issue #11457, opened 2021-12-16
[DCR]: NuGet allows version conflict resolution to cross major version boundaries potentially resulting in runtime errors


iconics-milan:
### NuGet Product(s) Affected

Visual Studio Package Management UI

### Current Behavior

Per semver, new major versions are allowed to introduce breaking changes in their API.
Yet NuGet will not inform of any potential problems and will download packages with a higher major version than specified by a dependency, unless the dependency explicitly specifies a maximum version in its PackageReference.

To demonstrate using purely MS packages:
1. in Visual Studio (currently using 2019), create a new project using the template "ASP.NET Core Empty"
2. in the wizard, check the box "Run on .NET Framework"
3. (optionally) upgrade the .NET Framework and all nugets in the project to their latest available versions
4. add the nuget Microsoft.Azure.Devices.Client 1.39.0 to the project
When run, the app immediately crashes. Now continue:
5. Remove Microsoft.Azure.Devices.Client again
6. Add the nuget Microsoft.Azure.NotificationHubs 4.1.0
Run again. It will not crash, but won't run either. After disabling "Just my code" in Visual Studio, you'll notice exceptions somewhere deep, preventing the app from running properly.

These are just two examples that are easy to reproduce, but we found more during our recent package upgrade. At this point, we are unable to tell confidently whether our application is fine, or if some new issue will appear when our app hits an uncommon execution path.

I know that NuGet would be able to warn about the potential incompatibilities if every PackageReference in every nuget was version locked to the current major version. But since MS does not do that, nobody will.

### Desired Behavior

At least, NuGet should print a warning.
This means: if any PackageReference does not contain an explicit maximum version, then assume that it is only compatible with the current major version, as specified by the minimum version of the PackageReference. (If the package author did include a maximum version, then no change in behavior is needed)
There could also be an option to turn this warning into an error, so that developers cannot ignore this.

Ideally though, NuGet should be able to suggest ways how to resolve these issues. If applied to the case described previously, NuGet should suggest to downgrade Microsoft.Azure.Devices.Client or Microsoft.Azure.NotificationHubs, and also determine to which version. Debian's apt system can do something like this.

### Additional Context

This isn't new, it is also known as the Diamond dependency problem and discussed here: https://codeblog.jonskeet.uk/2019/10/25/options-for-nets-versioning-issues/
That blog suggests other solutions to this problems, but printing a warning is I think the least NuGet can do, and would be sufficient in many cases.

[...]

PR #11549, opened 2022-02-02
dotnet audit & dotnet audit fix for NuGet packages.

JonDouglas:
Introduction to an auditing experience in the .NET CLI that allows one to view known vulnerabilities, deprecations, and outdated packages from a glance and a subcommand to quickly fix and resolve issues to the best of its ability or provide a message to manually intervene.

This proposal introduces two new commands to .NET CLI known as `dotnet audit` and `dotnet audit fix`.

[Rendered Spec](https://github.com/NuGet/Home/blob/0310d373a1c20240863ccbb0574aae481d45d873/proposed/2021/DotNetAudit.md)

Thank You 🎉


proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+![dotnet audit](/meta/resources/dotnetaudit/dotnetaudit.png)
+
+```
+$ dotnet audit
+
+Fetching package metadata from: 'https://api.nuget.org/v3/index.json'
+Loaded 23 security advisories from 'https://api.nuget.org/v3/index.json'
+Scanning ContosoUniversity.sln (36 NuGet dependencies)
+
+error: Vulnernable packages found!
```
loic-sharma:
```suggestion
error: Vulnerable packages found!
```

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+##### dotnet audit fix Exit Codes
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities, deprecations, or outdated packages were found *or* remediation was able to fix all issues.
+- 1 - The command will exit with a 1 exit code if a vulnerability, deprecation, or outdated package was found *and* remediation is not able to fix all issues.
+- 2 - The command will exit with a 2 exit code if it unexpectedly failed.
+- 3 - The command will exit with a 3 exit code if an unsupported project is detected.
+
+#### Vulnerabilities
+
+When vulnerable packages are detected, an error is thrown by default.
```
loic-sharma, line 222: Could you explain why an error is thrown instead of a warning for vulnerabilities?

A reason I ask is because vulnerabilities in the Javascript ecosystem often noise or very low impact.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+#### dotnet audit fix
+
+The `dotnet audit fix` command will provide a remediation that is calculated with an implicit `dotnet audit` to then apply directly to a resulting package graph. It can add packages, remove packages, and update packages depending on the problem it's attempting to resolve. It does not take into consideration downgrading to a compatible version if a higher one has already been specified.
+
+![dotnet audit fix](/meta/resources/dotnetaudit/dotnetauditfix.png)
+
+```
+$ dotnet audit fix
+
+Fixing vulnernable packages in ContosoUniversity.sln
```
loic-sharma:
```suggestion
Fixing vulnerable packages in ContosoUniversity.sln
```

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+              "advisoryUrl": "https://github.com/advisories/GHSA-mv2r-q4g5-j8q5"
+            }
+          ]
+        }
+      ]
+    },
+    "deprecations": {
+      "legacy": "1",
+      "critical-bugs": "0",
+      "other": "0",
```
loic-sharma, line 119: Should we remove these fields? From my understanding, you can get the same information by inspecting the `dependencies`'s `reason` property`.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+
+Run 'dotnet audit fix' to fix them.
+
+warning: Outdated packages found!
+[net5.0]:
+Top-level Package						            Resolved	Latest
+> anurse.testing.TestDeprecatedPackage	            1.0.0		2.0.0
+> UmbracoForms							            8.4.1	    8.7.1
+
+Found 2 top-level outdated package(s) in 36 scanned packages.
```
loic-sharma, line 73: Would this report transitive packages that are outdated? Would users have to hoist outdated transitive dependencies to update them to the latest version? That may be a frustrating experience for customers.

Perhaps we should only report outdated top-level dependencies. If so, consider adding a note on that limitation somewhere.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+      "low": "0",
+      "moderate": "2",
+      "high": "0",
+      "critical": "0",
+      "dependencies": [
+        {
+          "name": "UmbracoForms",
+          "requestedVersion": "8.4.1",
+          "resolvedVersion": "8.4.1",
+          "vulnerabilitySeverity": "Moderate",
```
loic-sharma, line 103: Line 109 uses property name `severity`. Should this line align to that?

```suggestion
          "severity": "Moderate",
```

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+To get a detailed audit report in a JSON format.
+
+```
+{
+  "audit": {
+    "vulnerabilities": {
+      "low": "0",
+      "moderate": "2",
+      "high": "0",
+      "critical": "0",
```
loic-sharma, line 97: Should we remove these fields? From my understand you can get the same information by inspecting the `severity` properties?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+      "dependencies": [
+        {
+          "name": "UmbracoForms",
+          "requestedVersion": "8.4.1",
+          "resolvedVersion": "8.4.1",
+          "vulnerabilitySeverity": "Moderate",
+          "advisoryUrl": "https://github.com/advisories/GHSA-8m73-w2r2-6xxj",
+          "transitiveDependencies": [
```
loic-sharma, line 105: What would this look like if a transitive dependency has a vulnerability but all top-level dependencies are OK?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should the command be named `audit` or `check`?
+  - `audit` is the more consistent name for package manager tooling with other ecosystems.
+  - `check` is the more consistent name with dotnet CLI.
+- Should this command only audit `vulnerabilities`? Or should it audit & be proactive as to `vulnerabilities`, `deprecations`, and `outdated` packages?
+- How much information should be present in the --json output?
+
+## Future Possibilities
```
loic-sharma, line 290: Could we make `dotnet pack` warn if one of your dependencies have a version range that accepts a vulnerability?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+The `dotnet audit` command will scan your project and solution's dependencies and look for known vulnerabilities, deprecations, and outdated packages. If any vulnerabilities, deprecations, or outdated packages are found, a remediation will be calculated to then `fix` which will be applied to the resulting package graph.
+
+Example:
+
+![dotnet audit](/meta/resources/dotnetaudit/dotnetaudit.png)
+
+```
+$ dotnet audit
+
+Fetching package metadata from: 'https://api.nuget.org/v3/index.json'
```
kartheekp-ms, line 42: I think the command should consider `package source mapping configuration` if available to decide which sources to fetch metadata from.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+- Should the command be named `audit` or `check`?
+  - `audit` is the more consistent name for package manager tooling with other ecosystems.
+  - `check` is the more consistent name with dotnet CLI.
+- Should this command only audit `vulnerabilities`? Or should it audit & be proactive as to `vulnerabilities`, `deprecations`, and `outdated` packages?
+- How much information should be present in the --json output?
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+- `dotnet audit` can be run on every `restore` which can throw warnings or errors to the user to take action against a vulnerable, deprecated, or outdated software supply chain.
```
erdembayar, line 293: Could customer suppress warnings/errors?  It could be low impact noise. Also, even package has vulnerability and there is no alternative then customer stuck with it and took other mitigation steps (isolate network from outside world), then keep throwing error could discourage further usage.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+
+##### Endpoints
+
+NuGet will use existing endpoints to optimize the speed of audit results.
+
+- [Deprecation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation)
+- [Vulnerability](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#vulnerabilities)
+- Outdated - No existing endpoint, will need to call a source.
+
+##### dotnet audit --json
```
nkolev92, line 86: If we're doing this, should it match the dotnet list package design?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+#### CLI Usage
+
+```
+dotnet audit --help
+dotnet audit [<PROJECT>|<SOLUTION>|<Directory.Packages.props>] [-v|--verbosity <LEVEL>] [--json] [--interactive]
+```
+
+```
+dotnet audit fix --help
+dotnet audit fix [<PROJECT>|<SOLUTION>|<Directory.Packages.props>] [-v|--verbosity <LEVEL>] [--dry-run] [--json] [--interactive]
```
kartheekp-ms, line 241: Consider the following scenario.
1. Customer runs `dotnet audit --json`
2. Few violations are reported
3. Customer runs `dotnet audit fix` command

This may be an implementation detail how `dotnet audit fix` command works but it could be great to add an option to optionally pass `--audited-json` file as input to the `fix` sub command so that it can act on those findings instead of finding the problems again.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+##### dotnet audit Exit Codes
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities, deprecations, or outdated packages were found.
+- 1 - The command will exit with a 1 exit code if a vulnerability, deprecation, or outdated package was found.
+- 2 - The command will exit with a 2 exit code if it unexpectedly failed.
+- 3 - The command will exit with a 3 exit code if an unsupported project is detected.
+
+#### dotnet audit fix
+
+The `dotnet audit fix` command will provide a remediation that is calculated with an implicit `dotnet audit` to then apply directly to a resulting package graph. It can add packages, remove packages, and update packages depending on the problem it's attempting to resolve. It does not take into consideration downgrading to a compatible version if a higher one has already been specified.
```
erdembayar, line 164: I assume it fails if there is no alternative, right?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+  
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should the command be named `audit` or `check`?
+  - `audit` is the more consistent name for package manager tooling with other ecosystems.
+  - `check` is the more consistent name with dotnet CLI.
+- Should this command only audit `vulnerabilities`? Or should it audit & be proactive as to `vulnerabilities`, `deprecations`, and `outdated` packages?
```
nkolev92, line 287: I think vulnerabilities and deprecations should be accounted for. 

Outdated packages is the one that gives me a pause (at least by default), due to the fact updating packages to the latest is not necessarily always the right solution.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
+# dotnet audit & dotnet audit fix
+
+- [Jon Douglas](https://github.com/JonDouglas/)
+- Start Date (2021-04-01)
+- [#8087](https://github.com/NuGet/Home/issues/8087)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+`dotnet audit` & `dotnet audit fix` helps you find, fix, and monitor known security vulnerabilities, deprecated packages, and outdated versions in your .NET projects & solutions.
```
kartheekp-ms, line 10: It may be crazy but finding unused packages would be awesome.

loic-sharma: Here's some a community project that does this: https://github.com/spectresystems/snitch

FYI this is a hard problem since a package may be used through reflection. But .NET trimming effort aligns well with this idea.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+  }
+}
+```
+
+##### dotnet audit Exit Codes
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities, deprecations, or outdated packages were found.
+- 1 - The command will exit with a 1 exit code if a vulnerability, deprecation, or outdated package was found.
+- 2 - The command will exit with a 2 exit code if it unexpectedly failed.
+- 3 - The command will exit with a 3 exit code if an unsupported project is detected.
```
loic-sharma, line 160: What's the value of splitting exit codes 2 and 3? It may be simpler to have a single exit code for "NuGet was unable to determine if there vulnerabilities, deprecations, or outdated packages. This may due to unexpected failure or unsupported projects"

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+- [NuGet Defense](https://github.com/digitalcoyote/NuGetDefense)
+  
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should the command be named `audit` or `check`?
+  - `audit` is the more consistent name for package manager tooling with other ecosystems.
+  - `check` is the more consistent name with dotnet CLI.
```
dominoFire, line 286: check is taken in `dotnet format --check` command

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+```
+dotnet audit fix --help
+dotnet audit fix [<PROJECT>|<SOLUTION>|<Directory.Packages.props>] [-v|--verbosity <LEVEL>] [--dry-run] [--json] [--interactive]
+```
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
```
dominoFire, line 246: Would be nice to outline how `dotnet audit fix` will fix outdated, deprecated, vulnerable packages:

- Pick the latest stable?
- Pick the closest next-stable, non-flagged package?
- Pick the alternate package in deprecated package?
- Should it include preview versions?

dominoFire (commented):
Thanks! Some comments.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+- [Jon Douglas](https://github.com/JonDouglas/)
+- Start Date (2021-04-01)
+- [#8087](https://github.com/NuGet/Home/issues/8087)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+`dotnet audit` & `dotnet audit fix` helps you find, fix, and monitor known security vulnerabilities, deprecated packages, and outdated versions in your .NET projects & solutions.
+
+## Motivation
```
zivkan, line 12: In my time in the NuGet team, there have been a few times when we've been looking at some old feature and trying to understand WHY it was done a particular way. So, in the interest of helping people in the future understand, can you please add something (anywhere in the document, not necessarily here) about why we're not just adding a `--fix` argument to `dotnet list package --deprecated --include-transitive`?

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+#### dotnet audit
+
+The `dotnet audit` command will scan your project and solution's dependencies and look for known vulnerabilities, deprecations, and outdated packages. If any vulnerabilities, deprecations, or outdated packages are found, a remediation will be calculated to then `fix` which will be applied to the resulting package graph.
+
+Example:
+
+![dotnet audit](/meta/resources/dotnetaudit/dotnetaudit.png)
```
zivkan, line 37: personal comment, I don't find the screenshot + text below useful. It just makes the document longer for me.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+[net5.0]:
+Top-level Package	    Requested	Resolved	Severity	Advisory URL
+> UmbracoForms		    8.4.1		8.4.1		Moderate	https://github.com/advisories/GHSA-8m73-w2r2-6xxj
+
+Transitive Package		Resolved	Severity	Advisory URL
+> Microsoft.Data.OData	5.2.0		Moderate	https://github.com/advisories/GHSA-mv2r-q4g5-j8q5
+
+Found 1 top-level Moderate severity vulnerability & 1 transtive Moderate severity vulnerability package(s) in 36 scanned packages.
+
+Run 'dotnet audit fix' to fix them.
```
kartheekp-ms, line 56: How about a scenario where top-level package has no vulnerability but a transitive one has a warning? AFAIK, we can't update the version of transitive dependency unless we add it as top-level dependency.

@zivkan answered this question in an offline review that, `making it a top-level dependency is the design for how to upgrade transitive packages.`

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+##### dotnet audit Exit Codes
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities, deprecations, or outdated packages were found.
+- 1 - The command will exit with a 1 exit code if a vulnerability, deprecation, or outdated package was found.
+- 2 - The command will exit with a 2 exit code if it unexpectedly failed.
+- 3 - The command will exit with a 3 exit code if an unsupported project is detected.
+
+#### dotnet audit fix
```
zivkan, line 162: I'm sure we have more duplicates of the same request, but `dotnet audit fix` just isn't going to work for a whole lot of customers, including ASP.NET Core customers using an LTS (not latest) version: https://github.com/NuGet/Home/issues/11457

Unless we're planning on implementing `dotnet audit fix` after implementing the above feature, I think this spec should document this known limitation.

proposed/2021/DotNetAudit.md
zivkan: Maybe we need to document this under a better heading, but we've also documented it publicly: https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#cousin-dependencies

proposed/2021/DotNetAudit.md
kartheekp-ms: Will other feeds such as `AzDO` have this vulnerabilities, deprecations, or outdated information?

proposed/2021/DotNetAudit.md
JonDouglas: I'll add more rationale in the proposal. I believe with our lower amount of known vulnerabilities, high importance of the types of vulnerabilities issued for .NET, and the nature of shifting left, these should be treated as errors for now.

https://overreacted.io/npm-audit-broken-by-design/

proposed/2021/DotNetAudit.md
JonDouglas: I don't think so. We want to use these as high level counts for the known reason category per dependency. Is there a way we can associate a number with the reason field?

proposed/2021/DotNetAudit.md
JonDouglas: It currently follows the outdated experience which doesn't include transitives. I'm also not so sure what value seeing outdated transitives would have unless one wanted to promote those explicitly. Will have to circle back here.

proposed/2021/DotNetAudit.md
JonDouglas: These are to aggregate the number from the audit. If there's another way to do that, we can. Wanted to give a summary of json at the top so it's clear where attention should be spent and not looking at each node.

proposed/2021/DotNetAudit.md
JonDouglas: Yeah if pack can resolve at the time of packing we should definitely do that.

proposed/2021/DotNetAudit.md
JonDouglas: Yes I think we can figure out something. I am not so sure what arguments that might be right now as this experience will require an internet connection or a local cache. It might be something like `--offline`, `--level high`, or `--features vulnerability`.

proposed/2021/DotNetAudit.md
JonDouglas: Yes, given that design is still open and being iterated on, whatever comes from it should be reflected here for consistency and used.

proposed/2021/DotNetAudit.md
JonDouglas: I do not believe they support these features. The MVP is designed around the central registry.

https://developercommunity.visualstudio.com/t/add-option-to-deprecate-nuget-package-in-azure-art/1316798

proposed/2021/DotNetAudit.md
JonDouglas: The `dotnet audit fix` command does an implicit `dotnet audit`. Thus I don't think this is needed. Perhaps we can do some caching to make it faster, but we should probably call out to the net often to ensure nothing changed.

proposed/2021/DotNetAudit.md
JonDouglas: If there's no alternative found, the tool would likely throw a manual review for the user to remove the dependency or ignore it in the future.

proposed/2021/DotNetAudit.md
JonDouglas: Makes sense. I'd like to see what other thoughts are about this 3-in-1 tool and then can iterate further.

proposed/2021/DotNetAudit.md
JonDouglas: Visual Studio does this today. I think it have much more context & tooling to do so.

image

Would be cool for sure! Not sure if feasible until a technical pass.

proposed/2021/DotNetAudit.md
JonDouglas: Personally I think unsupported vs. failure are different scenarios. Why have someone guess?

proposed/2021/DotNetAudit.md
JonDouglas: I'll add a flowchart concept to the proposal soon.

proposed/2021/DotNetAudit.md
JonDouglas: I hope to answer this further with a flowchart or similar. Thanks for bringing it up.

proposed/2021/DotNetAudit.md
JonDouglas: Sure thing.

proposed/2021/DotNetAudit.md
JonDouglas: This is quite a hard challenge indeed. I think we can do our best job designing this feature for the majority of packages that aren't tied directly to a .NET version in NuGet. Even then, there is value for when vulnerabilities or deprecations are found.

i.e. https://www.nuget.org/packages/System.DirectoryServices.Protocols/ may have vulnerabilities in NET5 & prompt a user to migrate to NET6 or higher. I don't think this tool will be the panacea of modernization, but it can help bring awareness to the benefit of doing so or even point users to the respective modernization tools if needed. 

If we do the work in the issue that allows for such functionality, we can port it over to this tool / try our best to detect it from the package lists to be smarter in these cases.

I'll add similar notes in the proposal.

JonDouglas:
Thank you everyone so far for the reviews!

PatGet:
Any idea how that would work with a nuget feed on azure devops server for example?

proposed/2021/DotNetAudit.md
nkolev92: I don't think the extra comment about the packages helps tbh. 
We should have just have a a `Sources used` statement and capture everything.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+
+Example:
+
+![dotnet audit](/meta/resources/dotnetaudit/dotnetaudit.png)
+
+```
+$ dotnet audit
+
+Fetching package metadata from: 'https://api.nuget.org/v3/index.json'
+Loaded 23 security advisories from 'https://api.nuget.org/v3/index.json'
```
nkolev92, line 43: What does 23 advisories mean in this case? 
23 packages in our graph had advisories? Or 23 packages were inspected? Or the source had 23 advisories.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+
+warning: Outdated packages found!
+[net5.0]:
+Top-level Package						            Resolved	Latest
+> anurse.testing.TestDeprecatedPackage	            1.0.0		2.0.0
+> UmbracoForms							            8.4.1	    8.7.1
+
+Found 2 top-level outdated package(s) in 36 scanned packages.
+
+Run 'dotnet audit fix' to fix them.
```
nkolev92, line 75: Say there are 10 packages with advisories. 

Does fix update them one by one? (this can be super slow, something we don't do in VS). 

What if the updates fial to bring us to a successful restore state?

I think the answer is we our best, and if we fail, say conflicts xyz and add details about potential actions the customer can take. 

Note that because of the graph flattening, failures to upgrade are more common compared to something like NPM.

proposed/2021/DotNetAudit.md
```diff
@@ -0,0 +1,296 @@
...
+```
+
+##### Endpoints
+
+NuGet will use existing endpoints to optimize the speed of audit results.
+
+- [Deprecation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation)
```
nkolev92, line 82: Does audit check nuget.org if nuget.org is *not* part of the list? Probably should call it out.

nkolev92 (commented):
A few comments I've been meaning to add for a while. They're largely stylistic, but things that I think are valuable to be called out.

GrabYourPitchforks:
Thanks Jon for getting this up! I have a few comments and questions.

I think the audit command should flag obsolete packages by default. This is somewhat implicit in the name: if you're auditing your code base, you care about being in compliance with some sort of application health policy. An outdated package has been explicitly marked by the maintainer as "I have abandoned this and will no longer provide actionable package health data," and such a statement is absolutely relevant to people dealing with compliance enforcement.

What about people who need to suppress warnings? Let's say that for whatever reason I absolutely must use vulnerable package _Foo.1.2.3.4_ - not any other version. I have mitigated my own risk by reviewing every single line of code involved in my app's data processing and I have high confidence that reported vulnerabilities do not apply to my use cases. What would the recommendation be here: put some _DoNotAudit_ flag in packages.config for this particular reference, have the audit command continue to warn but make some external reporting tool responsible for implementing the idea of "baselining" warnings, stop using the vulnerable package and copy the relevant code into my own project, or something else?

Similarly, when you implement the audit fix, you may need to introduce the concept of a "weak" package dependency, which can be thought of as a version redirect without the direct dependency. For example, let's say my app depends directly on _DirectDependency.2.0.0.0_, which depends on _TransitiveDependency.1.0.0.0_. There's a vulnerability in the transitive dependency, fixed in v1.0.1.0. The fix process could make my app take a weak dependency (not a direct dependency!) on _TransitiveDependency.1.0.1.0_. The behavior would be that if _TransitiveDependency_ \<1.0.1.0 appears anywhere in my package dependency graph as a strong dependency, it gets updated to 1.0.1.0; but if _TransitiveDependency_ does not appear anywhere in my package dependency graph, it does not ever get deployed alongside my app. This allows for a future where _DirectDependency.2.2.0.0_ completely drops its reference to _TransitiveDependency_. When my app eventually upgrades to _DirectDependency.2.2.0.0_, there remains no strong dependency to _TransitiveDependency_ anywhere in the entire graph, and my app's "weak" dependency is completely ignored. Perhaps it's even auto-deleted from packages.config.

Finally, do you need to introduce the concept of a "fork" dependency? Let's say that my package _Fork_ is a fork of some other package _Base_. I fork the code, make changes, and republish under my own banner. My package does not reference _Base_. However, because my code is so tightly intertwined with _Base_, any vulnerabilities in that package should also be reflected in my package so that my package's consumers are alerted to the problem, so I publish my package with a special attribute saying "vulnerabilities in _Base.1.2.3.4_ also apply to this package." The big issue here is that consumers cannot self-remediate after learning of a vulnerability, since they'd need to wait for the _Fork_ author to publish a new package version. (Apps can't simply reference the patched version of _Base_, since presumably they were depending on some custom behavior of _Fork_.) The only benefit this would confer is that consuming apps could learn about the problem early and take proactive mitigation steps - perhaps by disabling affected features of their app - while waiting for an official _Fork_ patch.

JonDouglas:
Thanks for the review Levi, I'll try my best to answer everything but do know that not everything has an answer today.

> I think the audit command should flag obsolete packages by default. This is somewhat implicit in the name: if you're auditing your code base, you care about being in compliance with some sort of application health policy. An outdated package has been explicitly marked by the maintainer as "I have abandoned this and will no longer provide actionable package health data," and such a statement is absolutely relevant to people dealing with compliance enforcement.

This should be covered by `deprecated` packages. Is `obsolete` a different concept in your eyes? i.e. the package is marked `deprecated` because it is `legacy`. Or do you mean that `deprecated` should be promoted to an error experience like `vulnerabilities` instead?

> What about people who need to suppress warnings? Let's say that for whatever reason I absolutely must use vulnerable package Foo.1.2.3.4 - not any other version. I have mitigated my own risk by reviewing every single line of code involved in my app's data processing and I have high confidence that reported vulnerabilities do not apply to my use cases. What would the recommendation be here: put some DoNotAudit flag in packages.config for this particular reference, have the audit command continue to warn but make some external reporting tool responsible for implementing the idea of "baselining" warnings, stop using the vulnerable package and copy the relevant code into my own project, or something else?

Definitely have thought about a `<NoWarn>` equivalent for suppressing vulnerable package ids/versions or advisory IDs. The answer right now is...no idea what is right. There's [guidance that all known vulnerabilities](https://www.whitehouse.gov/briefing-room/statements-releases/2022/03/21/fact-sheet-act-now-to-protect-against-potential-cyberattacks/) need to be addressed regardless of severity(even if it doesn't affect your code) and similar guidance you mention to doing everything within your control. I think we should definitely consider a way to suppress this information. I don't have an answer right now outside of not doing it. It just doesn't seem like a common practice in other tools I've researched thus far and why I left it out for now.

> Similarly, when you implement the audit fix, you may need to introduce the concept of a "weak" package dependency, which can be thought of as a version redirect without the direct dependency. For example, let's say my app depends directly on DirectDependency.2.0.0.0, which depends on TransitiveDependency.1.0.0.0. There's a vulnerability in the transitive dependency, fixed in v1.0.1.0. The fix process could make my app take a weak dependency (not a direct dependency!) on TransitiveDependency.1.0.1.0. The behavior would be that if TransitiveDependency <1.0.1.0 appears anywhere in my package dependency graph as a strong dependency, it gets updated to 1.0.1.0; but if TransitiveDependency does not appear anywhere in my package dependency graph, it does not ever get deployed alongside my app. This allows for a future where DirectDependency.2.2.0.0 completely drops its reference to TransitiveDependency. When my app eventually upgrades to DirectDependency.2.2.0.0, there remains no strong dependency to TransitiveDependency anywhere in the entire graph, and my app's "weak" dependency is completely ignored. Perhaps it's even auto-deleted from packages.config.

Absolutely. We will have to consider all our options for resolving the graph. Even considering the potential features of a [package overrides](https://github.com/npm/rfcs/blob/main/accepted/0036-overrides.md) type feature to fix problems in the graph at the transitive layer.

> Finally, do you need to introduce the concept of a "fork" dependency? Let's say that my package Fork is a fork of some other package Base. I fork the code, make changes, and republish under my own banner. My package does not reference Base. However, because my code is so tightly intertwined with Base, any vulnerabilities in that package should also be reflected in my package so that my package's consumers are alerted to the problem, so I publish my package with a special attribute saying "vulnerabilities in Base.1.2.3.4 also apply to this package." The big issue here is that consumers cannot self-remediate after learning of a vulnerability, since they'd need to wait for the Fork author to publish a new package version. (Apps can't simply reference the patched version of Base, since presumably they were depending on some custom behavior of Fork.) The only benefit this would confer is that consuming apps could learn about the problem early and take proactive mitigation steps - perhaps by disabling affected features of their app - while waiting for an official Fork patch.

Although this is a great feature idea and has been mentioned a number of times to our binary package ecosystem vs. source based one, we will be looking towards warning consumers with as much information as possible for them to resolve issues in their projects via upgrading, replacing, and even removing dependencies. If we can't do anything, we will give them a notice that we can't help and they will have to advise based on the problem. At this time, we are not considering extra features such as monkey patching, but using what we have today and working on-top of that.

steinarox:
Others already mentioning allowlist of specific CVE's. As to running audit in a pipeline; i really like the way[ Trivy is setup](https://aquasecurity.github.io/trivy/v0.24.2/getting-started/cli/fs/), where you can say `--severity CRITICAL --exit-code 0`  This way i can run multiple scans in a pipeline,  ex HIGH,CRITICAL which will block the pipeline, and another run for lower severitys that will only warn. 

Another question is of the json output could support https://cyclonedx.org/ format. It seems like there is a lot of tools popping up that will use this format. if all tools can support the same format, then managing scanresults/sboms from multiple development environments will become easier. could even take a cyclonedx-bom file as input in the future to have a quick way to check an old build.

Frulfump:
How will this work with transitive vulnerabilities I didn't see it mentioned in the spec.
(Maybe covered by https://github.com/NuGet/Home/issues/13372?)
If they are fixed by promoting it to a direct dependency can a comment be added to the .csproj, Directory.Build.props or Directory.Packages.props so that we later can remove it when it's no longer required.

Frulfump:
> How will this work with transitive vulnerabilities I didn't see it mentioned in the spec.
> (Maybe covered by https://github.com/NuGet/Home/issues/13372?)
> If they are fixed by promoting it to a direct dependency can a comment be added to the .csproj, Directory.Build.props or Directory.Packages.props so that we later can remove it when it's no longer required.

This comment has been unanswered for 8 months so not stale just waiting for feedback.

Frulfump:
> 

See previous comment

zivkan:
The main goal of this spec, to be able to find and update packages with known vulnerabilities, is available through some slightly different mechanisms than what's proposed here.

We shipped ["dotnet package update --vulnerable" in the .NET 10.0.100 SDK](https://learn.microsoft.com/nuget/concepts/auditing-packages#feature-availability), which has basic upgrade logic, and the [NuGet MCP Server](https://learn.microsoft.com/nuget/concepts/nuget-mcp-server) has a "solver" tool which can do a more complex package graph analysis to resolve version conflicts.

That said, this proposal has some ideas where there are still gaps in the current tooling:

1. A `--dry-run` option for `dotnet package update`. I create a new tracking issue: https://github.com/NuGet/Home/issues/14870
2. An option to output the upgrade's results in json. There's an existing issue to add this to `dotnet package update`: https://github.com/NuGet/Home/issues/14360
3. Being able to upgrade packages whose versions are deprecated. I created a new tracking issue: https://github.com/NuGet/Home/issues/14871
4. Warning when using deprecated packages. Arguably the existing `dotnet package list --deprecated` command already does this, it's just that it doesn't report both vulnerable and deprecated packages in the same command, you'd need to run the command twice. There's also an existing request to add deprecated package warnings to restore, or have VS automatically warn, rather than needing to manually check: https://github.com/NuGet/Home/issues/12244

If you're interested in any of these features, please add a 👍 reaction to the original comment in the issue, to help us prioritize what's important to our customers.

With that, I think this design spec is effectively superseded. Thanks to @JonDouglas for the great work in writing it up in the first place, which informed us on making `dotnet package update`, and to everyone else who contributed to the discussion. If I've missed anything, please search if there's an existing issue to upvote, or create a new issue with information on your request.

Issue #10389, opened 2020-12-17
[Feature]: Add option to allow versions of transitive dependencies to be overridden


marcin-krystianc:
Recently we've noticed a [PR](https://github.com/NuGet/NuGet.Client/pull/3719) that disables the transitive pinning behaviour for CPVM.
It worries us because we started to rely on that feature (or rather on the unintentional side effect of that feature).

So the problem we are struggling with is the slowness of the restore operation. For large solutions that contain many projects and reference many packages the time needed for restore operation becomes unacceptable (see https://github.com/NuGet/Home/issues/10030 for details). As it is explained in the linked issue, the slowness is caused by the exponential nature of the graph walking algorithm.

We've discovered that enabling the CPVM for such solutions makes it to restore **much** quicker. It turns out that transitive pinning helps to reduce the impact of exponential nature of the graph walking algorithm. Because of the transitive pinning the graph walking needs to traverse the dependency graph of each centrally managed package only once. It happens only when a package node is visited from the top level node and it is skipped when visited from all other nodes (due to the ["Nearest wins"](https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#nearest-wins) rule).

I know that transitive pinning is a controversial feature. Some users want it but others really don't. But do you consider bringing it back and making it an opt-in feature as it was suggested in the comments of https://github.com/NuGet/Home/issues/10115 ? It is really important for us, so I'm happy to work on making this change if you are willing to accept it.

rconard:
@marcin-krystianc, this is our current plan. The actual implementation may be a little rough as we get the pieces in place, but we're working on it. 👍

jimmylewis:
Throwing my vote on this too, hoping to see the priority bumped up.  Honestly, I don't know where else to file it, but PackageReference has been completely unusable in my team's repo for the last several years due to the complex (and disorganized) nature of our dependencies.  CPVM with transitive dependencies was our hope to finally make the move back to standardized tooling for our dependency management in a way that would not be grossly inconvenient for our developers.

For context, my team depends on several components coming from several other teams each in their own repos which all publish to a variety of feeds (our upstream list is impressive).  This results in a lot of complex dependency conflicts.  The options to resolve those are to:
(a) add any conflicting dependencies as primary to each project in order to force resolution of the package version (which really sucks, because it comes up all the time in new ways, and really devaules PackageReference in general), 
[...]

PR #11569, opened 2022-02-07
Package Version Overrides for Central Package Management

JonDouglas:
When specifying package versions in a central location, developers want the to be able to override versions for a particular project. This is useful when a project is not directly tied to the rest of the repo and needs to use a newer version of a package. Additionally, some projects have a need to specify an older version of a package when targeting an older version of .NET Framework.

This proposal introduces the ability to override a centrally defined package version on a <PackageReference /> item in a well documented and supported way.

[Rendered Proposal](https://github.com/NuGet/Home/blob/7636cd25480e694d53aab29ed06204728e78574c/proposed/2022/version-overrides.md)

Thank You 🎉

### DIFF
--- a/proposed/2022/version-overrides.md
+++ b/proposed/2022/version-overrides.md
@@ -1,0 +1,121 @@
+# Title
+
+- [Jeff Kluge](https://github.com/jeffkl) & [Jon Douglas](https://github.com/jondouglas)
+- Start Date (2022-02-07)
+- [11516](https://github.com/NuGet/Home/issues/11516)
+- [4426](https://github.com/NuGet/NuGet.Client/pull/4426)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+When defining package versions in a central location, developers want the control to be able to override versions in various scenarios or even disable the ability to override versions for their product. In some cases, you may need to override a version for a particular project. With multiple definitions of the package version to be used, it can lead to undesired behaviors and make diagnosing any build errors quite difficult.
+
+This proposal introduces the ability to override any centrally defined package version to ensure there's a single source of truth for the overridden version.
+
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+When interviewing a number of developers who currently use the NuGet provided central package management feature, we found that the majority of them were frustrated with the lack of a way to override a package version. Many of these individuals would use the MSBuildSdks CentralPackageVersions as an alternative to give them the flexibility they needed in their repositories
+
+The concept of dependency resolution overrides exists in many ecosystems and for NuGet, we want to ensure that there is a formal way to override package versions in the case that:
+
+- There is a bug that is yet to be fixed in a transitive dependency in the project's graph such as awaiting a proper bugfix to be published and we can override with the last known good version.
+- Any security vulnerability has been identified in a transitive dependency, but you're not able to upgrade the direct dependency pulling it in but rather the last good version.
+- Only allowing a single copy of a given dependency in the tree to discourage version conflicts and define it in one place
+- Being able to override a transitive dependency that may bring in an undesired behavior for a specific version being used with another dependency.
+
+In these cases, rules would apply to all top-level and transitive dependencies in the sense that any override would apply anywhere in the dependency graph. This makes the override more powerful, and simplifies the implementation in the process. On the other side, this also can prose a risk in which an override may apply to packages that the user did not intend it to.
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+First a user opts-in to Central Package Management with a `Directory.Packages.props` file in their directory tree.
+
+A user would specify package versions in `Directory.Packages.props`:
+
+```
+<Project>
+  <ItemGroup>
+    <PackageVersion Include="PackageA" Version="1.0.0" />
+    <PackageVersion Include="PackageB" Version="2.0.0" />
+  </ItemGroup>
+<Project>
+```
+
+A project would override the version with the `VersionOverride` metadata on a `<PackageReference />` item:
+
+```
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="PackageA" VersionOverride="3.0.0" />
+  </ItemGroup>
+<Project>
+```
+
+In this case, the project's restore graph would resolve `PackageA` to version `3.0.0`. Any project that references it would also get that version and a user would be responsible for handling an unresolved conflicts.
+
+Finally, a repo owner should be able to disable the ability for developers to override package version. This would be used for instance if someone wanted to ensure that all package versions are unified. This would be possible by setting the MSBuild property `EnablePackageVersionOverride` to `false` in a project or import like `Directory.Build.props`:
+
+```
+<Project>
+  <PropertyGroup>
+    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
+  </PropertyGroup>
+<Project>
+```
+
+When this is disabled, specifying a `VersionOverride` on a `<PackageReference />` would result in a restore error indicating that he feature is disabled.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
+There could be concerns regarding designing this at the version level and also at the central level rather than a more holistic approach for any package, it's version, and anywhere it may be defined in a project. Since MSBuild essentially allows users to do anything they want, we want to make overriding a package version simplistic and well documented.
+
+Although there may be different approaches to the current design, this is one that has been tested in the field with many projects and the feedback has matched our expectations of doing the job to be done in question.
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+The current concept that exists in the MSBuildSdks's CentralPackageVersions project has met the needs of many developers around the world who take it on as a dependency and it has been noted many times that this feature is missing in NuGet's first-class support of central package management.
+
+Although there could be a more flushed out design that includes complete package IDs & their respective versions similar to other ecosystems, we are working with what we know teams need today and will build upon it in the future. 
+
+If for any reason we decide to do a similar "package override" feature where a developer can replace any package on the dependency graph with another, we will allow for different levels of control such as replacing the package ID, the package version, and perhaps other selective mutations for a package in a graph.
+
+## Prior Art
+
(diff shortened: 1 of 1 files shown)


proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+
+```xml
+<Project>
+  <PropertyGroup>
+    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
+  </PropertyGroup>
+<Project>
+```
+
+When this is disabled, specifying a `VersionOverride` on a `<PackageReference />` would result in a restore error indicating that he feature is disabled.
```
zivkan:
```suggestion
When this is disabled, specifying a `VersionOverride` on a `<PackageReference />` would result in a restore error indicating that the feature is disabled.
```

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+In these cases, a project can override the centrally defined package version.  However, if that project is part of a transitive graph with version conflicts, a user would still need to resolve those conflicts before restore would succeed.  The main point of this feature is to give a well documented and supported way of overriding a package version for a particular project, not to make graph resolution easier if package conflicts exist.  
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+First a user opts-in to Central Package Management with a `Directory.Packages.props` file in their directory tree.
```
zivkan: Given that `VersionOverride` only works when CPM is enabled, should we warning/error if `VersionOverride` is used when CPM is not used? 

I'm thinking specifically of a scenario where someone new to .NET copy pastes lines from existing csproj files into a new project they created themselves, without understanding NuGet's features. If they did that with the current spec (no warning/error in this scenario), they will get a very obscure warning/error message which will not be clear at all what the root cause is.

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+
+Although there could be a more fleshed out design that includes complete package IDs & their respective versions similar to other ecosystems, we are working with what we know teams need today and will build upon it in the future.   Since the specifications of packages and versions are MSBuild-based, we feel its better to stick with current constructs that Microsoft developer ecosystem customers are familiar with rather than inventing a new process that could lead to confusion.
+
+If for any reason we decide to do a similar "package override" feature where a developer can replace any package on the dependency graph with another, we will allow for different levels of control such as replacing the package ID, the package version, and perhaps other selective mutations for a package in a graph.
+
+Users can also just specify a `<PackageVersion Update="" />` in an import or use an MSBuild property that represents a package version to override what's used in a particular project.
```
jimmylewis: The `<PackageVersion Update="" />` works fine in the project file itself, as Directory.Packages.props is evaluated first (through the SDK.props chain).  It doesn't need to be in an import.

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+When this is disabled, specifying a `VersionOverride` on a `<PackageReference />` would result in a restore error indicating that the feature is disabled.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+## Drawbacks
+
```
jimmylewis, line 93: One drawback not cited: this proposed method requires an explicit PackageReference.  Contrast with the `<PackageVersion Update="" />` method, where the version is overridden only if the PackageReference exists in the graph.  This difference becomes evident with a transitive dependency: in the scenario `project -> A -> B`, if I want to override the version of B *only if it is in this project's graph*, the proposed method *forces me to include it in the graph*.  If a future version of A removes that dependency, then my override (per the proposal) would still have B in the graph.

proposed/2022/version-overrides.md
nkolev92: I agree with this. 

Being in different modes has been something people that were trying to migrate to the originjal version of CPVM hit a lot.

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+The current concept that exists in the MSBuildSdks's CentralPackageVersions project has met the needs of many developers around the world who take it on as a dependency and it has been noted many times that this feature is missing in NuGet's first-class support of central package management.
+
+Although there could be a more fleshed out design that includes complete package IDs & their respective versions similar to other ecosystems, we are working with what we know teams need today and will build upon it in the future.   Since the specifications of packages and versions are MSBuild-based, we feel its better to stick with current constructs that Microsoft developer ecosystem customers are familiar with rather than inventing a new process that could lead to confusion.
```
nkolev92:
```suggestion
Although there could be a more fleshed out design that includes complete package IDs & their respective versions similar to other ecosystems, we are working with what we know teams need today and will build upon it in the future. Since the specifications of packages and versions are MSBuild-based, we feel its better to stick with current constructs that Microsoft developer ecosystem customers are familiar with rather than inventing a new process that could lead to confusion.
```

proposed/2022/version-overrides.md
nkolev92: That's a very good point. 

I know relying on the Update is super difficult for tooling, so I wonder if this should be a known limitaiton of this design.

We should explicitly ask for feedback from the people that rely on transitive pinning, to gauge their thoughts on this limitation. 

I'm really not sure there's any other solution that doesn't rely on the Update functionality.

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+In this case, the project's restore graph would resolve `PackageA` to version `3.0.0`. Any project that references it would also get that version and a user would be responsible for handling an unresolved conflicts.
+
+Finally, a repo owner should be able to disable the ability for developers to override package version. This would be used for instance if someone wanted to ensure that all package versions are unified. This would be possible by setting the MSBuild property `EnablePackageVersionOverride` to `false` in a project or import like `Directory.Build.props`:
+
+```xml
+<Project>
+  <PropertyGroup>
+    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
```
joelverhagen: Can this be set in `Directory.Packages.props` too? This is a guess but I'm thinking it might be more intuitive to keep this setting with the overrides themselves rather than in `Directory.Build.props` which might otherwise not exist in a repo.

jimmylewis: As long as it's any file imported by the SDK or an explicit import, it should be fine.  

IMO it would be nice if the NuGet docs would standardize on putting all CPVM related properties in one place (i.e. Directory.Packages.props) instead of recommending some there and some in Directory.Build.props.  The two files have no relation to each other, other than that both are imported via the SDK.props, and centralizing all CPVM concerns in one place seems simpler to manage.

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
```
joelverhagen, line 132: This might be a good place to put mentions of migration tool, VS/UI enhancements, etc here.

proposed/2022/version-overrides.md
jimmylewis: > I'm really not sure there's any other solution that doesn't rely on the Update functionality.

A possible alternative would be to introduce a `<PackageVersionOverride>` item that could be specified in the project but would only be applied if a matching package existed in the restore graph.  But at that point, it has the complexity of the `<PackageVersion Update="" />` (i.e. I need both a `<PackageReference>` and a `<PackageVersionOverride>` to override a primary reference), or it adds complexity to the overall feature design (i.e. implementing both VersionOverride as metadata and PackageVersionOverride as a separate item).

I'm not sure if that's worth the it vs. keeping the simpler VersionOverride metadata.  Maybe documenting both the VersionOverride and the Update method is sufficient, calling out the tradeoffs for each?

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
+# Title
```
agr: Should there be a proper title here?

proposed/2022/version-overrides.md
```diff
@@ -0,0 +1,119 @@
...
+```xml
+<Project>
+  <ItemGroup>
+    <PackageVersion Include="PackageA" Version="1.0.0" />
+    <PackageVersion Include="PackageB" Version="2.0.0" />
+  </ItemGroup>
+<Project>
+```
+
+A project would override the version with the `VersionOverride` metadata on a `<PackageReference />` item:
```
agr: Why is a special attribute needed for the feature? Why having just

    <PackageReference Include="PackageA" Version="3.0.0" />

would not suffice for the override?

proposed/2022/version-overrides.md
jeffkl: I'll put a note in the spec to emit a warning when `VersionOverride` is ignored du to CPM being disabled.

proposed/2022/version-overrides.md
jeffkl: Thanks I keep forgetting that `Directory.Packages.props` is imported before the project.  I'll update this proposal.

proposed/2022/version-overrides.md
jeffkl: Thanks I'll update the proposal to call this out.  I'm personally hoping that VersionOverride is a last resort and therefore a rarity.  Ideally a repo would be able to unify on a single version.  

As discussed here, `<PackageVersion Update="" />` is a viable option and we won't be doing anything to stop that pattern.  We won't be able to detect if these are used incorrectly since its transparent to NuGet that they ran.  I think we'll even place this pattern in the docs for even more advanced use cases.  Just be aware that built-in tooling will respect them when restoring but not always when updating them in a UI or other end-user experience (which is the case with any build customization).

proposed/2022/version-overrides.md
jeffkl: Yes you can set this property in individual projects, a common import for a set of projects, or a common import for the whole repository.  The docs will call out possible locations like `Directory.Packages.props` or `Directory.Build.props`.

proposed/2022/version-overrides.md
jeffkl: Thanks I'll update the title

proposed/2022/version-overrides.md
jeffkl: Unfortunately we need the ability to differentiate between accidental and deliberate actions by users.  When opting into using central package management, you do so by declaring `<PackageReference />` items with only a package ID in projects and then `<PackageVersion />` items in a single location for the whole repository.  NuGet will then tell you that you have let a package version creep into a project which means you are no longer centrally managing your package versions.

Since the `Version` metadata generates an error when using central package management, we'll need another attribute to convey that the user wants to override the centrally defined version and not emit an error since this is intentional.

nkolev92:
@JonDouglas @jeffkl 

Can you please cross post the design in https://github.com/NuGet/Home/issues/10389.

jeffkl (approved):

PR #11708, opened 2022-03-30
[Proposal] Pack `ProjectReference`s

maxkoshevoi:
Related discussion: #3891 

[Rendered Proposal](https://github.com/NuGet/Home/pull/11708/files?short_path=7d373cd)

Thank You 🎉


proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
+Currently there's no easy way to override this behavior except for ditching `crproj`, and  going back to `nuspec`, which becomes another source of truth and all that good stuff.
```
smkanadl: It is `csproj` :-)

matthewjones555 (approved):

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,73 @@
...
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
+Currently there's no easy way to override this behavior except for ditching `csproj`, and  going back to `nuspec`, which becomes another source of truth and all that good stuff.
+
+## Explanation
+
+### Functional explanation
+
+Proposed solution is to take into account `IsPackable` property of the referenced projects. If it's `True` or not present (to preserve backward compatibility) - consider the project as a NuGet dependency, if it's `False` - pack its artifacts in.
+
+Second part is introduce new project-wide `PackReferencedProjects` property that would switch pack behavior for all `ProjectReference`s inside a single project.
```
AraHaan: I would rather have it where each ``ProjectReference`` would instead have optional metadata ``Pack="true"`` or ``<Pack>true</Pack>`` inside it like how ``<None Include="..." Pack="true" ... />`` works.

Why? Because of unit test projects normally set themselves as non-packable. So with this it could be an option to include them in the package without being "package dependencies". However I do not see why a sane person would reference their unit test projects in their actual code though. However this could be extended to not only just ``ProjectReference`` like ``PackageReference`` as well if one wanted to do that too with the metadata option.

maxkoshevoi: > Why? Because of unit test projects normally set themselves as non-packable. So with this it could be an option to include them in the package without being "package dependencies"

This is exactly how it would work with `IsPackable` approach. If project `N` references unit test project `UT`, `UT`'s artifacts would be packed into `N`'s NuGet since `UT` has `IsPackable` set to `False`. Am I missing something in your scenario?

> However this could be extended to not only just ProjectReference like PackageReference as well if one wanted to do that too with the metadata option.

This is interesting, but I don't see why would anybody want to pack NuGet inside another NuGet. Do you have any use case for that?

AraHaan: Yes, for example I might have a package that depends on packages that are not on nuget.org (a different feed or is a version of the package that is not in nuget.org like packages in the dotnet-7 feed for example) and may want to explicitly tell those specific packages to be packed in as well so when my package deploys on nuget.org people do not complain about the missing dependency.

maxkoshevoi: Hm, makes sense, thanks.

Are you proposing adding `Pack="true"` to `ProjectReference` and that's it, or introducing `PackReferencedProjects` as well?
Or even maybe inferring default `ProjectReference.Pack` value based on `IsPackable`? 

All three might be to much for such feature, but `IsPackable` + `Pack` might be good. This way user don't need to add `Pack="true"` for projects that explicitly state that they are not packable.

Thinking about it, even just adding `ProjectReference.Pack` would be good MVP for this.

AraHaan: I was thinking about only adding ``Pack`` to both ``ProjectReference`` and ``PackageReference``.

JonDouglas:
Thank you so much for your contribution, @maxkoshevoi. I have edited the description to help out and will schedule an initial review with the NuGet team for next week.

Please feel free to update the proposal as questions & feedback come in, and I'll do my best to support.

Thanks again!

proposed/2022/pack-referenced-projects.md
maxkoshevoi: Sounds good. Updated the proposal, could you take a look please?

AraHaan: It looks good so far, thanks.

Updated [Rendered Proposal](https://github.com/NuGet/Home/pull/11708/files?short_path=7d373cd)

AraHaan (approved):

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Prior Art
+
+This design flaw have been a pain for the community for more than 5 years now. It was such an issue that even third party solution was created - [nugetizer](https://github.com/devlooped/nugetizer).
+
+I haven't used it, so don't have much to comment here, but community clearly [expressed a desire for out of the box solution](https://github.com/NuGet/Home/issues/3891#issuecomment-1080044314).
+
+## Unresolved Questions
+
+None
```
MeikTranel: One of the primary use cases of bundling third party package content within your own package is custom build logic - analyzers, custom msbuild tasks, bundling satellite applications. There will be certain pressure to not only resolve a single level of package but a xcopy ready representation of that dependency's own dependency tree. To give some examples:

- a custom build task i am working on at the moment uses `Microsoft.EntityFrameworkCore.Sqlite` to write out gathered information as a database file
  - that package incurs a slew of packages dependent on TFM & platform - Do we stop after bundling the upper node of the dependency tree specified?
- an analyzer package needs a parser library at runtime - where does that `ProprietaryFileFormatParserXLib` land without further instructions ?  A naive implementation would drop this under lib which would mean the analyzer wouldn't work and the analyzer consumer would reference that parser library to add insult to injury - `PackagePath` anyone?

A possible basic implementation of the feature would look like a 1:1 copy of `PackageDependencyA/lib/[TFM or any]/` to `PackageToBePacked/lib/[TFM or any]/` - this would mean that any library package added into a bundled package would end up being referenced by consumers of that bundled package - is that a desired outcome?

I would recommend adding the following questions:

```suggestion
- **How do we flow/assign TFM based `lib`/`content` items?** (Is `PackagePath` a desired parameter?)
- **How is conflict resolution going to work with consumers taking in a dependency already present in the bundle?**
- Should license flow/validation be considered right from the start?
- **Should a `PackageReference` or `ProjectReference` tagged with `Pack="true"` be resolved transitively?**
```

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
+Currently there's no easy way to override this behavior except for ditching `csproj`, and  going back to `nuspec`, which becomes another source of truth and all that good stuff.
```
MeikTranel:
```suggestion
Currently there's no easy way to override this behavior except for ditching `csproj`, and going back to `nuspec`. In many cases that involves piping lots of information from the `MSBuild` context into the `nuspec` evaluation via `key=value` parameters.  
Chaining together dozens of arguments to achieve parity with metadata usually automagically introduced by the SDK becomes tedious, which eventually weakens the otherwise rich set of information attached to a modern NuGet package.
```

proposed/2022/pack-referenced-projects.md
maxkoshevoi: cc: @AraHaan

AraHaan: 1. I think usually one can check that as an implementation detail using:
- ``$(TargetFramework)``

However there comes another task it would have to do as well:
- use the value of ``$(RuntimeIdentifier)`` if set in project (otherwise ignore it when it == ``any``).

2. I think if analyzers/generators are referenced in project and are tagged with ``Pack="true"`` that it should package everything under the analyzer's folder in it's package AS IS which means conflicts between internal deps of the analyzers could possibly happen.

3. As for the license flow, what it could do is add these to the package metadata to support it: ``licenses/<package name>`` (bundle them in that package path) where it would bundle the licenses from the included projects/packages.
4. That just depends entirely on a lot of factors, if the package is meant only for an generator / analyzer (nothing more) than I say that resolving them would not be needed, otherwise yes they would be needed (we do not want to resolve dependencies of analyzers that would never be used on the consuming project and so wastes disc space).

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,64 @@
...
+- GitHub PR (-)
+
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
```
dotMorten, line 15: I don’t really understand this. Why wouldn’t another package transitively be adding those resources instead? Packaging the contents could create conflicts when you want to reference a newer version of a transitive reference.
If there are issues with transitively adding resources, why not address this scenario better without packing things into the package?
Could you also add some user stories around these problematic scenarios, beyond just stating it is an issue?

proposed/2022/pack-referenced-projects.md
dotMorten: `<None Include="../otherproject/bin/$(Configuration)/x64/mydll.com" Pack="true" PackagePath="/runtimes/win-x64/native/" />` would do it right?

Is the idea that you would include the above in "otherproject" instead, and the referencing project would automatically pick that up if 'otherproject' isn't packable?
In my scenario my biggest problem is packing runtimes, which require packing multiple architectures into the runtimes folder. How will this be accomplished with this feature?

AraHaan: Honestly people separate architectures by RuntimeIdentifiers nowadays.

dotMorten: I’m not following? Runtime identifiers doesn’t really apply to nuget packages but to those consuming them. You first have to get those runtimes into the nuget

maxkoshevoi: @dotMorten What do you mean by "piping"? "Manually duplicating", or something else? Meaning of original wording was that we ending up storing the same information in two places (which complicates maintenance). Also the whole `nuspec` needs to be manfully created, even if most of the stuff in it can be inferred from `csproj`.

MeikTranel: I think the word piping was introduced by me in this convo so ill take this one - basically it is acknowledging that in an ideal world where people actually know about these properties you don't actually need to have an actual copy. 

For reference (one of my projects that would end up profiting from the overall feature):

https://github.com/MeikTranel/NXPorts/blob/master/src/NXPorts/NXPorts.csproj#L38-L48

maxkoshevoi: Cool, didn't know you could do that, thanks!

JonDouglas:
Hi all,

Just commenting that we did review this last Friday and I'm trying to find time to write-up some notes. I apologize as this week we had a few large releases as per our blogs.

Tratcher:
Do you also get to choose the location in the package? Here's a similar mess I had to figure out recently:
https://github.com/aspnet/AspNetKatana/blob/beb224c88712b08ce45f1d14bb8cf0cd9d4a8503/src/OwinHost/OwinHost.csproj#L31-L45

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,69 @@
...
+Currently there's no easy way to override this behavior except for ditching `csproj`, and going back to `nuspec`. In many cases that involves piping lots of information from the `MSBuild` context into the `nuspec` evaluation via `key=value` parameters.  
+Chaining together dozens of arguments to achieve parity with metadata usually automagically introduced by the SDK becomes tedious, which eventually weakens the otherwise rich set of information attached to a modern NuGet package.
+
+## Explanation
+
+### Functional explanation
+
+Proposed solution is to add new `Pack` property to `ProjectReference` (`<ProjectReference Include="..." Pack="True" />`) and `PackageReference`. If it's `False` or not present - consider the project (or package) as a NuGet's dependency, if it's `True` - pack its artifacts in.
```
maxkoshevoi:
```suggestion
Proposed solution is to add new `Pack` and `PackagePath` properties to `ProjectReference` (`<ProjectReference Include="..." Pack="True" PackagePath="tools" />`) and `PackageReference`. 
If `Pack` is `False` or not present - consider the project (or package) as a NuGet's dependency, if it's `True` - pack its artifacts in.
If `PackagePath` is not present - pack artifacts to the default location within the package. If it's present - pack them to the specified path.
```

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+If `PackagePath` is not present - pack artifacts to the default location within the package. If it's present - pack them to the specified path.
+
+For `PackageReference`s it's useful in case referenced package is from private feed, but NuGet that's being packed will be public.
+
+### Technical explanation
+
+When NuGet is being created from `csproj` file:
+
+1) If `ProjectReference` doesn't have `Pack` property, or has it set to `False` - preserve behavior that have existed before introduction of this feature (don't include artifacts of referenced projects into the NuGet, and consider them as NuGet-dependencies).
+2) If `ProjectReference` has `Pack` property and it's set to `True` - include it's artifacts into the NuGet.
```
maxkoshevoi:
```suggestion
2) If `ProjectReference` has `Pack` property and it's set to `True` - include it's artifacts into the NuGet (to the default location or one specified in `PackagePath`).
```

maxkoshevoi:
> Do you also get to choose the location in the package? Here's a similar mess I had to figure out recently: [aspnet/AspNetKatana@`beb224c`/src/OwinHost/OwinHost.csproj#L31-L45](https://github.com/aspnet/AspNetKatana/blob/beb224c88712b08ce45f1d14bb8cf0cd9d4a8503/src/OwinHost/OwinHost.csproj#L31-L45)

Good catch. Updated the proposal to allow to choose the location

shuebner:
Regarding the empty "Drawbacks" section:
Consider what happens when the assemblies contained in a nuget package's TFM folder have different TFMs themselves, e. g. one has netstandard2.0 and the other has net6.0 or net48. If the projects from which they were built have the same `PackageReference`, they can get different assembly reference versions because of their different TFMs. Those can then lead to runtime assembly loading errors.

I detailed a real-life scenario in my blog:
https://svenhuebner-it.com/nuget-packaging-series-2-beware-of-dll-version-conflicts/

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Technical explanation
+
+When NuGet is being created from `csproj` file:
+
+1) If `ProjectReference` doesn't have `Pack` property, or has it set to `False` - preserve behavior that have existed before introduction of this feature (don't include artifacts of referenced projects into the NuGet, and consider them as NuGet-dependencies).
+2) If `ProjectReference` has `Pack` property and it's set to `True` - include it's artifacts into the NuGet (to the default location or one specified in `PackagePath`).
+3) Do the same steps for `PackageReference`s
+
+## Drawbacks
```
JonDouglas: It may be worth to add that one of the best practices in NuGet today is to have one assembly per NuGet package. 

https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#decide-which-assemblies-to-package

https://docs.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects#recommended-one-assembly-per-package

https://docs.microsoft.com/en-us/nuget/create-packages/select-assemblies-referenced-by-projects#not-recommended-multiple-assemblies-in-one-package

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+- Author Name [maxkoshevoi](https://github.com/maxkoshevoi)
+- Start Date (2022-03-28)
+- GitHub Issue [3891](https://github.com/NuGet/Home/issues/3891)
+- GitHub PR (-)
+
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
```
JonDouglas: Great motivation. Thanks for writing it!

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+This design flaw have been a pain for the community for more than 5 years now. It was such an issue that even third party solution was created - [nugetizer](https://github.com/devlooped/nugetizer).
+
+I haven't used it, so don't have much to comment here, but community clearly [expressed a desire for out of the box solution](https://github.com/NuGet/Home/issues/3891#issuecomment-1080044314).
+
+## Unresolved Questions
+
+- **How do we flow/assign TFM based `lib`/`content` items?** (Is `PackagePath` a desired parameter?)
+- **How is conflict resolution going to work with consumers taking in a dependency already present in the bundle?**
+- Should license flow/validation be considered right from the start?
+- **Should a `PackageReference` or `ProjectReference` tagged with `Pack="true"` be resolved transitively?**
```
JonDouglas: We'll have to consider the ProjectReference side of this problem with a broader team. I can include them.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Use `IsPackable` property from referenced project to determine it it's supposed to be packed or be a NuGet dependency.
+
+It will be more automated, but won't fit some edge cases (also it will be a breaking change). And it isn't applicable to `PackageReferences`.
+
+- Introduce new project-wide `PackReferencedProjects` property that would switch pack behavior for all `ProjectReference`s inside a single project.
+
+This one is less flexible since it cannot switch on per-`ProjectReference` basis. And it isn't applicable to `PackageReferences`.
+
+- Another solution is to introduce a `IncludeReferencedProjects` switch for `nuget` CLI.
```
JonDouglas: Would this be for `dotnet` CLI instead?

JonDouglas:
I apologize in advance for forgetting about this proposal until now. I wanted to let people know that the earliest we can think about this work will start in September. Thanks again for writing it and everyone's contributions and upvotes thus far. It significantly helps us in our upcoming planning cycles.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+- GitHub PR (-)
+
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
```
nkolev92: I think the motivation needs to capture the particular scenarios where this is helpful. We'd appreciate a bigger focus on the scenarios.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+Currently there's no easy way to override this behavior except for ditching `csproj`, and going back to `nuspec`. In many cases that involves piping lots of information from the `MSBuild` context into the `nuspec` evaluation via `key=value` parameters.
```
nkolev92: I think it's an unfair characterization to say the only way to pack more than assembly is by going with a nuspec, as `None` items can glob in an equivalent way the nuspec files elements can.

Furthermore the issue body does link to a comment that contains a workaround:
https://github.com/nuget/home/issues/3891#issuecomment-377319939, that can be implemented with the documented pack extensibility. https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#advanced-extension-points-to-create-customized-package

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+## Summary
+
+Allow referenced projects' artifacts to be packed inside the NuGet.
+
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
+Currently there's no easy way to override this behavior except for ditching `csproj`, and going back to `nuspec`. In many cases that involves piping lots of information from the `MSBuild` context into the `nuspec` evaluation via `key=value` parameters.  
+Chaining together dozens of arguments to achieve parity with metadata usually automagically introduced by the SDK becomes tedious, which eventually weakens the otherwise rich set of information attached to a modern NuGet package.
```
nkolev92: Someone brought up:
`One of the primary use cases of bundling third party package content within your own package is custom build logic - analyzers, custom msbuild tasks, bundling satellite applications`

I think this what the motivation should cover.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Explanation
+
+### Functional explanation
+
+Proposed solution is to add new `Pack` and `PackagePath` properties to `ProjectReference` (`<ProjectReference Include="..." Pack="True" PackagePath="tools" />`) and `PackageReference`. 
+If `Pack` is `False` or not present - consider the project (or package) as a NuGet's dependency, if it's `True` - pack its artifacts in.
+If `PackagePath` is not present - pack artifacts to the default location within the package. If it's present - pack them to the specified path.
+
+For `PackageReference`s it's useful in case referenced package is from private feed, but NuGet that's being packed will be public.
```
nkolev92: nit: This needs to go into the motivation.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+## Motivation
+
+When SDK-style project is being packed as a NuGet, all projects that it references are considered as NuGet themselves that this NuGet depends on.
+This is not always a desired behavior, sometimes user needs artifacts from referenced project(s) to be packed into the NuGet.
+Currently there's no easy way to override this behavior except for ditching `csproj`, and going back to `nuspec`. In many cases that involves piping lots of information from the `MSBuild` context into the `nuspec` evaluation via `key=value` parameters.  
+Chaining together dozens of arguments to achieve parity with metadata usually automagically introduced by the SDK becomes tedious, which eventually weakens the otherwise rich set of information attached to a modern NuGet package.
+
+## Explanation
+
+### Functional explanation
```
nkolev92: I think multiple asks are being conflated into 1 here.

- Packaging multiple projects into 1 package.
- Packaging assemblies from packages into a different package.

These are obviously related, but the scenarios vary and the potential for misuses of a feature that does things automatically varies.

It's worth nothing that IncludeReferenceProjects handled projects only, so I don't think it'd be unfair to suggest that these features are not all combined into one.

proposed/2022/pack-referenced-projects.md
nkolev92: Additionally an aside, packing assemblies from packages published on private feeds only can lead to a lot of issues.

If a package is in a private feed, it must have been done so for a reason. Redistributing assemblies for packages you don't own might lead into some licensing violations.

There's no guarantee that this package currently on a private feed (maybe because it's in previews) does not make it to NuGet.org. Redistributing the `same assembly in multiple packages` is something that's known to cause a lot of problems for the consumers.
The build and runtime work best when only 1 reference of the same assembly is used in an application. NuGet ensures that you cannot have multiple versions of the package in the project, but that guarantee cannot be transferred to the assemblies themselves. If an assembly is in multiple packages, it's very possible that someone ends up referencing both. The more dependency issues can be resolved at restore time the  better.

The above concerns can be generalized to `Never include the same assembly in multiple library packages`.
Note the focus on library packages. In some scenarios, packages contain tools, and those often need the entry assembly + some of it's dependencies.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+For `PackageReference`s it's useful in case referenced package is from private feed, but NuGet that's being packed will be public.
+
+### Technical explanation
+
+When NuGet is being created from `csproj` file:
+
+1) If `ProjectReference` doesn't have `Pack` property, or has it set to `False` - preserve behavior that have existed before introduction of this feature (don't include artifacts of referenced projects into the NuGet, and consider them as NuGet-dependencies).
+2) If `ProjectReference` has `Pack` property and it's set to `True` - include it's artifacts into the NuGet (to the default location or one specified in `PackagePath`).
+3) Do the same steps for `PackageReference`s
```
nkolev92: I think the concerns about distributing the same assembly in multiple packages in a good case against PackageReference inclusion for libraries. In fact, it should *never* be done for libraries.

nkolev92: Does this design recommend only assembly built by projects are included?

What about the PackageReference side? There are different types of assemblies in a package, some for compile, some for runtime. We'd need to define the behavior for these further.

There might also be compatibility concerns due to the potential usage of AssetTargetFallback.

proposed/2022/pack-referenced-projects.md
nkolev92: I think some of the feedback around compatibiltiy and redistribution of packages should be listed in the drawbacks here.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Drawbacks
+
+None
+
+## Rationale and alternatives
+
+- Use `IsPackable` property from referenced project to determine it it's supposed to be packed or be a NuGet dependency.
```
nkolev92: I think this is a fine consideration.

proposed/2022/pack-referenced-projects.md
nkolev92: If a Project is being included into a joined package and never gets packed itself, it's dependencies would need to be something that has to be represented in the joined package.
Example:

Project 1 -> Project 2 -> Package A.

Say, Project 1 includes the reference project, and it is intended to be redistributed as a library, it's extremely likely that Project 1 needs Package A as a dependency.
This is a `reverse restore`. Instead of given the dependency, give me a resolved graph, it starts with a resolved graph and it tries to generate the correct top level dependencies. Note that we'd need to account for package include/excludes as well.

nkolev92: > Would this be for `dotnet` CLI instead?

Correct. It'd be for dotnet CLI and the msbuild pack target (which the dotnet pack is just a wrapper for)

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+This one is less flexible since it cannot switch on per-`ProjectReference` basis. And it isn't applicable to `PackageReferences`.
+
+- Another solution is to introduce a `IncludeReferencedProjects` switch for `nuget` CLI.
+
+Downsides are: Behavior of individual projects (if packing is performed in bulk) and `ProjectReference`s cannot be changed, and it's not obvious how to properly pack any specific project (I thought we are consolidating all the information in `csproj` now). And it isn't applicable to `PackageReferences`.
+
+## Prior Art
+
+This design flaw have been a pain for the community for more than 5 years now. It was such an issue that even third party solution was created - [nugetizer](https://github.com/devlooped/nugetizer).
```
nkolev92: I think it's unfair to characterize the feature ask as design flaw. It's a feature that hasn't been implemented.
This also suggests that nugetizer was created because of the lack of this feature, which isn't correct.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+I haven't used it, so don't have much to comment here, but community clearly [expressed a desire for out of the box solution](https://github.com/NuGet/Home/issues/3891#issuecomment-1080044314).
+
+## Unresolved Questions
+
+- **How do we flow/assign TFM based `lib`/`content` items?** (Is `PackagePath` a desired parameter?)
```
nkolev92: I'm not sure I fully understand this example. Can you clarify better what the ask is?

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+- **How is conflict resolution going to work with consumers taking in a dependency already present in the bundle?**
```
nkolev92: It unfortunately doesn't. The controls for assembly includes are on a package level. Furthermore, it's not something that NuGet tries to address today. The build itself does go through conflict resolution itself.

proposed/2022/pack-referenced-projects.md
nkolev92: This is a very good question, which is why I think the motivation isn't the correct one here. Understanding the scenarios would guide this answer (the answer would usually be yes.)

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+- Should license flow/validation be considered right from the start?
+- **Should a `PackageReference` or `ProjectReference` tagged with `Pack="true"` be resolved transitively?**
+- What happens to NuGets that are referenced in `PackageReference` tagged with `Pack="true"`? Do they become dependencies of the resulting NuGet or being packed into it.
```
nkolev92: This a reverse restore cases similar to the above one.

proposed/2022/pack-referenced-projects.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Future Possibilities
+
+This proposal would allow users more flexibility in choosing what is being packed as part of a NuGet, and how it's dependencies look like.
```
nkolev92: nit: More of a motivation thing.

nkolev92 (commented):
Apologies for the delayed review. 

I'd love for us to focus on the scenarios first. 

1 package (library), 1 assembly is a policy that tends to provide more predictable outcomes compared to packaging many assemblies into one package.

proposed/2022/pack-referenced-projects.md
baronfel: One specific motivating example for this feature is [MSBuild Tasks](https://docs.microsoft.com/en-us/visualstudio/msbuild/tutorial-custom-task-code-generation?view=vs-2022#bundle-dependencies-into-the-package). Happy to see this brought up!

joe-triscari (approved):

proposed/2022/pack-referenced-projects.md
MeikTranel: Referring to the conversation earlier - here: https://github.com/NuGet/Home/pull/11708#discussion_r844459807

nkolev92: Thanks for that link. In that case, I think the compat side would probably need to not be configurable. 

In other words, given a project reference today, the particular assembly chosen by NuGet/SDK is the best compatible framework one, so I think the switch would only really need to be copy for project, or don't. 

The conflict resolution between package items and references is more complex as that happens in a part that NuGet doesn't understand. So we'd need NuGet and the SDK to work through that in some way.

nkolev92: I understand there is *potentially* a better framework for some of the dependent assemblies, but accounting for that would make things too complicated imo. NuGet today chooses a full lib folder, so we can't really pick and choose from different folders.

maxkoshevoi:
Ping

igitur:
Ping

AraHaan:
ping, anyone going to review this again? This would be useful in the future.

donnie-msft:
Hi, we have removed our "proposed" folder, so please move this proposal to the "accepted" folder. 
Thanks!

AraHaan:
We still waiting.

AraHaan:
Also shouldn't this be moved to 2024 folder?

stevendarby:
What is the proposal for nested references?

For example, when we were manually creating nuspec files, the referenced projects that we weren't creating packages for (IsPackable=false) were included in the nuspec as files instead of dependencies, but the packages that those referenced projects themselves referenced _were_ added as dependencies - because they actually were packages.  

If we go down the route of Pack=false on a project reference, then I'd hope this wouldn't prevent it recursively checking the references within that project to determine what else should be added as a dependency and what as a file. But this may then require setting Pack=false against a lot more project references across a solution and becomes a bit messy and confusing. 

That's why I have to say I prefer it if we just used the IsPackable setting on the project. It just feels more intuitive and clearer and also means we don't need to set Pack=false everywhere that the project is referenced. It'd be a property of the project rather than of references to it.

AraHaan:
This PR is still needed.

Issue #10159, opened 2020-10-20
[Feature]: Add ability to designate a package reference as used by every project in the repo


zarenner:
We're looking at migrating from https://github.com/microsoft/MSBuildSdks/tree/master/src/CentralPackageVersions to the built-in central package versioning feature, and (unless I missed it) I didn't see an equivalent of `GlobalPackageReference`.

I'm guessing that for all of our (many) build/analyzer packages, we can do the equivalent and manually add both `PackageVersion` and `PackageReference` (with IncludeAssets and PrivateAssets set as `GlobalPackageReference` does)? Is there any reason we shouldn't add the `PackageReference`s directly in Directory.Packages.props?

aortiz-msft:
Also adding @zivkan

zivkan:
@zarenner I believe you're correct. I'm not aware of any equivalent feature. I agree that a `PackageReference` in `Directory.Packages.props` is a good alternative. You'll need to add the `PrivateAssets` and `IncludeAssets` yourself in this case.

@rconard I don't know if you've seen this feature before, but it sounds interesting. [There are a few public repos on github using it](https://github.com/search?q=GlobalPackageReference&type=code).

rconard:
@zarenner, thank you for raising this!

This is not currently in the design doc, but this feature is still very young.

If you could please give us some additional context on how you're using `GlobalPackageReference` and what you did prior to MSBuild SDK, that would be helpful.

We would also like to hear from more users of MSBuild SDK about how you're using `GlobalPackageReference`.

zarenner:
@rconard we have ~15 `GlobalPackageReferences`, the vast majority of which are code analyzers which we want to run against all projects.

We also have a few packages which use props/targets imports from the `build` folder to do things like enforce code structure, or provide some tools via nuget package and define their location as a property.

We introduced `Microsoft.Build.CentralPackageVersions` right around when we switched from packages.config to PackageReference style restore, so what we did prior isn't especially relevant. That said, at one point we did have a special project we called `Externals` which all projects depended on and triggered a package restore if necessary, before wiring in props/targets for the "global" tools/build-time packages.

Historically we've had a _way too much_ custom nuget-related code IMO (we implemented our own global packages cache at one point to work around nuget limitations), which is why I'd love to get as close as possible to "nuget as it's meant to be used".

aortiz-msft:
[...]

Issue #10746, opened 2021-04-09
Support for source generators that add non BCL code


AArnott:
In VS 16.9.3 (or as new as 16.10 Preview 3, 31206.385.main)...

1. Create a new C# class library:
```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <LangVersion>9</LangVersion>
  </PropertyGroup>
</Project>
```

2. Use the NuGet Package Manager to install Microsoft.Windows.CsWin32 0.1.422-beta to the project.
3. Add a type reference to `Span<byte>` somewhere in a .cs file

**Expected**

No compilation errors.

**Actual**

`Span<byte>` produces a compiler error because System.Memory.dll is not a reference that is passed to the compiler.

![image](https://user-images.githubusercontent.com/3548/114211552-052f8100-991e-11eb-83b1-d825089c95b3.png)

The nuget package reference was added with:

```xml
  <ItemGroup>
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.1.422-beta">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
    </PackageReference>
  </ItemGroup>
```

Note the `IncludeAssets` metadata is missing the `compile` value. Adding it manually (or removing the `IncludeAssets` metadata altogether) fixes the problem.

![image](https://user-images.githubusercontent.com/3548/114211589-0f517f80-991e-11eb-8398-a4bb56fa2751.png)

Note that the nuspec for Microsoft.Windows.CsWin32 includes the System.Memory dependency with `include="all"`:

```xml
<dependency id="System.Memory" version="4.5.4" include="All" />
```

Note also that our nuspec sets:
```xml
    <developmentDependency>true</developmentDependency>
```

The net effect we want is that Microsoft.Windows.CsWin32 itself not be a package dependency expressed in the consuming project's own package, but we *do* want `System.Memory` to be expressed. 

As originally reported at https://github.com/microsoft/CsWin32/issues/242

jnm2:
I suspect the answers will be "this is by design" or "it can't be `compile` because that has other consequences." In that case, please consider extending the design to make development dependency packages capable of adding runtime dependencies for consuming projects, or in some other way make the state of things more friendly for consumers of source generators.

AraHaan:
[...]

PR #11794, opened 2022-05-05
[Proposal]: Global Package Reference for Central Package Management

jeffkl:
# [Rendered](https://github.com/NuGet/Home/blob/dev-jeffkl-globalpackagereference/proposed/2022/globalpackagereference.md)

Proposal for https://github.com/NuGet/Home/issues/10159

Thank You 🎉

### DIFF
--- a/proposed/2022/globalpackagereference.md
+++ b/proposed/2022/globalpackagereference.md
@@ -40,7 +40,7 @@ default metadata to ensure that only package assets related to build purposes ar
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
 `<GlobalPackageReference />` will just be syntactical sugar for a `<PackageReference />` with the following metadata:
 
-* IncludeAssets = `Runtime; Build; Native; ContentFiles; Analyzers`
+* IncludeAssets = `runtime; build; native; contentfiles; analyzers`
 * PrivateAssets = `All`
 
 This ensures that when these repository-wide packages are consumed, they don't flow to downstream dependencies and only include assets related to
@@ -64,7 +64,7 @@ This will be achieved by copying the `GlobalPackageReference` items to `PackageR
       Add GlobalPackageReference items to the PackageReference item group with no version.  The PackageVersion items are added
       in the CollectCentralPackageVersions target.
 
-      Global package references only include the same assets as a development dependency (Analyzers;Build;BuildMultitargeting;BuildTransitive)
+      Global package references only include the same assets as a development dependency (runtime; build; native; contentfiles; analyzers)
       because those kind of packages are the best candidate for a global package reference.  They are generally packages that
       extend the build.
 
@@ -74,7 +74,7 @@ This will be achieved by copying the `GlobalPackageReference` items to `PackageR
     -->
     <PackageReference Include="@(GlobalPackageReference)"
                       Version=""
-                      IncludeAssets="Analyzers;Build;BuildMultitargeting;BuildTransitive"
+                      IncludeAssets="runtime; build; native; contentfiles; analyzers"
                       PrivateAssets="All" />
   </ItemGroup>
 </Target>
--- a/proposed/2022/globalpackagereference.md
+++ b/proposed/2022/globalpackagereference.md
@@ -40,7 +40,7 @@ default metadata to ensure that only package assets related to build purposes ar
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
 `<GlobalPackageReference />` will just be syntactical sugar for a `<PackageReference />` with the following metadata:
 
-* IncludeAssets = `Analyzers;Build;BuildMultitargeting;BuildTransitive`
+* IncludeAssets = `Runtime; Build; Native; ContentFiles; Analyzers`
 * PrivateAssets = `All`
 
 This ensures that when these repository-wide packages are consumed, they don't flow to downstream dependencies and only include assets related to
@@ -51,13 +51,13 @@ This will be achieved using an MSBuild [ItemDefinitionGroup](https://docs.micros
   <ItemDefinitionGroup Condition="'$(ManagePackageVersionsCentrally)' == 'true' And '$(RestoreEnableGlobalPackageReference)' != 'false'">
     <GlobalPackageReference>
       <!--
-        Default global package references to only consume the Analyzers and Build logic in a package.
-        This helps ensure that the package assets are not passed to the compiler or copied to the
-        output directory.  Having a compile-time reference to a package for all projects in a tree
-        is not recommended.  You should only have "global" references to packages that are used for
-        build.
+        Default global package references to only consume the Analyzers and Build logic in a package,
+        which is the same for development dependencies.  This helps ensure that the package assets
+        are not passed to the compiler or copied to the output directory.  Having a compile-time
+        reference to a package for all projects in a tree is not recommended.  You should only have
+       "global" references to packages that are used for build.
       -->
-      <IncludeAssets Condition="'$(RestoreEnableGlobalPackageReferenceIncludeAssetsBuildAnalyzers)' != 'false'">Analyzers;Build;BuildMultitargeting;BuildTransitive</IncludeAssets>
+      <IncludeAssets Condition="'$(RestoreEnableGlobalPackageReferenceIncludeAssetsBuildAnalyzers)' != 'false'">runtime;build;native;contentfiles;analyzers</IncludeAssets>
       <!--
         Default global package references to have all assets private.  This is because global package
         references are generally stuff like versioning, signing, etc and should not flow to downstream
--- a/proposed/2022/globalpackagereference.md
+++ b/proposed/2022/globalpackagereference.md
@@ -1,0 +1,125 @@
+# Global Package Reference for Central Package Management
+
+- [Jeff Kluge](https://github.com/jeffkl)
+- 2022-05-05
+- [[Feature]: CPVM equivalent of GlobalPackageReference? (#10159)](https://github.com/NuGet/Home/issues/10159)
+- GitHub PR (GitHub PR link)
+
+## Summary
+<!-- One-paragraph description of the proposal. -->
+A common pattern in repositories is to have every project in the tree reference packages that are used for development purposes only.  This includes
+packages that provide assembly versioning, signing, code analysis, etc.  We want to provide a built-in mechanism for users to define packages that
+are used by every project but only for development purposes.
+
+## Motivation 
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Users can manually reference the package in a common import like Directory.Build.props but there is no built-in safe guards to ensure that they don't
+accidentally consume more assets than just development assets.  For example, you don't want to have a package's compile-time assets be referenced by
+every assembly in your repository.  Instead, individual projects should express their build time dependencies and project references should receive
+them transitively.  But you should not need to add a package reference in order to have your whole tree use code analysis or assembly versioning.
+
+## Explanation
+
+### Functional explanation
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+We'll introduce a new MSBuild item group named `<GlobalPackageReference />` which will define packages that every project should reference but has
+default metadata to ensure that only package assets related to development purposes are used.
+
+**NOTE:** This functionality will only be enabled if a user has opt-ed into [Central Package Management (CPM)](https://docs.microsoft.com/nuget/consume-packages/central-package-management).
+
+```xml
+<Project>
+  <ItemGroup>
+    <GlobalPackageReference Include="Nerdbank.GitVersioning" Version="1.0.0" />
+  </ItemGroup>
+</Project>
+```
+
+### Technical explanation
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+`<GlobalPackageReference />` will just be syntactical sugar for a `<PackageReference />` with the following metadata:
+
+* IncludeAssets = `Analyzers;Build;BuildMultitargeting;BuildTransitive`
+* PrivateAssets = `All`
+
+This ensures that when these repository-wide packages are consumed, they don't flow to downstream dependencies and only include assets related to
+development.
+
+This will be achieved using an MSBuild [ItemDefinitionGroup](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-definitions):
+```xml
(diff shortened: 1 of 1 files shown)


proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+We'll introduce a new MSBuild item group named `<GlobalPackageReference />` which will define packages that every project should reference but has
+default metadata to ensure that only package assets related to development purposes are used.
+
+**NOTE:** This functionality will only be enabled if a user has opt-ed into [Central Package Management (CPM)](https://docs.microsoft.com/nuget/consume-packages/central-package-management).
+
+```xml
+<Project>
+  <ItemGroup>
+    <GlobalPackageReference Include="Nerdbank.GitVersioning" Version="1.0.0" />
```
nkolev92, line 34: I don't think the name `GlobalPackageReference` is correct. 

The `Global` part is a function of where this is written, not the value of the item itself. 

The itself is what was said above...a development dependency reference. 

So if we were to add this functionality, the name should be changed.

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+  </ItemGroup>
+</Project>
+```
+
+### Technical explanation
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+`<GlobalPackageReference />` will just be syntactical sugar for a `<PackageReference />` with the following metadata:
+
+* IncludeAssets = `Analyzers;Build;BuildMultitargeting;BuildTransitive`
```
nkolev92: Note that this is *different* from how `DevelopmentDependency` works in PackageReference.
If you read that thread, you'll discover that the feature is a `personal favorite` of mine :D 

If I could break things, I'd say development dependencies:

* Can't bring compile/runtime assets
* Have to be self contained, and can't have dependencies
* Are not part of the standard project graph, something theoretically addressed by not allowing them to have dependencies.

Note that what I described is effectively the guidance for writing msbuild task packages.

nkolev92 (commented):
Thanks for getting the conversation going @jeffkl. Really happy we're making progress on all these CPM polish tasks.

I don't think we have to completely match the functionality of the SDK. GlobalPackageReference is a function of what's used, but the name suggests a scope, so imo that needs addressed, and conveniently it'd address some of the unresolved questions.

The `challenge` with development dependency is that they can have dependencies. 
You can read more about it in:
https://github.com/NuGet/Home/issues/10746#issuecomment-819153729
The dependencies that *affect* the graph and it meaning gets lost quickly. 

Is there a plan for tooling support here?

proposed/2022/globalpackagereference.md
nkolev92: See details: https://github.com/NuGet/Home/pull/11794

proposed/2022/globalpackagereference.md
clairernovotny, line 34: If CPM is a requirement, then shouldn't we not put a version here and put that in the Directory.Packages.props? Or are we trying to put a harder line between the two types of packages?

jeffkl: The idea is that `Directory.Packages.props` contains all of your package versions and `GlobalPackageReference` items should really only be declared in this file.  The documentation will try to make this clear, `GlobalPackageReference` items should only be in `Directory.Packages.props` and not in individual projects themselves.

dominoFire: Naming suggestion: `CentralPackageReference`, to reflect Central Package Management usage.

dominoFire: Naming suggestion: `CentralDevPackageReference`, toe reflect CPM usage on a development dependency.

Nirmal4G: I would suggest `BuildOnlyPackageReference` or `DevOnlyPackageReference` but I suspect `GlobalPackageReference` will be an appropriate naming considering the purpose of it. Also `CentralPackageReference` doesn't sound right compared with the former. I can also suggest `SharedPackageReference` and `CommonPackageReference` but @jeffkl would've already pondered about it when he named it. So, it's better to ask him.

Nirmal4G:
I like the proposal overall but I would change one thing, separate central package versions from central (_global/common/shared_) packages. What I mean is, have two properties that individually enable each of the features and change how those features interact or complement each other when enabled/disabled individually.

There are projects that use `GlobalPackageReference` without using central versioning. They don't have to depend on each other.

proposed/2022/globalpackagereference.md
Nirmal4G: I would suggest `BuildOnlyPackageReference` or `DevOnlyPackageReference` but I suspect `GlobalPackageReference` will be an appropriate name considering the location and the **supposed** purpose of it. The default is development only dependencies but it's not stopping us from including other assets.

Also `CentralPackageReference` doesn't sound right compared with the former. I can also suggest `SharedPackageReference` and `CommonPackageReference` but @jeffkl would've already pondered about it when he named it. So, it's up to him.

proposed/2022/globalpackagereference.md
jeffkl: I'm going to remove all references to "development dependency" so there is no confusing.  This feature is about making sure repo-wide packages only consume _build_ assets.

proposed/2022/globalpackagereference.md
jeffkl: After two months of collecting feedback, `GlobalPackageReference` appears to be the most voted for naming based on internal and external surveys.

https://twitter.com/nuget/status/1545037921455153155

I think we should move forward with `GlobalPackageReference` at this point.

JonDouglas (approved):
Approving after returning from vacation & sickness.

We have done some pulse surveys and believe that the current name is in the best interest for this feature and existing migration. While we can spend more time on the name here, we have over 30 votes from various .NET OSS, contributors, and other influences. We should move forward unless there is a serious concern or alternate name that hasn't already been suggested.

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+* PrivateAssets = `All`
```
Nirmal4G:
```suggestion
* IncludeAssets = `Runtime;Build;Native;ContentFiles;Analyzers`
* ExcludeAssets = `Compile`
* PrivateAssets = `All`
```

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+      <IncludeAssets Condition="'$(RestoreEnableGlobalPackageReferenceIncludeAssetsBuildAnalyzers)' != 'false'">Analyzers;Build;BuildMultitargeting;BuildTransitive</IncludeAssets>
+      <!--
+        Default global package references to have all assets private.  This is because global package
+        references are generally stuff like versioning, signing, etc and should not flow to downstream
+        dependencies.  Also, global package references are already referenced by every project in the
+        tree so we don't need them to be transitive.
+      -->
+      <PrivateAssets Condition="'$(RestoreEnableGlobalPackageReferencePrivateAssetsAll)' != 'false'">All</PrivateAssets>
+    </GlobalPackageReference>
+  </ItemDefinitionGroup>
```
Nirmal4G:
```suggestion
  <ItemDefinitionGroup Condition="'$(UseGlobalPackageReferences)' == 'true'">
    <GlobalPackageReference>
      <!--
        Default global package references should only consume Development Dependency assets.
        This helps to ensure that other package assets are not passed to the compiler or copied to output.
        Having a compile-time reference to a package for all projects in a tree is not recommended.
        You should only have "global" references to packages that are used only in build-time.
      -->
      <IncludeAssets Condition="'$(EnableGlobalPackageReferenceDefaultIncludeAssets)' != 'false'">Analyzers;Build;ContentFiles;Native;Runtime</IncludeAssets>
      <!-- These assets are not recommended to be required at all, so they are excluded explicitly -->
      <ExcludeAssets Condition="'$(EnableGlobalPackageReferenceDefaultExcludeAssets)' != 'false'">Compile</ExcludeAssets>
      <!--
        Default global package references should have all assets private to the project.
        This is because "global" package references are generally stuff like versioning, signing, etc and
        they should not flow to downstream dependencies. Also, they are already referenced by every
        project in the tree; so, we don't need them to be transitive.
      -->
      <PrivateAssets Condition="'$(EnableGlobalPackageReferenceDefaultPrivateAssets)' != 'false'">All</PrivateAssets>
    </GlobalPackageReference>
  </ItemDefinitionGroup>
```

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+The following MSBuild properties can be set to change the functionality:
+
+| Property Name | Function | Default value|
+|---|---|---|
+| `ManagePackageVersionsCentrally` | Enables or disables central package management and all dependent features | `false` |
+| `RestoreEnableGlobalPackageReference` | Enables or disables just the concept of `GlobalPackageReference` | `true` |
+| `RestoreEnableGlobalPackageReferenceIncludeAssetsBuildAnalyzers` | Enables or disables defaulting the `IncludeAssets` metadata for global package references | `true` |
+| `RestoreEnableGlobalPackageReferencePrivateAssetsAll` | Enables or disables defaulting the `PrivateAssets` metadata for global package references | `true`|
```
Nirmal4G: I'm suggesting a new property to oversee all CPM-related features. i.e.: `ManagePackagesCentrally` though naming is up for debate.

```suggestion
| Property Name | Function | Default value |
|---|---|---|
| `ManagePackagesCentrally` | Enables or disables central package management and all dependent features | `false` |
| `ManagePackageVersionsCentrally` | Enables or disables managing of package versions in a centralized way | `$(ManagePackagesCentrally)` or `false` |
| `UseGlobalPackageReferences` | Enables or disables global package references | `$(ManagePackagesCentrally)` or `true` |
| `EnableGlobalPackageReferenceDefaultIncludeAssets` | Enables or disables defaulting the `IncludeAssets` metadata for global package references | **nil** |
| `EnableGlobalPackageReferenceDefaultExcludeAssets` | Enables or disables defaulting the `ExcludeAssets` metadata for global package references | **nil** |
| `EnableGlobalPackageReferenceDefaultPrivateAssets` | Enables or disables defaulting the `PrivateAssets` metadata for global package references | **nil** |
```

#### Other names:

* `EnableDefaultPrivateAssetsInGlobalPackageReference`
* `EnableDefaultPrivateAssetsForGlobalPackageReference`

The above default value implementation expressed in targets:

```xml
  <PropertyGroup>
    <UseGlobalPackageReferences Condition="'$(UseGlobalPackageReferences)' == ''">$([MSBuild]::ValueOrDefault('$(ManagePackagesCentrally)', 'true'))</UseGlobalPackageReferences>
    <ManagePackageVersionsCentrally Condition="'$(ManagePackageVersionsCentrally)' == ''">$([MSBuild]::ValueOrDefault('$(ManagePackagesCentrally)', 'false'))</ManagePackageVersionsCentrally>
  </PropertyGroup>
```

Nirmal4G:
@jeffkl @nkolev92 Does the `PackageReference` and `ProjectReference` support `Pack` and `Publish` metadata? If so, we should include them in the definition. Also, don't we need to exclude `Compile`, `Runtime` and `Native` assets? OR does it automatically exclude everything except the ones we include? I saw that the default for excludes is `None`.

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+
+`GlobalPackageReference` items will then be copied to the `PackageReference` item group along with their metadata:
+```xml
+<ItemGroup>
+  <PackageReference Include="@(GlobalPackageReference)" Version="" />
+  <PackageVersion Include="@(GlobalPackageReference)" Version="%(Version)" />
+</ItemGroup>
```
Nirmal4G:
```suggestion
<ItemGroup>
  <PackageReference Include="@(GlobalPackageReference)" Condition="'$(ManagePackageVersionsCentrally)' != 'true'" />
  <PackageReference Include="@(GlobalPackageReference)" Version="" Condition="'$(ManagePackageVersionsCentrally)' == 'true'" />
</ItemGroup>

<ItemGroup>
  <PackageVersion Include="@(GlobalPackageReference->ClearMetadata())" Version="%(Version)" />
</ItemGroup>
```

jeffkl:
Team Review: See how many development dependency packages have runtime assets, otherwise just use same assets from DevelopmentDependency

Nirmal4G (changes requested):
Simple names for properties

proposed/2022/globalpackagereference.md
jeffkl: Based on our conversation we discovered that there are a lot of packages labeled as "development dependencies" with runtime assets.  So I've adjusted the default IncludeAssets to match that.  Otherwise a package with a runtime asset would behave differently as a "global" package reference vs normal.

proposed/2022/globalpackagereference.md
jeffkl: We've decided to align this more with the default development dependency assets `runtime; build; native; contentfiles; analyzers` so that package references behave the same in both scenarios.

proposed/2022/globalpackagereference.md
jeffkl: Since the packages are PrivateAssets=All they won't participate in pack.  They'll participate in Publish if they have any runtime or contentfiles assets but that's expected.

proposed/2022/globalpackagereference.md
nkolev92: The Pack metadata doesn't apply to Package/ProjectReferences anyways. 

It's only a thing in some content items.

proposed/2022/globalpackagereference.md
Nirmal4G: Also, instead of opting-in, just as I mentioned detecting `Directory.Packages.props` to enable central package versions, do the same for "Global" package references. We could do this either by detecting "packages" props or by detecting "build" props and non-empty **GPR** items check.

proposed/2022/globalpackagereference.md
Nirmal4G: In response to <https://github.com/NuGet/Home/pull/11794#issuecomment-1161166465>

Nirmal4G:
Updated my review in response to changes/comments

proposed/2022/globalpackagereference.md
```diff
@@ -0,0 +1,125 @@
...
+      extend the build.
+
+      Global package references have all assets private because central package references are generally packages that provide
+      versioning, signing, etc and should not flow to downstream dependencies.  Also, central package references are already
+      referenced by every project in the tree so they don't need to be transitive.
+    -->
+    <PackageReference Include="@(GlobalPackageReference)"
+                      Version=""
+                      IncludeAssets="runtime; build; native; contentfiles; analyzers"
+                      PrivateAssets="All" />
```
Nirmal4G, line 78: With the item definition, you can override the GPC's metadata but this removes the ability to that. If we specify GPC with a custom metadata, those will get overriden here.

Issue #4103, opened 2016-12-13
support for bulk-updating references into csproj from commandline(s) - dotnet package update


rrelyea:
splitting off from #3751 - which now just covers the add side of this work

Initial spec: https://github.com/NuGet/Home/blob/dev/accepted/2024/package-update-command.md

Board: https://github.com/orgs/NuGet/projects/13/views/30

rrelyea:
since add will support "update", we'll prioritize this task as the lowest of the 4:
1) add 2) remove 3) list 4) update

dasMulli:
Would it make sense to allow updating all package references? For a lot of dependencies, it is very quick to do sth like `dotnet update pkg --all` to begin with when updating a project.

kjbetz:
I would highly suggest checking out how yarn has taken care of this with `yarn upgrade` and `yarn upgrade-interactive`

https://yarnpkg.com/lang/en/docs/cli/upgrade/

https://yarnpkg.com/lang/en/docs/cli/upgrade-interactive/

I would just use the same names as well... `dotnet upgrade` `dotnet upgrade-interactive` or possibly `dotnet package upgrade` `dotnet package upgrade-interactive`

rrelyea:
Notice that #6193 added the additional requests to support cross-project updates, etc...

mudrz:
Since this issue is 2 years old - are there any plans to create any basic cli commands for managing package versions?
Currently it is very inconvenient to manage packages outside of Visual Studio on Windows.

letmaik:
As part of continuous integration I'd like to auto-update dependencies to check that nothing breaks with newer versions. However, this seems impossible without this feature if you manage your dependencies in `.csproj` files. Any ETA for this? It seems like a basic feature to have.

mishra14:
@letmaik you can always run `dotnet add package <package_name>`. This command will find the latest version of the package and update it in your csproj.

letmaik:
Unfortunately the [VSTS .NET Core](https://docs.microsoft.com/en-us/vsts/pipelines/tasks/build/dotnet-core?view=vsts) task does not support `add` and if you use the "custom command" feature then it doesn't support authentication, which I would need since I use a private VSTS feed. `dotnet restore` supports authentication of VSTS feeds. Back to square one I think.

worldbeater:
Is there any news about this issue? `dotnet update %packagename%` command could be really useful.

jjgriff93:
[...]

PR #11812, opened 2022-05-11
dotnet update package for NuGet packages.

JonDouglas:
Introduction to an updating experience in the .NET CLI that allows one to update dependencies in projects and solutions while supporting helpful options to stay current based on an individuals update strategy.

This proposal introduces one new command to .NET CLI known as `dotnet update package`.

[Rendered Spec](https://github.com/NuGet/Home/blob/4ea15dd4f7f9431310e71432e8a31555d1b69517/proposed/2022/DotnetUpdatePackage.md)

Thank You 🎉


proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+dotnet update package Microsoft.*
+```
+
+- Displays what would be updated in the update operation for the current working directory.
+
+```
+dotnet update package --dry-run
+```
+
+#### Exit Codes
```
chrisraygill, line 127: Is it all or nothing? Is it possible for some packages to update while others fail?

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+- [dart pub upgrade](https://dart.dev/tools/pub/cmd/pub-upgrade)
+- [pip install -U](https://pip.pypa.io/en/stable/cli/pip_install/)
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should this command use `dotnet add package` under the hood for the scenarios that make sense such as finding the latest version of a specific package name?
+- What verb makes most sense? `update`, `upgrade`, `refresh`, etc? Update seems the most consistent for package managers such as cargo, npm, rubygems, and NuGet. Upgrade seems the most consistent for package managers such as pip.
```
chrisraygill, line 172: Worth noting that "update" is consistent with the current PMUI terminology.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+- --prerelease
+
+Allows prerelease packages to be installed. Available since .NET Core 5 SDK
+
+- -s|--source <SOURCE>
+
+The URI of the NuGet package source to use during the restore operation.
+
+- -v|--version <VERSION>
```
chrisraygill, line 85: Can you specify version for a glob of packages as well?

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+Version of the package. See NuGet package versioning.
+
+- --dry-run
```
chrisraygill, line 89: Will this display the changes in transitive packages as well?

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+    [-n|--no-restore] [--package-directory <PACKAGE_DIRECTORY>]
+    [--prerelease] [-s|--source <SOURCE>] [-v|--version <VERSION>]
+    [--dry-run] [--highest-minor] [--highest-patch]
+
+dotnet update package -h|--help
+```
+
+#### Arguments
+
+- PROJECT | SOLUTION
```
chrisraygill, line 43: Is there way to specify a directory of projects? i.e. I want to update all my packages in my \test directory?

chrisraygill (commented):
Looks awesome overall! Main thing it's missing is example outputs.

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Still thinking through it. Any suggestions?

proposed/2022/DotnetUpdatePackage.md
JonDouglas: We will need to lock down the update strategies first I think. In general, I don't think so unless people have a compelling reason to do this!

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Top-level for now since those would be the only "affected" packages per-say. But let's keep this open for more perspectives.

proposed/2022/DotnetUpdatePackage.md
JonDouglas: We'll need to limit the scope. So let's start with project or solution. Updating the world would be amazing, but very complex I think!

proposed/2022/DotnetUpdatePackage.md
chrisraygill: I remember having a conversation with @nkolev92 about how some packages need to have their version kept in sync, particularly in ASP.NET Core world. I believe a lot of those packages also share prefixes like these:

image

Azure packages are also versioned similarly:

image

In these cases `dotnet projectA update package Azure.Storage.* 12.10.0` might be pretty handy

proposed/2022/DotnetUpdatePackage.md
chrisraygill: IMO, top-level only by default but have a --verbosity option to include more info.

proposed/2022/DotnetUpdatePackage.md
chrisraygill: Is it? This design already necessitates the ability to "find a project" within the current directory when the `<PROJECT>` isn't specified. Might be oversimplifying, but isn't the functionality I'm describing just a loop to do that for all projects recursively within a directory?

Conceivably, one of the biggest users of this feature will be VS Code .NET devs who might not have a solution file to target. However, we know that many users prefer to keep all of their packages across all projects at the latest stable version.

It also adds a nice algorithmic component for an intern 😉

proposed/2022/DotnetUpdatePackage.md
JonDouglas: That should be supported with current spec right?

proposed/2022/DotnetUpdatePackage.md
JonDouglas: I believe it's possible. For example if you ran the command on the repo folder, it should work recursively.

proposed/2022/DotnetUpdatePackage.md
chrisraygill: Yes, unless explicitly disallowed. I just wanted to clarify the intention 🙂

proposed/2022/DotnetUpdatePackage.md
chrisraygill: Based on this line:

> The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution or project is found, an error is thrown.

The current design will error out if you run it on the repo folder because multiple projects will be found. My expectation would be that it would run recursively/ apply to all found projects.

proposed/2022/DotnetUpdatePackage.md
chrisraygill: IMO, it shouldn't be all or nothing by default - but may want to provide an option for it if users want to take advantage of the exit code - but that's a secondary need.

If I have a project with an older framework where the latest version of half my packages still support my TFM, but the latest version of the other half don't, an all or nothing design would prevent me from bulk updating altogether.

Based on feedback about version preferences from the HaTS survey, I think the typical user will likely want to update all compatible packages to the latest stable version, even if other will fail. I think it's probably a fairly small subset that will take advantage of the exit code for automation purposes.

Maybe `--atomic` or `--disable-partial` would make sense as an option for all or nothing 🤷‍♂️

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+- PACKAGE_NAME
+
+The package reference to add. Package name wildcards should be supported and update a subset of a glob-style pattern package name. i.e. `Microsoft.*` would update Microsoft packages. The commonly used symbols to support are:
+
+- `*` - zero or more characters.
+- `?` - single occurrence of any character.
+- `.` - literal "." character
+
+#### Options
```
chrisraygill, line 55: Any plan to include a `--compatible` or have it automatically update packages to the latest compatible version if the package latest stable version doesn't support the TFM?

It's referenced in this comment: https://github.com/NuGet/Home/issues/4103#issuecomment-802598457

It's probably beyond the scope an initial release, but a maybe a cool idea 🤷‍♂️

JonDouglas: Not yet. We don't have compat client tooling yet outside of one field. Once we can use the server side stuff, perhaps we can light up this.

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Let's change this language then!

Craige:
> To ensure compatibility with existing package definitions in various files, we should not overwrite or simplify XML elements, but rather only update package version attributes.

To be clear, please do not update the version string in the csproj file. This is what lock files are for. What I set in csproj should be a resolvable version constraint, not an exact pinned version.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+#### Arguments
+
+- PROJECT | SOLUTION
+
+The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution or project is found, an error is thrown.
+
+- PACKAGE_NAME
+
+The package reference to add. Package name wildcards should be supported and update a subset of a glob-style pattern package name. i.e. `Microsoft.*` would update Microsoft packages. The commonly used symbols to support are:
```
nkolev92, line 49: Do we know if there's an ask for this at this point?
Would a list of package ids be preferable instead?

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+- `*` - zero or more characters.
+- `?` - single occurrence of any character.
+- `.` - literal "." character
+
+#### Options
+
+- -f|--framework <FRAMEWORK>
```
nkolev92, line 57: This becomes tricky quickly in the solution case. 

For example, what if a project in a solution does not target that exact framework?

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+Adds a package reference only when targeting a specific framework.
```
nkolev92, line 59: Probably meant to say update.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+Allows the command to stop and wait for user input or action. For example, to complete authentication.
+
+- -n|--no-restore
+
+Adds a package reference without performing a restore preview and compatibility check.
+
+- --package-directory <PACKAGE_DIRECTORY>
+
+The directory where to restore the packages. The default package restore location is %userprofile%\.nuget\packages on Windows and ~/.nuget/packages on macOS and Linux. For more information, see Managing the global packages, cache, and temp folders in NuGet.
```
nkolev92, line 75: I wonder if we can not add this.
I know add package has it, but I'm wondering how commonly it's used if at all. 

Specifying the global packages folder makes more scene in commandline restore scenarios that add/update imo.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+- --prerelease
+
+Allows prerelease packages to be installed. Available since .NET Core 5 SDK
```
nkolev92, line 79: nit: Copied from add package? :D Probably remove the last part.

proposed/2022/DotnetUpdatePackage.md
nkolev92: Would it be preferrable to specify a list of ids instead?

proposed/2022/DotnetUpdatePackage.md
nkolev92: I think we should start with top level ones, but we can consider adding it later. 

It's probably worth considering whether we do deltas or maybe just display the new list of packages.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+
+The URI of the NuGet package source to use during the restore operation.
+
+- -v|--version <VERSION>
+
+Version of the package. See NuGet package versioning.
+
+- --dry-run
+
+Displays what would be updated, but doesn't actually do the operation.
```
nkolev92, line 91: Default is major? 

Wonder if we need an option that takes parameters instead?

proposed/2022/DotnetUpdatePackage.md
nkolev92: I think `update by tfm compatibility` is a special scenario. 
It's one of those few scenarios where a failure is that exact package only. 

The trickier one is what if the latest versions causes resolution conflicts, such as downgrades etc. 
It becomes a user decision what they'd want to do then, so something semi interactive.  
`--no-restore` is a an option that *skips* validations.
Right now the most common failures at, NU1605, NU1105, NU1201 and NU1104.

Back to the compat scenario, I think it's the one that we're most likely to solve. Worth noting that we'd have performance/implementation challenges for any source that isn't nuget.org.
Right now, search by tfm is a search thing only, we'd need to validate the perf issues. 

tldr; I feel the pain of the customers, I run into some of these myself, but solving this in the first iteration might be a reach.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+- [RubyGems update](https://rubygems.org/gems/rubygems-update)
+- [dart pub upgrade](https://dart.dev/tools/pub/cmd/pub-upgrade)
+- [pip install -U](https://pip.pypa.io/en/stable/cli/pip_install/)
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should this command use `dotnet add package` under the hood for the scenarios that make sense such as finding the latest version of a specific package name?
```
nkolev92, line 171: I think that's a technical question that'll likely be accounted for by the implementer. 
These 2 commands will share a bunch of common functionality.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+- What verb makes most sense? `update`, `upgrade`, `refresh`, etc? Update seems the most consistent for package managers such as cargo, npm, rubygems, and NuGet. Upgrade seems the most consistent for package managers such as pip.
+- Should `--highest-minor` and `--highest-patch` be replaced with a single `--dependency-version <version>` parameter to account for more scenarios like `Lowest(default)` and `Highest`?
```
nkolev92, line 173: The `Lowest` vs `Highest` is a larger feature than the update command itself. 
it makes it difficult for me to suggest a name based on what that feature may turn out to be.

proposed/2022/DotnetUpdatePackage.md
nkolev92: How many folders deep would you search for projects? 
All of it? 

I think deciding the `target` of a command should be something we involve the SDK team on given that `package` is a first class noun in dotnet.exe.

nkolev92 (commented):
This proposals brings up all the discussion points that we'll need to have when we implement this features.

Thanks for the initial write-up @JonDouglas 

Most of my comments are doc readiness, but a good amount are about the functionality itself as well.

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Yes for consistency sake!

proposed/2022/DotnetUpdatePackage.md
JonDouglas: The primary ask is updating all packages. The secondary ask is having control of updating. A list of package IDs can be preferable for sure. This shorthand might be helpful in our ecosystem given prefix/source mapping.

proposed/2022/DotnetUpdatePackage.md
JonDouglas: @baronfel Any idea how we can get more eyes from SDK team on these types of things? :)

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Sadly we don't have much of a signal of how much this is used. Does this still make sense for update scenario?

proposed/2022/DotnetUpdatePackage.md
JonDouglas: Open to perspectives! Right now yes, major default with two flags for minor and patch.

proposed/2022/DotnetUpdatePackage.md
baronfel: You can unfortunately can't tag @dotnet/dotnet-cli, so I'll poke a few here: @dsplaisted, @joeloff, @gkulin

I'd generally be in favor of globs/patterns for more SDK operations - I know a similar capability has been requested for many of the MSBuild-driving commands in the past.  It's common to use globs for this kind of thing - `**/*Test.*proj` to grab every test project in a solution recursively, for example.

proposed/2022/DotnetUpdatePackage.md
```diff
@@ -0,0 +1,179 @@
...
+dotnet update package -h|--help
+```
+
+#### Arguments
+
+- PROJECT | SOLUTION
+
+The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution or project is found, an error is thrown.
+
+- PACKAGE_NAME
```
baronfel, line 47: going to throw out a crazy idea here that might consolidate a few flags/options.

`PACKAGE_NAME` can be
* pattern as described below, or
* a list of 'package with version spec's, where 'package with version spec' means a string of the format `{PACKAGE_NAME}@{npm-style version expression}`

The NPM(and also gem/paket/etc)-style expressions allow you to easily express bounds for the update operation on a package-by package basis, unambiguously, in a way that `PACKAGENAME -version VERSION` can't scale to.  It also removes the need for `--higest-minor`, `--highest-patch`, and `--prerelease`, as those can be specified with appropriate syntax in the correct positions.

This does come with a downside of a) documenting these and b) writing code to parse them, however.

proposed/2022/DotnetUpdatePackage.md
OskarKlintrot: Another "fun" case when updating by tfm compatibility that I run in to is that dotnet-ef supports something like dotnet core 2.1 in the latest version so if you update everything to the latest compatible version the EF Core packages might get out-of-sync. I really whish to be to update all packages to the latest compatible version but it's not as easy as it might seem.

Frulfump:
What's the status of this, some of it is implemented I think?

zivkan:
I need to review the PR and see what ideas it proposed that are missing from the already implemented `dotnet package update` command, and create issues to track them.  But this PR will get closed, as https://github.com/NuGet/Home/blob/dev/accepted/2024/package-update-command.md supersedes this PR.

Frulfump:
> 

See previous comment. Can we get the do not auto close label please?

Frulfump:
Waiting for review and then this PR getting closed as per https://github.com/NuGet/Home/pull/11812#issuecomment-4027439686 but if we let it get closed it will most likely fade into the ether

zivkan:
Oops, I forgot to close this out.  When I reviewed, I/copilot found two features this spec proposed without tracking issues, so I created them:

* https://github.com/NuGet/Home/issues/14870
* https://github.com/NuGet/Home/issues/14871

Copilot said there was at least one other feature this spec proposed that is not implemented in the current `dotnet package update`, but there was already an issue for it. Unfortunately I can't remember what issue that is.

Maybe we should consider adding an `Area:DotnetPackageUpdate` label, to make it easier to find issues specific to it, but for now what I do is [put "dotnet package update" in quotes, and also use the `in:title` filter](https://github.com/NuGet/Home/issues?q=is%3Aissue%20state%3Aopen%20in%3Atitle%20%22dotnet%20package%20update%22)

Frulfump:
> Oops, I forgot to close this out. When I reviewed, I/copilot found two features this spec proposed without tracking issues, so I created them:
> 
>     * [Add `--dry-run` to `dotnet package update` #14870](https://github.com/NuGet/Home/issues/14870)
> 
>     * [Add `--deprecated` option to `dotnet package update` #14871](https://github.com/NuGet/Home/issues/14871)
> 
> 
> Copilot said there was at least one other feature this spec proposed that is not implemented in the current `dotnet package update`, but there was already an issue for it. Unfortunately I can't remember what issue that is.
> 
> Maybe we should consider adding an `Area:DotnetPackageUpdate` label, to make it easier to find issues specific to it, but for now what I do is [put "dotnet package update" in quotes, and also use the `in:title` filter](https://github.com/NuGet/Home/issues?q=is%3Aissue%20state%3Aopen%20in%3Atitle%20%22dotnet%20package%20update%22)

Ah no worries, thanks! Area label sounds good and maybe even an epic to track all of them? Appreciate the search query in the meantime

Issue #11547, opened 2022-02-02
[Feature]: Show updates/issues in VS Solution Explorer for packages


timheuer:
### NuGet Product(s) Involved

Other/NA

### The Elevator Pitch

When working within VS, the packages I use may have updates or may have issues (security issues, deprecations, etc) that I may want to be aware of.  Right now the only way to see that is launching the Package Management UI, but it isn't a normal thing I would do if nothing is alerting me to do that.

I propose that the Packages (and thus roll-up to the dependency node) node in VS should have different states for the packages listed...some examples:

- ⚠️ Warning: you should check this out and the hover tooltip would tell you something "Packages have information"
- ⬆️ or ℹ️ Information/Upgrade: your packages have updates available
- ⛔ Error: for security check or deprecation message (tooltip would show)

Example:
![image](https://user-images.githubusercontent.com/4821/152074883-43644453-f288-4bb6-9396-9a01752946a3.png)

### Additional Context and Details

Part of this exists in VSMac already today.

Issues: 

- https://github.com/NuGet/Home/issues/3292
- https://github.com/NuGet/Home/issues/8956

nkolev92:
Thanks for filling this. 

Duplicate of https://github.com/nuget/home/issues/8956.

timheuer:
@nkolev92 problem is the dupe explicitly says this is not in scope. So when the dupe is resolved then this will still be a request

nkolev92:
Sure, I can reopen. 

We'll likely consolidate at some point, but for now, I'll let it go through triage on its own.

zivkan:
@timheuer I had a brief chat with @drewnoakes (from project system team) about this. He tells me that the project system's dependencies tree doesn't have the capabilities to achieve this (although it's not clear to me if it's a dotnet/project-system feature, or CPS). In any case, this is going to take cross team collaboration and scheduling of priorities to achieve. I don't know who should take the lead, but FYI given you're a VS PM.

timheuer:
@zivkan Understood -- but gotta start somewhere who owns the customer :-)...give us the requirements 😜

zivkan:
triage: @chgill-MSFT to talk to CPS team about scheduling work to provide extensibility for overlay icons

JonDouglas:
I've drafted an initial spec with motivation and scenarios here:

https://github.com/NuGet/Home/pull/11838

nkolev92:
[...]

Issue #3292, opened 2016-08-10
Add NuGet Package update indicators in Solution Explorer


SupporterMan:
Now we need to open the NuGet Package Manager to check whether there has any package updates for the installed packages. Is it possible to let the updates notification added into the Visual Studio Notifications windows, which is more easily be found?

nkolev92:
Not sure if we want to make this such a prominent thing, as it might be interpreted as NuGet recommending this update.

The VS extensions updates are normally recommended changes and not expected to break the experience. 
Package update can break the experience and until we have a better package compatibility story, I don't think this should be added. 

//cc @rrelyea @anangaur 
Thoughts here?

clabough:
@nkolev92 Clicking on the notification should only take you to the NuGet -> Updates tab so then the user can determine whether or not to apply the update. It would be even better if this could be set per package because there are some that I never update - and I wouldn't want to be notified that those are still outdated - but others I update as soon as I can.

nkolev92:
> Clicking on the notification should only take you to the NuGet -> Updates tab so then the user can determine whether or not to apply the update

Regardless of the action, VS notifications are used for higher priority things which basically say "hey, you should do this". That's not always correct with packages.

clabough:
@nkolev92 Yes, I agree. I would take the notification as "hey, you should check your NuGet available updates".

jepl4052:
Some sort of notification of available updates would be really handy. Currently I'm working in a microservices architechture where each microservice provides a NuGet package with some models, interfaces etc. Whenever one microservice has its NuGet package update with a new release, it would be super handy if there was a possibility in a depending microservice to be able to setup a custom "notification/subscription" or similiar on at least some specified NuGets.

anangaur:
@jepl4052 Since we are talking about VS here, we have the [updates tab on NuGet Package Manager](https://docs.microsoft.com/en-us/nuget/consume-packages/install-use-packages-visual-studio#update-a-package) that has a notification notion to let users know if updates are available.
![image](https://user-images.githubusercontent.com/14800916/65628734-e27d8780-df86-11e9-8724-facf87a2feb3.png)

I think the idea is that developers should be aware of package updates without having to open the NuGet package manager. We have a way to [show available updates on VS Mac](https://docs.microsoft.com/en-us/visualstudio/mac/nuget-walkthrough?toc=%2Fnuget%2Ftoc.json&view=vsmac-2019#updating-packages). 
[...]

PR #11838, opened 2022-05-26
Package Status Indicators in Visual Studio Solution Explorer

JonDouglas:
Introduction to an experience in Visual Studio in which a developer can view the current package statuses in the solution explorer.

This proposal introduces a concept of package status indicators in Visual Studio.

[Rendered Spec](https://github.com/NuGet/Home/blob/package-status-indicators/proposed/2022/package-status-indicators.md)

Thank You 🎉


proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+Outdated packages would still show the indicators if the packages node is expanded however.
+
+Currently, NuGet only uses two icons to display status indicators:
+
+- A warning triangle (vulnerability and deprecation).
+![](../../meta/resources/PackageStatusIndicators/StatusWarning.png)
+- An up arrow (outdated).
+![](../../meta/resources/PackageStatusIndicators/StatusUpdateAvailable.png)
+
+Hovering over the package and icon will display the respective information for the status.
```
drewnoakes, line 63: Supporting tool tips in Solution Explorer would require work across three teams: in Solution Explorer (in its hierarchy provider), in CPS and in the .NET Project System.

Currently only tool tip text is displayable for source control provider status. The APIs would need to be extended to allow other kinds of data to be included in tool tips.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+
+- **Deprecated** - Please don't use this for a disclosed reason and please use a suggested alternative(if any).
+- **Vulnerable** - Please don't use this due to a known vulnerability and update/replace/remove/do nothing accordingly.
+- **Outdated** - Please consider using a newer version.
+
+Building upon Tim's initial mockup, there should be a UI affordance that demonstrates the truth of each package's status in the solution explorer representing these three words.
+
+![](../../meta/resources/PackageStatusIndicators/package-status-indicators.png)
+
+For the sake of not annoying developers, we should only bubble up a warning affordance to the top-level node (Packages and Dependencies) if there is a known top-level vulnerability or deprecated package in their package graph.
```
drewnoakes, line 52: The current bubble-up logic is based on priority. At each level, the "maximum" level's icon is shown. The order is: error, warning, none. Doing something different here would require some rewriting within the dependencies tree.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+- An up arrow (outdated).
+![](../../meta/resources/PackageStatusIndicators/StatusUpdateAvailable.png)
+
+Hovering over the package and icon will display the respective information for the status.
+
+- Deprecated - "This package has been deprecated as it is `<deprecation category>`."
+- Vulnerable - "This package has at least one vulnerability with `<severity category>` severity."
+- Outdated - "This package has version `<latest version>` available.
+
+When right clicking a package, a developer should be able to get to the "Manager NuGet Packages..." context menu action like they can with the "Packages" and "Dependencies" nodes.
```
drewnoakes, line 69: It would be great if we could have more targeted gestures in these menus for packages. "Manage" is vaguer than "See updates" or "See vulnerabilities".

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+
+Because this feature may need some experimentation to get the right signal from the noise and the potential for exponential transitive-level vulnerabilities, developers will need a way to opt-out from the experience as a whole, and options to opt-in to varying levels of information.
+
+This may need to include a new options panel inside the `Projects and Solutions > SDK-Style Projects` regarding the Package Settings where there's a top-level checkbox for the feature, and either a drop-down indicating a package status level or multiple checkboxes for each status mentioned above.
```
drewnoakes, line 73: We receive regular requests to make the Dependencies tree general purpose for other kinds of projects, so I try to not think of it as belonging to SDK-style projects, despite the .NET Project System team owning the feature. That said, I don't know offhand of a better options page for such a setting, nor do I think we want to add another page for just a single option.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+<!-- ### Technical explanation -->
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
+While developers continue to seek this information in more locations, it can also be quite a bit of noise towards the general development experience. Developers tend to compare visual issues in the solution explorer akin to ["DLL Hell"](https://en.wikipedia.org/wiki/DLL_Hell) which we would like to avoid the perception as much as possible while providing valuable information about dependencies.
+
+This feature can potentially collide with existing experiences in which package dependencies may not exist on disk and/or assets file are incorrect and require a restore to resolve. There can also be false positives/negatives with the current state of the project system and the need to reload projects to refresh the package status in realtime.
```
drewnoakes, line 84: Could you explain more about false positives that would require reloading projects? I'm not aware of such a situation, for projects opened by the .NET Project System at least.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+- NuGet.org, dotnet CLI, and NuGet Package Manager inside Visual Studio provide much of this functionality already, but they are multiple clicks and commands away from developers seeing them.
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+- NuGet provides status indicators in the NuGet Package manager experience in Visual Studio.
+- Visual Studio for Mac provides "outdated" status indicators in the solution explorer.
```
drewnoakes, line 102: Can we include a screenshot of how this looks in VS Mac?

proposed/2022/package-status-indicators.md
drewnoakes: I see you've called this out under "Future Possibilities" below.

drewnoakes (commented):
I wasn't able to make today's meeting as it was scheduled outside of my working hours (I'm in Australia) and I haven't yet watched the recording, so apologies if these have already been discussed.

JonDouglas:
Moving this proposal to ready as @mckennabarlow is currently working on it.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+- Vulnerable - "This package has at least one vulnerability with `<severity category>` severity."
+- Outdated - "This package has version `<latest version>` available.
+
+When right clicking a package, a developer should be able to get to the "Manager NuGet Packages..." context menu action like they can with the "Packages" and "Dependencies" nodes.
+
+Because this feature may need some experimentation to get the right signal from the noise and the potential for exponential transitive-level vulnerabilities, developers will need a way to opt-out from the experience as a whole, and options to opt-in to varying levels of information.
+
+This may need to include a new options panel inside the `Projects and Solutions > SDK-Style Projects` regarding the Package Settings where there's a top-level checkbox for the feature, and either a drop-down indicating a package status level or multiple checkboxes for each status mentioned above.
+
+<!-- ### Technical explanation -->
```
nkolev92, line 75: I'd love to see a technical explanation that focuses on the feasibility of some of these asks. That might allow to define the stages for this work.

Vulnerabilities are going to always be up to date. 

What happens with deprecation & update, which are expensive to calculate for example.

Things I'm curious about. 

- How often are they going to be done? How frequent should the refresh be?
- Any correctness concerns, given that this simply can't happen at every restore, it'd be an expensive perf hit right now.

proposed/2022/package-status-indicators.md
```diff
@@ -0,0 +1,120 @@
...
+
+Hovering over the package and icon will display the respective information for the status.
+
+- Deprecated - "This package has been deprecated as it is `<deprecation category>`."
+- Vulnerable - "This package has at least one vulnerability with `<severity category>` severity."
+- Outdated - "This package has version `<latest version>` available.
+
+When right clicking a package, a developer should be able to get to the "Manager NuGet Packages..." context menu action like they can with the "Packages" and "Dependencies" nodes.
+
+Because this feature may need some experimentation to get the right signal from the noise and the potential for exponential transitive-level vulnerabilities, developers will need a way to opt-out from the experience as a whole, and options to opt-in to varying levels of information.
```
nkolev92, line 71: This sounds like a perfect candidate to add in the preview features in VS.

proposed/2022/package-status-indicators.md
nkolev92: Is this where we'd want to open on the `Installed` tab and and have the package in question selected potentially?

nkolev92:
@JonDouglas 
https://github.com/NuGet/Home/issues/3292 and https://github.com/NuGet/Home/issues/11547 are almost the same thing. 
Should we close one in favor of the other one? 

btw, it'd be great to have the issues assigned to the right folks :)

proposed/2022/package-status-indicators.md
drewnoakes: That's the kind of thing I was thinking, yeah. It'd be useful regardless of the status indicators. Today, right clicking on a package node doesn't provide any way to get to the package manager. If you click on "Packages" you get "Manage NuGet packages". It'd be great to have a command on individual package nodes that opens the package manager with the package pre-selected.

proposed/2022/package-status-indicators.md
drewnoakes: Data about deprecations/vulnerabilities/updates/etc could be handled async relative to the currently project data flow. In VS, the total set of packages in the solution could be gathered and this information queried. Once available, it could be pushed to projects, which could then update project state. 

If the data was cached, then solution-load could efficiently load the data during solution load so that we show that state immediately.

The component that manages that state could also be responsible for keeping it updated via whatever means makes sense, such as polling a remote server.

One area I'm curious about is vulnerable transitive dependencies. The project references A which references B, and B is vulnerable. We would want to show A as vulnerable too. Therefore this component would need to model dependencies between packages and push any such state back up the graph to referencing packages, recursively.

nkolev92: > One area I'm curious about is vulnerable transitive dependencies. The project references A which references B, and B is vulnerable. We would want to show A as vulnerable too. Therefore this component would need to model dependencies between packages and push any such state back up the graph to referencing packages, recursively.

Transitive dependencies is definitely an area where we need lots of work. 
By default the .NET 8 SDK will report on directs, but transitive dependencies graphs can be really large.

drewnoakes: Given there are commonly many more indirect dependencies than direct dependencies, I wouldn't be surprised if the majority of exposure to vulnerable packages occurs through indirect dependencies.

nkolev92:
Just tidying up here improving the naming :)

JonDouglas:
idk why this was closed. stupid robots

LockTar (approved):

Frulfump:
Still waiting for this!

Frulfump:
> idk why this was closed. stupid robots

😂 . Not stale

Frulfump:
Unsure what the actual status is, some of it is possible to see in the packages node, not sure if the warning icon without any explanation is still there.

Frulfump:
Not stale, would be ideal to at least get a cursory review again

Frulfump:
Might be vacation period so bumping a bit again. Still want this!

Frulfump:
Still something that would be beneficial for the users.

Frulfump:
Hoping this will eventually get some love

Issue #11807, opened 2022-05-10
[Feature]: dotnet.exe add package support for projects onboarded onto central package management


kartheekp-ms:
### NuGet Product(s) Involved

dotnet.exe

### The Elevator Pitch

The [dotnet add package command](https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-package) provides a convenient option to add or update a package reference in a project file. If the repo has been onboarded onto [`Central Package Management`](https://docs.microsoft.com/en-us/nuget/consume-packages/central-package-management) feature then the said command should add package reference to project (no version) and add a package version to the props file. Few scenarios to consider during implementation.
- If the version already exists in the. props file then the command should update it.
- If no version is specified for add package command, then add latest version to CPM props file.
- In case of update the tooling changes the version number in .props file but no edits should be made to the project file.

### Epic details
- [x] https://github.com/NuGet/Home/issues/11858
- [x] https://github.com/NuGet/Home/issues/11890

pragnya17:
When using the CLI command `dotnet add package` on a project that is already onboarded to Central Package Management the following error is thrown: `error: NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items: [PackageName].`

The aim of this project will be to resolve this by enabling the dotnet CLI command to be used, regardless of whether the project has been onboarded to CPM or not.

darthkurak:
When we can expect a release of this functionality? :)

kartheekp-ms:
> When we can expect a release of this functionality? :)

cc @martinrrm

constantinmuraru:
Hey @martinrrm @kartheekp-ms - is this working? I see a few PRs merged, but can't seem to make it work with `dotnet 7.0.102`:

https://stackoverflow.com/questions/75213427/how-to-make-dotnet-add-work-with-central-package-management

When I run `dotnet add`, a `PackageReference` is added in the .csproj file, which contains the version.
Example:
```xml
<PackageReference Include="Microsoft.EntityFrameworkCore" Version="7.0.2" />
```

This upsets CPM on `dotnet restore`:

> error NU1008: Projects that use central package version management should not define the version on the PackageReference items but on the PackageVersion items

kartheekp-ms:
[...]

Issue #11516, opened 2022-01-20
[Feature]: Allow overriding a centrally defined package version


jeffkl:
### NuGet Product(s) Involved

NuGet.exe, MSBuild.exe, dotnet.exe

### The Elevator Pitch

Central Package Management (CPM) allows users to specify versions for their packages in one place.  However, there are edge cases where a project might want to use a different version.  For example, a prototype project could need a newer version than a repo is using so it would opt into newer versions.

There is an existing implementation, Microsoft.Build.CentralPackageVersions, which uses a `VersionOverride` attribute to indicate that a `<PackageReference />` should use a different version than is centrally defined.

This design has a few advantages:
1. Makes it obvious when a project is using its own package version(s)
2. Ideally, an entire repo would use the same package version which is why users would define it in one place.  Users who want to use a different version would need to fully understand the ramifications of specifying the `VersionOverride` attribute.
3. Migration from Microsoft.Build.CentralPackageVersions would be easier.

Other options for users that would still be possible are:
1. Specify a `<PackageVersion />` in a project
2. Define an MSBuild property that represents the package version and set it in a project

Since MSBuild essentially lets users do anything they want, the built-in mechanism for overriding a package version should be simple and well documented.  

# Design

A user would specify package versions in `Directory.Packages.props`:
```xml
<Project>
  <ItemGroup>
    <PackageVersion Include="PackageA" Version="1.0.0" />
    <PackageVersion Include="PackageB" Version="2.0.0" />
  </ItemGroup>
<Project>
```

A project would override the version with the `VersionOverride` metadata on a `<PackageReference />` item:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="PackageA" VersionOverride="3.0.0" />
  </ItemGroup>
<Project>
```

In this case, the project's restore graph would resolve `PackageA` to version `3.0.0`.  Any project that references it would also get that version and a user would be responsible for handling an unresolved conflicts.

Finally, a repo owner should be able to disable the ability for developers to override package version.  This would be used for instance if someone wanted to ensure that all package versions are unified.  This would be possible by setting the MSBuild property `EnablePackageVersionOverride` to `false` in a project or import like Directory.Build.props:

```xml
<Project>
  <PropertyGroup>
    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
[...]

PR #11849, opened 2022-06-01
Proposal: Adding dotnet CLI support for projects onboarded to CPM

pragnya17:
Tracking issue https://github.com/NuGet/Home/issues/11858

[Rendered Design Spec](https://github.com/NuGet/Home/blob/dev-pragnya17-CPMDotnetCLISupport/proposed/2022/CPMDotnetCLISupport.md)

Proposal for [11807](https://github.com/NuGet/Home/issues/11807)


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=11849) <br/>All CLA requirements met.

erdembayar:
@pragnya17 
Your spec looks great. Probably need to rename to `cpm-dotnet-cli-add-package-support.md` or something like that to express what problem it's solving, otherwise `CPMDotnetCLISupport.md` is too generic. Also, please remove image files since you're using embedded codes.

proposed/2022/CPMDotnetCLISupport.md
```diff
@@ -0,0 +1,134 @@
...
+2. Project2 will evaluate the Directory.Packages.props file in the Repository\ directory.
+
+***Sourced from https://devblogs.microsoft.com/nuget/introducing-central-package-management/
+
+### Other Scenarios that can be considered
+
+While the above cases must be considered to solve the issue, there are a few other cases that can be considered to make the product more usable:
+
+1. If the project has the <ManagePackageVersionsCentrally> set to true, but the `Directory.packages.props` file has not been created yet the file should be created. At this point, package references can be updated according to the cases discussed above.
+2. If the `Directory.packages.props` file has been created but the <ManagePackageVersionsCentrally> property has not been set to true, it should be set to true.
```
erdembayar: In my opinion, it's out of scope unless there is special option to enable CPM, most likely it's cause ton of error if customer already have packagereferences with version in csproj files.

proposed/2022/CPMDotnetCLISupport.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Unresolved Questions
+1. If there are `Directory.packages.props` files inherit other props files, which props file should be modified when `add package` is executed?
```
erdembayar: Most likely it would be the closest one enabled CPM. Of course, calculating that one could be tricky.

jeffkl (approved):

proposed/2022/CPMDotnetCLISupport.md
```diff
@@ -0,0 +1,134 @@
...
+        <Nullable>enable</Nullable>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <PackageReference Include="Newtonsoft.Json"/>
+    </ItemGroup>
+</Project>
+```
+
+### 2. The package reference does exist
```
kartheekp-ms: @jeffkl - What if the package has version override defined in the `.csproj` file? I guess the command should update the `VersionOveride` attribute instead of `Directory.packages.props` correct?

Incase if team members need more information please refer to https://github.com/NuGet/Home/issues/11516

proposed/2022/CPMDotnetCLISupport.md
```diff
@@ -0,0 +1,134 @@
...
+1. Project1 will evaluate the Directory.Packages.props file in the Repository\Solution1\ directory.
+2. Project2 will evaluate the Directory.Packages.props file in the Repository\ directory.
+
+***Sourced from https://devblogs.microsoft.com/nuget/introducing-central-package-management/
+
+### Other Scenarios that can be considered
+
+While the above cases must be considered to solve the issue, there are a few other cases that can be considered to make the product more usable:
+
+1. If the project has the <ManagePackageVersionsCentrally> set to true, but the `Directory.packages.props` file has not been created yet the file should be created. At this point, package references can be updated according to the cases discussed above.
```
kartheekp-ms: My understanding is that GitHub ignores content specified in "<>" brackets. Hence the property name was not rendering in the markdown file.

``` suggestion
1. If the project has the `ManagePackageVersionsCentrally` set to true, but the `Directory.packages.props` file has not been created yet the file should be created. At this point, package references can be updated according to the cases discussed above.
```

proposed/2022/CPMDotnetCLISupport.md
pragnya17: After discussing with Jeff, we decided it would be best to consider all imported files when adding package references. This way if there is a package reference in an imported file that needs to be updated, then it can be updated accordingly.

proposed/2022/CPMDotnetCLISupport.md
pragnya17: I agree that this is probably out of the scope of my project, but these are just additional considerations I thought would be worth mentioning after discussing with Jon and Kartheek.

proposed/2022/CPMDotnetCLISupport.md
pragnya17: Updated the spec accordingly, thanks for noticing it!

proposed/2022/CPMDotnetCLISupport.md
jeffkl: Yes good point.  If `VersionOverride` is specified, then we should just update that.

pragnya17: Ok, the spec has been updated accordingly. Thank you!

pragnya17:
The spec has been reviewed and changes have been made according to the feedback received.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -0,0 +1,134 @@
...
+    <PropertyGroup>
+        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    </PropertyGroup>
+    <ItemGroup>
+    <PackageVersion Include="Newtonsoft.Json" Version="12.0.1"/>
+    </ItemGroup>
+</Project>
+```
+
+#### After `add package` is executed:<br>
```
nkolev92, line 111: I think this behavior might not be super obvious to the end user. 

If I am already managing a package centrally, it feels like the version that should be installed is the existing one, and ideally not update it. 

Version updates should only happen explicitly or through an update command.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -0,0 +1,134 @@
...
+    <PropertyGroup>
+        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    </PropertyGroup>
+    <ItemGroup>
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.1"/>
+    </ItemGroup>
+</Project>
+```
+
+In case there are multiple `Directory.packages.props` files in the repo, all of them should be searched for a matching package reference. Once a matching package reference is found, its package version can be updated appropriately.
```
nkolev92, line 123: I imagine the intention here is if multiple Directory.Packages.props file apply correct? 

This is something that the user can only do manually.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -0,0 +1,134 @@
...
+    </PropertyGroup>
+    <ItemGroup>
+    <PackageVersion Include="Newtonsoft.Json" Version="13.0.1"/>
+    </ItemGroup>
+</Project>
+```
+
+In case there are multiple `Directory.packages.props` files in the repo, all of them should be searched for a matching package reference. Once a matching package reference is found, its package version can be updated appropriately. 
+
+### Other Scenarios that can be considered
```
nkolev92, line 125: This is probably something best handled in a `Future possibilities` section.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -0,0 +1,134 @@
...
+
+While the above cases must be considered to solve the issue, there are a few other cases that can be considered to make the product more usable:
+
+1. If the project has the `<ManagePackageVersionsCentrally>` set to true, but the `Directory.packages.props` file has not been created yet the file should be created. At this point, package references can be updated according to the cases discussed above.
```
nkolev92, line 129: It's worth calling out what would actually happen here today, even if we're not promising/committing to  the improvements themselves.

nkolev92 (commented):
Nice job @pragnya17!

It would be great if we can clarify the second scenario about the updating package version. 

I wonder if @JonDouglas had a chance to look over that. 

I also have some minor spec comments.

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: We went down this path because `dotnet add package` command currently updates the package version if the package refence already exists. 

If it is not ideal for `dotnet add package` command to update the package version then should we raise a warning instead of updating the version in the `directory.packages.props` file for projects onboarded onto CPM?

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: > I imagine the intention here is if multiple Directory.Packages.props file apply correct?

You are correct. Incase if others are curious, as per [`Central Package Management rules`](https://devblogs.microsoft.com/nuget/introducing-central-package-management/#central-package-management-rules), only the closest `Directory.Packages.props` file is evaluated for a given project. 
```
Repository
 |-- Directory.Packages.props
 |-- Solution1
     |-- Directory.Packages.props 
     |-- Project1
 |-- Solution2
     |-- Project2
```
However, I think customers can specify an `Import` statement in `Solution1\Directory.Packages.props` file to import properties from `Repository\Directory.Packages.props`

```
Repository
 |-- Directory.Packages.props
 |-- Solution1
     |-- Directory.Packages.props (specifies an import statement pointing to the props file in the parent folder)
     |-- Project1
 |-- Solution2
```
The current spec suggests that `dotnet add package` command should walk through the hierarchy of `directory.packages.props` file (using `Import` statements) and update the package version in the appropriate props file.

> This is something that the user can only do manually.

Once we get clarity on https://github.com/NuGet/Home/pull/11849/files#r890672222 conversation then we consider this scenario as out of scope / future possibilities for this work.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -0,0 +1,134 @@
...
+2. If the `Directory.packages.props` file has been created but the `<ManagePackageVersionsCentrally>` property has not been set to true, it should be set to true.
+3. If the project has the `<ManagePackageVersionsCentrally>` set to false, the project should not be considered as onboarded to CPM. Therefore the `Directory.packages.props` file should be deleted if it exists and only `.csproj` files should be modified.
```
kartheekp-ms, line 131:
``` suggestion
3. If the project has the `<ManagePackageVersionsCentrally>` set to false, the project should not be considered as onboarded to CPM. Therefore the `Directory.packages.props` file in the project folder should be deleted if it exists and only `.csproj` files should be modified
```

proposed/2022/cpm-dotnet-cli-add-package-support.md
nkolev92: There might be some implementational complexity with multiple files. 
Example, what if both files have it? 
You'd then need to know where it's coming from, which is probably more complicated than a V1. 

I'd encourage explicitly saying that scenarios with multiple Directory.Packages.props are out of scope for now.

kartheekp-ms: Discussed about this scenario offline with @jeffkl. Yes, scenarios with multiple `Directory.Packages.props` are out of scope for now. Thanks Nikolche. Created a PR with revised proposal https://github.com/NuGet/Home/pull/11873

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: I added a table in the revised proposal https://github.com/NuGet/Home/pull/11873 explaining current and new behavior for every scenario.

PR #11856, opened 2022-06-03
Proposal of consolidating user wide NuGet.Config files on Linux/Mac

heng-liu:
Design for https://github.com/NuGet/Home/issues/4413, https://github.com/NuGet/Home/issues/5555

I could not reopen https://github.com/NuGet/Home/pull/10364 so I create this new PR.

### DIFF
--- a/designs/Consolidate-NuGetConfig_OnLinux.md
+++ b/designs/Consolidate-NuGetConfig_OnLinux.md
@@ -1,0 +1,134 @@
+
+# Spec Name
+
+* Status: In Review
+* Author(s): [Heng Liu](https://github.com/heng-liu)
+* Issue: [4413](https://github.com/NuGet/Home/issues/4413) [MacOS, Linux] NuGet writes the NuGet.Config in ~/.config folder, but dotnet restore reads from ~/.nuget 
+
+## Problem Background
+
+On Linux/Mac, NuGet has inconsistent paths of user wide NuGet.Config file for different NuGet tools. Dotnet CLI uses hard coded ~/.nuget folder (~/.nuget/NuGet/NuGet.Config), while Mono uses the folder from environment variable "APPDATA", (by default, it's ~/.config/NuGet/NuGet.Config). This is confusing and we should consolidate the two into one location. 
+After discussing in NuGet team and checking with Visual Studio Mac/Mono team, we believe we should change the dotnet CLI path to match that of Mono’s, for the following reasons:
+ * On Linux/Mac, users usually get application data from environment variable APPDATA, which has the default value of ~/.config. Getting application data from ~/.nuget will cause inconsistency user experience between NuGet and other applications.
+ * Getting application data from environment variable APPDATA on Linux/Mac will make it consistent with the behavior on Windows.
+
+For the sake of simplicity, we will refer to ~/.nuget/NuGet/NuGet.Config as legacy path, and refer to ~/.config/NuGet/NuGet.Config as consolidated path in the following.
+
+## Who are the customers
+Dotnet CLI users who work on Linux and Mac.
+
+## Goals
+Make dotnet CLI user-wide config file path consistent with Mono on Linux/Mac and break as less dotnet CLI users as possible.
+
+## Non-Goals
+Completely get rid of the legacy path.
+
+## Background knowledge
+On Linux/Mac, user wide NuGet.Config file is created when trying to get the user wide NuGet.Config file at the beginning phase of running dotnet restore command(and many other dotnet commands with implicit restore if not specifying --no-restore option, like dotnet new, dotnet build, dotnet run, dotnet test, dotnet publish, and dotnet pack), or dotnet msbuild -t:restore command, and found the user wide NuGet.Config file does not exist. Then the user wide NuGet.Config will be created with package source of nuget.org, as the only default package source. 
+
+## Solution overview 
+Add following steps when trying to get user-wide config file path by running dotnet CLI on Linux/Mac:
+(If it's not on Linux/Mac, or it's not in .NET core code path, we will not add those steps. So that the mono and dotnet CLI on Windows will not be affected.)
+Check if the following conditions are satisfied.
+•	The legacy path does not exist.
+•	The consolidated path exists.
+If yes, NuGet will get user wide NuGet.Config path from consolidated path.
+If above condition is not satisfied, it should be in one of the following conditions:
+
+1.Neither legacy path nor consolidated exists.
+In this scenario, user has never called NuGet in dotnet CLI or Mono before.
+NuGet will create a default NuGet.Config file in consolidated path.
+NuGet will get user wide NuGet.Config file from consolidated path. 
+
+2.Legacy path exist, but consolidated path does not exist.
+In this scenario, user has called NuGet in dotnet CLI before, but has never called NuGet in Mono.
+NuGet will copy the NuGet.Config file from legacy path to consolidated path, then delete the one in legacy path. 
+NuGet will get user wide NuGet.Config file from consolidated path.
+
+3.Both legacy path and consolidated path exist.
+In this scenario, user has called NuGet in both dotnet CLI and Mono before.
+NuGet could not determine for users to consolidate the contents of the two NuGet.Config files, as there might be conflicting config.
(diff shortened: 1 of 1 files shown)


heng-liu:
Updated since the symbolic APIs are available in net6.0.

jeffkl (approved):

nkolev92 (commented):
This is a great and detailed write-up, nicely done @heng-liu 

As I was going through it, it wasn't obvious what direction we're leaning to? Did I miss it?

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+Make dotnet CLI user-wide config file path consistent with Mono on Linux/Mac and break as less dotnet CLI users as possible.
+
+## Non-Goals
+Completely get rid of the legacy path.
+
+## Background knowledge
+On Linux/Mac, user wide NuGet.Config file is created when trying to get the user wide NuGet.Config file at the beginning phase of running `dotnet restore` command(and many other dotnet commands with implicit restore if not specifying `--no-restore` option, like `dotnet new`, `dotnet build`, `dotnet run`, `dotnet test`, `dotnet publish`, and `dotnet pack`), or `dotnet msbuild -t:restore` command, but the user wide NuGet.Config file does not exist. Then the user wide NuGet.Config will be created with package source of nuget.org, as the only default package source. 
+
+## Potential Solutions overview 
+### Solution 1: Use only consolidated path
```
zivkan: There's another option, very similar to this Solution 1:  Output a warning to the customer when the legacy path exists, and let them take manual action, if they want to. We should include a URL to a docs page in the warning, and that docs page can suggest manually merging the two nuget.config files into one (that was we don't have to write complex code that will try to do it automatically), and suggest setting up a symbolic link if they use global.json.

This has the least effort on our part, allowing us to implement is the most quickly.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+
+NuGet will create a default NuGet.Config file in consolidated path.
+NuGet will get user wide NuGet.Config file from consolidated path. 
+No warning will be shown.
+
+**2. Legacy path exist, but consolidated path does not exist.**
+
+NuGet will copy the NuGet.Config file from legacy path to consolidated path, then delete the one in legacy path. 
+NuGet will get user wide NuGet.Config file from consolidated path.
+No warning will be shown.
```
zivkan: I think an informational message should be displayed is we're moving the config file from a path that older tools use to a path that older tools will not use.  The message might get lost if it's in a long restore log. But silently moving files that might affect other tools doesn't feel like a great idea to me.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+
+**3. Both legacy path and consolidated path exist.**
```
zivkan: As I wrote above, another option is to just warn the customer and let them manually resolve it. Consolodating the two config files might be tricky to ensure we get everything right, but will hopefully be easy for customers to do themselves.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+
+#### Cons: 
+Legacy path still exists for all scenarios, even if the NuGet.Config file is only in consolidated path, legacy path will be created as a symbolic link.
+
+If customer sets legacy path wrongly by themselves(not following instructions in warning of scenario 4.2 and let NuGet to do that for them), e.g. a symbolic link pointing to a wrong path, NuGet is able to detect that, but NuGet won't know if it's on purpose or it's a wrong setting. So NuGet won't break customers in this situation.
+
+### Scenarios from the developer side
+**1. Visual Studio for Mac** 
+
+  It's in .NET Framework code path so it will not be affected by this change.
```
zivkan: Is this still true?  I thought I heard that VS 2022 for Mac now runs on .NET (Core?), and shells out to mono for non-SDK style projects?

Unfortunately this would be bad for us, since that would mean a whole lot more customers have nuget.config files in the legacy path.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+  On windows, it will not be affected by this change.
+  On Linux/Mac, this change will be applied.
+
+**4. CI/CD (Headless CLI)**
+
+  On windows, it will not be affected by this change.
+  On Linux/Mac, this change will be applied. But considering most users will put the URLs in a NuGet.Config file in the repository, so the user wide NuGet.Config in both legacy path and consolidated path should be the default content in most cases. Not sure if there is any data we could use to prove this assumption.
+
+
+### Challenges
```
zivkan: This section still appears to have a lot of open questions, which makes me feel that the spec is not ready to be merged.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+## Future Work
+Get rid of the legacy path completely. 
+
+## Open Questions
+
+**Which dotnet version are we going to bring in this change?**
+
+The version of .NET with this change will show warning to users who still have NuGet.Config in legacy path, encouraging them to move it to consolidated path.
+So it's good if more people could get this warning and get prepared to get rid of legacy path compleletly for future versions.
+For now, .NET 7 is the earlest possible version we could have this change.
```
zivkan: I don't believe this is still our plan. If we're planning to implement as soon as possible, then this "which version" question can be removed from the spec. Once the spec is merged, it will no longer provide valuable information to anyone reading it, in my opinion.  Same as the "port to net6" question below.

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,216 @@
...
+
+**Shall we insert this change into dotnet 6.x as servicing patch?**
+
+6.x is a LTS version. If 6.x doesn't have this change, it will still use the legacy path, we could not get rid of the legacy path completely if there are still people using 6.x.
+But at the same time, the bar for servicing patch should be high and for fixing security issues. 
+
+**When can we get rid of the legacy path completely?**
+
+This version: the .NET version with this change. It will show warning to users who still have NuGet.Config in legacy path, encouraging them to move it to consolidated path. If customers react to the warning, then they migrate to the consolidated path.
```
zivkan: depends on which solution is chosen. There are many scenarios where this spec proposes silently doing something, so no warning will be displayed. In fact, from the proposals, it looks like the only scenario where a warning will be displayed is when the legacy and consolodated nuget.config files both exist, and can't be automatically merged.

zivkan (commented):
A spec with multiple options is a great conversation starter. But in my opinion, in order to be approved and merged, it should explain which option is approved, and list the other options for people trying to understand the decision making processs.  But I don't think people who are reading the approved and merged spec should have questions about what is going to be implemented.

heng-liu:
Sorry for taking so long to make this PR ready for review!
After discussing with VS Mac/Mono team, I changed the spec from proposing consolidate the two paths, to telling users to consolidate manually if they want.
This PR is now ready for review. 
//cc @NuGet/nuget-client @JonDouglas

nkolev92 (approved):

zivkan (approved):

designs/Consolidate-NuGetConfig_OnLinux.md
```diff
@@ -0,0 +1,112 @@
...
+**Pros:** 
+The updated document will have a higher visibility when it's link is in warnings. Not sure about if it's just an info in the log. 
+
+**Cons:** 
+Users who have already consolidated the two paths by following the instructions in the updated document will still see the warning/information. It might be annoying.
+
+We can only warn/inform users when they use NuGet tools based on .NET Core code path.
+
+#### **Solution 2:**
+If NuGet detects two paths both exists, and neither of them is the symbolic link path, warn users and link the updated document in the warning.
```
KalleOlaviNiemitalo, line 70: Another option would be to read both files and warn if they are not byte for byte identical. That would allow symbolic links and even hard links, and would not need any new .NET API.

Issue #11363, opened 2021-11-06
Package Source Mapping UI - Allow the Creation/Removal of package source mappings in the NuGet Options UI


nkolev92:
Blocked on https://github.com/NuGet/Home/issues/11362

Refer to https://github.com/nuget/home/issues/11361 and https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md

donnie-msft:
Rough estimate is you can look for the introduction of this feature in 17.5 Preview 1

Issue #11364, opened 2021-11-06
Package Source Mapping UI - Allow a source to be pinned as default (ie. set `*`) in the NuGet Options -> Package Sources view.


nkolev92:
Refer to https://github.com/nuget/home/issues/11361 and https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md

donnie-msft:
Note: There's a long discussion about why we shouldn't implement the topic of this issue.
See: https://github.com/NuGet/Home/pull/11865#discussion_r899650092

donnie-msft:
Closing based on comment https://github.com/NuGet/Home/issues/11364#issuecomment-1158434597

PR #11865, opened 2022-06-10
Create spec for package source mapping in VS options

mcnallyella:
This is the spec for the VS options portion of https://github.com/NuGet/Home/issues/11363

### DIFF
--- a/proposed/2022/Package-Source-Mapping-VS-Options.md
+++ b/proposed/2022/Package-Source-Mapping-VS-Options.md
@@ -89,7 +89,7 @@ If the user removes one of these package source mappings, it will be deleted fro
 
 If the user clicks the _Clear_ button on the package source mappings options page, then all package sources will be removed from the configuration, and package source mapping will be disabled. 
 
-If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, this will also be a message to the user.
+If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, they will get a message that this source has pre-existing mappings in the config that will be applied once the source is added. 
 
 If the user adds `*` as the package ID, then the selected sources will be pinned on the package sources page. This is because if the user adds `*` as the package ID, it will be written to the solution level config with `pattern=*` for the selected source(s). See `Example 2` below.
 
--- a/proposed/2022/Package-Source-Mapping-VS-Options.md
+++ b/proposed/2022/Package-Source-Mapping-VS-Options.md
@@ -1,0 +1,133 @@
+# Package Source Mapping in Visual Studio Options Dialog
+
+* Start Date: 06-10-2022
+* Authors: Ella McNally
+* Issue: https://github.com/NuGet/Home/issues/11363
+
+## Summary
+
+Currently there is no support for package source mapping in the Visual Studio options dialog. I will be creating a feature where users can pin default sources and choose a specific source(s) for certain package source mappings from sources already provided in package sources page. This feature is a portion of what was discussed in a previous spec. https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md
+
+## Motivation 
+
+Adding support for package source mapping in the VS options dialog will allow user to pin sources and add/remove package source mappings with more ease.
+
+## Explanation
+
+### Functional Explanation
+
+#### Package Source Mappings Page 
+
+When the user opens the package source mapping page in the options dialog, they will see a list of all the package source mappings they have already configured. This page will have three buttons: _Add_ , _Remove_, and _Clear_.
+
+* Add: Brings user to a pop-up where they can add package source mappings to their project.
+
+* Remove: Lets user remove selected package source mapping from project.
+
+* Clear: Lets user clear all package source mappings from current project.
+
+The popup when the user clicks the add button will have a textbox where a user can type in a package ID. All sources that are defined in package sources page of VS options dialog will appear in a `ComboBox` below. The user checks a `CheckBox` for the sources they would like to use for that package ID. 
+
+The add window has two buttons: _Add_ and _Close_.
+
+* Add: Add button adds package source mapping to project only using selected sources, and the pop-up will disappear. 
+
+* Close: Will close pop-up without adding any package source mappings.
+
+#### Package Sources Page
+
+I will add a feature that allows a user to pin default sources for all package source mappings that are not specified on the package source mapping page. There will be a pin button, that when clicked, creates a default mapping for all package IDs to the pinned source. A pin icon will appear next to all pinned sources.
+
+### Technical Explanation
+
+#### Package Source Mappings Page
+
+When the user opens the package source mapping page in the options dialog, it will automatically display all previous package source mapping configurations. 
+
+When the user adds a package source mapping to their project, the NuGet config will be updated when the user clicks the okay button. The specified package ID will be mapped to the selected source(s) in the config. 
+
+**Example** 
+
+```xml
+<PackageReference Include="Microsoft.A" Version="1.0.0"/>
+<PackageReference Include="Newtonsoft.Json" Version="11.0.0"/>
+<PackageReference Include="Serilog" Version="11.0.0"/>
+```
+
+```xml
+<packageSourceMapping>
+    <packagesource key="nuget.org">
+        <package pattern="Serilog" />
+        <package pattern="Microsoft.Extensions.Options" />
+        <package pattern="Newtonsoft.Json" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+**Result:**
+
+In this example, the user added `Serilog`, `Microsoft.A`, and `Newtonsoft.Json`, separately, on the package source mappings page and selected `nuget.org` as the source from the `ComboBox` for all of them. All three package source mappings will be configured to be installed from `nuget.org`. 
+
+If the user removes one of these package source mappings, it will be deleted from the configuration as well. For example, if the user removes `Serilog`, the above code becomes
+
+```xml
+<packageSourceMapping>
+    <packagesource key="nuget.org">
+        <package pattern="Microsoft.Extensions.Options" />
+        <package pattern="Newtonsoft.Json" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+If the user clicks the _Clear_ button on the package source mappings options page, then all package sources will be removed from the configuration, and package source mapping will be disabled. 
+
+If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, this will also be a message to the user.
+
+#### Package Sources Page
+
+When the user opens the package sources page, it will automatically display a pin icon with all previously pinned default sources.
+
+When the user pins a package source(s), the NuGet config will be updated. Unless otherwise specified in the package source mappings page, all package IDs will be pinned to this source because the `pattern=*` in the config. 
+
+**Example**
+
+```xml
+<packageSourceMapping>
+    <packagesource key="contoso">
+        <package pattern="*" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+**Result:**
+
+In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern.
+
+### Accessibility 
+
+To ensure screen-reader I will run Accessibility Insights to find any accessibility faults.
+
+
+## Drawbacks 
+
+Currently, mappings to sources are not deleted in the config file when the source they are mapped to is deleted. It is complicated to remove this code because there are multiple different config files this mapping could be in.
+
+
+## Rationale and Alternatives 
+
+There were multiple ways to design the package source mappings page (e.g. use a table to display package source mappings/sources), but I found a `ListView` of the package source mappings to be the cleanest look. 
+
+## Prior Art
+
+
+## Unresolved Questions 
+
+
+* Should package IDs be validated? <https://github.com/NuGet/Home/issues/10783>
+
+
+## Future Possibilities 
+
+* Package source mapping in PMUI
+
+* Delete package source mappings from config
--- a/proposed/2022/Package-Source-Mapping-VS-Options.md
+++ b/proposed/2022/Package-Source-Mapping-VS-Options.md
@@ -54,7 +54,7 @@ When the user opens the package source mapping page in the options dialog, it wi
 
 When the user adds a package source mapping to their project, the NuGet config will be updated when the user clicks the okay button. The specified package ID will be mapped to the selected source(s) in the config. 
 
-**Example** 
+**Example 1** 
 
 ```xml
 <PackageReference Include="Microsoft.A" Version="1.0.0"/>
@@ -91,13 +91,15 @@ If the user clicks the _Clear_ button on the package source mappings options pag
 
 If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, this will also be a message to the user.
 
+If the user adds `*` as the package ID, then the selected sources will be pinned on the package sources page. This is because if the user adds `*` as the package ID, it will be written to the solution level config with `pattern=*` for the selected source(s). See `Example 2` below.
+
 #### Package Sources Page
 
 When the user opens the package sources page, it will automatically display a pin icon with all previously pinned default sources.
 
 When the user pins a package source(s), the NuGet config will be updated. Unless otherwise specified in the package source mappings page, all package IDs will be pinned to this source because the `pattern=*` in the config. 
 
-**Example**
+**Example 2**
 
 ```xml
 <packageSourceMapping>
@@ -109,7 +111,7 @@ When the user pins a package source(s), the NuGet config will be updated. Unless
 
 **Result:**
 
-In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern.
+In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern. Alternatively, the user could have added `*` as the package ID and selected `contoso` as the source on the package source mappings page. These two options would be written to the solution level config and read by the package sources page in the same way.
 
 ### Accessibility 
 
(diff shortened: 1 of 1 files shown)


ghost:
[CLA assistant check](https://cla.dotnetfoundation.org/NuGet/Home?pullRequest=11865) <br/>All CLA requirements met.

aortiz-msft:
Great start! Would it be possible to add a section that clarifies that source mappings can come from a hierarchy of files, that we will display all mappings throughout the hierarchy, but that we will only add to the lowest-level file?

aortiz-msft:
Also, could we please copy relevant screenshots/wireframes of the new UI as needed?

aortiz-msft:
Should we allow to pin more than one source? Is that even allowed? @nkolev92

mcnallyella:
> Should we allow to pin more than one source? Is that even allowed? @nkolev92

Yeah it seems that Jon has it in his spec that multiple sources can be pinned. 

image

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,133 @@
...
+In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern.
+
+### Accessibility 
+
+To ensure screen-reader I will run Accessibility Insights to find any accessibility faults.
+
+
+## Drawbacks 
+
+Currently, mappings to sources are not deleted in the config file when the source they are mapped to is deleted. It is complicated to remove this code because there are multiple different config files this mapping could be in.
```
dominoFire, line 125: Would it be possible to show in any of the options pages the filepath of the modified nuget.config file?

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Rationale and Alternatives 
+
+There were multiple ways to design the package source mappings page (e.g. use a table to display package source mappings/sources), but I found a `ListView` of the package source mappings to be the cleanest look. 
+
+## Prior Art
+
+
+## Unresolved Questions 
+
```
dominoFire, line 136: Is there any planned telemetry to measure feature usage, feature success? 

If this is out of scope, should we add a telemetry issue for package source mapping options?

cc @donnie-msft

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,133 @@
...
+* Clear: Lets user clear all package source mappings from current project.
+
+The popup when the user clicks the add button will have a textbox where a user can type in a package ID. All sources that are defined in package sources page of VS options dialog will appear in a `ComboBox` below. The user checks a `CheckBox` for the sources they would like to use for that package ID. 
+
+The add window has two buttons: _Add_ and _Close_.
+
+* Add: Add button adds package source mapping to project only using selected sources, and the pop-up will disappear. 
+
+* Close: Will close pop-up without adding any package source mappings.
```
dominoFire, line 37: Would you please a link to Mockup section in Package Source Mapping to indicate which window refers each section on this spec?

dominoFire (approved):
Looks good.

proposed/2022/Package-Source-Mapping-VS-Options.md
donnie-msft: What would you like to know in this telemetry? VS already knows the frequency of page openings for any VS options page.

I don't know if feature success has been defined. Adding a mapping via VS doesn't tell us about the entire mapping usage.  -@nkolev92 thoughts?

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Package Sources Page
+
+I will add a feature that allows a user to pin default sources for all package source mappings that are not specified on the package source mapping page. There will be a pin button, that when clicked, creates a default mapping for all package IDs to the pinned source. A pin icon will appear next to all pinned sources.
```
kartheekp-ms: AFAIK, to pin a package source for all NuGet operations there should be a source mapping configured with pattern `*`. Did I miss something here?

``` xml
<packageSourceMapping>
    <packagesource key="pinnedSource">
        <package pattern="*" />
    </packageSource>
    <packagesource key="nuget.org">
        <package pattern="Serilog" />
        <package pattern="Microsoft.Extensions.Options" />
        <package pattern="Newtonsoft.Json" />
    </packageSource>
</packageSourceMapping>
```

proposed/2022/Package-Source-Mapping-VS-Options.md
dominoFire: I would like to know how many customers use options dialog to modify package source mappings. If this can be inferred elsewhere, I would like to take look at those data points.

proposed/2022/Package-Source-Mapping-VS-Options.md
donnie-msft: Kartheek, is the example under Technical Explanation what you're looking for? `pattern=*` is described there. https://github.com/NuGet/Home/pull/11865/files/a43a3e807fa606013bf97413088c8407724498f0#diff-f9e22802f1ca486df02ad7e0668bf53e5c4c087c6c5adad061653045811fdccaR86

dominoFire (approved):

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+Adding support for package source mapping in the VS options dialog will allow user to pin sources and add/remove package source mappings with more ease.
+
+## Explanation
+
+### Functional Explanation
+
+#### Package Source Mappings Page 
+
+When the user opens the package source mapping page in the options dialog, they will see a list of all the package source mappings they have already configured and the filepath to the solution level config file. There is a hierarchy of NuGet configs levels: machine, user, and solution. Package source mappings will be read from all of the levels, but only written to the solution level config. This page will have three buttons: _Add_ , _Remove_, and _Clear_.
```
nkolev92: I think this could probably use a clarification. 
What if there's no solution level config? We don't really recognize a solution level config, it's just another config for us, so they may prove to be challenging to implement. 

Additionally, it seems like we'll only display the mappings if they are enabled, so maybe a lot of concerns about where we write mappings is moot.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+* Add: Brings user to a pop-up where they can add package source mappings to their project.
```
nkolev92: to their configuration instead of project. 

The configuration applies to all projects.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+* Remove: Lets user remove selected package source mapping from project.
```
nkolev92: What happens when the last mapping is removed for a configuration? Does the package source mapping tab go away?

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+* Clear: Lets user clear all package source mappings from current project.
```
nkolev92: Is there value in having a removal all? Do we think this is something that'd be common? 

Probably simplest if we  just have an add and remove options for now. 
It's unclear (pun not intended) that clear is going to be a viable scenario.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+The popup when the user clicks the add button will have a textbox where a user can type in a package ID. All sources that are defined in package sources page of VS options dialog will appear in a `ComboBox` below. The user checks a `CheckBox` for the sources they would like to use for that package ID.
```
nkolev92:
```suggestion
The popup when the user clicks the add button will have a textbox where a user can type in a package pattern. All sources that are defined in package sources page of VS options dialog will appear in a `ComboBox` below. The user checks a `CheckBox` for the sources they would like to use for that package pattern.
```

proposed/2022/Package-Source-Mapping-VS-Options.md
nkolev92: fwiw, I know clear is in the original spec, but personally I don't think there's harm in removing it from the V1 version if enough people agree (basically don't remove it based on solely my feedback here).

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+The popup when the user clicks the add button will have a textbox where a user can type in a package ID. All sources that are defined in package sources page of VS options dialog will appear in a `ComboBox` below. The user checks a `CheckBox` for the sources they would like to use for that package ID. 
+
+The add window has two buttons: _Add_ and _Close_.
+
+* Add: Add button adds package source mapping to project only using selected sources, and the pop-up will disappear. 
+
+* Close: Will close pop-up without adding any package source mappings.
+
+Below are the mockups for the package source mappings page. The top left is the package source mappings page, and the bottom left is the popup that appears when the user clicks the add button.
```
nkolev92, line 39: How are the mappings going to be sorted? 

I think it allowing them to be sortable by both source name and pattern might be a neat feature. 
We did find that some people do prefer the mental model of source first, but others did like the pattern first approach instead. 
This might require unmerging some options.

In my opinion, the default should be sorted by source first, mostly because that's what the configuration model looks like. 

If we wanted to, we could even consider grouping them under a source node instead the same way the configuration does it.

proposed/2022/Package-Source-Mapping-VS-Options.md
nkolev92: I'm not 100% sure the meaning of the `pin` is something easily understood. 
I understand this is partially a comment on the original mock-ups but I believe it's something up for discussion. 

I'm not sure I have another proposal for it, beyond maybe just not having it.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+![Options 2](../../meta/resources/PackageSourceMapping/VSOptions1.png)
+
+### Technical Explanation
+
+#### Package Source Mappings Page
+
+When the user opens the package source mapping page in the options dialog, it will automatically display all previous package source mapping configurations. 
+
+When the user adds a package source mapping to their project, the NuGet config will be updated when the user clicks the okay button. The specified package ID will be mapped to the selected source(s) in the config.
```
nkolev92: worth reviiewing the full doc, but we're usually referring to package patterns here.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+```xml
+<packageSourceMapping>
+    <packagesource key="nuget.org">
+        <package pattern="Microsoft.Extensions.Options" />
+        <package pattern="Newtonsoft.Json" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+If the user clicks the _Clear_ button on the package source mappings options page, then all package sources will be removed from the configuration, and package source mapping will be disabled.
```
nkolev92, line 92: See other clear comments, but is the idea that clear clears the configuration for that source or the general configuration?

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+    <packagesource key="nuget.org">
+        <package pattern="Microsoft.Extensions.Options" />
+        <package pattern="Newtonsoft.Json" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+If the user clicks the _Clear_ button on the package source mappings options page, then all package sources will be removed from the configuration, and package source mapping will be disabled. 
+
+If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, they will get a message that this source has pre-existing mappings in the config that will be applied once the source is added.
```
nkolev92, line 94: I love this :)
Worth calling out in what matter the message will be shown.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+<packageSourceMapping>
+    <packagesource key="contoso">
+        <package pattern="*" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+**Result:**
+
+In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern. Alternatively, the user could have added `*` as the package ID and selected `contoso` as the source on the package source mappings page. These two options would be written to the solution level config and read by the package sources page in the same way.
```
nkolev92: I'm being particular in this comment here. 

A `pin` is just another name for a package source mappiong configuration for a source with only 1 pattern, `*`. 

So technically, source mappings do exist. 

Also, note that `*` doesn't mean that all packages will come from that source. 
There is a source mapping precedence, so the longest pattern matched is going to be considered. 

Example:

nuget.org - * 

contoso - nuget.* 

If I only have `nuget.` id packages in my  project, thne they'll be downloaded from contoso and not nuget.org.

proposed/2022/Package-Source-Mapping-VS-Options.md
nkolev92: The overall effort's goal is increase the adoption of source mapping, which is something that we're tracking by different means so we shouldn't need to account for that here.

I think we should add some telemetry to track the usage of the package source mapping dialog as @dominoFire suggests. 

The frequency of package source mapping changes, add vs remove etc that could be helpful for us.
Think of adding mappings equivalent to adding packages for example. We could benefit from some minimal telemetry here.

nkolev92 (commented):
Great write-up!

proposed/2022/Package-Source-Mapping-VS-Options.md
mcnallyella: @nkolev92  That's a fair point. I feel like a tooltip for the pin icon would be helpful.

nkolev92: I would propose not even having it in the first place. 
I feel like the concept isn't worth special casing:

Say you have the following: 

nuget.org - *

contoso - special.id
contoso - Microsoft.*
contoso - nuget.*
contoso - contoso.*

It's super likely that most packages are getting restored from contoso, but nuget.org will have the pin, and contoso wouldn't. 

Having something telling us about nuget.org having the fallback version can even be misleading in the package sources option.

nkolev92: I wonder if we should have something saying a mapping exists for a certian package source? That might be more helpful.

donnie-msft: I think of the pin as a graphical representation that this exists:
`nuget.org - *`

Are you thinking that's not worth representing in a UI, or are can you elaborate if you're saying more than that?

donnie-msft: > It's super likely that most packages are getting restored from contoso, but nuget.org will have the pin, and contoso wouldn't.

If Ella's proposal for the tooltip on the pin emphasizes the meaning of Default, would that alleviate the concern?
I think we're going this direction because once we modify PMUI to support Pinning, the concepts will amplify the point of a pin icon.

zivkan: > Are you thinking that's not worth representing in a UI, or are can you elaborate if you're saying more than that?

In my opinion, the package sources page should not have any representation about which source is default. This is assuming that the package source mapping config is shown in a different page.  The package source mapping page having an icon or something to show it's the default seems reasonable.

Going off on a tangent, [the feature spec for package source mapping](https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md) uses the word "pin" in the context that a package that has a mapping is now pinned to a source. Reusing the same word, pin, to mean something different (package source used when no other package prefix matches) risks confusion and bad SEO, assuming we as a team talk to customers using this terminology, and customers also use it. It would be ideal if we could agree on standard terminology and use it consistently to improve searchability and reduce confusion.

nkolev92: Another option for the package sources dialog:

Add an indicator for each source that has a mapping defined. 

That indicator could have a tooltip that could call out the default/fallback concept if we wanted to.
It can also just say `see the package source mapping page for mapping information`.

zivkan: Reading other comments in this PR, it reminded me that a package source with prefix `*` is a fallback source, not really a default source. The `*` prefix will be used if no other prefix matches (longest prefix wins). To me, that makes the package source feel less important, so maybe it doesn't warrant to be pointed out more clearly than other package sources?

donnie-msft: > , [the feature spec for package source mapping](https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md) uses the word "pin" in the context that a package that has a mapping is now pinned to a source. Reusing the same word, pin, to mean something different (package source used when no other package prefix matches) risks confusion and bad SEO,

If you scroll down to the Mockups section of the file you linked, you'll see the pin icon and the terminology of pin used there. We didn't deviate from the feature spec here in the UI spec. Whether we want to change the feature spec's direction is debatable. 

For now, we'll save changing the sources page to last. I think that's not blocking anything on this spec, and it's more easily changeable as we iterate.

nkolev92: > Reading other comments in this PR, it reminded me that a package source with prefix `*` is a fallback source, not really a default source. The `*` prefix will be used if no other prefix matches (longest prefix wins). To me, that makes the package source feel less important, so maybe it doesn't warrant to be pointed out more clearly than other package sources?

100% agree with this. Better summary of my concerns.

proposed/2022/Package-Source-Mapping-VS-Options.md
donnie-msft: @mcnallyella if you can get to this during your project, go for it. Otherwise, I'll keep it on the back burner.
https://github.com/NuGet/Client.Engineering/issues/1691

proposed/2022/Package-Source-Mapping-VS-Options.md
donnie-msft: @nkolev92 wrote up the tracking issue for Package Source options changes for default/* in https://github.com/NuGet/Home/issues/11364. 

To reiterate:

> For now, we'll save changing the sources page to last. I think that's not blocking anything on this spec, and it's more easily changeable as we iterate.

nkolev92: the merged spec doesn't really call out the fact that we're not going to add the pin thumbthack in V1. 
It's not a big deal, given that we have a tracking issue for it, but we can always just link it.

Issue #11858, opened 2022-06-06
Spec for dotnet.exe add package support for projects onboarded onto central package management


kartheekp-ms:

nkolev92:
@kartheekp-ms Just a reminder to label issues according to https://github.com/NuGet/Client.Engineering/blob/main/docs/creating-github-issues.md#minimum-requirements.

PR #11873, opened 2022-06-13
Revised proposal for Adding dotnet CLI support for projects onboarded to CPM

kartheekp-ms:
Tracking issue https://github.com/NuGet/Home/issues/11858

[Rendered Design Spec](https://github.com/NuGet/Home/blob/dev-kartheekp-ms-cpmaddpackagesupport/proposed/2022/cpm-dotnet-cli-add-package-support.md)

Proposal for [11807](https://github.com/NuGet/Home/issues/11807)

Here is what changed when compared to the initial proposal:
1. Added a table explaining the scenario, current behavior, new behavior and in/out of scope as per current plan.
2. Added an example for `VersionOverride` scenario


proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -1,29 +1,46 @@
...
 
-## Solutions
+## Solution
+
 When `dotnet add package` is executed in a project onboarded to CPM (meaning that the `Directory.packages.props` file exists) there are a few scenarios that must be considered.
 
+| Scenario # | PackageReference exists? | Directory.Packages.Props file exists? | Is ManagePackageVersionsCentrally property set to true? | Is VersionOverride? | Current behavior | New behavior in dotnet CLI | In Scope |
+| ---- |-----| -----|-----|----------|----------|----------|----------|
+| 1 | ❌ | ✔️ | ✔️ (in Directory.Packages.Props or .csproj file) | ❌ | Restore failed with NU1008 error and NO edits were made to the csproj file (same in VS and dotnet CLI) | `PackageReference` should be added to .(cs/vb)proj file and `PackageVersion` should be added to the closest `Directory.Packages.Props` file | ✔️ |
+| 2 | ❌ | ❌ | ✔️ in (cs/vb)proj file | ❌ | Restore failed with NU1008 error and NO edits were made to the csproj file (same in VS and dotnet CLI) | `PackageReference` and `Version` should be added to .(cs/vb)proj file.| ✔️ |
```
nkolev92, line 28: So for this one, we're basically saying if you only have 1 of the 2 reqs, you're not fully enabled and we proceed as it CPM wasn't a thing?

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -1,29 +1,46 @@
...
+| 3 | ❌ | ✔️ | ❌ | ❌ | `PackageReference` and `Version` added to .(cs/vb)proj file. | In addition to the current behavior `Directory.packages.props` file should be deleted if it exists in the project folder | ❌ |
```
nkolev92, line 29: I don't think we should be deleting files. 
Maybe warn about them, but deleting them seems too aggressive imo.

proposed/2022/cpm-dotnet-cli-add-package-support.md
```diff
@@ -1,29 +1,46 @@
...
+| 4 | ✔️ | ✔️ | ✔️ | ❌ | **dotnet CLI** - Restore failed with NU1008 error and NO edits were made to the csproj file. **VS** - Clicked on Update package in PM UI. New `PackageVersion` added to csproj file and `Directory.packages.props` file was not updated | No changes should be made to the `PackageReference` item in .(cs/vb)proj file but `PackageVersion` item should be updated with the new version in the appropriate `Directory.Packages.Props` file.|✔️ [More Info](https://github.com/NuGet/Home/pull/11849#discussion_r890639808) |
```
nkolev92: I don't think add package should update versions. 
This is something that's not readily apparent to the user. 

Imo, the behavior should be either version override or an explciit command to update in CPM depending on whether --version has been passed it or not.

nkolev92 (commented):
Thanks for updating the spec @kartheekp-ms. 

A few comments.

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: Yeah, that is my understanding after discussing with @jeffkl.

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: Sure, I will update the spec, but this scenario is out of scope for version 1.

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: We went down on this path because `dotnet add package` currently updates the package version if the reference already exists. In the case of CPM, updating version in the `Directory.Packages.Props` file is an option.

Your suggestion to add `VersionOverride` if the package reference already exists is also a good one. @jeffkl what do you think?

jeffkl: In my opinion, its just more work for a user to use a different gesture for add vs update.  They might be running it on a bunch of projects via automation and they'd have to determine whether not that package is already referenced.  So I think add package should just update the version for them.

I'd rather not make `VersionOverride` easier to use personally.  Its an escape hatch/last restore and I'd much rather people try to unify their graph.  If we wanted to use VersionOverride, I would see that as a command-line argument for you opt into that functionality.

proposed/2022/cpm-dotnet-cli-add-package-support.md
jeffkl: This is correct.  You need to have created the file wherever you think it should be and you have to set the property to `true` in order for CPM to work currently.  We could create the file for you but I think that would need to be a different command-line argument so you could specify where to place the file because determining that on our own would be very error prone.  For now we could just fall back to existing behavior until people onboard and then later on we could add a command-line argument if we get feedback that they want `dotnet add package` to onboard them.

nkolev92: That's perfect. I agree with that.

Onboarding should be intentional.

proposed/2022/cpm-dotnet-cli-add-package-support.md
nkolev92: > In my opinion, its just more work for a user to use a different gesture for add vs update. They might be running it on a bunch of projects via automation and they'd have to determine whether not that package is already referenced. So I think add package should just update the version for them.

The reason why I bring it up is because we have issues and customer interviews that have suggested otherwise. 
There's also an ask for a `dotnet update package` command, which is something we'd want to do eventually as well. 

I think dotnet add package should match the VS experience for install package.

When you go to an individual project and run `add package`, you don't expect the package to be updated for other projects as well. 

You could introduce potential compatibility issues but you wouldn't know about that until you get to your build, because there's no prevalidation for that.

proposed/2022/cpm-dotnet-cli-add-package-support.md
nkolev92: I totally missed the scope column. 

Maybe 2 tables would've been easier, but it's probably just me.

kartheekp-ms: Good idea. I will split the table (one for in scope work and another for out of scope work)

proposed/2022/cpm-dotnet-cli-add-package-support.md
kartheekp-ms: > I think dotnet add package should match the VS experience for install package.

Current behavior in VS - "Clicked on Update package in PM UI. New `PackageVersion` added to csproj file and `Directory.packages.props` file was not updated". I think it is a bug because the tooling adds `PackageVersion` XML element not the `VersionOverride`.

**EDIT** - When I launched @jeffkl's PR branch https://github.com/NuGet/NuGet.Client/pull/4642 in an experimental instance, I noticed that updating package version in Visual Studio for CPM project didn't update .csproj file also. But the updated version is present in the assets.json file and solution explorer dependencies node except the props file.

image

nkolev92: The VS experience is not really implemented fully yet.

jeffkl: I filed an issue for the `<PackageVersion />` being added to the project: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1544117

kartheekp-ms: @jeffkl - Hope you are okay with changing the logic to add `VersionOverride` when add package is trying to update the package version. I will update the proposal shortly.

cc @pragnya17

kartheekp-ms: I updated the spec to capture the summary of this conversation. No changes should be made to the `Directory.Packages.Props` file. Add `VersionOverride` attribute to the existing `PackageReference` item in .(cs/vb)proj file.

jeffkl: Yes if VersionOverride is already specified, we should update that value.

kartheekp-ms: > Yes if VersionOverride is already specified, we should update that value.

I agree. Updated the spec.

Can I get approval on the updated spec?

jeffkl (approved):

nkolev92 (approved):

LeftTwixWand:
Can recommend this dotnet tool for automatic migration to CPM:
https://github.com/Webreaper/CentralisedPackageConverter

Issue #11782, opened 2022-04-28
[Feature]: Add tracing to NuGet to identify all possible sources of package downloads


ericstj:
### NuGet Product(s) Involved

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, MSBuild.exe, dotnet.exe, NuGet SDK

### The Elevator Pitch

Component Governance is important to help identify potential vulnerabilities in a repository.

Today component governance detects direct references, but also side-effects of restore -- the assets file, packages downloaded, binaries in the output.  These are all good indicators but sometimes are hard to trace to the root cause.

One example we've noticed that's exceptionally hard for folks to identify is eclipsed packages.  NuGet will download packages in the graph if it cannot "know" at the time of the download if that package will eventually win or not.  These packages will appear in a local packages folder and *could* be used by the repository, so they are flagged by component governance.

Take one example which we hit in some dotnet repositories.  Imagine package C1 is vulnerable and flagged by component governance.
```
Project > A > B1 > C1
Project > D > B2 > C2
```

In this case there is nothing inside the repository's intermediates that tells the repo owner "why" NuGet downloaded C1.  They could find the reference in other packages downloaded and try to walk back up to some package which does appear in intermediate assets files, but that's a ton of work and not really feasible for the majority of our users.

Can we make it *crystal clear* why a package was downloaded?  Potentially through opt-in tracing?  NuGet knows this information and can make it available through some simple logging.  Doing so would save repo owners everywhere many hours of time and make the ecosystem better by helping folks make more educated decisions in their repos around dismissing issues, working around them, or asking dependencies to update.

nkolev92:
Adding partner labels as some of the dotnet partners have been affected by this.

wtgodbe:
👀

nkolev92:
For reference, this is where the unresolved packages are downloaded: https://github.com/NuGet/NuGet.Client/blob/2d93510efe2537fb26b3d1057a05e0cc9a6f07c5/src/NuGet.Core/NuGet.Commands/RestoreCommand/RestoreTargetGraph.cs#L166-L169

nkolev92:
A different solution to the original problem could be an opt in to avoid installing rejected packages. 
A means to avoid even considering rejected packages would be to use lock files.

aortiz-msft:
Another alternative would be to create a separate folder, or a sub-folder, for packages downloaded to determine the graph but are not actually used for building the code.

ericstj:
[...]

PR #11446, opened 2021-12-09
Spec for machine readable json output for dotnet list package

erdembayar:
## Bug

Fixes: https://github.com/NuGet/Home/issues/11449
Related: https://github.com/NuGet/Home/issues/7752

## Description

Many organization are required by [regulation](https://www.whitehouse.gov/briefing-room/presidential-actions/2021/05/12/executive-order-on-improving-the-nations-cybersecurity/) to audit packages that they're using in repository. It is difficult to parse the output of dotnet list package.  Without having a good way to parse the output of this command, it makes it difficult to structure the list of packages in a way that can be fed into another auditing system.
In this spec design proposal, I tried to address `json` output part, for `parsable` format output it needs own spec.
For this purpose, I added below 2 new options for all `dotnet list package` commands.

* `--format <FORMAT>`
`<FORMAT>` - Allowed values as part of spec is `json`. Also `text` is acceptable input value too, it'll just output current cli output. (In the future `parseable`, `csv`, `yaml`, `xml` could be candidates.)

```dotnetcli
dotnet list [<PROJECT>|<SOLUTION>] package [--config <SOURCE>]
    [--deprecated]
    [--framework <FRAMEWORK>] [--highest-minor] [--highest-patch]
    [--include-prerelease] [--include-transitive] [--interactive]
    [--outdated] [--source <SOURCE>] [-v|--verbosity <LEVEL>]
    [--vulnerable]
    [--format <FORMAT>]

dotnet list package -h|--help
```

[Rendered view](https://github.com/NuGet/Home/blob/fc6a7376bcaf69a5e69ac1c093ee748dd7bd8864/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md)

To do:  
* Rename spec md file to `DotnetListPackageMachineReadableJsonOutput.md`
* Get community input https://twitter.com/nuget/status/1470426335705239557
* Get explicit approval from stake holders
* packages.lock.json format looks very similar what we want, code/schema could be reused.
* Single json schema

PR #11875, opened 2022-06-13
dotnet nuget why proposal

### LINKED ISSUE #11874: spec for dotnet nuget why command

kartheekp-ms:
Fixes https://github.com/NuGet/Home/issues/11874

[Rendered spec](https://github.com/NuGet/Home/blob/dev-kartheekp-ms-nugetwhy-proposal/proposed/2022/dotnet-nuget-why-proposal.md)

Tracking https://github.com/NuGet/Home/issues/11782

//cc @chgill-MSFT


proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+Version of the package. See NuGet package versioning.
+
+- -v|--verbosity <VERSION>
+
+Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. For more information, see [LoggerVerbosity](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).
+
+#### Examples
+
+- List depdency graph of a package given package id.
```
erdembayar:
```suggestion
- List dependency graph of a package given package id.
```

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+Project 'projectNameA' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
+   [net472] Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
+
+Project 'projectNameB' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.1.0) > Microsoft.ML.Util (1.1.0) > pacakgeA (1.1.0)
+   [net472] Microsoft.ML (1.1.0) > Microsoft.ML.Util (1.1.0) > pacakgeA (1.1.0)
+```
+
+- List depdency graph of a package given `package id` and `version`.
```
erdembayar:
```suggestion
- List dependency graph of a package given `package id` and `version`.
```

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+```
+dotnet nuget why packageA --version 1.0.0
+
+Project 'projectNameA' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
+   [net472] Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
+```
+- List depdency graph of a package given `pacakge id` and `target framework`.
```
erdembayar:
```suggestion
- List dependency graph of a package given `pacakge id` and `target framework`.
```

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+dotnet nuget why packageA -f net6.0
+
+Project 'projectNameA' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
+
+Project 'projectNameB' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.1.0) > Microsoft.ML.Util (1.1.0) > pacakgeA (1.1.0)
+```
+
+- List depdency graph of a package given `pacakge id`, `version` and `target framework`.
```
erdembayar: same here

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Problem Background
+
+Developers are often confused with top-level & transitive dependencies in their project. A developer should be able understand where every package originated from. A user should be able to promote a transitive dependency quickly & easily to a top-level dependency in the case there is a security concern on the package version depended on or if promoting to a top-level package will resolve a conflict.
+
+Knowing how a package is being included in a project helps a developer diagnose many problems. It also provides them with an opportunity to understand complex dependency chains they may have within their projects. There is not a great solution that exists to understand the nature of top-level packages and their transitive dependencies. The solution explorer in Visual Studio does the best job at allowing a user to dive into each NuGet package & see all transitive dependencies from the top-level dependency.
+
+![](../../meta/resources/TransitiveDependencies/SolutionView.png)
+
+The dotnet CLI does not provide the insight into “why” a transitive dependency is listed, although it does list everything that has been resolved.
```
erdembayar: Maybe instead of `why` could `explain` would be better word?  For example [sql explain](https://www.geeksforgeeks.org/explain-in-sql/) does similar  action for sql.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+#### Examples
+
+- List depdency graph of a package given package id.
+
+```
+dotnet nuget why packageA
+
+Project 'projectNameA' has the following dependency graph for 'pacakgeA'
+   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > pacakgeA (1.0.0)
```
erdembayar:
```suggestion
   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > packageA (1.0.0)
```

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+The contents of `{projectName}.nuget.dgspec.json` and `project.assets.json` will be helpful to understand the dependency graph of a given package but there are few limitations.
+
+-  NuGet restore operation downloads packages during dependency resolution which may not part of the final graph The absence of a package id and version combination in `project.assets.json` file signals that it was downloaded during dependency resolution but not part of the final dependency graph.
+- Packages acquired through `PackageDownload` are not tied to the project in any way beyond acquisition. These packages are not recorded in the final graph that `project.assets.json` maintains.
+
+`dotnet nuget why` command prints the dependency graph of a given package only if it is part of the final graph.
+
+```
+dotnet nuget why [<PROJECT>|<SOLUTION>] package <PACKAGE_NAME>
```
erdembayar: Why only 1 package? Could customer lookup more than 1 package?
Could customer do `dotnet nuget why [<PROJECT>|<SOLUTION>] packages package1, package2`
Or `dotnet nuget why [<PROJECT>|<SOLUTION>] package 'nuget.*'`?

Could be out of scope: Some cli tools like `git` make really good graphs, instead of just listing path as text probably visualize as graph tree so help better for customers reason why this package pulled into graph, it's quite useful in case we're looking up multiple packages same time.
https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/How-to-use-the-git-log-graph-command

erdembayar (commented):
Looks good start, let's start with some comments.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+Prints out a description of how to use the command.
+
+- --version <VERSION>
+
+Version of the package. See NuGet package versioning.
+
+- -v|--verbosity <VERSION>
+
+Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. For more information, see [LoggerVerbosity](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).
```
erdembayar: It's not clear exactly what kind of role verbosity going to play here.
I assume we don't print any errors right so does it print different verbosity graph tree or something else?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+Search package depdnency graph only for a specific target framework.
+
+- -?|-h|--help
+
+Prints out a description of how to use the command.
+
+- --version <VERSION>
+
+Version of the package. See NuGet package versioning.
```
erdembayar: I added versioning for[ json output spec](https://github.com/NuGet/Home/pull/11446), but none of the other text output has any versioning. Why we're adding versioning this time?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+- -v|--verbosity <VERSION>
+
+Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. For more information, see [LoggerVerbosity](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).
+
+#### Examples
+
+- List dependency graph of a package given package id.
+
+```
+dotnet nuget why packageA
```
erdembayar, line 146: I believe correct syntax for this case would be:
`dotnet nuget why package packageA`

Please fix for other examples too.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
+# dotnet nuget why command
+
+- Status: **Draft**
```
zivkan: The Status field made sense (kind of) when we used to put specs in the wiki. However, now that they're markdown files in the repo, that use pull requests, I think we should remove this.

Similarly, when this PR is approved and merged, it's no longer "proposed" (file path), it's approved.  In my opinion, there should be no `proposed/` folder in the repo root.

proposed/2022/dotnet-nuget-why-proposal.md
zivkan: Given that `why` will be used to show information about the resolved graph, and that NuGet only allows a single version of a package to be selected in the project, I don't understand the use-case for `--version`.  Lets say that PackageA version 1.2.3 is selected in the project, what happens if the customer does `--version 1.0.0`?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+- --version <VERSION>
+
+Version of the package. See NuGet package versioning.
+
+- -v|--verbosity <VERSION>
+
+Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. For more information, see [LoggerVerbosity](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.loggerverbosity).
+
+#### Examples
```
zivkan, line 141: An example with a diamond dependency is missing.  When a package is brought in by more than one path, what does the output look like?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+- -?|-h|--help
+
+Prints out a description of how to use the command.
+
+- --version <VERSION>
+
+Version of the package. See NuGet package versioning.
+
+- -v|--verbosity <VERSION>
```
zivkan:
```suggestion
- -v|--verbosity <LEVEL>
```

proposed/2022/dotnet-nuget-why-proposal.md
zivkan: Does the .NET SDK team have a "rule" that all dotnet cli commands must have msbuild level verbosity?

proposed/2022/dotnet-nuget-why-proposal.md
kartheekp-ms: Good question. `dotnet nuget why package` can be executed on a solution level just like `dotnet list package` command.

My understanding is that if multiple projects that are unrelated have different version of a same package installed then it would be convenient for customers to search for a specific version of a package. 

image

proposed/2022/dotnet-nuget-why-proposal.md
kartheekp-ms: I had a chat with @chgill-MSFT about this and the feedback is use to `dotnet nuget why` command. What does everyone think?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
+# dotnet nuget why command
+
+- Status: **Draft**
+- Author: [Kartheek Penagamuri](https://github.com/kartheekp-ms)
+- GitHub Issue [11782](https://github.com/NuGet/Home/issues/11782)
+
+## Problem Background
+
+Developers are often confused with top-level & transitive dependencies in their project. A developer should be able understand where every package originated from. A user should be able to promote a transitive dependency quickly & easily to a top-level dependency in the case there is a security concern on the package version depended on or if promoting to a top-level package will resolve a conflict.
```
nkolev92: This is being pedantic, but I'd remove this part: 

> Developers are often confused with top-level & transitive dependencies in their project.

Developers want to know why a package is installed, that's the motivation, the confusion part leaves a bad taste in my mouth personally.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: Isn't `why` play on the command name here?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+![](../../meta/resources/TransitiveDependencies/SolutionView.png)
+
+The dotnet CLI does not provide the insight into “why” a transitive dependency is listed, although it does list everything that has been resolved.
+
+![](../../meta/resources/TransitiveDependencies/DotNetCLI.png)
+
+We could have considered a minor tweak to the existing experience of “dotnet list package --include-transitive” to provide the user with a sense of where the package came from. This could be a new column next to “Resolved” which says “Transitively Referenced” or “Referenced” and has a list of the top-level packages that requested the dependency. Tracking issue can be found here: https://github.com/NuGet/Home/issues/11625. Unfortunately, this solution doesn't provide an easy way to identify the dependency graph `(starting from project -> top-level package -> transitive-dependency (1) -> ..-> transitive dependency (n))` for a particular package.
```
nkolev92: This sohuld be in the rationale and alternatives, not in the problem background.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+![](../../meta/resources/TransitiveDependencies/TransitiveDotNetCLI.png)
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will additionally be a header titled “Top-level Packages” where the current experience is today for top-level packages.
```
nkolev92: This is prior art, or something related, not background.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
+![](../../meta/resources/TransitiveDependencies/TransitiveVSPMUI.png)
+
+Sadly, we don’t have any further CLI or visualizing tool today to answer `Running 'dotnet restore' on a (large) .sln file restores a certain problematic package. How do I know which project(s) in the solution is causing that package to be restored?` question.
+
+## Goals
+The main goal is to add `dotnet nuget why` command to .NET CLI to help customers understand the dependency graph of a given package for PackageReference projects.
+
+Projects that use `packages.config` format are out of scope for this effort because the XML file contains a flat list of both direct and transitive package dependencies. Customers cannot visualize the dependency graph from the file itself but it would be easy to update the package version of transitive depedency incase if there is any security vulnerability in the current version.
```
nkolev92: Appreciate the detail :) 
Considering using only 1 sentence to describe a specific goal. 

Also consider making this under a non-goals or out of scope subsection.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Customers
+We are building this for .NET users to be successful with understanding dependency graph for a given package.
+
+## Solution
```
nkolev92: The more I read it, I don't think we're using the correct template for this proposal: https://github.com/NuGet/Home/blob/dev/meta/template.md.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+NuGet restore operation generates following files under `obj` folder for `PackageReference` style projects.
+- `project.assets.json` file that maintains a project's dependency graph, which is used to make sure that all necessary packages are installed on the computer.
+- `{projectName}.nuget.dgspec.json` file that maintains a project's top-level dependencies along with other metadata.
```
nkolev92: I would delete this. 
You're not supposed to use it in the implementation.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+- `{projectName}nuget.g.targets` and `{projectName}.nuget.g.props` files define various properties, such as the path to where the packages cache is on the local machine. They also include any MSBuild imports that NuGet packages referenced by the project need.
```
nkolev92: We don't refer to the global packages folder as the packages cache.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: This is becoming a restore internals description, but only really the assets file is relevant. 
I'd keep it scoped to the problem.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+The contents of `{projectName}.nuget.dgspec.json` and `project.assets.json` will be helpful to understand the dependency graph of a given package but there are few limitations.
```
nkolev92: {projectName}.nuget.dgspec.json has only the top declared dependencies. 
the project.assets.json has the effective data from nuget.dgspec.json, so we probably don't need it.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+
+-  NuGet restore operation downloads packages during dependency resolution which may not part of the final graph The absence of a package id and version combination in `project.assets.json` file signals that it was downloaded during dependency resolution but not part of the final dependency graph.
+- Packages acquired through `PackageDownload` are not tied to the project in any way beyond acquisition. These packages are not recorded in the final graph that `project.assets.json` maintains.
```
nkolev92: They are recorded, just not in the targets/library section.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: > My understanding is that if multiple projects that are unrelated have different version of a same package installed then it would be convenient for customers to search for a specific version of a package.

I think we should start with a more minimal solution before we add things like version.

nkolev92 (commented):
This is a great start, thanks Kartheek!

Most of my comments are really about doc style/completeness more so than the actual command proposal. 

Feel free to reach out if you have individual questions.

proposed/2022/dotnet-nuget-why-proposal.md
erdembayar: I'm proposing `explain` is better command name than `why`.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Based on other comments, I agree. I can change the format to more closely resemble the template you linked.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+    [--version <VERSION>]
+    [-v|--verbosity <LEVEL>]
+
+dotnet nuget why -h|--help
+```
+#### Arguments
+
+- PROJECT | SOLUTION
+
+The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution or project is found, an error is thrown.
```
pragnya17: I believe no error should be thrown if more than one project is found.

kartheekp-ms: That's a good catch. I guess we can change this statement to if more than one solution or **no** project is found, an error is thrown.

Please feel free to update the spec.

nkolev92: > I believe no error should be thrown if more than one project is found.

We should match the behavior of equivalent commands such as `dotnet list package`, `dotnet restore`. 

I think those fail when multiple projects are seen (ambiguity), but best way to figure it out would be just test it out :D

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: If explain is already used in sql, I personally think it would be best to follow the same naming convention here too. Are there any other suggestions/things to consider?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I think both of these suggestions are great, but for the first minimal version we should probably exclude them for now. However, I will definitely document these in the spec.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: > Does the .NET SDK team have a "rule" that all dotnet cli commands must have msbuild level verbosity?

Is there somewhere where I can check the rules? If there is no rule, then I don't think the verbosity option is necessary.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: The `why` name comes from npm's why command, which is the exact same problem space as the one we're solving.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Since the only possible parameter that can be given to this command is a package name, I think it makes sense to keep it as `dotnet nuget why`.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: > The Status field made sense (kind of) when we used to put specs in the wiki. However, now that they're markdown files in the repo, that use pull requests, I think we should remove this.
> 
Agreed, will remove this. 
> Similarly, when this PR is approved and merged, it's no longer "proposed" (file path), it's approved. In my opinion, there should be no `proposed/` folder in the repo root.

I think this is probably a larger conversation, but I do agree with the reasoning.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: If we don't add the version in the minimal solution, I'm unsure what should happen in the case that Kartheek mentioned. Would we just arbitrarily choose one of the matching package references to print out the dependency tree for?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: This is a great point, I will look into it more and update the spec accordingly.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Agreed, I can change the wording.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: Packages are managed on a project level. 

If running it on a solution, we just show project number of outputs. 

Running on a solution just means running it for each project separately.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: > > I believe no error should be thrown if more than one project is found.
> 
> We should match the behavior of equivalent commands such as `dotnet list package`, `dotnet restore`.
> 
> I think those fail when multiple projects are seen (ambiguity), but best way to figure it out would be just test it out :D

Agreed, after testing it out I'll update the spec.

proposed/2022/dotnet-nuget-why-proposal.md
JonDouglas: Should try to remain consistent with other package managers such as 'why' or 'tree'. SQL explain is a good prior art though

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Thanks for adding that info. Just wondering if the PackageDownloads are not recorded in the targets/library section, where would they be recorded?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Ok, I am convinced. Also based on this https://github.com/NuGet/Home/pull/11875#discussion_r897458421, it seems like @kartheekp-ms and @chgill-MSFT agree that `dotnet nuget why` would be best.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: The section is: 
project -> frameworks -> {frameworkName} -> downloadDependencies

Try looking at an assets file such as NuGet.CommandLine in our repo. 

artifacts\NuGet.Commandline\obj\project.assets.json

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Makes sense, I agree we can remove the version part for the first minimal version.

proposed/2022/dotnet-nuget-why-proposal.md
erdembayar: > Does the .NET SDK team have a "rule" that all dotnet cli commands must have msbuild level verbosity?

@baronfel
Could you be able to clarify on this for us? Could we create dotnet cli command without verbosity options if it's not really necessary?

proposed/2022/dotnet-nuget-why-proposal.md
erdembayar: But it doesn't leave us room to grow next time. For example, if we ever want to add `dotnet nuget config why`, `dotnet nuget xxx why` (I made up them for now) then we can't use it anymore, in the future nuget going to add more actions and functionality and it may need clarification in the future.

proposed/2022/dotnet-nuget-why-proposal.md
baronfel: We have design guidance for CLIs (especially those that operate in the context of the .NET SDK CLI) [here](https://docs.microsoft.com/en-us/dotnet/standard/commandline/syntax#design-guidance).  

Specifically for the [verbosity option](https://docs.microsoft.com/en-us/dotnet/standard/commandline/syntax#the---verbosity-option), we suggest that _if_ a command supports variable verbosity, then use the 'standard' 5 levels. For a command like `why`, it may not make sense for there to be different levels of output, so dropping `--verbosity` would be acceptable. If/when you _do_ need variable verbosity, then you can decide what levels you need within the 5 standard levels.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,137 @@
...
+```
+dotnet nuget why packageA
+
+Project 'projectNameA' has the following dependency graph for 'packageA'
+   [net6.0]: Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > packageA (1.0.0)
+   [net472] Microsoft.ML (1.0.0) > Microsoft.ML.Util (1.0.0) > packageA (1.0.0)
+
+Project 'projectNameB' has the following dependency graph for 'packageA'
+   [net6.0]: Microsoft.ML (1.1.0) > Microsoft.ML.Util (1.1.0) > packageA (1.1.0)
+   [net472] Microsoft.ML (1.1.0) > Microsoft.ML.Util (1.1.0) > packageA (1.1.0)
```
erdembayar: Personally, I feel `->` as separator is more readable for end users. What do you think?
image

nkolev92:
@pragnya17 whenever you are done and ready for review, can you please let us know by clicking the re-request review.
Thanks!

image

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Makes sense, I've updated the explanation to reflect this.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Agreed, I removed the parts that were not relevant to the problem.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: After testing out `dotnet list package`, I found that if there were multiple project it would list the packages for each project. So, I will update the spec to reflect similar behavior for the `dotnet nuget why` command.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: That's a good point, I agree!

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I see where you're coming from @erdembayar . Any additional thoughts @kartheekp-ms and @chgill-MSFT ?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Based on this comment, I think we have agreed to removing the --version for this first minimal version of the feature.

pragnya17:
There are a few more comments here that I have left unresolved because I'm not quite sure how to answer them. I hope to get some more clarity on them soon, and we can also talk about them during the spec review meeting I set up.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Explanation
+
+### Functional Explanation
+
+The NuGet restore operation generates a `project.assets.json` file under the `obj` folder for `PackageReference` style project. This file maintains a project's dependency graph. Therefore, this file can be used in order to produce a dependency graph when the `dotnet nuget why` command is called. 
+
+### Technical Explanation
+
+The `dotnet nuget why` command will print out the dependency graph of a given package only if it is part of the final graph.
```
zivkan, line 113: My concern about the command/verb `why` is that it sounds so generic that it might confuse customers. For example, if a customer is having a network connection issue to a package source, can `dotnet nuget why` help?  What if they're having TLS/encryption issues?  What about authentication issues? What about signed package root certificate trust issues?  How about restore failed (or maybe just had warnings), and now the customer wants hints on how to fix those problems. `dotnet nuget why` certainly sounds like it might help.

I know it's easier for customer to use short commands, so it's going to be impossible to beat `why`, but something like `dotnet nuget package path`, or `dotnet nuget tree` might be names that are more clearly relevant to dependency graphs.

nkolev92: `dotnet package why`? :D 

I don't think we'd be able to make it a first class verb.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,124 @@
...
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will be an additional header titled “Top-level Packages”.
+
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
+![](../../meta/resources/TransitiveDependencies/TransitiveVSPMUI.png)
+
+
+## Unresolved Questions
+
+Projects that use `packages.config` format are out of scope for this effort because the XML file contains a flat list of both direct and transitive package dependencies. Customers cannot visualize the dependency graph from the file itself but it would be easy to update the package version of a transitive depedency if there are any security vulnerabilities in the current version.
```
zivkan: The .NET CLI (and therefore `dotnet restore`) doesn't support packages.config, so I don't think we should even consider supporting it in other `dotnet nuget` commands. I'd consider it completely out of scope, rather than an unresolved question.

pragnya17: Sure, that makes sense, I'll remove it from the spec.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: After looking at it, I think it would make most sense to display all possible paths if a package is brought in by more than one path. I have updated the spec with an example to reflect this.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: That's a great point Andy. @erdembayar also echoed similar concerns in this comment: https://github.com/NuGet/Home/pull/11875#discussion_r897028332 . Based on all of your comments it seems like `dotnet package why` might be the best solution?

Also @chgill-MSFT, do you have any feedback on the naming for this command?

nkolev92: cc @JonDouglas and @baronfel as well.


@pragnya17 @kartheekp-ms maybe we need to have a separate discussion about this.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: fwiw, I think every possible path is the only reasonable behavior. 

Say I want to get rid of a certain package in my graph for whatever reason. 

1. Run dotnet nuget why,figure out a top level dependency
2. Tweak the version to get to one that excludes that package it.
3. dotnet list package/dotnet nuget why, I see that same damn package *again*. 
4. dotnet nuget why...

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+
+- PACKAGE_NAME
+
+The package for whom the dependency graph has to be identified. Package name wildcards are not supported.
+
+#### Options
+
+- -f|--framework <FRAMEWORK>
+
+Search package dependency graph for a specific target framework.
```
nkolev92: I understand what you're trying to say with his, but something about it just reads off. 
Curious if it's just me or others feel the same way. 

Maybe just removing the package word fixes it,. 
Alternatively  `Search for a package in the dependency graph for a specific target framework`.


Basically I'm thinking that the focus should be on the package itself, and the specific target framework. Right now, I sort of read it like we're search for the specific target framework.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: We're honestly kind of all over right now, that I'm not sure there's a 100% perfect solution.

I think if we're gonna be a adding a new verb, we'd either be specific to nuget only, or consider whether it makes sense as a first class verb in the SDK altogether. 
So extending something to dotnet nuget config why feels wrong, but we also have dotnet nuget sources already.
Illustrated with examples: 

dotnet nuget why - This is a nuget specific command. `why` applies to whatever seems to be the default for nuget (think dotnet nuget sign).
dotnet nuget package why - This is a nuget specific command and a package specific command? package is a first class noun in dotnet.exe, having it be under nuget explicitly is probably unnecessary. On the other, it's flexible as @erdembayar said.
dotnet package why - this makes sense if package is what we care about the most, but does why apply to other nouns?

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+```
+
+- List dependency graph of a package given `package id` when there is a diamond dependency (a package is brought in by more than one path).
+
+```
+dotnet nuget why packageA
+
+Project 'projectNameA' has the following dependency graph for 'packageA'
+   [net6.0]: Microsoft.ML (1.0.0) -> Microsoft.ML.Util (1.0.0) -> packageA (1.0.0)
+   [net6.0] Microsoft.ML (1.0.0) -> Microsoft.ML.SampleUtils (1.0.0) -> packageA (1.0.0)
```
nkolev92:
```suggestion
   [net6.0] Microsoft.ML (1.0.0) -> Microsoft.ML.SampleUtils (1.0.0) -> packageA (1.0.0)
```

Should net6.0 be a more of a top level thing and then we nest?

I don't have too strong of an opinion here. 

Ie. should we have:

```
   net6.0:
     Microsoft.ML (1.0.0) -> Microsoft.ML.Util (1.0.0) -> packageA (1.0.0)
     Microsoft.ML (1.0.0) -> Microsoft.ML.SampleUtils (1.0.0) -> packageA (1.0.0)
```

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+
+Project 'projectNameB' has the following dependency graph for 'packageA'
+   [net6.0]: Microsoft.ML (1.1.0) -> Microsoft.ML.Util (1.1.0) -> packageA (1.1.0)
+```
+
+## Drawbacks 
+
+The contents of the `project.assets.json` will be helpful in understanding the dependency graph of a given package but there are a few limitations:
+
+-  There may be some packages in the file that are not part of the final graph because the NuGet restore operation downloads other packages during dependency resolution. The absence of a package id and version in the `project.assets.json` file indicates that a package was downloaded during dependency resolution but is not part of the final dependency graph.
```
nkolev92: I'm not sure this first bullet point is truly a drawback. 

The fact that packages that aren't explicitly used are reported in CG is unrelated to what this feature aims to do. 
The installation of packages in the GPF != usage of package in the project. 

What is a consideration is potentially printing the graph *with* rejected package versions, but that's probably something difficult to be done outside of restore itself.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Rationale and Alternatives
+
+We could have considered a minor tweak to the existing experience of “dotnet list package --include-transitive” to provide the user with a sense of a package's dependencies. This could have been a new column next to “Resolved” called “Transitively Referenced” that had a list of the top-level packages that requested the dependency. The corresponding tracking issue can be found here: https://github.com/NuGet/Home/issues/11625. Unfortunately, this solution doesn't provide an easy way to identify the dependency graph `(starting from project -> top-level package -> transitive-dependency (1) -> ..-> transitive dependency (n))` for a particular package.
+
+![](../../meta/resources/TransitiveDependencies/TransitiveDotNetCLI.png)
+
+## Prior Art
+
+Within Visual Studio, there will be a new panel within the Visual Studio Package Manager UI named “Transitive Packages” in which all transitive packages will be displayed to the user. For the sake of not confusing the user, there will be an additional header titled “Top-level Packages”.
```
nkolev92: The VS one is also lacking in the fact that it hides the transitive dependency hidden through projects. 
dotnet nuget why doesn't have that limitation.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+
+When a user highlights a transitive package, they will see a pop-up that displays how the transitive dependency originated & what top-level package(s) are bringing it in.
+
+![](../../meta/resources/TransitiveDependencies/TransitiveVSPMUI.png)
+
+## Additional improvements
+
+Add a [--version <VERSION>] option to the command if the user wants to print dependency graphs for a specific version of the package. See NuGet package versioning for more information.
```
nkolev92: NuGet currently `flattens`, so it only allows one version of a package per framework to be resolved. Could be helpful to add it there.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,131 @@
...
+
+Allow the customer to look up transitive dependencies of more than one package. For example: `dotnet nuget why [<PROJECT>|<SOLUTION>] packages package1, package2` or `dotnet nuget why [<PROJECT>|<SOLUTION>] package 'nuget.*'`.
+
+Allow the customer to transitive dependencies of more than one framework. For example: `--framework net6.0 --framework netstandard2.0`.
```
nkolev92, line 209: dotnet add package supports multiple frameworks. 

I don't think you have to do it in the first project, but I don't think it'd be worst to add it as well, but I also don't think it's necessary. 

Personally, I can't imagine myself having any issues even if the framework option did not exist.

nkolev92 (commented):
Looks pretty good to me.
Sounds like there's some naming feedback to resolve.

Great job overall though! I'm really excited about this feature.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Sure, this way it's not as redundant as it is right now.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: > dotnet add package supports multiple frameworks.

Looking at other nuget commands is actually why I added this in the additional improvements section. I think I will leave it as is for the first project since I only have 4 weeks left in my internship. However, I wanted to document it just in case for future reference.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I see what you mean, I changed the wording which hopefully makes it clearer.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,138 @@
...
+   [net6.0]:
+      Microsoft.ML (1.1.0) -> Microsoft.ML.Util (1.1.0) -> packageA (1.1.0)
+```
+
+## Drawbacks 
+
+The contents of the `project.assets.json` will be helpful in understanding the dependency graph of a given package but there are a few limitations:
+
+-  There may be some packages in the file that are not part of the final graph because the NuGet restore operation downloads other packages during dependency resolution. The absence of a package id and version in the `project.assets.json` file indicates that a package was downloaded during dependency resolution but is not part of the final dependency graph.
+- Packages acquired through `PackageDownload` are not tied to the project in any way beyond acquisition. These packages are not recorded in the final graph that `project.assets.json` maintains, however they are recorded under project -> frameworks -> {frameworkName} -> downloadDependencies.
```
pragnya17: Should the packages acquired through `PackageDownload` be included in the final dependency graph?

pragnya17: After an offline conversation with @kartheekp-ms we agreed this part could be moved to additional improvements for a future version of this project.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: That's a great point, I have added it!

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Could you expand on what you mean by this? Are you referring to the lack of details in the PMUI view or the solution explorer view (or potentially both)?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I reworded this section a bit to reflect what you are saying. Hopefully that helps clarify things!

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+      Microsoft.ML (1.0.0) -> Microsoft.ML.Util (1.0.0) -> packageA (1.0.0)
+
+Project 'projectNameB' has the following dependency graph for 'packageA'
+   [net6.0]:
+      Microsoft.ML (1.1.0) -> Microsoft.ML.Util (1.1.0) -> packageA (1.1.0)
+```
+
+## Limitations 
+
+The NuGet restore operation downloads other packages during dependency resolution which are not part of the final dependency graph that the `dotnet nuget why` command prints out. The absence of a package id and version in the `project.assets.json` file indicates that a package was downloaded during dependency resolution but is not part of the final dependency graph.
```
ericstj: I think this limitation will make the feature insufficient for many of the component governance cases I have helped people with.  I think that this verb should do its own graph walk rather than rely on the one produced by restore, or in addition to.  That way it can identify other versions that nuget may have downloaded and why.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+
+![](../../meta/resources/TransitiveDependencies/SolutionView.png)
+
+However, in comparison to Visual Studio, the dotnet CLI does not provide insight into “why” a transitive dependency is listed, although it does list everything that has been resolved as seen below.
+
+![](../../meta/resources/TransitiveDependencies/DotNetCLI.png)
+
+## Motivation
+
+A developer should be able to understand where every package in a solution/project originated from. This can help them understand complex dependency chains they may have within their projects. Having this understanding can help a developer promote a transitive dependency quickly & easily to a top-level dependency in case there is a security concern or conflict.
```
JonDouglas, line 18: Great motivation here.

proposed/2022/dotnet-nuget-why-proposal.md
JonDouglas: https://twitter.com/nuget/status/1546878617199804418

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+```
+#### Arguments
+
+- PROJECT | SOLUTION
+
+The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution is found, an error is thrown. If more than one project is found, the dependency graph for each project is printed. 
+
+- PACKAGE_NAME
+
+The package for whom the dependency graph has to be identified. Package name wildcards are not supported.
```
JonDouglas: Could we simplify saying "exact package name / id"?

proposed/2022/dotnet-nuget-why-proposal.md
JonDouglas: See https://github.com/NuGet/Home/pull/11875/files#r919109645

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+Prints out a description of how to use the command.
+
+#### Examples
+
+- List dependency graph of a package given `package id`.
+
+```
+dotnet nuget why packageA
+
+Project 'projectNameA' has the following dependency graph for 'packageA'
```
JonDouglas, line 148: Great example! Very clear.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+      Microsoft.ML (1.1.0) -> Microsoft.ML.Util (1.1.0) -> packageA (1.1.0)
+```
+
+## Limitations 
+
+The NuGet restore operation downloads other packages during dependency resolution which are not part of the final dependency graph that the `dotnet nuget why` command prints out. The absence of a package id and version in the `project.assets.json` file indicates that a package was downloaded during dependency resolution but is not part of the final dependency graph.
+
+## Rationale and Alternatives
+
+We could have considered a minor tweak to the existing experience of “dotnet list package --include-transitive” to provide the user with a sense of a package's dependencies. This could have been a new column next to “Resolved” called “Transitively Referenced” that had a list of the top-level packages that requested the dependency. The corresponding tracking issue can be found here: https://github.com/NuGet/Home/issues/11625. Unfortunately, this solution doesn't provide an easy way to identify the dependency graph `(starting from project -> top-level package -> transitive-dependency (1) -> ..-> transitive dependency (n))` for a particular package.
```
JonDouglas, line 188: I think I put this in the original transitive dependencies spec as a CLI enhancement. I'm not sure it will scale well when there's more than a couple of transitively referenced parents. I think a separate command is the right approach here. Good callout!

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,139 @@
...
+
+Within Visual Studio, there is a new panel in the Package Manager UI called “Transitive Packages” in which all transitive packages are displayed to the user. There is also an additional header titled “Top-level Packages”.
+
+When a user highlights a transitive package, they will see a pop-up that displays what top-level package(s) are bringing it in.
+
+![](../../meta/resources/TransitiveDependencies/TransitiveVSPMUI.png)
+
+However, this still does not provide the user with enough detail about how transitive packages originate, therefore we propose the `dotnet nuget why` command.
+
+## Additional improvements
```
JonDouglas, line 202: Great improvements list. While they all can add value, the spec is in a good state to finalize the first iteration in my opinion!

JonDouglas (approved):
Wonderful spec. Great write-up. I'm approving the design for now although there is still outstanding discussion on the name. I hope I can help bring clarity to the name very shortly.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Since I have limited time as an intern I don't think I will be able to implement the feature such that it does its own graph walk and includes all the packages that were downloaded during dependency resolution. It would require a more extensive implementation since those packages are not recorded in the assets file. 
However, I can definitely document this under future improvements so that future iterations of this command can potentially address this concern.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Thanks to @kartheekp-ms for coming up with it!

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: It was @kartheekp-ms 's idea to include this, so thanks Kartheek!

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: I get why people picked `tree`, but in reality, it's a graph. It feels kind of incorrect to use the incorrect data structure.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: We're also going to be looking at the other side of the picture, in particular the reporting of these packages by CG. 

We don't have details what that might lead to just yet, but it is conversation that's been started.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Yes, others that I've discussed this with also expressed similar concerns about using `dotnet nuget tree`. After also sending out an internal poll, the results of which are below, @JonDouglas and I decided it would be best to use `dotnet nuget explain.`

If anyone else has any feedback, I would love to hear it!
MicrosoftTeams-image

erdembayar: @pragnya17 
Great. 
Are you going to use `dotnet nuget explain package newtonsoft.json` or `dotnet nuget explain newtonsoft.json`?

baronfel: @erdembayar I think you have the same syntax twice, did you mean to compare between `dotnet nuget explain package newtonsoft.json`and `dotnet nuget explain newtonsoft.json`?

I think explain is a fine term - I have a personal preference towards `why` based on existing commands like `npm why`, `yarn why`, `pnpm why`, `paket why`, etc, but explain definitely usable. Out of curiosity, what were the `Other` entries in that survey?

One thing I worry about is the similarity between `dotnet nuget explain` and a potential future command I made up just now called `dotnet nuget info <package Id>` that dumps a formatting of the package's metadata in the console. How would you differentiate between these terms if we had such a command? It's a bit of a made-up exercise, because as far as I know no-one has asked for `dotnet nuget info` but 'explain' is a fairly wide design space so I wanted to call that out.

erdembayar: @baronfel 
You're right, I corrected it.

nkolev92: I like `why` better than `explain` as well, due to what the commands in other ecosystems are named. 
I totally forgot about `paket why` which is a tool for .NET projects.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: In the PM UI view, transitive packages that are transitive because they come from a project are not shown in the PM UI. 

Which isn't great imo, but it was a design decision in the PM UI. 

This command wouldn't have that issue.

proposed/2022/dotnet-nuget-why-proposal.md
nkolev92: both `why` and `explain` carry a lot of ambiguity, so I think it's ok if we lean towards what's known. 

I don't think the 1 extra vote for explain is a strong enough signal to lean that way.

proposed/2022/dotnet-nuget-why-proposal.md
```diff
@@ -0,0 +1,222 @@
...
+```
+#### Arguments
+
+- PROJECT | SOLUTION
+
+The project or solution file to operate on. If not specified, the command searches the current directory for one. If more than one solution is found, an error is thrown. If more than one project is found, the dependency graph for each project is printed. 
+
+- PACKAGE_NAME
+
+The exact package name/id for which the dependency graph has to be identified. Package name wildcards are not supported.
```
kartheekp-ms, line 129: Should we also accept multiple packages or a regular expression pattern to match package names?

Most of the existing `dotnet nuget` commands accepts multiple packages that includes a regular expression pattern. For example, [dotnet nuget verify](https://docs.microsoft.com/dotnet/core/tools/dotnet-nuget-verify#arguments) command.

proposed/2022/dotnet-nuget-why-proposal.md
baronfel: Here's some context on the `paket why` command, for comparison/inspiration in case folks haven't seen it before :)

Paket manages all dependencies at a solution level in a file that's essentially equivalent to a combination of CMPV _and_ Package Source Mapping. After the resolution step, all resolved versions are written to a lock file, and both of these files are intended to be checked in to source control.  As a result, `paket why` is pretty fast. 

The command takes a single argument, which is a dependency that's either a direct dependency, a transitive dependency, or an unknown dependency. The following output is taken from the fsharp/fsautocomplete repository.

Direct dependency:

```
➜ dotnet paket why Serilog
NuGet Serilog - 2.11 is a direct (paket.dependencies) dependency.
It is part of following dependency chains:

-> Destructurama.FSharp - 1.2
  -> Serilog - 2.11

-> Serilog.Sinks.Async - 1.5
  -> Serilog - 2.11

-> Serilog.Sinks.Console - 4.0.1
  -> Serilog - 2.11

-> Serilog.Sinks.File - 5.0
  -> Serilog - 2.11
```

This calls out the kind of dependency (direct/transitive), the location of the file that introduced the dependency, and then all of the dependencies that influenced the resolution of the requested dependency.

Transitive dependency:

```
➜ dotnet paket why System.Security.Cryptography.Pkcs
NuGet System.Security.Cryptography.Pkcs - 6.0.1 is a transitive dependency.
It is part of following dependency chains:

-> FSharp.Analyzers.SDK - 0.11
  -> FSharp.Compiler.Service - 41.0.5
    -> Microsoft.Build.Tasks.Core - 17.2
      -> System.Security.Cryptography.Pkcs - 6.0.1

-> FSharp.Formatting - 14.0.1
  -> FSharp.Compiler.Service - 41.0.5
    -> Microsoft.Build.Tasks.Core - 17.2
      -> System.Security.Cryptography.Pkcs - 6.0.1

-> FSharpLint.Core - 0.21.2
  -> FSharp.Compiler.Service - 41.0.5
    -> Microsoft.Build.Tasks.Core - 17.2
      -> System.Security.Cryptography.Pkcs - 6.0.1
```

Here you just get the dependency kind and the chains.

Missing/unknown dependency:

```
➜ dotnet paket why MyCoolPackage
NuGet MyCoolPackage was not found in paket.lock
```

pragnya17: After @baronfel 's comment specifically, I think it would be best to go back to `dotnet nuget why` because it is very similar to paket why in terms of printing out dependency paths. Also I agree with @nkolev92 that the survey results can't be taken at face value. Therefore my vote is for `dotnet nuget why`. 

I know @JonDouglas was leaning towards `dotnet nuget explain`, do you have any further thoughts?

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I think this is a great idea! Given the limited time of my internship I don't think I will be able to implement it, but I will definitely document it in the spec under additional improvements.

proposed/2022/dotnet-nuget-why-proposal.md
JonDouglas: For the sake of time, I think moving forward with the original proposal is ideal. At a later date we can alias to `explain` or another verb entirely.

While my personal leaning is opposite due to my surprise of people's first impressions (i.e. start with `explain` and alias later to `why`), it really isn't a concern from my end.

I believe the best verbs here to describe the functionality as written in this proposal are `why` and `explain`. I believe that `tree` and `graph` would not be justified. What was clear to me when we first proposed this work is that majority of other package ecosystems have a "why" command with an alias(i.e. `explain`, `depends`, etc). Let's start here.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: I forgot, but this is actually already documented here: https://github.com/NuGet/Home/pull/11875/files#diff-79e5d3f65439f3d176fb83b648e933018d283daf2bb4b4fef783785b4ed79249:~:text=Allow%20the%20customer%20to%20look%20up%20transitive%20dependencies%20of%20more%20than%20one%20package.%20For%20example%3A%20%60dotnet%20nuget%20why%20%5B%3CPROJECT%3E%7C%3CSOLUTION%3E%5D%20packages%20package1%2C%20package2%60%20or%20%60dotnet%20nuget%20why%20%5B%3CPROJECT%3E%7C%3CSOLUTION%3E%5D%20package%20%27nuget.*%27%60.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Thanks for explaining, I have updated the spec to reflect that.

proposed/2022/dotnet-nuget-why-proposal.md
pragnya17: Based on this conversation: https://github.com/NuGet/Home/pull/11875#discussion_r924901011 , we have decided to move forward with `dotnet nuget why`.

nkolev92 (approved):

Issue #5926, opened 2017-09-22
Consume pdbs from packages in PackageReference - add Debug Symbols to LockFileTargetLibrary


wli3:
https://github.com/dotnet/sdk/issues/1458 depends on this

To include PDB to _CopyLocal_, SDK need _NuGet.ProjectModel_ to categorize PDB and XML file to an ideally different category like "debug symbol" or "related file". 

So SDK can get all the symbol files by something like `var symbolPath = lockFileTargetLibrary.debugSymbols` , just like CompileTimeAssemblies, RuntimeAssemblies today

https://github.com/NuGet/NuGet.Client/blob/efae789c75899ca0bdd063f34a4b73cd41510d56/src/NuGet.Core/NuGet.ProjectModel/LockFile/LockFileTargetLibrary.cs#L27

- [x] https://github.com/nuget/home/issues/11350
- [x] https://github.com/nuget/home/issues/11351
- [x] https://github.com/nuget/home/issues/11352

nkolev92:
@wli3 @nguerrera 

I know we discussed this in a sync-up meeting. 
If there's any sort of urgency for this issue, let us know so we can account for that in our triaging. 

fyi @aortiz-msft

nguerrera:
It was on me to write up a more complete proposal and I got caught up with release issues since. I still intend to push forward on this. It would be good to have it on your radar for soon. There are a lot of pings on related sdk issues for quite some time. I think we should try to prioritize it.

nkolev92:
@wli3 
Let us know when this becomes a priority for you.

nguerrera:
Here is a small proposal to push this forward:

For any given asset, if there are files next to it that differ only by extension AND those files are not themselves assets of another type, then add a `"related"` array underneath the asset listing the extensions of these files.

``` diff
{
  "version": 3,
  "targets": {
    ".NETCoreApp,Version=v5.0": {
      "Newtonsoft.Json/12.0.3": {
        "type": "package",
        "compile": {
-          "lib/netstandard2.0/Newtonsoft.Json.dll": {}
+          "lib/netstandard2.0/Newtonsoft.Json.dll": {
+             "related": [".pdb", ".xml"]
+          }
        },
        "runtime": {
-          "lib/netstandard2.0/Newtonsoft.Json.dll": {}
+          "lib/netstandard2.0/Newtonsoft.Json.dll": {
+             "related": [".pdb", ".xml"]
+          }
      }
    }
  },
```

[...]

PR #11884, opened 2022-06-14
Adds Package Scoring v1

JonDouglas:
This proposal introduces a concept known as package scoring or net score for short. This is a pagerank-like score that depends on qualities of the whole NuGet package & dependencies while accounting for the risks of today's modern, connected world and all of the security implications they have on package managers. This is the first iteration building on-top of the work here:

https://github.com/dotnet/designs/pull/216

[Rendered Proposal](https://github.com/NuGet/Home/blob/package-scoring-v1/proposed/2022/package-scoring-v1.md)

Thank You 🎉


jeffkl (dismissed):

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+
+As a means to scale package scoring to it's potential through continuous iteration, each category will have its own set of issues that a user can be aware of when browsing through the package. Keeping the same idea of **being in control** or **not in control** also applies to package issues.
+
+For issues which can be addressed, an issue will include an empowering message regarding how one can take action to help address the issue.
+
+EX: A package might be missing a README which a developer may not be able to easily get started with the package. They might be able to then contribute a new issue or even the inclusion of the README file to the NuGet package if it's open source.
+
+**Issue Examples:**
+
+- Invalid .nuspec / .csproj
```
joelverhagen: .csproj is not known by NuGet.org and is lost at the time of "pack". I don't see how this could be a dimension considered by NuGet.org or by a CLI command that operates offline, against the .nupkg file.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Missing dependency
+- Missing README
+- No example
+- Not enough public API documentation
+- Bad semver
+- Not v1+
+- No website
+- No repository
+- Unmaintained
+- No bug tracker
```
joelverhagen, line 72: We don't have dedicated metadata for bug tracker URL. And for the life of me I couldn't find an issue tracking such a feature request. We could potentially render an "issues" URL if a GitHub repository URL is provided by this isn't even perfect. You can turn off issues at the repo level in GitHub (see NuGet/NuGet.Client repo) so we can't even assume `{GitHub repo}/issues` is a valid link.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Has CVE (Critical, High, etc)
+- Deprecated license
+- Missing license
+- Non SPDX license
+- Unsafe copyright
+- Deprecated
+- Empty package
+- Malware
```
joelverhagen: Malware should be removed, not shown as available with a lower score.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- etc
+
+#### Popularity
+
+Popularity is the more unique of the other three categories which are more issue based. Popularity is calculated as how many projects depend on a package over the past 30 days. This is scored as a percentile up to 100% (most used) to 0% (least used).
+
+Since this is not possible with current data pipelines, we will work on providing such an experience in the future and use two proxy values such as total weekly downloads and total count of packages depending on the package to be scored up to a percentile of up to 100%.
+
+- Total Weekly Downloads
+- Number of Dependents
```
joelverhagen, line 89: Could this be tuned to dependents _owned by other others_? Or perhaps number of distinct owners depending on the package with their own package?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+#### Quality
+
+Quality is the combination of:
+
+- Follow NuGet conventions
+  - Provides a valid .csproj / .nuspec
+  - Provides a valid README
+- Provide documentation
+  - Provides an example
+  - 20% or more of the public API has xml doc comments
```
joelverhagen, line 101: This may be fixed soon (per https://github.com/NuGet/Home/issues/5926) but I believe XML docs don't work with <PackageReference> today. It's unclear whether this is a useful quality metric while the E2E is not working. Perhaps @zivkan or @heng-liu can correct me here.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Platform support
+  - Supports all possible modern platforms i.e. [.NET 5+/.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-5-and-net-standard) (iOS, Android, Windows, MacOS, Linux, etc)
+- Pass static analysis
+  - Code has no errors, warnings, lints, or formatting issues.
```
joelverhagen, line 105: Static analysis would be performed on compiled assemblies, not the input source, right? Or would we need to correlate source and run static analysis on that?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+
+Total Score - 100
+
+#### Maintenance
+
+Maintenance is the simplest of them all as it pertains to whether the package has been updated in a reasonable timeframe. Packages that have not been updated in more than a year may be unmaintained and any problems with the package may go unaddressed.
+
+- Package updated in the last year.
+
+The score may also be shown as a decayed value over the year approaching the year mark where it goes stale.
```
joelverhagen, line 120: I wonder how a package owner can attest "this package is still maintained but it hasn't needed an update in over 1 year". Or is this a case we don't think is common?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+Since this is not possible with current data pipelines, we will work on providing such an experience in the future and use two proxy values such as total weekly downloads and total count of packages depending on the package to be scored up to a percentile of up to 100%.
+
+- Total Weekly Downloads
+- Number of Dependents
+
+Total Score - 100
+
+#### Quality
+
+Quality is the combination of:
```
joelverhagen, line 95: What about:

- sources available
- symbols available
(likely SourceLink)

i.e. the package is easily debuggable.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+EX: A package might be missing a README which a developer may not be able to easily get started with the package. They might be able to then contribute a new issue or even the inclusion of the README file to the NuGet package if it's open source.
+
+**Issue Examples:**
+
+- Invalid .nuspec / .csproj
+- Missing dependency
+- Missing README
+- No example
+- Not enough public API documentation
+- Bad semver
```
joelverhagen, line 67: What does bad semver mean?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+
+We then asked how people install a package with the following emphasis in order.
+
+1. Package solves my problem.
+2. Package has enough downloads.
+3. Package is open source.
+4. Package has high quality documentation.
+5. Package is maintained by a notable author or organization.
+6. Package has been updated recently and is updated regularly.
+7. Package has few dependencies.
```
joelverhagen, line 173: I wonder if it's worth analyzing or surfacing the total size of a package's dependency graph or just consider the top level. If the later, maybe a score based on this could be gamed/unintentionally obfuscated by having a single meta-package dependency.

proposed/2022/package-scoring-v1.md
nkolev92: I'd probably remove this altogether. You can't publish a package with an invalid nuspec.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+For issues which can be addressed, an issue will include an empowering message regarding how one can take action to help address the issue.
+
+EX: A package might be missing a README which a developer may not be able to easily get started with the package. They might be able to then contribute a new issue or even the inclusion of the README file to the NuGet package if it's open source.
+
+**Issue Examples:**
+
+- Invalid .nuspec / .csproj
+- Missing dependency
+- Missing README
+- No example
```
nkolev92, line 65: Can you elaborate on this?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Not enough public API documentation
+- Bad semver
+- Not v1+
+- No website
```
nkolev92, line 69: does this refer to the projecturl metadata?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- No repository
+- Unmaintained
+- No bug tracker
+- Has CVE (Critical, High, etc)
+- Deprecated license
+- Missing license
+- Non SPDX license
+- Unsafe copyright
```
nkolev92, line 77: what would this be?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+
+#### Issues
+
+As a means to scale package scoring to it's potential through continuous iteration, each category will have its own set of issues that a user can be aware of when browsing through the package. Keeping the same idea of **being in control** or **not in control** also applies to package issues.
+
+For issues which can be addressed, an issue will include an empowering message regarding how one can take action to help address the issue.
+
+EX: A package might be missing a README which a developer may not be able to easily get started with the package. They might be able to then contribute a new issue or even the inclusion of the README file to the NuGet package if it's open source.
+
+**Issue Examples:**
```
nkolev92, line 61: This is a great list. 

Wonder if we should take a look at how feasible all the individual bullet points are

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Number of Dependents
+
+Total Score - 100
+
+#### Quality
+
+Quality is the combination of:
+
+- Follow NuGet conventions
+  - Provides a valid .csproj / .nuspec
```
nkolev92: I'd remove this. Only valid packages can be created.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+  - Provides a valid README
```
nkolev92, line 98: What would a valid readme mean? Something that renders correctly?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+- Provide documentation
+  - Provides an example
+  - 20% or more of the public API has xml doc comments
+- Platform support
```
nkolev92, line 102: We should be mindful about how we handle this for packages that are dedicated to a specific platform only, example mac only. 
There could be an android matching one that's a separate package potentially and they have the same dependencies. 

Not sure if we can easily capture any of my comments in automation, but just wanted to make sure we don't miss out on that perspective :)

proposed/2022/package-scoring-v1.md
nkolev92: I think some of these are better applicable to eco-systems where the code is shipped rather than compiled assemblies. 
Example formatting.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+  - Supports all possible modern platforms i.e. [.NET 5+/.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-1-0#net-5-and-net-standard) (iOS, Android, Windows, MacOS, Linux, etc)
+- Pass static analysis
+  - Code has no errors, warnings, lints, or formatting issues.
+- Support up-to-date dependencies
+  - All of the package dependencies are supported in the latest version
```
nkolev92, line 107: Not sure I understand this. 
Because of how compatibility works, if a package supports a certain framework, then all future versions of that framework are likely to be supported as well. 

If a package claims to support a framework but their dependencies don't then that's a problem package. 
Maybe that's what we should focus on instead, ensure that the package is installable in all of it's declared frameworks.

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,222 @@
...
+  - Package supports the latest stable .NET SDK
```
nkolev92, line 108: How would this work?
.NET SDK support is not really a package construct right now.

nkolev92 (commented):
This is great write-up, well done!

proposed/2022/package-scoring-v1.md
maxkoshevoi, line 95: What about "Opted in to Nullable reference types"?

https://pub.dev has this as a metric and even adds this cool badge to the package

image

proposed/2022/package-scoring-v1.md
clairernovotny: We know this with the compiler flags metadata included in the pdb's

nkolev92: I'm not sure that answers the question. 

`Supports SDK` is not really a thing. Supporting certain frameworks is what's expressed in NuGet. 

Metadata about what SDK was used to build probably isn't helpful either.

proposed/2022/package-scoring-v1.md
joelverhagen: This is a great idea, and a huge win for package consumers if implemented more broadly.

joelverhagen: /cc @loic-sharma 😃

proposed/2022/package-scoring-v1.md
zivkan: I know that different people have different preferences, but I'm on the more risk adverse side. I highly value mature and stable packages. Packages with high churn increase risk of bugs, so I strongly dislike punishing (reducing the score) for "older" packages.

Consider Newtonsoft.Json for example. Excluding the recent prerelease version, the current stable version (13.0.1) is more than 12 months ago. The time between it and the previous stable version (12.0.3) was more than 12 months.  Are we really willing to say that this package is unmaintained and deserves a reduced score?

I understand the intent is to encourage packages that have bug fixes, but I think a simple "how long since the last publish date" is too simplistic and will have worse unintended consequences than the benefits it hopes to bring.

proposed/2022/package-scoring-v1.md
joelverhagen: /cc @chgill-MSFT, @lyndaidaii - we've talked a bit about README quality in other contexts. I think _any_ valid, linked readme file should meet the criteria. Perhaps we can basically do a `string.IsNullOrWhiteSpace`. Once authors have onboarded to the feature, it's very easy to incrementally improve it. It's a better baseline to be in than just a plain text description.

Imagine if you see someone else's repo you want to open a PR against -- it's certainly more approachable to edit a single Markdown file in the PR than do all of the onboarding and potential plumbing to set it in the first place.

Kuinox:
I'm worried a lot of these metrics will lead to libraries that spend a lot of time on the marketing to be ranked better, instead of spending time building a great library.

While having a great test coverage is time consuming, it's definitely better for the package quality and their users than spending time marketing the library.

In some other ecosystems, some manages to get a lot of popularity only by spending tons of their time on marketing, while the library itself doesn't bring or solve anything new, or better.

I also fear that a lot of metrics will enforce current practice and hinder new innovative and potentially better practices.

I saw a "is the code correctly formatted" metric:
image

While for a simple static analyser standpoint, this code is not correctly formatted, such layout allow better code readability, but will penalise the scoring.


Mono.Cecil, and some other great .NET librairies would rank badly on a lot of the metrics listed, but are very valuable package for the .NET ecosystem.

In additions, a lot of these metrics are easily to cheat on. There are xml docs generators that write basic english sentences from the code, allowing to reach 100% of useless documentation instantly. You can start trying to programmatically check for the docs quality, but then someone will start generating the documentation with github copilot.

While I welcome new packages metrics, I don't think a package score, where its purpose is to help package choice, should be made from an aggregate of metrics computed on the package, it will incentivise cheating on, and these metrics will become meaningless. Worse, ranking too high on some of these metrics could rise suspicion instead of being a proof of quality.

On the topic of metrics being cheated on, it doesn't have to come from the library author themselves, for exemple, [npm audit is spammed of regex DoS](https://overreacted.io/npm-audit-broken-by-design/) and I remember reading a blogpost about CVE spam that happened to boost spammer's resume, but I sadly can't find any link to it.

Something that I didn't seen proposed, and would be far easier to implement (than a lot of static analysis), is a nuget package rating/comment systems on nuget.org.

While it causes other issues, like angry devs swearing at benevolent package maintainer, requiring moderation, heated discussion between a user that lowly rated a package and it's maintainer, etc, I think it has a lot of benefits:
- It's harder to cheat on: skewing metrics is evil, botting reviews is illegal, or at least break multiple ToS.
- I have more trust on the ratings of fellow developers, especially if I have high esteem for them.
- People searching for a package will have input of other developers that used this library and will be able to make more informed decision from that, for example "Easy to use API but too slow", well if speed is not important for your use case, you won't see a problem using this library.
- Library author will have more feedback: it's usually not well seen to come at someone github repo and write how I will not use their package because XYZ.
- Such system can be easily improved with various features. (follow people to get their review on top, tags system like you can see on various reviews systems, etc)

Thank you if you have read everything :D

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,219 @@
...
+
+Total Score - 100
+
+#### Security
+
+Finally, Security represents the trust indicators representing the package in the current state of the supply chain risk. This includes known vulnerabilities, license risk, and any issues surrounding secure supply chain risk.
+
+- No known security vulnerabilities (Critical, High, Moderate, Low) in top-level or transitive dependencies.
+- Specifies a valid license (SPDX, embedded)
+- Not deprecated.
```
loic-sharma, line 130: Should packages with installation scripts get flagged for security purposes?

proposed/2022/package-scoring-v1.md
aortiz-msft: +1. Not sure what static analysis means exactly.

proposed/2022/package-scoring-v1.md
joelverhagen: I think the concept of "flagged" in this scoring conversation is very interesting. We can have "clearly good" things like having a license and "clearly bad" things like having no content or dependencies. But what about "watch out!" things? Scripts/MSBuild tasks are not necessarily bad but they should probably be noted to the consumer.

But if we do that, then we have to think about transitivity... a sneaky package author could hide a flagged attribute in a dependency package. It gets complicated fast.

proposed/2022/package-scoring-v1.md
loic-sharma: Perhaps this could validate that the README uses HTTPS for all links and images?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,219 @@
...
+
+A high quality or "healthy" package is one that follows the following characteristics:
+
+- It is actively maintained. Either with recent commits or an annual update/notice that the package is up-to-date.
+- It is documented. It provides enough documentation to install, get started, and has public API documentation on it's members.
+- It has a way to report bugs. It provides a centralized location or repository in which issues are regularly triaged & resolved in future releases.
+- It resolves security flaws quickly. It fixes known vulnerabilities & releases an unaffected release quickly.
+- It is not deprecated. It is in an active state meeting all the criteria above.
+
+This proposal introduces the first iteration of package scoring. While it derives from the original proposal [last year on this topic](https://github.com/dotnet/designs/pull/216), it provides the minimum requirements for a v1.0 of implementing this on NuGet.org.
```
zivkan, line 22: I assume "on NuGet.org" means that Visual Studio and dotnet CLI are out of scope. Therefore, there are also no proposed changes to the NuGet protocol (that clients use to talk to servers).

Is it worth making any of this more explicit?

proposed/2022/package-scoring-v1.md
```diff
@@ -0,0 +1,219 @@
...
+- Empty package
+- etc
+
+#### Popularity
+
+Popularity is the more unique of the other three categories which are more issue based. Popularity is calculated as how many projects depend on a package over the past 30 days. This is scored as a percentile up to 100% (most used) to 0% (least used).
+
+Since this is not possible with current data pipelines, we will work on providing such an experience in the future and use two proxy values such as total weekly downloads and total count of packages depending on the package to be scored up to a percentile of up to 100%.
+
+- Total Weekly Downloads
```
erdembayar, line 88: Why weekly downloads? It can be gamed too. Because during XMAS down time someone run bunch of bots to generate 10k downloads for own package, then it it's placed at higher in ranking then it gets more downloads in following weeks. To reduce weight of this kind of spike in finance use [moving average](https://en.wikipedia.org/wiki/Moving_average) last 2-3 months. Actually, implementing moving average is not complicated, simple arithmetic.

proposed/2022/package-scoring-v1.md
KalleOlaviNiemitalo: Nullable reference types require C# 8, which is not _supported_ on .NET Framework. If NRT is scored, then it will lure developers to an unsupported scenario.

maxkoshevoi: Well that's not entirely true. You can use C# 8+ features in .Net Framework as long as they don't require some new types from SDK. 

NRT only requires a handful of attributes which can be created manually, or added via something like https://www.nuget.org/packages/Nullable#readme-body-tab

After that NRT can be fully used even in .Net Framework

proposed/2022/package-scoring-v1.md
maxkoshevoi: Heard a great quote about the Rust packages (crates) today: https://youtu.be/Z3xPIYHKSoI?t=145

> They are not abandoned, they are done

I feel like this should be something to strive towards, instead of expecting packages to be constantly updated

proposed/2022/package-scoring-v1.md
agr: How about making scripts/msbuild tasks more visible on a package details page? Like have a tab with its text, so one could evaluate the contents without downloading a package?

ghost:
This PR has been automatically marked as stale because it has no activity for **30 days**. It will be closed if no further activity occurs **within another 15 days** of this comment, unless it has a "Status:Do not auto close" label. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

PR #11916, opened 2022-06-23
Export FabricBot configuration to fabricbot.json file and add it to home repo

heng-liu:
Fixes: https://github.com/NuGet/Client.Engineering/issues/1695

Regression? Last working version: N/A

Description
As FabricBot recently removed the "save" button and move FabricBot to be a [config-as-code-only](https://1esdocs.azurewebsites.net/datainsights/merlinbot/extensions/Bot-config-as-code.html) platform, we need to export existing NuGet/Home repo fabricbot configurations to a config file named .github\fabricbot.json.

For adding new rulesets into the configuration in the future, we will need to wait for the "inflight mode" feature. So that we could test the new rulesets before merging PRs(changing this .github\fabricbot.json file).

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,271 @@
+{
+  "version": "1.0",
+  "tasks": [
+    {
+      "taskType": "trigger",
+      "capabilityId": "IssueResponder",
+      "subCapability": "IssuesOnlyResponder",
+      "version": "1.0",
+      "config": {
+        "conditions": {
+          "operator": "and",
+          "operands": [
+            {
+              "name": "isAction",
+              "parameters": {
+                "action": "opened"
+              }
+            },
+            {
+              "operator": "or",
+              "operands": [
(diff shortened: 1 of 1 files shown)


nkolev92 (approved):
Do we have configuration from our test rules? 
the portal seems to have removed them, so it seems our work might be lost.

erdembayar:
It looks there is overlap between your PR and https://github.com/NuGet/NuGet.Client/pull/4694.
Just wondering is going to cause merge conflicts on `.github/fabricbot.json` file?

nkolev92:
Can we merge this?
I think this is ready correct?

nkolev92:
@erdembayar That's NuGet.Client, so they're different repos. 
We'll have different bots for different repos.

erdembayar (approved):

heng-liu:
Sorry for the late reply! Yes, it's ready to merge.

Issue #11361, opened 2021-11-06
Package Source Mapping UI - Add Package Source Mapping in the NuGet Options UI


nkolev92:
This issue captures *all* the additions for `Package Source Mapping` in the NuGet Options UI. 

Note that this does not include the package manager UI, that's a different task. 

This issue captures creating clarity and ensuring the spec is up to date with whatever gets implemented. 

Refer to the design here: https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md

As of right now, the design doc seems pretty complete. 

Some potential questions.

* How do you onboard to source mapping using the UX (ex. enabling the feature).

cc @JonDouglas 

Design primarily captured in https://github.com/NuGet/Home/issues/11377

- [x] https://github.com/nuget/home/issues/11362
- [x] https://github.com/nuget/home/issues/11363
- [x] https://github.com/nuget/home/issues/11364
- [x] https://github.com/NuGet/Client.Engineering/issues/1834
- [x] https://github.com/NuGet/Client.Engineering/issues/1691
- [x] https://github.com/NuGet/docs.microsoft.com-nuget/issues/2799
- [x] https://github.com/NuGet/Client.Engineering/issues/2034
- [x] https://github.com/NuGet/Home/issues/12271
- [ ] https://github.com/NuGet/Home/issues/12155
- [ ] https://github.com/NuGet/Home/issues/13355

nkolev92:
Closing this as complete.

PR #11922, opened 2022-06-27
Spec for PMUI package source mapping

### LINKED ISSUE #11378: Design Package Source mapping package installation experiences in VS PM UI
Child of https://github.com/NuGet/Home/issues/10730

mcnallyella:
Fixes: https://github.com/NuGet/Home/issues/11378


proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
+# PMUI 
+* Start Date: 6/25/22
+* Authors: Ella McNally
+* Issue
```
JonDouglas: Should be able to correlate to one of these issues! https://github.com/NuGet/Home/issues/11361

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Summary
+
+Currently there is no support for package source mapping in the PMUI. I will add a feature that allows the user to choose a source to map a package to when they install it and a feature that shows all package source mappings in the preview. This feature is a portion of what was discussed in a previous spec. https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md
```
JonDouglas: Great summary. Perhaps two paragraphs to separate each problem being solved here.

1. Mapping a source when installing a package.
2. Managing mapped sources in the options.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Motivation
+
+Adding support for package source mapping in PMUI will allow user to make/view package source mappings with more ease. It will also help the user be more aware of where their packages are coming from and what mappings they have already configured
```
JonDouglas, line 12: Great motivation!

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Explanation
+
+### Functional Explanation
+
+The user will be able to choose whether or not they want to make a mapping to the source in the package source combobox in the details pane. There will be a new checkbox below the package version dropdown which the user can check if they want to map the package they are installing to the source they've selected above. If the user has `All` sources selected, then the checkbox will be grayed out because it doesn't make sense <!--(reword)--> to let the user make mappings from a package to all configured sources.
```
JonDouglas: Maybe a screenshot showing the package source combobox to help the reader? 

Also a quick mockup of the new checkbox or affordance may help the reader visually see how to opt-in to the feature.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+ 
+The existing preview dialog that shows when the user hits `Install` or `Update` will be modified to list the mapping they just made under the version. The user already has the option to disable this preview dialog, so if they choose to disable it, there will be nothing telling them what mappings they made.
```
JonDouglas: Great attention to detail! Love the re-use of this.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+<!--add screenshot-->
+
+#### Error Messages
+
+There are a few errors the user could make while trying to install or update a package with a mapping that would make restore fail. I will address those errors with a popup telling the user what did not work. First, there could be issues with the transitive packages. Maybe there are transitive packages that are already mapped to a different source or the source that the user is trying to map a package to does not support some of the transitive packages. In either case, a popup would appear telling the user that the package source mapping could not be made. The popup would have a grid with a column showing the top level package and all of the transitive packages, a column showing if each package already has a mapping (if so what source it is mapped to previously), and whether the source supports the package. <!-- <- are all of these things even doable? -->
```
JonDouglas: Great description. A quick mockup of how this might look would speak loudly! Or even some markdown code block that gives the idea of the error message and empowering action the user can take.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+Two more scenarios could arise when a user tries to update a package if that package has an exisiting mapping. First, the mapping could be to a different source than the one chose from the package source dropdown when `Update` is clicked. In this case, the user would get a popup error message telling them that it cannot update because the package already has a mapping to a different source and ask if the user would like to update using the other source instead. The user could click `OK` to update using the other source or `Cancel`. The second scenario would be if the user tries to update a package that is already mapped to a different source (like the first scenario), but the source it is mapped to does not support the version they are trying to update to. In this case, the same popup error message would tell the user that the package is mapped to a different source, and that source does not support the version they are trying to install. The user would not be able to click `OK` in this case because restore would not be able to make this update.
```
JonDouglas: Good catch!

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,72 @@
...
+
+**Result:**
+
+In this example, the user had `nuget.org` selected from the source dropdown. They clicked on the `Serilog` package on the installed tab, chose version `11.0.0`, and selected the checkbox to make a mapping from `Serilog` to `nuget.org`.
+
+## Drawbacks
+
+## Rationale and Alternatives
+
+On the previous spec, there was a pin icon next to the sources dropdown to pin a source instead of a checkbox in the details pane. I thought the pin was confusing since there was nothing to explain what it did. Also the same pin icon is used to different actions already in VS like to pin a file at the top of the screen. It would be confusing to have the same button do different tasks in different parts of VS.
```
JonDouglas, line 66: The older spec was based on an older name of "package namespaces & source pinning". We can use whatever affordance makes sense to "map a source".

Here are many other icons that are fair game to use: https://d3assets.azurewebsites.net/icons/Vswin2022

Anything surrounding the metaphor for a "relationship", "association", or even "map" may be worth pursuing. A checkbox in the details pane may have some cons to discoverability and UX of mapping batch packages on install/update.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### Package Source Mappings Page 
+
+When the user opens the package source mapping page in the options dialog, they will see a list of all the package source mappings they have already configured and the filepath to the solution level config file. The mappings will be able to be sorted by package pattern or by source. There is a hierarchy of NuGet configs levels: machine, user, and solution. Package source mappings will be read from all of the levels, but only written to the solution level config. This page will have three buttons: _Add_ , _Remove_, and _Clear_.
+
+* Add: Brings user to a pop-up where they can add package source mappings to the config.
+
+* Remove: Lets user remove selected package source mapping from the config.
+
+* Clear: Lets user clear all package source mappings from the current config.
```
JonDouglas: Do we have any worry that a user will hit this button out of curiosity and be burdened to lost work? Should we rather allow a mass selection and keep the remove metaphor instead? On re-thinking this detail, I think the design was mostly parity of similar screens. Do we want that here though?

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+* Close: Will close pop-up without adding any package source mappings.
+
+Below are the mockups for the package source mappings page. The top left is the package source mappings page, and the bottom left is the popup that appears when the user clicks the add button.
+
+![Options 1](../../meta/resources/PackageSourceMapping/VSOptions.png)
+
+#### Package Sources Page
+
+I will add a feature that allows a user to pin default sources for all package source mappings that are not specified on the package source mapping page. There will be a pin button, that when clicked, creates a default mapping for all package patterns to the pinned source. A pin icon will appear next to all pinned sources. There will be a tooltip for the pin button at the top of the page. This tooltip will explain that pinning a source makes it the default source for package patterns for which the user does not make specific mappings on the package source mappings page.
```
JonDouglas: To stay consistent with the other design here, if we're not going to use the pin icon, let's try to find an affordance that will be consistent with the checkboxes or other chosen icon.

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+    <packagesource key="nuget.org">
+        <package pattern="Microsoft.Extensions.Options" />
+        <package pattern="Newtonsoft.Json" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+If the user clicks the _Clear_ button on the package source mappings options page, then all package sources will be removed from the configuration, and package source mapping will be disabled. 
+
+If the user removes a source from the package sources page, any mappings to that source will not be deleted in the config. The user will get a message that they are removing a source that still has mappings to it. If a user adds a source that has already has mappings to it in the config, they will get a message that this source has pre-existing mappings in the config that will be applied once the source is added.
```
JonDouglas: Love this detail!

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+<packageSourceMapping>
+    <packagesource key="contoso">
+        <package pattern="*" />
+    </packageSource>
+</packageSourceMapping>
+```
+
+**Result:**
+
+In this example, the user pinned the source `contoso` and did not specify any specific package source mappings. All packages will come from the source `contoso` due to matching the `*` pattern and since no other sources are pinned. There is a source mapping precedence, so if there is more than one source pinned, the one with the longest pattern matched will have precedence. Alternatively, the user could have added `*` as the package pattern and selected `contoso` as the source on the package source mappings page. These two options would be written to the solution level config and read by the package sources page in the same way.
```
JonDouglas: love it

proposed/2022/Package-Source-Mapping-VS-Options.md
```diff
@@ -0,0 +1,145 @@
...
+
+### Accessibility 
+
+To ensure screen-reader I will run Accessibility Insights to find any accessibility faults.
+
+
+## Drawbacks 
+
+Currently, mappings to sources are not deleted in the config file when the source they are mapped to is deleted. It is complicated to remove this code because there are multiple different config files this mapping could be in.
```
JonDouglas: It's also something we shouldn't worry about too much. Sometimes sources can get lost with bugs, bad check-ins, etc. We should preserve their source of truth. This would require users to refactor their code to remove old mappings. Our tooling I don't think can be smart enough to do this on behalf of people as it may lead to undesired behavior.

JonDouglas (approved):
Great specs! Added some comments to consider.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,89 @@
+# PMUI
```
dominoFire:
```suggestion
# Package Source Mapping in PM UI
```

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,89 @@
...
+|Not Checked | Package source mapping is not enabled. Restore works as normal| No new mappings are made. Package only has previous mappings| |
+
+#### Error Messages
+
+There are a few errors the user could make while trying to install or update a package with a mapping that would make restore fail. I will address those errors with a popup telling the user what did not work. First, there could be issues with the transitive packages. Maybe there are transitive packages that are already mapped to a different source or the source that the user is trying to map a package to does not support some of the transitive packages. In either case, a popup would appear telling the user that the package source mapping could not be made. The popup would have a grid with a column showing the top level package and all of the transitive packages, a column showing if each package already has a mapping (if so what source it is mapped to previously), and whether the source supports the package. This error message will look like:
```
dominoFire: Instead of creating a new pop-up window for source mapping errors, can we show an error entry in VS Error Window?

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Motivation
+
+Adding support for package source mapping in PMUI will allow user to make/view package source mappings with more ease. It will also help the user be more aware of where their packages are coming from and what mappings they have already configured
+
+## Explanation
+
+### Functional Explanation
+
+The user will be able to choose whether or not they want to make a mapping to the source in the package source combobox in the details pane. There will be a new checkbox below the package version dropdown which the user can check if they want to map the package they are installing to the source they've selected above. The checkbox will be grayed out and will not allow the user to make a mapping if `All` sources are selected from the dropdown. It would not make sense for a user to make a mapping from one package to all sources. The added checkbox will make the details pane look like:
```
dominoFire: Are we going to add a 'pin' icon to mapped, installed packages on "Installed" tab?

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,89 @@
+# PMUI 
+* Start Date: 6/25/22
+* Authors: Ella McNally
+* Issue: https://github.com/NuGet/Home/issues/11797
+
+## Summary
+
+Currently there is no support for package source mapping in the PMUI. I will add a feature that allows the user to choose if they want to map the package to the source chosen when they install. I will also add error messages that tells the user if restore has problems with the mapping. This feature is a portion of what was discussed in a previous spec. https://github.com/NuGet/Home/blob/dev/implemented/2021/PackageSourceMapping.md
```
dominoFire: s this spec only for "Browse" tab?

proposed/2022/PMUIPackageSourceMapping.md
mcnallyella: I think right now we are not going to add a pin icon at all. It seems confusing to use it in one place but not another since we already ruled it out for the options UI.

proposed/2022/PMUIPackageSourceMapping.md
mcnallyella: It is for the browse and update tabs

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,80 @@
...
+|---|---|---|---|
+|`Add additional mapping to source` | Package source mapping is enabled. New mapping is created. Package is now mapped to only the new source| New mapping is created. Package is now mapped to the new source in addition to previous mappings | New mapping is created. Package is now mapped to only the new source.|
+|`Use only existing mappings` | Not possible since this button will be greyed out (since there are no existing mappings)| No new mappings are made. Package only has previous mappings| Not possible since this button will be greyed out (since there are no existing mappings). User must select `Add additional mapping to source` to install. |
+|Neither is checked| Package Source Mapping is not enabled. Restore works as normal | Not possible since `Use only existing mappings` will be chosen by default once package source mapping is enabled| Not possible since `Use only existing mappings` will be chosen by default once package source mapping is enabled (unless there are no existing mappings in which case the user must select the other option to `Install`)|
```
nkolev92, line 32: We should probably just hide it if the package source mapping is not enabled.

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,80 @@
...
+
+![PMUI 1](../../meta/resources/PackageSourceMapping/PMUI_details_pane.png)
```
nkolev92, line 20: What happens when the source selected is all?

What does `add mapping` do?
Which source would it add mapping for?

proposed/2022/PMUIPackageSourceMapping.md
```diff
@@ -0,0 +1,80 @@
...
... if they want to add a mapping to the package from the selected source when they `Install`/`Update` or if they only want to use existing mappings. These `RadioButtons` will always be shown even if package source mapping is not enabled, but neither button will have to be selected to install. Package source mapping can be enabled by the user choosing the `RadioButton` that adds a new mapping. Once package source mapping is enabled, the user must select one of the two `RadioButtons` to install. The `Use only existing mappings` option will be selected by default once package source mapping is enabled. The existing mappings will not be shown, but there will be a link to the package source mapping options page so the user can view the mappings. If package source mapping is enabled, but there are no existing mappings, then the `Use only existing mappings` button will be disabled. A label will also appear under the `Use only existing mappings` button if it is disabled that says there are no existing mappings. The link will be a settings icon next to the `RadioButton` (similar to the settings button next to the sources dropdown in the top right of the PMUI). The details pane will look like:
```
dominoFire, line 18: Will the source mapping ratio button group be visible for packages.config projects?

What happens in Project PM UI ?

I am assuming this feature will only work with PackageReference projects in Promect PM UI.

dominoFire (approved):
Looks good. Only one question that can be answered later.

kartheekp-ms:
Team Triage: Assigning this PR @donnie-msft to drive this work to completion.

donnie-msft:
Merging and will start a clean PR modify it with recent decisions: https://github.com/NuGet/Home/issues/11378

PR #11961, opened 2022-07-11
Add stale waiting for customer bots

nkolev92:
Fixes https://github.com/NuGet/Client.Engineering/issues/1038

1.	Client people manually add "WaitingForCustomer" label if needs input from customer.
2.	When customer replies/ or if any response if it’s an issue transferred from other repo/VSFeedback, remove the "WaitingForCustomer"  label and add "WaitingForClientTeam"  label.
3.	When no response after an issue adding a "WaitingForCustomer" label for 14 days, add a "Status: No recent activity"  label, add a comment, e.g.
4.	Close the issue if there is no activity for 14 days after .
Remove any "WaitingForCustomer"  or  "WaitingForClientTeam" label when the issue is closed and add `Resolution:Need more info`

This was tested in 
 https://github.com/NuGet/bot-testing/issues/13 and https://github.com/NuGet/bot-testing/issues/12
 
 Currently on 5 issues have the potential to be reported by these bots: 
 
 https://github.com/NuGet/Home/issues?q=is%3Aopen+is%3Aissue+label%3AWaitingForCustomer

I will monitor these individual issues. 
4 of these issues sohuld get the `Status:No recent activity` label immediately.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -130,7 +130,7 @@
         "eventNames": [
           "issue_comment"
         ],
-        "taskName": "[Manage \"WaitingFor\" labels] [4-2] Replace tag \"WaitingForCustomer\" with \"WaitingForClientTeam\" when the author comments on an issue. Also remove `Status: No Recent Activity` if it's been set.",
+        "taskName": "[Manage \"WaitingFor\" labels] [4-2] Replace tag \"WaitingForCustomer\" with \"WaitingForClientTeam\" when the author comments on an issue. Also remove `Status:No recent activity` if it's been set.",
         "actions": [
           {
             "name": "removeLabel",
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -821,7 +821,7 @@
             }
           }
         ],
-        "taskName": "[Managed stale WaitingForCustomer] Search for WaitingForCustomer issues with no activity over 14 days and warn.",
+        "taskName": "[Manage stale WaitingForCustomer issues] Search for WaitingForCustomer issues with no activity over 14 days and warn.",
         "actions": [
           {
             "name": "addLabel",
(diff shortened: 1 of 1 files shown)


jeffkl (approved):
Does the bot need to run every 3 hours?  Why not just once a day?

nkolev92:
> Does the bot need to run every 3 hours? Why not just once a day?

I think it can. 
There are some other bots that work in lock step with this one (such as the bots that act on issue actions to add/remove labels), so I tried to keep a similar cadence with those, but looking into it now, I don't think it's necessary.

nkolev92:
Fixed. 
The bot will now run once a day at 6am, PST.

.github/fabricbot.json
```diff
@@ -686,6 +686,204 @@
...
+        ],
+        "searchTerms": [
+          {
+            "name": "hasLabel",
+            "parameters": {
+              "label": "WaitingForCustomer"
+            }
+          },
+          {
+            "noActivitySince": "14",
```
heng-liu: Since there is no documentation introducing syntax for fabricbot.json file yet, I tried to export from bot-testing repo and the exported json file has:
`"name": "noActivitySince",`
May I know if you tried to save it in config portal in home repo and compare it with exported json file? (The save button is reenabled)

heng-liu (commented):
LGTM. Just have one question.

heng-liu:
> Does the bot need to run every 3 hours? Why not just once a day?

I learnt from FabricBot team that, if there are many records in search results, it might be good to set several times a day, as each search will only process a few hundreds of records in the searching results. I'm not sure if that's still the case as there is no documentation explaining this.
Since there should be a few records in the search results in this case, once a day should be good.

zivkan:
Do any of our bots have queries that return hundreds of results?

Also, since GitHub's search results are deterministic, unless the bot modifies every issue it sees in search results so that it will not be returned the next time the bot runs, I don't see how running the bot multiple times a day will help. It will just re-evaluate the same set of issues, and ignore the same set as last run.

I don't feel like there's benefit for any of our bots to run multiple times a day, but I haven't been keeping a close eye on what bots we actually use.

.github/fabricbot.json
nkolev92: That's a mistake. Not sure how it happened.

I did not use the portal at all

nkolev92: I exported the config in bot-testing.
Comparing it with my changes now.

nkolev92: Looks correct now. PTAL.

erdembayar (approved):

heng-liu (approved):

heng-liu:
> Do any of our bots have queries that return hundreds of results?
> 
> Also, since GitHub's search results are deterministic, unless the bot modifies every issue it sees in search results so that it will not be returned the next time the bot runs, I don't see how running the bot multiple times a day will help. It will just re-evaluate the same set of issues, and ignore the same set as last run.
> 
> I don't feel like there's benefit for any of our bots to run multiple times a day, but I haven't been keeping a close eye on what bots we actually use.

I don't think it will just re-evaluate the same set of issues. In this PR, we add label "Status:No recent activity" to the issues for the first search, then for the second search, since we set the condition: with no label "Status:No recent activity", the first search results will be excluded for the second search. So it's like doing a search incrementally. 
One benefit of multiple search I could think of would be, it could apply changes more frequently. E.g. when we do a search and need to leave a comment to the author of a newly opened issue, if we do searching multiple times a day, the author will probably  see the comment on the same day they create the issue.

nkolev92:
I think the search frequency can vary from bot to bot. 
Given that this is a stale bot, not searching frequently is ok here, as we only do stuff if we have inactivity.

PR #11983, opened 2022-07-26
Add errors handling scenario for json output spec

### LINKED ISSUE #11449: Create spec for machine readable json output for "dotnet list package"
Sub issue of https://github.com/NuGet/Home/issues/7752, `json` output spec only.

Many organizations are required by regulation to audit packages that they're using in a repository.

Currently there's no easy way to produce a Software Bill of Material (SBOM) output which can be consumed by another auditing system or kept for records.
To do:
* Rename spec md file to `DotnetListPackageMachineReadableJsonOutput.md`

erdembayar:
Fixes: https://github.com/NuGet/Home/issues/11449
With my previous https://github.com/NuGet/Home/pull/11446 spec proposal, I didn't take into account if there is an error, it only covers happy paths only.
Here I'm adding section for unhappy paths by creating `problems` section, probably it can be named to `errors` instead. 

[Rendered](https://github.com/NuGet/Home/blob/04b3ebec7868c09054f800af8996148e4e3b15c8/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md)

### DIFF
--- a/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
+++ b/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
@@ -1062,8 +1062,9 @@ No assets file was found for `C:\Users\userA\repos\MainApp\src\lib\MyProjectB.cs
   "parameters": "",
   "problems": [
     {
+        "level" : "error",
         "project": "src/lib/MyProjectB.csproj",
-        "error": "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
+        "text": "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
     }
   ],
   "projects": [
@@ -1112,8 +1113,9 @@ Could not find file or directory 'C:\Users\userA\repos\MainApp\src\lib\NonExisti
   "parameters": "",
   "problems": [
     {
+        "level" : "error",
         "project": "src/lib/NonExisting.csproj",
-        "error": "Could not find file or directory 'C:/Users/userA/repos/MainApp/src/lib/NonExisting.csproj'"
+        "text": "Could not find file or directory 'C:/Users/userA/repos/MainApp/src/lib/NonExisting.csproj'"
     }
   ]
 }
@@ -1142,8 +1144,9 @@ The project `C:\Users\userA\repos\MainApp\src\lib\MyProjectB.csproj` uses packag
   "parameters": "",
   "problems": [
     {
+        "level" : "error",
         "project": "src/lib/MyProjectB.csproj",
-        "error": "The project `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`` uses package.config for NuGet packages, while the command works only with package reference projects."
+        "text": "The project `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`` uses package.config for NuGet packages, while the command works only with package reference projects."
     }
   ],
   "projects": [
@@ -1201,7 +1204,8 @@ Json output for above case.
   "parameters": "--highest-minor",
   "problems": [
     {
-        "warning": "The command option(s) '--include-prerelease', '--highest-minor', and '--highest-patch' are ignored by this command."
+        "level" : "warning",
+        "text": "The command option(s) '--include-prerelease', '--highest-minor', and '--highest-patch' are ignored by this command."
     }
   ],
   "projects": [
@@ -1230,7 +1234,8 @@ Warning for `Http source used` is included in json.
   "parameters": "--vulnerable",
   "problems": [
     {
-    	"warning": "You are running the 'list package' operation with an 'HTTP' source, 'nugettest [http://apidev.nugettest.org/v3-index/index.json]'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source."
+      "level" : "warning",
+      "text": "You are running the 'list package' operation with an 'HTTP' source, 'nugettest [http://apidev.nugettest.org/v3-index/index.json]'. Non-HTTPS access will be removed in a future version. Consider migrating to an 'HTTPS' source."
     }
   ],
   "projects": [
--- a/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
+++ b/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
@@ -473,6 +473,118 @@ Project `MyProjectB` has the following vulnerable packages
 }
 ```
 
+#### `> dotnet list package --vulnerable`
+There's no vulnerable package in a project.
+
+```dotnetcli
+The following sources were used:
+   https://api.nuget.org/v3/index.json
+   https://apidev.nugettest.org/v3-index/index.json
+
+The given project `MyProjectC` has no vulnerable packages given the current sources.
+```
+
+### `> dotnet list package --vulnerable --format json`
+There's no vulnerable package in a project.
+
+```json
+{
+  "version": 1,
+  "parameters": "--vulnerable",
+   "sources": [
+    "https://api.nuget.org/v3/index.json",
+    "https://apidev.nugettest.org/v3-index/index.json"
+  ],
+  "projects": [
+    {
+      "path": "src/lib/MyProjectC.csproj",
+      "frameworks": [
+        {
+          "framework": "netcoreapp3.1",
+          "topLevelPackages": [
+            ]
+        },
+        {
+          "framework": "net5.0",
+          "topLevelPackages": [
+            ]
+        }
+      ]
+    }
+  ]
+}
+```
+
+#### `> dotnet list package --vulnerable`
+For multi-target framework a project only 1 framework has a vulnerability.
+
+```dotnetcli
+The following sources were used:
+   https://api.nuget.org/v3/index.json
+   https://apidev.nugettest.org/v3-index/index.json
+
+Project `MyProjectC` has the following vulnerable packages
+   [netcoreapp3.1]:
+   Top-level Package      Requested   Resolved   Severity   Advisory URL
+   > DotNetNuke.Core      6.0.0       6.0.0      High       https://github.com/advisories/GHSA-g8j6-m4p7-5rfq
+                                                 Moderate   https://github.com/advisories/GHSA-v76m-f5cx-8rg4
+                                                 Critical   https://github.com/advisories/GHSA-x8f7-h444-97w4
+                                                 Moderate   https://github.com/advisories/GHSA-5c66-x4wm-rjfx
+                                                 High       https://github.com/advisories/GHSA-x2rg-fmcv-crq5
+                                                 High       https://github.com/advisories/GHSA-j3g9-6fx5-gjv7
+                                                 High       https://github.com/advisories/GHSA-xx3h-j3cx-8qfj
+                                                 Moderate   https://github.com/advisories/GHSA-5whq-j5qg-wjvp
+
+   [net5.0]: No vulnerable packages for this framework.
+```
+
+### `> dotnet list package --vulnerable --format json`
+For multi-target framework a project only 1 framework has a vulnerability.
+
+```json
+{
+  "version": 1,
+  "parameters": "--vulnerable",
+   "sources": [
+    "https://api.nuget.org/v3/index.json",
+    "https://apidev.nugettest.org/v3-index/index.json"
+  ],
+  "projects": [
+    {
+      "path": "src/lib/MyProjectC.csproj",
+      "frameworks": [
+        {
+          "framework": "netcoreapp3.1",
+          "topLevelPackages": [
+              {
+                "id": "DotNetNuke.Core",
+                "requestedVersion": "6.0.0",
+                "resolvedVersion": "6.0.0",
+                "vulnerabilities" : [
+                  {
+                      "severity":"High",
+                      "advisoryurl":"https://github.com/advisories/GHSA-g8j6-m4p7-5rfq"
+                  },
+                  {
+                      "severity":"Moderate",
+                      "advisoryurl":"https://github.com/advisories/GHSA-v76m-f5cx-8rg4"
+                  },
+      ...
+                  ]
+              }
+            ]
+        },
+        {
+          "framework": "net5.0",
+          "topLevelPackages": [
+           ]
+        }
+      ]
+    }
+  ]
+}
+```
+
 #### `> dotnet list package --include-transitive`
 
 ```dotnetcli
--- a/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
+++ b/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
@@ -1052,7 +1052,7 @@ No assets file was found for `C:\Users\userA\repos\MainApp\src\lib\MyProjectB.cs
   "problems": [
     {
         "project": "src/lib/MyProjectB.csproj",
-        "message": "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
+        "error": "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
     }
   ],
   "projects": [
@@ -1102,7 +1102,7 @@ Could not find file or directory 'C:\Users\userA\repos\MainApp\src\lib\NonExisti
   "problems": [
     {
         "project": "src/lib/NonExisting.csproj",
-        "message": "Could not find file or directory 'C:/Users/userA/repos/MainApp/src/lib/NonExisting.csproj'"
+        "error": "Could not find file or directory 'C:/Users/userA/repos/MainApp/src/lib/NonExisting.csproj'"
     }
   ]
 }
@@ -1132,7 +1132,7 @@ The project `C:\Users\userA\repos\MainApp\src\lib\MyProjectB.csproj` uses packag
   "problems": [
     {
         "project": "src/lib/MyProjectB.csproj",
-        "message": "The project `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`` uses package.config for NuGet packages, while the command works only with package reference projects."
+        "error": "The project `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`` uses package.config for NuGet packages, while the command works only with package reference projects."
     }
   ],
   "projects": [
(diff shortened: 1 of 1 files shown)


proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +773,138 @@ Outputs json for format version 1, if it's not specified then latest version'll
 }
 ```
 
+### Error handling
+
+In case of an error it would be written into `problems` section.
```
erdembayar, line 778: `npm -ls json` does same thing it creates `problems section` at top of json output.

erdembayar:
@JonDouglas  @baronfel
Please review.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,139 @@ Outputs json for format version 1, if it's not specified then latest version'll
 }
 ```
 
+### Error handling
+
+In case of an error it would be written into `problems` section. But in case there is parameter, options error then it may defaults back to console output instead of json output.
```
erdembayar: `npm -ls json` does same thing it creates `problems section` at top of json output. Only problem with this approach is json output needs to be completed internally before it displays on console.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,139 @@ Outputs json for format version 1, if it's not specified then latest version'll
...
+
+#### `> dotnet list package --format json`
+
+```json
+{
+  "version": 1,
+  "parameters": "",
+  "problems": [
+    "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
+  ],
```
zivkan: should it be an array of objects, rather than array of strings?

```json
[
  {
    "project": "C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj",
    "message": "No assets file was found for `C:/Users/userA/repos/MainApp/src/lib/MyProjectB.csproj`. Please run restore before running this command."
  },
  {
    "project": "C:/whatever/my.csproj",
    "source":  "https://private/index.json",
    "message": "HTTP timeout after 100 seconds"
 }
]
```

JonDouglas:
My general feedback:

I don't quite understand why we would have a `problems` section here in JSON. Why wouldn't the command just fail with the problem/error message and until that is resolved, the command will not work?

See https://clig.dev/#errors

nkolev92:
I agree with @JonDouglas, maybe we just fail the operation and error. 

Customers have asked for exit codes to represent the actual status of the operation, so writing an output when the op fails feels like too much.

zivkan:
With web APIs, it's very common to return errors in data structures. Given this is about `--format json`, also known as "machine readable", I think it makes sense to make errors machine readable/json as well.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
baronfel: Agree strongly with this - if the different _kinds_ of errors are known or can be anticipated, consider also having a `code` field for these problem structures - a code is often something that can be searched on, and often can be put into urls for easy user lookup: `aka.ms/nuget/listpackage#errors_<code>` for example.  This isn't a 100% match, but the HTTP Problem Details RFC ([RFC 7807](https://www.rfc-editor.org/rfc/rfc7807#section-3.1)) is a good example of this (where the `type` field maps to the `code` field I describe).

baronfel:
Echoing Jon/Nikolche/Andy - in the case of an error when structured output is requested, the structure returned should be the 'array of errors' structure (or a structure that is a wrapper around the array of errors), and the exit code should represent that there are multiple errors. You could also special case that if there was only one error the exit code is the code for that single error, but that seems unlikely to me :)

I agree that it doesn't make sense to have the problems in the 'success' response schema.

nkolev92:
Weirdly enough, `dotnet list package` doesn't really set a proper exit code when it fails. 

Should that be changed? I don't know if the exit code was an intentional decision.

baronfel:
@nkolev92  yes please! There are a number of issues in the SDK repo (or were before they were moved) about missing exit codes for package-related operations. These are _vital_ for scripting use cases.

erdembayar:
> @nkolev92 yes please! There are a number of issues in the SDK repo (or were before they were moved) about missing exit codes for package-related operations. These are _vital_ for scripting use cases.

This feedback echo https://github.com/dotnet/sdk/issues/16852, but that one is for if there is any vulnerable/outdated dependencies.
@baronfel We closed that one because we don't want to break existing customers if we change exit code. 

But for current case it seems reasonable to change exit code.
Could I return `1` as exit code if couldn't generate json output or if there is any error?

baronfel:
@erdembayar there are some semi-standards around choosing exit codes (https://tldp.org/LDP/abs/html/exitcodes.html for example). The templating team recently revamped their exit codes (PR [here](https://github.com/dotnet/templating/pull/4787)) and one thing they did was make sure there were aka.ms urls set up for each exit code. E.G for the CLI output:

> For details on the exit code, refer to https://aka.ms/templating-exit-codes#106

To me, it's most important that you choose the exit codes and provide easy ways to look up docs on each error code.

erdembayar:
> My general feedback:
> 
> I don't quite understand why we would have a `problems` section here in JSON. Why wouldn't the command just fail with the problem/error message and until that is resolved, the command will not work?
> 
> See https://clig.dev/#errors

Currently with cli console if there's error at least it displays partial result(see below), if we just fail without json then customer wouldn't see partial result at all. Also, intention of json result is for historical backup(SBOM) having at least partial data is better than nothing in case of investigating issue from the past. So, I tilt to return json array for problems. 
Same time we return non-zero exit code to signal there is an issue.

**MyProjectB.csproj was not restored.**

```
Project 'MyProjectA' has the following package references
   [netcoreapp3.1]:
   Top-level Package                      Requested             Resolved
   > Microsoft.Extensions.Primitives      [1.0.0, 5.0.0]        1.0.0
   > NuGet.Commands                       4.8.0-preview3.5278   4.8.0-preview3.5278
   > Text2Xml.Lib                         [1.1.2, 2.0.0)        1.1.2

No assets file was found for `C:\Users\userA\repos\MainApp\src\lib\MyProjectB.csproj`. Please run restore before running this command.
```

nkolev92:
@erdembayar When that results, what's the error code? 
I did some testing and it was 0.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,139 @@ Outputs json for format version 1, if it's not specified then latest version'll
 }
 ```
 
+### Error handling
+
+In case of an error it would be written into `problems` section. But in case there is parameter, options error then it may defaults back to console output instead of json output.
+Also in case of runtime there's error then it may not show json output at all because it needs all the data for json output before display.
+
```
dominoFire, line 1041: Two questions:

- Is JSON output printed in STDOUT or STDERR?
- Can we add a non-zero process exit code when an error occurs?

zivkan:
> To me, it's most important that you choose the exit codes and provide easy ways to look up docs on each error code.

I think that makes a lot of sense when the stdout and/or stderr output is not "machine parsable", but when the output is json, I don't think it matters.

For example, say two projects in the solution failed, for different reasons. What error code would you return? Returning a json array or error objects allows the command to give lots of context.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: >> Is JSON output printed in STDOUT or STDERR?

See Andy's [response](https://github.com/NuGet/Home/pull/11983#issuecomment-1196228788).

>> Can we add a non-zero process exit code when an error occurs?

Please check latest changes. Yes, we'll return non-zero exit code. For current implementation it would be general error code `1`.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: @zivkan @baronfel 
I added array of objects instead of array of strings into spec. Please review latest spec.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,173 @@ Outputs json for format version 1, if it's not specified then latest version'll
...
+Since output version `3` is not available, then it'll default to latest available version, and log about unsupported format version request into json result.
+
+```json
+{
+  "version": 1,
+  "parameters": "",
+  "problems": [
+    {
+        "project": "",
```
erdembayar: For general error, should I skip `project` property or not? Maybe put place holder `*` for value?

nkolev92: I think not including the project property when empty is ok.

erdembayar: Ok. Addressed, not including empty property.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,172 @@ Outputs json for format version 1, if it's not specified then latest version'll
 }
 ```
 
+### Error handling
+
+In case of an error it would be written into `problems` section and return non-zero error exit code to indicate there had been some error to help scripting use cases.
+For current implementation we'll returns general error code `1` from application, more specific error code for each error scenario is not considered in this spec scope.
```
nkolev92: A way to word this in a more generic fashion would be to say, `non-0 exit code` instead.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,172 @@ Outputs json for format version 1, if it's not specified then latest version'll
...
+In case there is parameter, options error then it may defaults back to console output instead of json output.
```
nkolev92: Can we do a list of things that we know would error out?

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: Sure, resolved.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: I'm still discovering those, I feel it's more of implementation details, before listing them out we need to agree on whether we error out on console and exit or include errors in `problems` section of json?

nkolev92: It was a formatting suggestion, not a content one.

You have sentences enumerating those failure cases. Making them a list would be more readable

nkolev92: Rereading it, I'm not sure I fully understand what these sentences are trying to say. 


When there's an error, the error details will be written into a `problems` section and a non-0 error code will be returned. 

In case there bad parameter input, such as options error the json might not be returned?
Do we have an error and no json output generated?

erdembayar: For example, `dotnet list package -include-transitive --format json`, here `-include-transitive` missing additional `-` in front. But parameter validation happens way before json rendering, so we couldn't return json result even though customer requested one.

```
Specify --help for a list of available options and commands.
error: Unrecognized option '-include-transitive'


Usage: NuGet.CommandLine.XPlat.dll package list [arguments] [options]

Arguments:
  <PROJECT | SOLUTION>  A path to a project, solution file or directory.
```

nkolev92: I see. Ok that makes sense.

What about dotnet list package --framework where the framework doesn't exist?

erdembayar: > I see. Ok that makes sense.
> 
> What about dotnet list package --framework where the framework doesn't exist?

It wouldn't fail before generating json, it'll return empty json. I manually tested with `net7.0` framework option.

nkolev92: ok. I see. 

I think those are the types of scenarios that'd be great to enumerate. 

It doesn't have to be a complete list, but largely what we discussed, something to give folks a good sense about what happens.

KalleOlaviNiemitalo:
I'd somewhat prefer a JSON object to stdout, plus MSBuild-compatible error and warning messages to stderr. So that if I run `dotnet list package --format json` during a build on Jenkins and capture the JSON from stdout, I can let stderr flow to the build log, from which a plugin of Jenkins is already able to parse the diagnostics and display them nicely. It's okay if this requires some opt-in flag (`--stderr-diagnostics`) or if you write the same diagnostics to the `"problems"` array in the JSON object as well.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: Ok, I added above known scenarios. We can add more later even this PR is merged as we discover.

erdembayar:
> I'd somewhat prefer a JSON object to stdout, plus MSBuild-compatible error and warning messages to stderr. So that if I run `dotnet list package --format json` during a build on Jenkins and capture the JSON from stdout, I can let stderr flow to the build log, from which a plugin of Jenkins is already able to parse the diagnostics and display them nicely. It's okay if this requires some opt-in flag (`--stderr-diagnostics`) or if you write the same diagnostics to the `"problems"` array in the JSON object as well.

It's interesting idea, but not planned for scope of this spec.

KalleOlaviNiemitalo:
For this spec, please clarify whether the JSON goes to stdout or stderr or explicitly to console (e.g. /dev/tty).

erdembayar:
> For this spec, please clarify whether the JSON goes to stdout or stderr or explicitly to console (e.g. /dev/tty).

Please see [comment](https://github.com/NuGet/Home/pull/11983#issuecomment-1196228788) by Andy, it'll just write json text to console.

`I think that makes a lot of sense when the stdout and/or stderr output is not "machine parsable", but when the output is json, I don't think it matters.`

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +772,173 @@ Outputs json for format version 1, if it's not specified then latest version'll
 }
 ```
 
+### Error handling
+
+In case of an error it would be written into `problems` section and return non-0 error exit code to indicate there had been some error to help scripting use cases.
+
+In case of runtime there's error then it may not show json output at all because it needs all the data for json output before display.
+
+In case there is parameter, options error then it may defaults back to console output instead of json output, for example: `dotnet list package -include-transitive --format json`, because that option/parameter validation for `-include-transitive` happen way before `list package` code execution. But `dotnet list package -framework net7.0` wouldn't fail even though `net7.0` is not valid framework at this moment, it would simply return empty result.
```
nkolev92: I feel like my original formatting suggestion hasn't been addressed, but not sure if that was on purpose or by accident. 

Enumerating things in separate lists, or having separate paragraphs would make it easy for people to parse what's going on. 

The content is correct, but it's super difficult to understand what it's actually saying. 

I'll approve the PR, but hoping this can be addressed before merging.

nkolev92 (approved):

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: Just having 2nd thought about printing json output on console. Since it's machine-readable output instead of trying to write json to console, just writing to json file would have been cleaner and display all the error and message as console window does as now. Then it would be more consistent, expected experience for end users for both console and json output scenario.

ToneyAlexander:
Are there any updates on this or https://github.com/NuGet/NuGet.Client/pull/4743?

erdembayar:
> Are there any updates on this or [NuGet/NuGet.Client#4743](https://github.com/NuGet/NuGet.Client/pull/4743)?

Sorry, I was out for a vacation, I'm pick up where I left and going to continue.

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
erdembayar: @nkolev92 
I added summary table for `Error/Warning` handling section, please check.

nkolev92 (approved):

KalleOlaviNiemitalo:
> it'll just write json text to console.

That is a bad decision. The JSON output should go to stdout so that it can be redirected to a file or pipe without having to call [CreatePseudoConsole](https://learn.microsoft.com/windows/console/createpseudoconsole) or [posix_openpt](https://www.man7.org/linux/man-pages/man3/posix_openpt.3.html), which are cumbersome in build scripts.

ffontecha080 (approved):

proposed/2022/DotnetListPackageMachineReadableJsonOutput.md
```diff
@@ -772,6 +1020,240 @@ Outputs json for format version 1, if it's not specified then latest version'll
...
+| 1 | Passing unsupported targetframe | `dotnet list package -framework net9.0` wouldn't fail even though `net9.0` is not valid framework at this moment, it would simply return empty result.  | 0
+| 1 | Asset file missing | Error would be in problems section of json, see below. | 1
+| 1 | Non-existing csproj file path | Error would be in problems section of json, see below. | 1
+| 1 | Package.config project | Error would be in problems section of json, see below. | 1
+| 1 | Incompatible option combination | Warning would be in problems section of json, see below. | 0
+| 1 | Use of not secure http source | Warning would be in problems section of json, see below. | 0
+| 1 | Unsupported output format | Defaults back to console output, for example: `dotnet list package --format yaml` | 1
+| 1 | Unsupported output version | if it's for json output then no json output, error out unsupported output version , see below | 1
+| 1 | Unsupported output version | if it's for console output then it's ignored, versioning is not considered for console output | 0
```
erdembayar: @NuGet/nuget-client 
Should we warn if customer passes unsupported output version for console output? Currently console output can just ignore any value in output version option.

nkolev92: imo we should error for unsupported versions. 

The customers wants something very specific and they can't get it with that version of tooling so they should know that.

erdembayar: Sure, we need to error out, updated the spec doc, I'll make change for `console output error out` implementation with https://github.com/NuGet/Home/issues/12167

erdembayar:
Spec is implemented here: https://github.com/NuGet/NuGet.Client/pull/4855

Issue #7778, opened 2019-02-10
Show error when using nuget.exe to pack SDK csproj


loic-sharma:
According to [this](https://github.com/NuGet/Home/issues/7453#issuecomment-435715917) and [this](https://github.com/NuGet/Home/issues/4491) discussions, NuGet.exe is not supported for packing SDK projects. The current experience is confusing as it seems like NuGet.exe pack "works" but ignores properties set in the project. Instead, NuGet.exe should show a helpful error message that guides users to use dotnet.exe instead.

*Design* - https://github.com/NuGet/Home/pull/12088

IanKemp:
This is the only logical solution, since the NuGet team apparently isn't interested in adding critical functionality to their app - merely making it look like it works when it actually doesn't, then ignoring the fact that it doesn't actually work for the next two-and-a-half years.

loic-sharma:
@rrelyea @anangaur Thoughts? This is a "breaking" change but would greatly help newcomers. For example: https://github.com/NuGet/docs.microsoft.com-nuget/pull/1700

stefanluchian:
The problem with "nuget pack" is known since 2017.
One could do the trick with "msbuild -t:pack".
One could use "dotnet pack".
But the others that rely on "nuget pack" loose many hours till they figure out what is going on.

@Microsoft: Will you please do something about it?

aortiz-msft:
@zkat - https://github.com/NuGet/Home/issues/4491, which points to this issue, has 47 upvotes. Might want to consider for Sprint 179.

nkolev92:
Proposal for this change is out in https://github.com/NuGet/Home/pull/12088.

Issue #7453, opened 2018-10-30
nuget.exe pack ignores SDK project's properties


loic-sharma:
If I `nuget.exe pack .\symbols.csproj` the following csproj:

```
<Project Sdk="Microsoft.NET.Sdk">
 <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
   
    <PackageId>Loshar.Symbols</PackageId>
    <PackageVersion>1.0.0</PackageVersion>
  </PropertyGroup>
</Project>
```

The generated package will have the id `symbols` even though the specified package id is `Loshar.Symbols`. Packing this using `dotnet pack` works as expected.

/cc @cristinamanum @shishirx34

rohit21agrawal:
nuget.exe pack isn't supposed to work with sdk style projects.

rrelyea:
Project metadata for Packing, is supported by dotnet pack...not the old nuget pack. (as Rohit says)
our goal is to remove any gaps by adding more features to dotnet pack, and then retire nuget.exe.

cristinamanum:
The nuget.exe ignores the version passed in the Properties command line as well. I tried 4 versions of nuget.exe . All behave the same, they do not add the version. The version is always 1.0.0.

49\nuget.exe pack MyProject.csproj -Properties version="3.1.0"
481\nuget.exe pack MyProject.csproj -Properties version="3.1.0" 
470\nuget.exe pack MyProject.csproj -Properties version="3.1.0"
451\nuget.exe pack MyProject.csproj -Properties version="3.1.0"

loic-sharma:
If that's the case, should we make the NuGet client show an error message if you try to pack an SDK project? The current experience seems like the pack operation half "worked".

PR #12088, opened 2022-09-09
Design for erroring when NuGet.exe pack is used for an SDK based project

### LINKED ISSUE #4491: nuget.exe fails to pack NETCore project
Using nuget.exe 4.0.0 build 2266

### Steps
1. Create a NETstandard class library
1. ``nuget.exe pack <path to csproj>``

### Expected
Package will be created

### Actual
```
Unable to cast object of type 'System.String' to type 'NuGet.Frameworks.NuGetFramework'.
System.InvalidCastException: Unable to cast object of type 'System.String' to type 'NuGet.Frameworks.NuGetFramework'.
   at NuGet.ProjectManagement.NuGetProject.GetMetadata[T](String key)
   at NuGet.ProjectManagement.PackagesConfigNuGetProject..ctor(String folderPath, Dictionary`2 metadata)
   at CallSite.Target(Closure , CallSite , Type , Object , Dictionary`2 )
   at System.Dynamic.UpdateDelegates.UpdateAndExecute3[T0,T1,T2,TRet](CallSite site, T0 arg0, T1 arg1, T2 arg2)
   at NuGet.CommandLine.ProjectFactory.AddDependencies(Dictionary`2 packagesAndDependencies)
   at NuGet.CommandLine.ProjectFactory.ProcessDependencies(PackageBuilder builder)
   at NuGet.CommandLine.ProjectFactory.CreateBuilder(String basePath, NuGetVersion version, String suffix, Boolean build
IfNeeded, PackageBuilder builder)
   at NuGet.Commands.PackCommandRunner.BuildFromProjectFile(String path)
   at NuGet.CommandLine.PackCommand.ExecuteCommand()
   at NuGet.CommandLine.Command.ExecuteCommandAsync()
   at NuGet.CommandLine.Command.Execute()
   at NuGet.CommandLine.Program.MainCore(String workingDirectory, String[] args)
```

nkolev92:
Design for https://github.com/NuGet/Home/issues/7778. [Rendered](https://github.com/NuGet/Home/blob/dev-nkolev92-nugetExeWarnWhenSDKpack/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md)

Completing that issue will allow us to close https://github.com/NuGet/Home/issues/4491 as well. 

The proposal is implemented in https://github.com/nuget/nuget.client/tree/dev-nkolev92-warnNuGetExePacksSDKProject.

### DIFF
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -5,13 +5,13 @@
 
 ## Summary
 
-NuGet.exe does support packing SDK-based, or PackageReference projects in general.
+NuGet.exe does not support packing SDK-based, or PackageReference projects in general.
 If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
 The proposal is to `error` whenever a SDK-based csproj pack is attempted, instructing the user to use `msbuild /t:pack` or `dotnet.exe pack` instead.
 
 ## Motivation
 
-NuGet.exe does support packing SDK-based, or PackageReference projects in general.
+NuGet.exe does not support packing SDK-based, or PackageReference projects in general.
 If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
 The pack SDK experience for PackageReference projects knows how to automatically manage the dependencies and makes multitargeted projects extremely easy to pack.
 It is recommended that SDK projects are managed by using `dotnet.exe` instead.
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -96,4 +96,4 @@ N/A
 
 ## Future Possibilities
 
-- We can consider *warning* when `NuGet.exe pack` is being used on csproj PackageReference projects. Given that you need to undergo a manual migration to PackageReference, this isn't likely to be a priority scenario.
+- We can consider *warning* or *erroring* when `NuGet.exe pack` is being used on csproj PackageReference projects. Given that you need to undergo a manual migration to PackageReference, this isn't likely to be a priority scenario.
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -6,13 +6,13 @@
 ## Summary
 
 NuGet.exe does not support packing SDK-based, or PackageReference projects in general.
-If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
+If you run NuGet.exe pack on an SDK-based csproj, it might pack, but it will do it incorrectly.
 The proposal is to `error` whenever a SDK-based csproj pack is attempted, instructing the user to use `msbuild /t:pack` or `dotnet.exe pack` instead.
 
 ## Motivation
 
 NuGet.exe does not support packing SDK-based, or PackageReference projects in general.
-If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
+If you run NuGet.exe pack on an SDK-based csproj, it might pack, but it will do it incorrectly.
 The pack SDK experience for PackageReference projects knows how to automatically manage the dependencies and makes multitargeted projects extremely easy to pack.
 It is recommended that SDK projects are managed by using `dotnet.exe` instead.
 
@@ -60,9 +60,10 @@ The parsing fails early.
 When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
 
 ```console
-Error NU5049: The `pack` command for SDK-style projects is not supported, use `dotnet pack` to pack this project instead.
-
+Error NU5049: The `pack` command for SDK-style projects is not supported, use `dotnet pack` to pack this project instead. You may set the ENABLE_LEGACY_NUGETEXE_CSPROJ_PACK environment variable to revert to the previous packing behavior.
 ```
+
+This *breaking change* will be announced via blog well before the NuGet.exe carrying the change ships.
 
 ### Technical explanation
 
@@ -79,9 +80,9 @@ To determine whether a project is `.NET SDK` based, we will evaluate the project
 
 - Do nothing. This would mean just acknowledging that SDK-based projects support in NuGet.exe would not be added and not do anything to help migrate users to `dotnet.exe pack` and `msbuild /t:pack`.
 - Add full pack SDK support into NuGet.exe. While not impossible, this creates maintainability considerations. dotnet.exe has a more involved support for the SDK-based PackageReference projects, and as such is considered to be the `right tool for the job`.
-- Call `msbuild /t:pack` from NuGet.exe. This has the same considerations as the previos approach that `dotnet.exe` sohuld be recommended as the right tool for the job.
+- Call `msbuild /t:pack` from NuGet.exe. This has the same considerations as the previos approach that `dotnet.exe` should be recommended as the right tool for the job.
 This is also likely to quietly change the behavior for certain users and change the type of package they're generating.
-A change like this likely has a lot more merit in 2018, than 2022 given how long dotnet.exe has been out.
+A change like this likely has a lot more merit in 2018, than 2022 given how long dotnet.exe has been the recommended tool for SDK based projects..
 - Add a warning and proceed packing.
   - A behavior change to an error in a minor version can be considered disruptive. This is one of those special cases where the disruption might be warranted.
   Instead of an error, we could add a warning instead. Given that NuGet.exe pack by default will usually raise a bunch of warnings, it is likely that if warnings were good enough, the customers would have reviewed said warnings and figured out that all those warnings cannot be addressed correctly.
@@ -92,8 +93,6 @@ A change like this likely has a lot more merit in 2018, than 2022 given how long
 N/A
 
 ## Unresolved Questions
-
-- Is it ok if we error in a NuGet.exe minor version? Scenario was likely broken already.
 
 ## Future Possibilities
 
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -60,7 +60,7 @@ The parsing fails early.
 When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
 
 ```console
-Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet.exe pack` to pack this project instead.
+Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet pack` to pack this project instead.
 ```
 
 ### Technical explanation
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -79,6 +79,9 @@ To determine whether a project is `.NET SDK` based, we will evaluate the project
 
 - Do nothing. This would mean just acknowledging that SDK-based projects support in NuGet.exe would not be added and not do anything to help migrate users to `dotnet.exe pack` and `msbuild /t:pack`.
 - Add full pack SDK support into NuGet.exe. While not impossible, this creates maintainability considerations. dotnet.exe has a more involved support for the SDK-based PackageReference projects, and as such is considered to be the `right tool for the job`.
+- Call `msbuild /t:pack` from NuGet.exe. This has the same considerations as the previos approach that `dotnet.exe` sohuld be recommended as the right tool for the job.
+This is also likely to quietly change the behavior for certain users and change the type of package they're generating.
+A change like this likely has a lot more merit in 2018, than 2022 given how long dotnet.exe has been out.
 - Add a warning and proceed packing.
   - A behavior change to an error in a minor version can be considered disruptive. This is one of those special cases where the disruption might be warranted.
   Instead of an error, we could add a warning instead. Given that NuGet.exe pack by default will usually raise a bunch of warnings, it is likely that if warnings were good enough, the customers would have reviewed said warnings and figured out that all those warnings cannot be addressed correctly.
@@ -90,7 +93,6 @@ N/A
 
 ## Unresolved Questions
 
-- Error message wording. Can be addressed in the pull request making the change instead.
 - Is it ok if we error in a NuGet.exe minor version? Scenario was likely broken already.
 
 ## Future Possibilities
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -60,7 +60,8 @@ The parsing fails early.
 When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
 
 ```console
-Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet pack` to pack this project instead.
+Error NU5049: The `pack` command for SDK-style projects is not supported, use `dotnet pack` to pack this project instead.
+
 ```
 
 ### Technical explanation
--- a/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
+++ b/proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
@@ -1,0 +1,97 @@
+# Show error when using NuGet.exe to pack an SDK csproj
+
+- [Nikolche Kolev](https://github.com/nkolev92)
+- GitHub Issue <https://github.com/NuGet/Home/issues/7778>
+
+## Summary
+
+NuGet.exe does support packing SDK-based, or PackageReference projects in general.
+If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
+The proposal is to `error` whenever a SDK-based csproj pack is attempted, instructing the user to use `msbuild /t:pack` or `dotnet.exe pack` instead.
+
+## Motivation
+
+NuGet.exe does support packing SDK-based, or PackageReference projects in general.
+If you run NuGet.exe pack on an SDK-based csproj, it will pack, but it will do it incorrectly.
+The pack SDK experience for PackageReference projects knows how to automatically manage the dependencies and makes multitargeted projects extremely easy to pack.
+It is recommended that SDK projects are managed by using `dotnet.exe` instead.
+
+## Explanation
+
+### Functional explanation
+
+#### How packing and SDK-based project with NuGet.exe behaves today
+
+Scenario 1: Single target framework SDK based projects, with .NET SDK installed on the machine.
+
+```console
+Attempting to build package from 'SDK.csproj'.
+MSBuild auto-detection: using msbuild version '17.4.0.37601' from 'C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\amd64'.
+Packing files from 'C:\Code\Temp\SDK\bin\Debug\net6.0'.
+WARNING: NU5115: Description was not specified. Using 'Description'.
+WARNING: NU5115: Author was not specified. Using 'Roki2'.
+WARNING: NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location. Consult the list of actions below:
+- Add a dependency group for net6.0 to the nuspec
+Successfully created package 'C:\Code\Temp\SDK\SDK.1.0.0.nupkg'.
+```
+
+The command packages the single assembly, but does not generate the correct dependencies or any other metadata in the same way that the pack SDK would.
+
+Scenario 2: Multiple target framework SDK based projects, with .NET SDK installed on the machine.
+
+```console
+Attempting to build package from 'SDK.csproj'.
+MSBuild auto-detection: using msbuild version '17.4.0.37601' from 'C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\amd64'.
+Error NU5012: Unable to find 'bin\Debug\SDK\bin\Debug\'. Make sure the project has been built.
+```
+
+NuGet.exe does not understand the concept of multitargetting and just merges the output paths.
+
+Scenario 3: SDK-based project, without the .NET SDK installed
+
+```console
+The SDK 'Microsoft.NET.Sdk' specified could not be found.  C:\Code\Temp\SDK\SDK.csproj
+```
+
+The parsing fails early.
+
+### Proposal
+
+When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
+
+```console
+Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet.exe pack` to pack this project instead.
+```
+
+### Technical explanation
+
+As described above, whenever there's no SDK installed, there's simply no accurate way for NuGet to detect that a project is SDK-based.
+We can do some XML parsing, but it is likely to be error prone. `NuGet.exe pack` is not the only command that will fail, so the customer will likely figure out that they need to install the .NET SDK. After that the same scenarios with the installed .NET SDK will follow.
+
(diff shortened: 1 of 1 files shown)


proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+The parsing fails early.
+
+### Proposal
+
+When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
+
+```console
+Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet.exe pack` to pack this project instead.
```
zivkan:
```suggestion
Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet pack` to pack this project instead.
```

The `dotnet` cli only has `.exe` extension on Windows, not on Mac or Linux. I realize this is for nuget.exe, but that runs on Mono on those platforms. On Windows, it's not necessary to specify the `.exe` extension, so I think it's easiest to just leave `.exe` out of any messages.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+### Technical explanation
+
+As described above, whenever there's no SDK installed, there's simply no accurate way for NuGet to detect that a project is SDK-based.
+We can do some XML parsing, but it is likely to be error prone. `NuGet.exe pack` is not the only command that will fail, so the customer will likely figure out that they need to install the .NET SDK. After that the same scenarios with the installed .NET SDK will follow.
+
+To determine whether a project is `.NET SDK` based, we will evaluate the project (part of the normal pack operation), and check for the `UsingMicrosoftNETSDK` property being set. If the property is set, then the project is presumed to be .NET SDK based and will be blocked from packing.
```
zivkan, line 73: Should we check `PackageRestoreStyle` == `PackageReference`, so that legacy csproj PackageReference projects also get captured?

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
+# Show error when using NuGet.exe to pack an SDK csproj
```
zivkan, line 1: Can we stop using the `proposed` folder, please?  This PR is the proposal, and if it gets merged then it shows the proposal is accepted.  Given we're now using pull requests, not the wiki, for designed, I see no reason to have a `proposed` folder any longer.

nkolev92: Feel free to make a proposal for that. I'm not sure they are very helpful either. 

We have 4 folders but 2 are completely empty. We also don't tend to `accept` a design and then let is sit there and not be implemented. We tend to rereview the design right before the implementation begins that way avoiding any stale design issues.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: I think it'd be easy to add that too. I have it in `Future Possiblities` right now. 
I'm pretty split on it, curious what others think.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+
+To determine whether a project is `.NET SDK` based, we will evaluate the project (part of the normal pack operation), and check for the `UsingMicrosoftNETSDK` property being set. If the property is set, then the project is presumed to be .NET SDK based and will be blocked from packing.
+
+## Drawbacks
+
+- The proposal here would be `breaking` for customers, but it's more than likely that the scenario for which they used nuget.exe pack on an SDK based project never worked correctly.
+
+## Rationale and alternatives
+
+- Do nothing. This would mean just acknowledging that SDK-based projects support in NuGet.exe would not be added and not do anything to help migrate users to `dotnet.exe pack` and `msbuild /t:pack`.
```
jeffkl, line 81: Since NuGet.exe locates MSBuild already, does it make sense to consider having NuGet.exe just call `msbuild.exe /t:pack`?  It could still log a warning telling the user to just run `msbuild /t:pack` themselves next time but at least it would do the correct thing.

nkolev92: I think if this was being done years ago, I'd say that's a valid thing we should consider. 
I have my hesitations about doing something like today, years after the original PackageReference release. 

Another motivation behind erroring is to guide users to the right tool for the job, and dotnet.exe has many commands for SDK based projects that NuGet.exe does not and will not have.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+- Add full pack SDK support into NuGet.exe. While not impossible, this creates maintainability considerations. dotnet.exe has a more involved support for the SDK-based PackageReference projects, and as such is considered to be the `right tool for the job`.
+- Add a warning and proceed packing.
+  - A behavior change to an error in a minor version can be considered disruptive. This is one of those special cases where the disruption might be warranted.
+  Instead of an error, we could add a warning instead. Given that NuGet.exe pack by default will usually raise a bunch of warnings, it is likely that if warnings were good enough, the customers would have reviewed said warnings and figured out that all those warnings cannot be addressed correctly.
```
loic-sharma, line 88: > Given that NuGet.exe pack by default will usually raise a bunch of warnings

Is that true? From my recollection, it's easy to not get any warnings which is why this behavior is so confusing. For example, it's easy to believe that packing symbols is broken when using nuget.exe: https://github.com/NuGet/Home/issues/7453

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+## Drawbacks
+
+- The proposal here would be `breaking` for customers, but it's more than likely that the scenario for which they used nuget.exe pack on an SDK based project never worked correctly.
+
+## Rationale and alternatives
+
+- Do nothing. This would mean just acknowledging that SDK-based projects support in NuGet.exe would not be added and not do anything to help migrate users to `dotnet.exe pack` and `msbuild /t:pack`.
+- Add full pack SDK support into NuGet.exe. While not impossible, this creates maintainability considerations. dotnet.exe has a more involved support for the SDK-based PackageReference projects, and as such is considered to be the `right tool for the job`.
+- Add a warning and proceed packing.
+  - A behavior change to an error in a minor version can be considered disruptive. This is one of those special cases where the disruption might be warranted.
```
loic-sharma, line 87: How many customers use nuget.exe to pack SDK projects? There's a surprising amount of upvotes ([1](https://github.com/NuGet/Home/issues/4491#issuecomment-527152856), [2](https://github.com/NuGet/Home/issues/7778#issue-408582969)) for this scenario, this may be a non-trivial disruption.

nkolev92: I'd venture a guess that the upvotes in both issues are from people that tried it and realized it doesn't really work properly and thus are not using anymore. 

It's difficult to know how many people are actually using `nuget.exe pack` for SDK projects successfully, since the package created does not really declare the dependencies (that's at least 1 thing that doesn't work).

The primary audience would be customers that try to pack an SDK based project with nuget.exe for the first time.

loic-sharma: Agreed, but I'm concerned by the potential impact of making this an error. An error is reasonable only if we know the disruption is minimal. Imagine we release this and we find out there are many folks using nuget.exe to pack SDK based projects - would we need to revert this error? How can we mitigate that risk? Perhaps it'd be better to play it safe and just make this a warning for now? We can always promote this to an error later once we feel more confident with the impact.

nkolev92: You summarized the concerns well, it's something hoping to get a good decision about through more feedback.

> An error is reasonable only if we know the disruption is minimal. 

That's fair. 
Another scenario where disruption may be warranted is if the scenario is irreparably broken, which this command may qualify for. 

By default, nuget.exe pack on an SDK project raises a bunch of warnings. Going through said warnings would have likely led people to figure out that nuget.exe pack is not really working the way they need it to.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: Things have changed in later versions. 
This is what I get with that project today: 

```console
Attempting to build package from 'SDK.csproj'.
MSBuild auto-detection: using msbuild version '17.4.0.37601' from 'C:\Program Files\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin\amd64'.
Packing files from 'C:\Code\Temp\SDK\bin\Debug\netstandard2.0'.
WARNING: NU5115: Description was not specified. Using 'Description'.
WARNING: NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location. Consult the list of actions below:
- Add a dependency group for .NETStandard2.0 to the nuspec
```

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
JonDouglas: The only reason we have a proposed folder is that things might not be accepted to be worked on even if a PR is ready to merge said design. We can change that and just use PRs though. I just based the initial flow off other popular RFC processes.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+The parsing fails early.
+
+### Proposal
+
+When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
+
+```console
+Error NU5049: NuGet.exe does not support packing SDK-based projects, use `dotnet pack` to pack this project instead.
```
JonDouglas:
```suggestion
Error NU5049: The `pack` command for SDK-style projects is not supported, use `dotnet pack` to pack this project instead.

```

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+- Add an information message and proceed warning. Given that the information this tool is not the right one for the job is critical, this is not likely to move the needle enough.
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions
+
+- Error message wording. Can be addressed in the pull request making the change instead.
+- Is it ok if we error in a NuGet.exe minor version? Scenario was likely broken already.
```
JonDouglas: We won't really know without shipping it. Given this is nuget.exe only, I think we can ship in a minor release with little disruption and plenty of feedback if we do disrupt.

JonDouglas (approved):
Great spec! Added some comments for consideration.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,97 @@
...
+As described above, whenever there's no SDK installed, there's simply no accurate way for NuGet to detect that a project is SDK-based.
+We can do some XML parsing, but it is likely to be error prone. `NuGet.exe pack` is not the only command that will fail, so the customer will likely figure out that they need to install the .NET SDK. After that the same scenarios with the installed .NET SDK will follow.
+
+To determine whether a project is `.NET SDK` based, we will evaluate the project (part of the normal pack operation), and check for the `UsingMicrosoftNETSDK` property being set. If the property is set, then the project is presumed to be .NET SDK based and will be blocked from packing.
+
+## Drawbacks
+
+- The proposal here would be `breaking` for customers, but it's more than likely that the scenario for which they used nuget.exe pack on an SDK based project never worked correctly.
+
+## Rationale and alternatives
```
JonDouglas, line 79: Thanks for listing out the alternatives. I think this approach makes plenty sense to help people use the right tool for the job.

I don't like doing nothing, because it doesn't help our or customer's problem at all.

I don't see full pack SDK support being something we can do nor would want to do for the LTS. It is an opportunity if it's the final thing we may need to add to NuGet.exe though, so not off the table.

The warning scenario is interesting though. I just worry people will absolutely ignore it and just continue doing what they've been doing though.

The disruption could be seen as potentially impactful to various CI platforms using "nuget" pre-built actions. The scarier thought is thinking how many people are using nuget to pack incorrectly in the first place for SDK-style scenarios. I think we ought to correct the behavior and bite the bullet with some blog messaging prior to release? That way people have some time to know it's coming.

nkolev92: I think announcing it/blogging preemptively if we add a warning is *a great idea*. 

Given the timeline, I think targeting 6.5 would give us plenty of time to post about that. 
Very difficult to know if we'd reach the right audience though :/

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: Even better, thanks!

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
loic-sharma: Nice! Marking this as resolved 👍

EDIT: For whatever reason, GitHub won't let me resolve this ¯\_(ツ)_/¯

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
kartheekp-ms: Does NuGet.exe pack work today for legacy csproj `PackageReference` projects? What will be the value of `UsingMicrosoftNETSDK` property for legacy `PackageReference` projects? If the value is `No` then I would assume the current implementation will raise an error. Is my understanding correct?

nkolev92: > Does NuGet.exe pack work today for legacy csproj PackageReference projects

It does not. `NuGet.exe in general won't work well with PackageReference projects`. 

> What will be the value of UsingMicrosoftNETSDK property for legacy PackageReference projects?

It will be empty. csproj PackageReference projects *do not* have the pack targets by default, so you'd need to install a package in order to make the project packable. 

NuGet.exe will complete the packing, pack the assembly, but the metadata and dependencies will not be correct.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: We'll blog/announce the breaking change.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,99 @@
...
+
+- Do nothing. This would mean just acknowledging that SDK-based projects support in NuGet.exe would not be added and not do anything to help migrate users to `dotnet.exe pack` and `msbuild /t:pack`.
+- Add full pack SDK support into NuGet.exe. While not impossible, this creates maintainability considerations. dotnet.exe has a more involved support for the SDK-based PackageReference projects, and as such is considered to be the `right tool for the job`.
+- Call `msbuild /t:pack` from NuGet.exe. This has the same considerations as the previos approach that `dotnet.exe` should be recommended as the right tool for the job.
```
Nirmal4G, line 83: Why this won't work? The `dotnet pack` also calls `dotnet msbuild -t:Pack` although it's Core MSBuild. There shouldn't be any differences between Core/Full MSBuild. They are just targeting different runtime, that's all.

_There are proposals to MSBuild to break away from compat. But it's unlikely to happen anytime soon. May be after .NET Framework 4.9 or .NET 9_ 😇

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions
+
+## Future Possibilities
+
+- We can consider *warning* or *erroring* when `NuGet.exe pack` is being used on csproj PackageReference projects. Given that you need to undergo a manual migration to PackageReference, this isn't likely to be a priority scenario.
```
Nirmal4G, line 99: Since `nuget pack` does not support `PackageReference` why not error out with the escape hatch similar to this proposal and make `nuget pack` call Pack target for both PR and SDK-style projects if pack targets are imported in both of them.

I think this would be a better solution for customers. This solves two scenarios in one go. It'll also help people migrating to .NET (Core).

We could add a suggestion to use `dotnet pack` as a warning later on if MSBuild tries to break compat. It will have the least impact as it would be tied in with MSBuild or Visual Studio announcement.

Nirmal4G (changes requested):
A better solution, maybe!?

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: I think it'd be trivial to handle the old csproj PackageReference scenario once we have established a pattern here. 
By keeping it scoped, it increases the chance of this getting accepted. 

> We could add a suggestion to use dotnet pack as a warning later on if MSBuild tries to break compat. It will have the least impact as it would be tie in with MSBuild or Visual Studio announcement.

There's really no difference between `msbuild /t:pack` and `dotnet pack`. 
The difference is really between old csproj and SDK csproj, as the SDK one automatically includes the pack targets, while the old style does not, thus it does not have the same servicing as SDK projects.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: We thought quietly switching out the implementation doesn't really inform users that `dotnet` is what they should use with SDK based projects.

It's a 2 fold effort given that `dotnet` has extra commands for SDK projects that are only available there.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
Nirmal4G: I don't think my message is getting through. What I mean is that there's a chance of getting this proposal to work for both PR and SDK-style by checking for NuGet Pack targets import. There's property `ImportNuGetBuildTasksPackTargetsFromSdk` but that doesn't really tell us if the user has disabled Pack targets or if the targets has been imported.

We could add a new property `NuGetPackTargetsHasBeenImported` to Pack targets and check this for Pack target availability. If the Pack targets are available then we would invoke `msbuild -t:Pack` and call it a day. Bonus points for checking if the project is using `packages.config` or `PackageReference` restore style.

If the property is not available (_it could be older targets_) then there are 3 ways to deal with them...

1. Check for `ImportNuGetBuildTasksPackTargetsFromSdk` property or `NuGet.Build.Tasks.Pack` package. If it is true then proceed to `msbuild -t:Pack` else continue.

2. If the project has set `UsingMicrosoftNETSdk` but disabled `UsingNETSdkDefaults`, it is likely a legacy project using .NET SDK (_yeah, seen those lying around in the GitHub_). If that is the case then show the suggestion to enable Pack targets and error out with an escape hatch.

3. Finally, if the project is legacy and using `packages.config` then proceed to `nuget pack`. If it is using `ProjectReference` then show suggestion to enable Pack targets and error out with an escape hatch.

The escape hatch should always lead to `nuget pack`. This is just a starting point but things could change during actual implementation.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
Nirmal4G: True but if we want to inform users, we could say...

```
> nuget pack ...
Using NuGet Pack targets (msbuild -t:Pack)
```

```
> nuget pack ...
Using NuGet Pack CLI
```

We're only going to map available commands so as to not break existing CI/CD pipelines/workflows. With the above message people can decide when they want to move out of `nuget pack`. Besides you can call this out in the blog post. Sure, it might not have the same effect of seeing on the CLI on every invoke. Still a solid way of doing things properly.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
Nirmal4G: I don't think my message is getting through. What I mean is that there's a chance of getting this proposal to work for both PR and SDK-style by checking for NuGet Pack targets import. There's property `ImportNuGetBuildTasksPackTargetsFromSdk` but that doesn't really tell us if the user has disabled Pack targets or if the targets has been imported.

We could add a new property `NuGetPackTargetsHasBeenImported` to Pack targets and check this for Pack target availability. If the Pack targets are available then we would invoke `msbuild -t:Pack` and call it a day. Bonus points for checking if the project is using `packages.config` or `PackageReference` restore style.

If the property is not available (_it could be older targets_) then there are 3 ways to deal with them...

1. Check for `ImportNuGetBuildTasksPackTargetsFromSdk` property or `NuGet.Build.Tasks.Pack` package. If it is true then proceed to `msbuild -t:Pack` else continue.

2. If the project has set `UsingMicrosoftNETSdk` but disabled `UsingNETSdkDefaults`, it is likely a legacy project using .NET SDK (_yeah, seen those lying around in the GitHub_). If that is the case then show the suggestion to enable Pack targets and error out with an escape hatch.

3. Finally, if the project is legacy and using `packages.config` then proceed to `nuget pack`. If it is using `ProjectReference` then show suggestion to enable Pack targets and error out with an escape hatch.

The escape hatch should always lead to `nuget pack`. This is just a starting point but things could change during actual implementation.

zivkan: > There's really no difference between msbuild /t:pack and dotnet pack.

Disclaimer needed. `msbuild` will load msbuild from Visual Studio's install directory, and therefore all the props/targets available in VS/msbuild will be available. `dotnet` will load MSBuild's props/targets from the .NET SDK directory. Therefore, there's a bunch of VS projects which will not load/evaluate in the dotnet cli, but will with msbuild.

If we want to scope the nuget.exe pack message/experience to SDK style projects only, then ok `msbuild -t:pack` and `dotnet pack` are interchangeable. But if we want to handle non-SDK style projects, we really need to use msbuild only.

nkolev92: @Nirmal4G 
I get your point, we're just explicitly choosing *not* to do it in this change.  That's why it's listed in future possibilities. 

> There's property ImportNuGetBuildTasksPackTargetsFromSdk but that doesn't really tell us if the user has disabled Pack targets or if the targets has been imported.

Yeah, there's no such property, but there's a capability that we might be able to use instead.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
nkolev92: > True but if we want to inform users, we could say...

I don't think that'd be good enough given what we're trying to communicate. 
nuget.exe pack would already do weird things with an SDK project today. Customers that were observing the output would've moved away anyways as the warnings would be something that'd be difficult to fix. 

It's similar motivation as the whole, why not do a warning thing I mentioned in the alternatives:

>   Instead of an error, we could add a warning instead. Given that NuGet.exe pack by default will usually raise a bunch of warnings, it is likely that if warnings were good enough, the customers would have reviewed said warnings and figured out that all those warnings cannot be addressed correctly.

proposed/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
Nirmal4G: See NuGet/NuGet.Client#1368. This PR add the `$(ImportNuGetBuildTasksPackTargetsFromSdk)` property. And it's been there ever since.

nkolev92: `NugetTaskAssemblyFile` is an even older one. We have options for sure.

nkolev92: Resolving this thread.

nkolev92:
Thank you all for the feedback. 

At this point we're limiting the scope to what's in the document. 
Special thanks to @Nirmal4G for their engagement! 

I'd be happy if we expanded the check in the future.

Issue #5154, opened 2017-05-02
Treat TargetFramework(s) values as aliases


emgarten:
NuGet should treat the values from ``TargetFramework`` and ``TargetFrameworks`` as aliases.

When setting the property ``TargetFramework=(alias)`` the actual target framework value can be evaluated from ``TargetFrameworkIdentifier`` and ``TargetFrameworkVersion``.

See for examples and more details: https://github.com/dotnet/project-system/issues/1938

*Update by @nkolev92*

NuGet in 5.8 and .NET 5.0 SDK now treats TargetFramework as an alias, however duplicate framework entries are not supported yet (ex. You can't target `.NETCoreApp,Version=3.0` twice.)

davkean:
Yes, see this guideline I just wrote: https://github.com/dotnet/project-system/blob/master/docs/repo/coding-conventions.md#data

natidea:
RE: https://github.com/dotnet/project-system/issues/615 
Nominate currently passes both TF and TFM. TF is the key used to identify each `TargetFrameworkInfo` object (though the api calls it "TargetFrameworkMoniker"), and TFM is stored in the `Properties` bag. I'll close https://github.com/dotnet/project-system/issues/615 since we are already passing all the info NuGet needs to make this switch.

nkolev92:
@rrelyea we should discuss this and tackle it for .NET 5.0.

zkat:
Moving under the .NET 5 epic (#9090), as this is probably something we want to look into prioritizing for that. It'll require a lot of partner work/sync. Most likely the next step for this is to collect some partner/user stories to get a sense of the impact the lack of this feature is having, so we can better prioritize it in our roadmap. (so if you're reading it, feel free to tell us how this impacts you!)

nkolev92:
As me, @zkat, @zivkan and others discussed during the sync-up, I think this is larger than a week's work and thus not quality week appropriate. 

Affected areas: 

* Pack (What happens when we see duplicate target frameworks). 
* ProjectReference..selecting the best framework would be ambiguous when a build products a build for the same target framework. 
* Lock file.  

cc @anangaur

nkolev92:
Another affected component is the solution explorer dropdown menu. 

See: https://github.com/NuGet/NuGet.Client/pull/3392#discussion_r430934193

ViktorHofer:
> so if you're reading it, feel free to tell us how this impacts you!

[...]

PR #12124, opened 2022-10-03
Design for allowing restore of multiple equivalent frameworks (Target Frameworks as Aliases)

nkolev92:
Design for https://github.com/NuGet/Home/issues/5154 

[Rendered](https://github.com/NuGet/Home/blob/dev-nkolev92-tfmaliases/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md)

cc @dsplaisted

### DIFF
--- a/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
+++ b/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
@@ -1,0 +1,462 @@
+# Support for handling multiple equivalent framework
+
+- [Nikolche Kolev](https://github.com/nkolev92), [Andy Zivkovic](https://github.com/zivkan)
+- GitHub Issue ([5154](https://github.com/NuGet/Home/issues/5154)) - Treat Target Framework as Aliases
+
+## Summary
+
+.NET SDK projects are capable of multi targeting based on a target framework where you can have different dependencies for each framework.
+In some cases, it is convenient to multi target based on different pivots such as target application version or maybe the specific runtime, as packages can contain runtime assets.
+This proposal covers the steps that need to be taken to enable that capability.
+
+## Motivation
+
+Allowing the same framework to be using multiple times in restore would allow customers to:
+
+- Use the same project to generate runtime specific assemblies based on the same target framework.
+  - Enable support for building for multiple RIDs, similar to how TargetFrameworks works - Evidence: <https://github.com/dotnet/sdk/issues/9795>
+- Allow the same project to target different Visual Studio versions, like for example 17.x and 18.x of Visual Studio.
+- Make it easier to run benchmarks, by allowing the same framework to reference 2 different package versions in 2 different build legs
+
+## Explanation
+
+### Functional explanation
+
+#### Background - TargetFramework values are already aliases
+
+NuGet, and the .NET SDK more generally, uses `TargetFrameworkMoniker` and `TargetPlatformMoniker` for project and package compatibility.
+Each of the two `*Moniker` properties are composed of `*Identifier` and `*Version` properties, which NuGet doesn't use, but might be used by other components, so should also be set correctly.
+If these properties do not have values, the .NET SDK will infer the values by parsing the `TargetFramework` property, which is what we typically see defined in project files.
+However, if the `TargetFrameworkMoniker` property is defined explicitly (`TargetPlatformMoniker` is allowed be be undefined), NuGet (and several other components, like the .NET SDK and Visual Studio's dotnet/project-system) will already treat these values as aliases.
+
+For example, the following is a valid project that works before this spec is implemented.
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>banana</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'banana' ">
+    <TargetFrameworkIdentifier>.NETFramework</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v4.7.2</TargetFrameworkVersion>
+    <TargetFrameworkMoniker>.NETFramework,Version=v4.7.2</TargetFrameworkMoniker>
+  </PropertyGroup>
+
+</Project>
+```
+
+```console
+C:\Code\Temp\Aliases [main]> dotnet restore
+  Determining projects to restore...
+  Restored C:\Code\Temp\Aliases\Aliases.csproj (in 47 ms).
+
+C:\Code\Temp\Aliases [main]> dotnet build --framework banana
+MSBuild version 17.3.0+92e077650 for .NET
+  Determining projects to restore...
+  All projects are up-to-date for restore.
+  Aliases -> C:\Code\Temp\Aliases\bin\Debug\banana\Aliases.dll
+
+Build succeeded.
+    0 Warning(s)
+    0 Error(s)
+
+Time Elapsed 00:00:00.27
+
+C:\Code\Temp\Aliases [main]> dotnet publish --framework banana
+MSBuild version 17.3.0+92e077650 for .NET
+  Determining projects to restore...
+  All projects are up-to-date for restore.
+  Aliases -> C:\Code\Temp\Aliases\bin\Debug\banana\Aliases.dll
+  Aliases -> C:\Code\Temp\Aliases\bin\Debug\banana\publish\
+```
+
+However, what doesn't currently work is when multiple TargetFramework values resolve to the same `*Moniker` property value.
+
+#### The proposal
+
+Only .NET SDK projects will support aliasing, as they are the only ones that support multiple frameworks.
+Aliasing requires all the tooling (.NET SDK, Visual Studio) to be aliasing aware, since there are some breaking changes in the files that these components depend on.
+In .NET 10.0.100, you would get an error if you tried to create a project with 2 of the same framework.
+If someone tries to target duplicate frameworks using a Visual Studio version that's aliasing aware, but pins a .NET SDK version that is not aliasing aware, an error will be raised indicating that the scenario requires a tooling upgrade.
+
+The changes proposed in this document will enable the following hypothetical scenarios.
+
+Note that projects and SDKs are responsible for setting the `TargetFramework*` and `TargetPlatform*` properties, which NuGet uses as the canonical framework to use for compatibility checks.
+Therefore, the exact `TargetFrameworks` value is not important to NuGet, as it is either the SDK author, or the project owner, to set any value that is reasonable to them.
+
+Firstly, a single project that creates multiple platform specific assemblies, all targeting the same .NET version.
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>linux;mac</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFrameworkIdentifier>.NETCoreApp</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v6.0</TargetFrameworkVersion>
+    <TargetFrameworkMoniker>.NETCoreApp,Version=v6.0</TargetFrameworkMoniker>
+
+    <DefineConstants>$(DefineConstants);$(TargetFramework)</DefineConstants>  
+    <AssemblyName>$(MSBuildThisFileName).$(TargetFramework)</AssemblyName>
+  </PropertyGroup>
+
+</Project>
+```
+
+Another example would be a Visual Studio Extension project targeting multiple versions of VS could look similar to the following.
+In this example, the Visual Studio Extensibility SDK would be responsible for setting `TargetFrameworkIdentifier`, `TargetFrameworkVersion` and `TargetFrameworkMoniker` properties.
+
+```xml
+<Project Sdk="Microsoft.VisualStudio.Extensibility.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>vs18;vs17.14</TargetFrameworks>
+  </PropertyGroup>
+
+</Project>
+```
+
+- When a project reference is made to a project with aliases, the TargetFramework will serve as a secondary pivot in case the standard get nearest target framework returns more than one option.
+The technical explanation has [the gory details](#project-to-project-references)
+- Packages lock file will also be extended to support aliasing. The file format will need to be updated, but it's already version and that work should be straightforward. [Packages Lock File technical details](#lock-file-changes).
+- Visual Studio Package Manager UI does not have great support for multi-targeting and no new work is being proposed here. This is a specific flavor of multi-targeting, and as such no difference from the challenges we have today. [Visual Studio challenges](#visual-studio-challenges)
+- dotnet package add, dotnet package update and dotnet package list will all be updated to support per framework operations in case they do not support it already.
+
+### Pack and duplicate frameworks
+
+There is no intuitive solution to the pack scenario.
+Given a project with multiple versions of the same framework, pack will fail by default.
+
+NuGet will hinge off off IncludeBuildOutput, IncludeContentInPack and SuppressDependenciesWhenPacking to decide to skip packing for a framework, effectively allowing a framework to be completely excluded.
+
+Pack can only succeed when one of the 2 frameworks doesn't have both of the following `IncludeContentInPack == true or IncludeBuildOutput == true or SuppressDependenciesWhenPackaging == false`.
+
+The pack experience likely requires some learning which may not be possible until we see more scenarios utilizing duplicate frameworks.
+
+> Note that as part of this proposal, there may be some scenarios where alias usage in per framework commands will be validated and potentially fixed. This will discovered during the implementation and won't be contained in detail in the current design.
+
+### Technical explanation
+
+There are many changes required to implement this feature.
+
+1. [Restore output (assets file) changes](#restore-output-assets-file-changes)
+1. [Components iterating target frameworks](#components-iterating-target-frameworks)
+1. [Lock file changes](#lock-file-changes)
+1. [ProjectReference changes](#project-to-project-references)
+1. [Visual Studio Challenges](#visual-studio-challenges)
+
+The changes listed are largely what NuGet needs to implement in code we own, but there will be impacts to other components as well.
+In particular the .NET SDK may need some changes to be able to use the assets file correctly.
+Majority of this had fortunately been done in past iterations and the .NET SDK currently depends on are NuGet APIs.
+Other tools, like various Visual Studio components, might have assumptions about TFM uniqueness in the project, which this feature changes.
+Therefore, while the NuGet team works very closely with the .NET SDK team and we will ensure that building projects work, there will increased risk of other features and tools breaking, which would normally not be affected by NuGet changes.
+
+### Restore output (assets file) changes
+
+In PackageReference, NuGet has a contract with the .NET SDK where NuGet writes the assets file and the .NET SDK consumes it.
+The .NET SDK uses NuGet libraries for the parsing of this file and as such, there'll be no code on the .NET SDK to consume it, but rather all the scenarios can be handled within the different NuGet code.
+
+PackageReference is also supported by non-SDK style projects, which use [dotnet/NuGet.BuildTools](https://github.com/NuGet.BuildTools) which, despite the name, is owned by the non-SDK project system team, not NuGet.
+The changes to the assets file that affect legacy projects will be done in a non-breaking way and as such, changes should not be required there.
+NuGet will utilize the SDKAnalysisLevel property when it writes out an assets file with breaking changes, ensuring that the .NET SDK will be able to read the assets file for the build.
+
+The assets file will be updated in a bunch of location to use the alias as a key instead of the target framework, effectively allowing duplicate frameworks.
+
+The assets file contains a `version` property since it was first created, and is always the first property in the file.
+The current assets file version is 3, and this feature will increment the version to 4.
+
+```diff
+-"version": 3
++"version": 4
+```
+
+By default for non-SDK style projects, we will always use version 3.
+Non-SDK style project only have a single target, so those project types won't see a change.
+
+For simplicity of implementation, whenever SDKAnalysisLevel is set to 10.0.200 or more, NuGet will use the V4 assets file.
+Whenever the SDKAnalysisLevel is set to a lower version, then NuGet will use the V3 assets file.
+Scenarios where the .NET SDK is newer of the Visual Studio version are not a concern.
+The same considerations extend to NuGet.exe restore, which will also be dependent on the SDKAnalysisLevel property.
+
+#### Target Framework Pivots
+
+There are multiple places in the assets file where the "NuGet Framework" is used as a JSON object property key.
+There are 3 top level sections where changes will be needed. The `targets`, `projectFileDependencyGroups` and `project` sections.
+
+Take a project using `<TargetFramework>production</TargetFramework>`, `<TargetFrameworkMoniker>.NETCoreApp,Version=v8.0</TargetFrameworkMoniker>`, and `<RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>`.
+
+Note that the effective framework itself will not be part of the serialized part, but it'll cleaned up by the reader by looking up the target framework based off of the "project" section of the assets file.
+
+1. `$/targets`
+
+    ```diff
+      "targets": {
+    -    "net8.0": {},
+    -    "net8.0/linux-x64": {},
+    +    "production": {},
+    +    "production/linux-x64": {},
+      },
+    ```
+
+For the projectFileDependencyGroup, the type already uses a string as the pivot, which simplifies things quite a bit, <https://github.com/NuGet/NuGet.Client/blob/f764caebef5937cff2802ba65208e8f43afc6a45/src/NuGet.Core/NuGet.ProjectModel/ProjectFileDependencyGroup.cs#L20>.
+
+The way the .NET SDK uses this is non breaking. It's never parsed since ProjectFileDependencyGroup already uses FrameworkName as a framework, so as long as this value and the targets name are changed together, it's good enough.
+
+1. `$/projectFileDependencyGroups`
+
+    ```diff
+      "projectFileDependencyGroups": {
+    -    "net8.0": [],
+    +    "production": [],
+      },
+    ```
+
+##### "project" PackageSpec changes
+
+The PackageSpec type and it's serialized form is used in multiple places in the NuGet code.
+It is a part of the assets file, but it is also used for the dgspec and the no-op scenarios.
+
+The PackageSpec is not versioned and at this point, I'm not proposing adding a version.
+
+The changes being made here can be divided into 2 parts.
+
+- Additive changes
+- Breaking changes
+
+The additive changes will be an always thing. Every newly serialized PackageSpec will make this change. This should not affect any of the existing scenarios, because at worse the new property is not parsed.
+
+The following 2 changes will be made on the writer side:
+
+- Add "framework" to the packageSpec, while keeping targetAlias for simplicity sake.
+This is an additive change and will be written out for every new package spec for simplicity.
+- Change the key from framework to alias. In most cases, really almost all non-alias cases, this is likely to be the same and cause no problems.
+The only default cases where the alias and the effective tfm are different are MAUI and platform scenarios.
+
+1. `$/project/restore/frameworks`
+
+    ```diff
+    "frameworks": {
+    -  "net8.0": {
+    -    "targetAlias": "production",
+    +  "production": {
+    +    "framework": "net8.0",
+    +    "targetAlias": "production",
+      }
+    }
+    ```
+
+1. `$/project/frameworks`
+
+    ```diff
+    "restore": {
+      "frameworks": {
+      -  "net8.0": {
+      -    "targetAlias": "production",
+      +  "production": {
+      +    "framework": "net8.0",
+      +    "targetAlias": "production",
+        }
+      }
+    }
+    ```
+  
+Given that there is no version for the package spec at this point, we can do a simple trick to infer the type of PackageSpec to read.
+
+- When reading a framework, if both targetAlias and framework, then this is equivalent to V4 and a .NET SDK scenario
+- If only targetAlias is set, this was a V3 writer and an SDK
+- If there's no targetAlias and no framework, this was a V3 writer and legacy project.
+
+There are 3 scenarios where NuGet writes out a package spec.
+
+- DependencyGraphSpec serialization for no-op.
+- DependencyGraphSpec write for unloaded project scenarios
+- PackageSpec written within the assets file in the "project" section.
+
+For simplicity sake, in the first 2 cases, NuGet will write-out a v4 version, regardless of whether we are dealing with an SDK or legacy project.
+In the dg spec for unloaded project scenario, this may cause a problem in theory if the .NET SDK is newer than the Visual Studio version, but that is unlikely and with the Visual Studio's modern lifecycle changes, at worst, we will have a month where VS is unable to read the package spec from the updated SDK.
+
+The PackageSpec written within the assets file of the project section will be switch between V3 and V4 based on the SDKAnalysisLevel property as previously mentioned.
+
+### Components iterating target frameworks
+
+There are some components that either read the assets file directly, or interact with project `TargetFrameworks` in another way.
+Some examples are:
+
+- The .NET SDK, as part of build, as previously discussed.
+- Any `dotnet` CLI command with a `--frameworks` or `-f` argument, such as `dotnet build`, `dotnet publish`, `dotnet test`.
+- `dotnet list package`.
+- Numerous features in Visual Studio.
+  - Test Explorer.
+  - Text editor (left-most dropdown, for TFM specific Intellisense).
+  - Solution Explorer's Dependencies node.
+
+### Lock file changes
+
+NuGet's "repeatable build" feature adds a package lock file to the project directory.
+The restore has a lock file has a similar schema to the assets file, so that schema would also need to be amended.
+Fortunately when Central Package Management and its transitive pinning was introduced we introduced the concept of a PackagesLockFile version successfully.
+We'd just add a version 3.
+
+NuGet would add version 3 of the packages lock file, which will [pivot on the target framework alias, just as assets files will](#target-framework-pivots).
+This lock file has a single dependencies section and the key there would be changed to the alias instead.
+
+From:
+
+```json
+{
+  "version": 1,
+  "dependencies": {
+    ".NETCoreApp,Version=v10.0": {
+    }
+  }
+}
+```
+
+To:
+
+```json
+{
+  "version": 3,
+  "dependencies": {
+    "net10.0": {
+    }
+  }
+}
+```
+
+### Visual Studio challenges
+
+Currently the PM UI does not have any support for multi targeting.
+This isn't anything new, but it can affect the experience of customers moving to SDK projects for the first time.
+Especially for customers who are not comfortable hand editing XML or MSBuild files.
+
+Another concern are all the APIs for displaying the dependencies, transitive dependencies and Get Installed Packages APIs in NuGet.VisualStudio.Contracts.
+All of these APIs may require an incremental addition.
+
+### Project to Project references
+
+MSBuild has a [`ProjectReference` protocol](https://github.com/dotnet/msbuild/blob/25fdeb3c8c2608f248faec3d6d37733ae144bbbb/documentation/ProjectReference-Protocol.md), which explains how `ProjectReferences` work.
+MSBuild calls the `GetReferenceNearestTargetFrameworkTask`, which NuGet implements, in order to obtain best TFM matching.
+
+Currently `GetReferenceNearestTargetFrameworkTask` is implemented by running NuGet's "nearest match" algorithm.
+This feature would extend it to first try nearest match, and if there is more than one matching `TargetFramework`, then look for a `TargetFramework` that has an exact name match in both projects.
+If there are more than one "nearest" framework, but no exact name match, then for the first version of this feature, NuGet will report an error.
+Depending on customer feedback, build customization can be added in the future.
+
+Today the nearest selection uses the following algorithm
+
+```mermaid
+graph TD
+    Start{Is there a nearest framework compatible with my project's target framework?}
+    Start -- Yes --> Yes1[Return that framework]
+    Start -- No --> No1{Are there any fallback frameworks defined in my project?}
+    No1 -- Yes --> Yes2[Is there a nearest framework compatible with my project's fallback frameworks?]
+    No1 -- No --> No2[No match found]
+    Yes2 -- Yes --> Yes3[Return that framework]
+    Yes2 -- No --> No3[No match found]
+```
+
+This algorithm does not account for multiple matches, as right now it is not allowed that projects have the same framework twice.
+To enable scenarios such as having test projects for these aliased project, the following enhancement will be added to the get nearest protocol.
+
+```mermaid
+graph TD
+    Start{Is there a nearest framework compatible with my project's target framework?}
+    Start -- Yes --> Yes1{Is it a single framework?}
+    Yes1 -- Yes --> Yes2[Return that framework]
+    Yes1 -- No --> No2{Do any of the multiple equivalent framework have the exact same `TargetFramework` value as the current project's.}
+    Start -- No --> No1{Are there any fallback frameworks defined in my project?}
+    No1 -- Yes --> Yes3[Is there a single nearest framework compatible with my project's fallback frameworks?]
+    Yes3 -- Yes --> Yes4[Return that framework]
+    Yes3 -- No --> No3[No match found]
+    No1 -- No --> No4[No match found]
+    No2 -- Yes --> Yes5[Return that framework]
+    No2 -- No --> No5[No match found]
+```
+
+To achieve this an enhacement needs to be made to the [GetReferenceNearestTargetFrameworkTask](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/GetReferenceNearestTargetFrameworkTask.cs), by adding a `TargetFramework` parameter. The code change will be made in the [MSBuild](https://github.com/dotnet/msbuild/blob/main/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1886-L1906) repo.
+
+As part of this work, there'll be an addition in the NuGet.Client's GetNearest libraries, but that's something that will be figured out during the implementation phase.
+
+## Drawbacks
+
+This feature just adds additional scenarios. All the risk it carries is implementational one.
+The new pack scenarios are pretty complex, but the scenario it enables are a good trade-off.
+
+## Rationale and alternatives
+
+The alternatives to this feature is using multiple projects, which is the current behavior.
+There are no other technical alternatives.
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+
+- Disabling pack requires setting three properties. Should we aim to reduce this number?
+- Are there any concerns regarding the changes to the [ProjectReference](#project-to-project-references) protocol?
+
+### TargetFramework alias to block `/` character
+
+As previously shown in [the assets file pivots changes](#target-framework-pivots), NuGet uses `tfm/rid` as the property name in the `$/targets` object.
+Therefore, a `TargetFramework` that includes a `/` will cause parsing challenges.
+NuGet's libraries could attempt to do the split on the last `/` character.
+
+### Other TargetFramework alias naming restrictions
+
+Should we block characters that need to be encoded in JSON files, like `"`, or non-ASCII characters?
+
+Since the `TargetFramework` value is used by the .NET SDK as the directory name for build output, should we block invalid characters on Windows file systems? (Linux doesn't prevent any characters, with the possible exception of `/`, since that's used as the directory separator).
+
+### More significant assets file schema changes?
+
+Should we consider more significant schema changes in the assets file?
+
+Under `$/libraries`, each package lists every file in the package.
+However, the performance penalty of needing to parse that list every time the assets file is read may be worse than any benefit it provides from avoiding enumerating the filesystem when it is needed.
+The assets file primary purpose is to tell the .NET SDK and dotnet/NuGet.BuildTools what package assets are used by the project, and they don't need or use the `$/libraries` section at all.
+While reading the assets file might be a small percentage of the overall build process, if we consider reading the assets file in isolation, there could be a fairly significant performance increase by removing this entire section of the assets file.
+
+`$/projectFileDependencyGroups` appears to be a duplication of the information available in `$/project/frameworks`.
+
+`$/project/originalTargetFrameworks` duplicates information already available in `$/project/restore/frameworks`
+
+Since the .NET SDK has to read the assets file after NuGet writes it, time and memory allocations spent on parts of the assets file not relevant to asset selection reduces performance.
+Additionally, MSBuild binlogs embed the assets file, which bloats the binlog size.
+
+Removing the duplicated data should be non-controversial.
+But the `$/libraries` section is useful for diagnosing customer complaints.
+Whether impacting the performance of every build is worth occasional benefit in helping debug customer issues is subjective.
+An alternative is we add a `<RestoreVerboseAssetsFile>` property to include it, but otherwise have it off by default for performance reasons.
+Although the .NET SDK already have a cache, so they should only read the assets file when the assets file changes.
+
+#### Add List Lengths
+
+This is not required for this feature, but taking advantage of a breaking change in the JSON schema, we may be able to improve performance.
+When deserialized, many parts of the assets file become some kind of collection, such as a `List<T>` or a `Dictionary<Tkey, Tvalue>`.
+During deserialization, if the collection size is not known, then .NET will create a backing array with a default size, and every time the collection exceeds the current capacity, it will need to double the backing array size and copy the old array data into the new array.
+This is a known cause of performance degradation.
+This is particularly important in large collections, where the resize will happen multiple times, which includes `$/targets/*` and `$/libraries` in the assets file.
+
+I will not enumerate every location that could benefit from a count, I consider that an implementation detail that will absolutely be hidden by the Nuget.ProjectModel APIs.
+But as for some examples, I propose `$/librariesCount` to specify the number of property keys under `$/libraries`.
+`$/targets/tfm1/@count` could be the first property in `$/target/tfm1`, since `@` is an invalid character for a package ID.
+
+### More significant lock file schema changes
+
+I don't want to make the scope of this feature so large that it takes an unreasonably long time to implement.
+However, breaking changes are rarely made in NuGet, so if we have to make a breaking change to enable a feature, it's tempting to take advantage and propose other changes at the same time.
+
+Since the lock file contains the package's content hash, and the package's `nuspec` is included in that content hash, this means that the dependencies in the the lock file are a redundant duplication of information available elsewhere.
+
+## Future Possibilities
+
+- [Enable support for building for multiple RIDs, similar to how TargetFrameworks works](https://github.com/dotnet/sdk/issues/9795)
--- a/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
+++ b/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
@@ -404,6 +404,8 @@ There are no other technical alternatives.
 - Disabling pack requires setting three properties. Should we aim to reduce this number?
 - Are there any concerns regarding the changes to the [ProjectReference](#project-to-project-references) protocol?
   - Should the alias matching require the frameworks to be fully equivalent? My instinct is no, since we'd potentially like to allow the test projects to have a newer framework than the libraries.
+- Should target alias based matching be consider when applying AssetTargetFallback as well?
+Example: P1 targets net10.0 with alias apple, P2 targets net472 (apple) and net472 (banana). Should the matching allow net10.0 to match net472 (apple) or should it fail due to ambiguity?
 
 ### TargetFramework alias to block `/` character
 
--- a/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
+++ b/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
@@ -409,13 +409,11 @@ There are no other technical alternatives.
 
 As previously shown in [the assets file pivots changes](#target-framework-pivots), NuGet uses `tfm/rid` as the property name in the `$/targets` object.
 Therefore, a `TargetFramework` that includes a `/` will cause parsing challenges.
-NuGet's libraries could attempt to do the split on the last `/` character.
-
-### Other TargetFramework alias naming restrictions
-
-Should we block characters that need to be encoded in JSON files, like `"`, or non-ASCII characters?
-
+NuGet's libraries could attempt to do the split on the last `/` character, but that may lead to ambiguity.
 Since the `TargetFramework` value is used by the .NET SDK as the directory name for build output, should we block invalid characters on Windows file systems? (Linux doesn't prevent any characters, with the possible exception of `/`, since that's used as the directory separator).
+
+Note that in the current iteration, something like `banana/3` is allowed as an alias and it seems to work.
+Adding an error here would be a potential breaking change. It is very unlikely to affect people, but something that would probably need to be documented.
 
 ### More significant assets file schema changes?
 
@@ -425,6 +423,7 @@ Under `$/libraries`, each package lists every file in the package.
 However, the performance penalty of needing to parse that list every time the assets file is read may be worse than any benefit it provides from avoiding enumerating the filesystem when it is needed.
 The assets file primary purpose is to tell the .NET SDK and dotnet/NuGet.BuildTasks what package assets are used by the project, and they don't need or use the `$/libraries` section at all.
 While reading the assets file might be a small percentage of the overall build process, if we consider reading the assets file in isolation, there could be a fairly significant performance increase by removing this entire section of the assets file.
+Another notable thing is that there are known users of the libraries section currently, so this would be a disruptive change.
 
 `$/projectFileDependencyGroups` appears to be a duplication of the information available in `$/project/frameworks`.
 
@@ -461,3 +460,6 @@ Since the lock file contains the package's content hash, and the package's `nusp
 ## Future Possibilities
 
 - [Enable support for building for multiple RIDs, similar to how TargetFrameworks works](https://github.com/dotnet/sdk/issues/9795)
+- Allow for custom matching in the project reference protocol when duplicate frameworks are detected. See discussion in <https://github.com/NuGet/Home/pull/12124#discussion_r2615620013>.
+There are some scenarios, where some further control over the duplicate reference will be useful, especially for scenarios in the runtime repository, such as `net10.0-linux` test project would want to reference a `net10.0-unix` project.
+The potential complexity of the solution here is why it's something that'll be addressed after the MVP.
--- a/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
+++ b/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
@@ -81,10 +81,10 @@ Aliasing requires all the tooling (.NET SDK, Visual Studio) to be aliasing aware
 In .NET 10.0.100, you would get an error if you tried to create a project with 2 of the same framework.
 If someone tries to target duplicate frameworks using a Visual Studio version that's aliasing aware, but pins a .NET SDK version that is not aliasing aware, an error will be raised indicating that the scenario requires a tooling upgrade.
 
-The changes proposed in this document will enable the following hypothetical scenarios.
-
 Note that projects and SDKs are responsible for setting the `TargetFramework*` and `TargetPlatform*` properties, which NuGet uses as the canonical framework to use for compatibility checks.
-Therefore, the exact `TargetFrameworks` value is not important to NuGet, as it is either the SDK author, or the project owner, to set any value that is reasonable to them.
+Therefore, the exact `TargetFrameworks` value is not important to NuGet, as it is the responsibility of either the SDK author, or the project owner, to set a value that is reasonable to them.
+
+The changes proposed in this document will enable the following hypothetical scenarios.
 
 Firstly, a single project that creates multiple platform specific assemblies, all targeting the same .NET version.
 
@@ -137,7 +137,7 @@ Pack can only succeed when one of the 2 frameworks doesn't have both of the foll
 
 The pack experience likely requires some learning which may not be possible until we see more scenarios utilizing duplicate frameworks.
 
-> Note that as part of this proposal, there may be some scenarios where alias usage in per framework commands will be validated and potentially fixed. This will discovered during the implementation and won't be contained in detail in the current design.
+> Note that as part of this proposal, there may be some scenarios where alias usage in per framework commands will be validated and potentially fixed. This will be discovered during the implementation and won't be contained in detail in the current design.
 
 ### Technical explanation
 
@@ -163,7 +163,7 @@ PackageReference is also supported by non-SDK style projects, which use [dotnet/
 The changes to the assets file that affect legacy projects will be done in a non-breaking way and as such, changes should not be required there.
 NuGet will utilize the SDKAnalysisLevel property when it writes out an assets file with breaking changes, ensuring that the .NET SDK will be able to read the assets file for the build.
 
-The assets file will be updated in a bunch of location to use the alias as a key instead of the target framework, effectively allowing duplicate frameworks.
+The assets file will be updated in a bunch of locations to use the alias as a key instead of the target framework, effectively allowing duplicate frameworks.
 
 The assets file contains a `version` property since it was first created, and is always the first property in the file.
 The current assets file version is 3, and this feature will increment the version to 4.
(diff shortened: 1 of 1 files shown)


proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### TargetFramework values are aliases today
+
+NuGet currently treats `TargetFramework` as aliases. For example, the following is a valid project.
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>banana</TargetFramework>
```
Nirmal4G: May be banana is a new framework either based on .NET Core CLR or a custom FrameworkReference or even a new CLR with custom libs. This works in this case.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+
+The missing part is allowing the same framework to be targeting among these aliases.
+
+The following scenarios would be enabled:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>latestnet-linux;latestnet-mac</TargetFrameworks>
```
Nirmal4G: Here too, when the runtime team is making the new version of the framework or a fork for internal testing of runtime features or simply pointing to the latest available in the SDK, this is OK!

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+  </Project>
+```
+
+Alternatively, the VS extension scenario would be something like:
+
+```xml
+<Project Sdk="Microsoft.VisualStudio.Extensibility.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>vs17.2;vs17.3</TargetFrameworks>
```
Nirmal4G: And this is where it gets confusing. The `vs17` is neither a framework nor a runtime. This is a custom target profile based on pre-defined `TargetFramework` that doesn't change along with some support libs targeting specifically that version.

This is going to confuse a lot of people, not advanced ones (_even they could confuse themselves at first_) but mostly beginners. I know this an advanced scenario but still, this could be useful in lot of ways even for smaller open-source libraries ***targeting multiple profiles*** at the same time.

**For immediate use-cases and for advanced users, this is OK, I think!** But my takeaway here is that we need to introduce a new property (***sooner or later***) for this, I propose `TargetProfile` or `TargetBuildProfile` (_name is just an example_) and the plural form could be its multi-targeting property. I already proposed this when .NET 5 design was proposed but they wanted to keep the same.

See <https://github.com/dotnet/designs/pull/92#discussion_r389243230>.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+Normally, this works. NuGet rarely makes non-additive changes to the assets file.
+
+Sample part of an assets file:
+
+```json
+  "targets": {
+    ".NETFramework,Version=v4.7.2": {},
+    ".NETFramework,Version=v4.8.0": {}
+  },
+
```
Nirmal4G:
```suggestion
  "version": 3
  "targets": {
    ".NETFramework,Version=v4.7.2": {},
    ".NETFramework,Version=v4.8.0": {},
    "net5.0": {},
    "net6.0": {}
  },
  "projectFileDependencyGroups": {
    ".NETFramework,Version=v4.7.2": {},
    ".NETFramework,Version=v4.8.0": {},
    "net5.0": {},
    "net6.0": {}
  }
  "project": {
    "version": 1.0.0
    "frameworks": {
      "net472": {
        "targetAlias": "net472"
      },
      "net480": {
        "targetAlias": "net480"
      },
      "net5.0": {
        "targetAlias": "net5.0"
      },
      "net6.0": {
        "targetAlias": "net6.0"
      }
    }
  }
```

Updated the assets file with interesting properties that refers target framework and aliases.

Looking at this, it seems like the support is already here. Like .NET 5+ targets are already being encoded in aliases. Sure V4 is necessary but only minimal changes are required at this point.

But to bring in machine readable (_and human readable_) format to this, a huge refactor seems required! It could be a chance to bring two features at once at the cost of compat!?

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+NuGet's assets file is written pivoting on the actual target framework.
+
+To solve this, NuGet would need to:
+
+- Add vNext for the assets file format, version 4. Version 4 would either a dictionary based on the `TargetFramework` value instead of the effective target framework, built by coming `TargetFrameworkMoniker` and `TargetPlatformMoniker`.
```
Nirmal4G: You could just leave it as the value of `$(TargetFramework)` itself. You are already using aliases for .NET 5+ targets. So, Won't that work?

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+- Make a decision what would be a satisfactory behavior for combining newer/older versions of `NuGet.exe` and `.NET SDK`. This is largely just confirming some of the recent decisions.
+
+To solve this, the .NET SDK would need:
+
+- Be able to `build` with both versions 3 and version 4 of the assets file. The NuGet dependencies would naturally flow into the .NET SDK and the work is to just differentiate between the 2 different versions.
+  - The amount of work needed to be done here would likely be affected by the exact shape of the NuGet APIs. These changes may be seamless.
+
+### Restore challenges
+
+In addition to the assets file, restore has a lock file which has a similar schema, that schema would need to be amended. Fortunately when Central Package Management and its transitive pinning was introduced we introduced the concept of a PackagesLockFile version succesfully. We'd just add a version 3.
```
Nirmal4G: Is it not possible to unify the assets file, dgspec file and the lock file into two distinct verions? One for the project itself and another for the dependencies (_projects or packages_)?

This could make way for `ProjectReference` <-> `PackageReference`!

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+- Add version 3 of the packages lock file.
+
+NuGet's intra restore caching is based on the existing of a single framework. It is difficult to predict the amount of work that'd be required to implement this correctly, but it'll certainly involve public API changes to NuGet libraries. (NuGet ships libraries, NuGet.exe, VS and dotnet.exe)
+
+### Pack challenges
+
+Multi-targetted pack only really works when each build leg (TargetFramework value) has 1 target location such as `lib/net5.0`.
+Packing projects that do target the same framework multiple may be helpful in some scenarios, so completely blocking it is likely not a great idea, but having a coherent experience would require better understanding of the scenarios at question.
+
+At the minimum, the pack command should be able to detect the scenario at hand and *not succeed queitly* as there's no guarantee what gets packed.
```
Nirmal4G: One solution on top of my head is mapping folders as per the aliases from the user-defined root (`\tools\vs17.2` or `\lib\banana`) or just user-defined mapping (`\lib\net5.0\Design\vs17.2`).

```xml
<PackageFolderForTarget Include="banana" Folder="lib"/>
<PackageFolderForTarget Include="vs" Folder="tools"/>
<PackageFolderForTarget Include="vs17" Folder="lib" RelativePath="Design\<alias>"/>
<PackageFolderForTarget Include="vs17.3" Folder="lib" RelativePath="Design" FileName="<original>.<alias>.<ext>"/>
```

The last two seems un-imaginative but you get the point, right? You can't use `TargetFramework` or `Identity` here because it could include version-less and limited versioned targets too. In the example above, `vs` could match all `vsXX.X` and is overridden by the following two with similar rules.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+#### Visual Studio challenges
+
+Currently the PM UI does not have any support for multi targeting. This isn't anything new, but it can affect the experience of customers moving to SDK projects for the first time.
+
+Another concern are all the APIs for displaying the dependencies, transitive dependencies and Get Installed Packages APIs in NuGet.VisualStudio.Contracts.
+All of these APIs may require an incremental addition.
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
```
Nirmal4G: It's pain to maintain and support. 😂

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,220 @@
...
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
```
Nirmal4G: Unifying all those json files along with providing a solution for <https://github.com/NuGet/Home/issues/11449> could be side-effect of this refactor.

Nirmal4G (commented):
Some ideas and suggestions. Feel free to ignore it. 😅🙃

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Note that I'm  not really suggesting `latestnet-linux` becomes a magic work. I'm just trying to communicate that you can put whatever makes to you as the user there.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: I think my doc needs work as it probably reads as I'm suggesting the values here. It is really up to the project author/SDK author to supply values that imply the correct thing from what they are attempting to do. 

The idea is that NuGet and other .NET build component do not need to interpret that value

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: It's unfortunately not minimal. 

The key is the NuGetFramework.GetShortFolderName, so we need to tweak that.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: We want to use the value of `TargetFramework` as the key, but it's not done everywhere. I'll update the doc to more clearly call that out.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: The 3 files have different purposes today. 

dgspec -> Internal NuGet generated and NuGet consumed file. Used as a cache to enable working with unloaded VS projects and something that contains all the inputs affecting restore. Think of this as the restore input, a intepretation of the csproj that affects restore.
assets file -> Nuget/SDK build contract. Think of this as the restore output.
packages.lock.json - This is the lock files that's something in addition to the csproj file.


I'm not sure we'd need to unify the files themselves, but where approrpiate our reader/writer code is the same.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: I think we need to differentiate between something being part of the NuGet package contract and being part of the build contract. 

I'd imagine any solution for would likely need some way to specify the target folders, the items etc as you suggested. Not sure what they looks like just yet, but as we iterate on this we can update.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: I think in a way, the future possibilities are the same as the motivation part. 

This feature is likely not going to be used in defaults often, but it gives *a lot* of flexibility to power users.

nkolev92:
Thanks for taking a look @Nirmal4G 

For context, this is design I'm still working on, I'd say it's at about 50%.

Hoping to add more details about the scenarios and just in general provide more concrete proposals soon.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
Nirmal4G: Hypothetical scenario. I'm talking as an example usage. Not to be taken literally. High 10-foot point of view of the feature. In Design terms.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
Nirmal4G: Hypothetical scenario. I'm talking as an example usage. Not to be taken literally. High 10-foot point of view of the feature. In Design terms.

Nirmal4G:
@nkolev92 These are some of my observations and preliminary suggestions, ideas and takeaways from this draft. I'm not taking this as a concrete proposal. I'm just viewing this from a 10-foot view (_or design perspective_) of the proposal to see how it could impact normal as well as advanced users. In that mindset, read through my comments again. You would then see what I was talking about.

Nirmal4G:
> Hoping to add more details about the scenarios and just in general provide more concrete proposals soon.

Sure, Will look though once you mark this ready for review. Sorry for the uninvited comments. I'll stop now.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
Nirmal4G: I meant not in the source code that handles the Json file but the schema itself.

nkolev92:
> > Hoping to add more details about the scenarios and just in general provide more concrete proposals soon.
> 
> Sure, Will look though once you mark this ready for review. Sorry for the uninvited comments. I'll stop now.

Oh no worries. I appreciate the feedback as this is a large doc so getting feedback early on is always great.

My comment was more about that fact that I'll be adding even more info to the doc, setting expectations about when something like this would even be close to happening :)

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
Nirmal4G: I know what purpose they serve. I just asked, technically, would it be possible to separate a project state and dependency states?

The way I think about packages is something you load it externally since you won't have it in source-control. And what the packages provide are the package assets that projects consume. That is also a project state. It doesn't need to know that an asset comes from a project or a package. The dependency state is where all your network laden or external resources you refer to. The project state is created from the project but depends on the dependency state to resolve the assets completely.

In this way you could refer to a package, a project or even a project from an external repo which I think is one of the scenarios that people asked. I forgot what issue it is.

nkolev92: I think that's orthogonal to this work. 

For the purposes of the project-package duality, you only really care about project file + assets file. You can ignore the lock file as it is about repeatability, and the dgspec is just a summary of what Nuget wants from the project file.

JonDouglas:
@ViktorHofer & @ericstj

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: Hey, I'm taking over this spec design. I don't understand what might be actionable based on this comment, and it's an old comment now, so I'm resolving it. If you think the spec should change somehow, please try re-explaining your comment in a different way.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: Similar to the previous comment, I don't see anything actionable here, so I'm resolving this. If this thread is proposing a change to the spec, please help me understand what the proposal is.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: While [msbuild static graph](https://github.com/dotnet/msbuild/blob/main/documentation/specs/static-graph.md) introduces MSBuild properties `InnerBuildProperty` and `InnerBuildPropertyValues`, so in theory the VSIX SDK could use `<TargetVS>vs17.8;vs16.11</TargetVS>`, the [msbuild reserved and well known properties docs](https://learn.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2022) explicitly states  that this is only applicable for graph builds. Therefore, this doesn't appear to be usable general feature (for example, `dotnet build` without the `-graphBuild` argument).

Additionally, many dotnet CLI commands already have a `--framework` argument on many commands, like `dotnet build -f net8.0`, `dotnet publish -f net8.0`, `dotnet test -f net8.0`.  Even if `<TargetFramework>` isn't a perfect fit for things like platforms/runtimes/VS versions, by using `TargetFrameworks`, all these commands will probably work without modification. On the other hand, if we use some new MSBuild property instead, then there will be a bunch of questions about what CLI arguments should be used when one inner build is desired, what to name the argument, etc.  Then it will be confusing why ProjectA needs a different CLI argument to build/test/whatever, compared to ProjectB.

So, re-using `TargetFrameworks` might not be ideal, but it simplifies several things, so from a practical point of view, I feel like it's better than any alternative.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: I'm resolving this, as the line before the code snippet says "part of an assets file". To me this makes it clear that

1. this is not a complete assets file
2. It's intended to highlight something specific, and the highlighted point is easily missed when there's too much other content, such as in a complete assets file.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: I'm resolving this conversation as out of scope for this spec. While I've seen multiple customers assume that lock files can remove the need for a restore, or request nuget consolidate into fewer files, we need to be able to ship this spec's feature in a reasonable time, and trying to expand the scope will make finishing the implement much slower.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: When I update the spec, I'm going to make pack, where the resolved `TargetFrameworkMoniker` and `TargetPlatformMoniker` is not unique as a future possibility, out of scope of this spec.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: I don't think unifying the json files is related to this spec, except that both require changes to the json file schema. Neither change would bring the other one closer to implementation. So, I'm resolving this conversation as out of scope.

proposed/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,289 @@
...
+Should we "simply" replace the TFM in property keys with the Target Framework alias?
+For example, change ".NETFramework,Version=4.7.2/win-x64" with "vs17.7/win-x64"?
+But there will be parsing problems when a `TargetFramework` contains a `/` character.
+While "something/another thing/win-x64" can determine that "win-x64" is a RID, when we get the RID-less "something/another thing" string, it's not clear that what follows the `/` is a RID, unless the assets file parser starts knowing the entire RID graph.
+
+### How ProjectReferences will work
+
+Currently ProjectReferences (P2P) do a compatibility check, both by MSBuild and NuGet (although I'm sure MSBuild delegates to NuGet APIs).
+When a project has multiple `TargetFramework`s that map to a single "NuGet Framework", which one should MSBuild and NuGet use?
+MSBuild needs to choose in order to know which dll to pass to the compiler (in case different aliases have different APIs), and so test projects actualy test the desired product binary.
```
zivkan: Turns out MSBuild asks NuGet which `TargetFramework` to use: https://github.com/dotnet/msbuild/blob/25fdeb3c8c2608f248faec3d6d37733ae144bbbb/documentation/ProjectReference-Protocol.md?plain=1#L62

Therefore, it appears to be in Nuget's control to determine how all of this works. Although, since it affects how the end to end build experience will be (particularly projects with unit tests), we need to include partners in discussions on the final design.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+This feature will require the following changes to the asset file:
+
+1. [Increment Version Number](#assets-file-version)
+1. [Target Framework Pivots](#target-framework-pivots)
+1. [Add List Lengths](#add-list-lengths)
+
+However, PackageReference is also supported by non-SDK style projects, which use [dotnet/NuGet.BuildTools](https://github.com/NuGet.BuildTools) which, despite the name, is owned by the non-SDK project system team, not NuGet.
+Their implementation parses the JSON file directly, rather than using NuGet.ProjectModel.
+Therefore, if we can implement the assets file changes without breaking changes to the NuGet.ProjectModel APIs, the .NET SDK may "just work" when the newer version of NuGet is inserted into the .NET SDK.
+However, the non-SDK project system will fail.
```
dsplaisted: If we say that it's a non-goal to support TargetFramework aliases from non SDK-style projects (and hence from NuGet.BuildTools), does that simplify things?  NuGet.BuildTools might generate an error if it encounters a newer assets file format without any changes to it, so we might not need as much synchronization.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+The multi-rep synchronization insertion would have a lot of risk and difficulty, so allowing NuGet.ProjectModel to support multiple versions of the assets file schema seems the better choice.
+
+There are still two ideas for how this can be implemented.
+
+1. Assets file version as a restore input
+
+   NuGet could require read a `RestoreAssetsFileVersion` property on each project, defaulting to 3 when not specified.
+   This allows project systems that have not yet adapted to the new assets file schema to keep working as before.
+
+1. Use the newer assets file automatically when a project multi-targets
```
dsplaisted: Could we use the new format only when the project multi-targets and there are multiple aliases that map to the same TargetFrameworkMoniker/TargetPlatformMoniker combination?  That might avoid existing projects breaking with older tools.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+### TargetFramework alias to block `/` character
```
dsplaisted, line 420: I think we should try to support `/` in the TargetFramewrok alias.  One of the scenarios we hope to eventually support is multi-targeting over RuntimeIdentifiers.  It would be natural to express that with the `/` character, for example `<TargetFrameworks>net8.0-windows10.0.19041.0/x64;net8.0-ios/arm64;net8.0-android/arm64</TargetFrameworks>`.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+All of these APIs may require an incremental addition.
+
+### Project to Project references
+
+MSBuild has a [`ProjectReference` protocol](https://github.com/dotnet/msbuild/blob/25fdeb3c8c2608f248faec3d6d37733ae144bbbb/documentation/ProjectReference-Protocol.md), which explains how `ProjectReferences` work.
+MSBuild calls the `GetReferenceNearestTargetFrameworkTask`, which NuGet implements, in order to obtain best TFM matching.
+
+Currently `GetReferenceNearestTargetFrameworkTask` is implemented by running NuGet's "nearest match" algorithm.
+This feature would extend it to first try nearest match, and if there is more than one matching `TargetFramework`, then look for a `TargetFramework` that has an exact name match in both projects.
+If there are more than one "nearest" framework, but no exact name match, then for the first version of this feature, NuGet will report an error.
```
ericstj, line 358: I'm interpreting this as supporting a case where you have a project that uses aliases referencing a project that's not using alias.

So if I have project 1 with TFs: `net6.0-a;net6.0-b` (where both are aliases for `net6.0`) and it referenced project 2 with TFs: `net6.0;netstandard2.0` project1 could reference project 2 and select `net6.0` because there is a single nearest framework.

If that's true this is a great middle ground for enabling aliases in a way that isn't viral.

ViktorHofer: The spec would benefit from an example to make the intent clear. I agree with your observation. Here are other examples that would work or not work with this algorithm:

### Example 1
project1 `net6.0-a;net6.0-b` references project2 that references the same TFMs. Even though the `GetReferenceNearestTargetFrameworkTask` algorithm returns two matches, there's a single direct name match so it would choose `net6.0-a` from project2 for `net6.0-a` from project1 and same for `net6.0-b`.

### Example 2
project1 `net6.0` references project2 that references `net6.0-a;net6.0-b`. The `GetReferenceNearestTargetFrameworkTask` would algorithm returns two matches, there's no direct name match so the algorithm would fail and NuGet would report an error.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+
+### Extending Pack
+
+It seems feasible that if a single project can target multiple versions of Visual Studio, through a new Visual Studio Extensibility SDK, that someone will also want to create a package that supports multiple versions of Visual Studio.
+
+Designing how this could work is out of scope for this spec, and can be introduced in a different spec.
```
ViktorHofer: Just writing down what we discussed in yesterday's meeting. For runtime to not be broken by this change, we need to keep supporting what works today.

In runtime, we set `IncludeBuildOutput=false` and `SuppressDependenciesWhenPacking=true` for custom TFMs (one that use a TPM not known by the SDK -> i.e. `net8.0-unix`).

Now if NuGet starts emitting an error for TFM aliases it would break runtime. It would be good to only emit an error when the TFM is included in the package (either the build output or the dependencies section). So in technical terms, only emit an error when `TFM == alias && (IncludeBuildOutput == true or SuppressDependenciesWhenPackaging == false)`

zivkan:
oops, vacation and other priorities has kept this feature on the backburner, but it's still planned.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>linux;mac</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <TargetFrameworkIdentifier>.NETCoreApp</TargetFrameworkIdentifier>
+    <TargetFrameworkVersion>v6.0</TargetFrameworkVersion>
+    <TargetFrameworkMoniker>.NETCoreApp,Version=v6.0</TargetFrameworkMoniker>
+
+    <DefineConstants>$(DefineConstants);$(TargetFramework)</>
```
jviau: small nit:

```suggestion
    <DefineConstants>$(DefineConstants);$(TargetFramework)</DefineConstants>
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,397 @@
...
+Should we block characters that need to be encoded in JSON files, like `"`, or non-ASCII characters?
+
+Since the `TargetFramework` value is used by the .NET SDK as the directory name for build output, should we block invalid characters on Windows file systems? (Linux doesn't prevent any characters, with the possible exception of `/`, since that's used as the directory separator).
+
+### More significant assets file schema changes?
+
+Should we consider more significant schema changes in the assets file?
+
+Under `$/libraries`, each package lists every file in the package.
+However, the performance penalty of needing to parse that list every time the assets file is read may be worse than any benefit it provides from avoiding enumerating the filesystem when it is needed.
```
jviau, line 436: For a functions extensibility scenario, I am considering having extension packages include a well-known marker file in their nupkg. I was going to leverage the fact that the assets file already has all nupkg file contents listed in it to speed up that process.

With this change, is the suggestion I would need to manually scan the file system of every nuget package? Or do you have an idea of a more general way I can add metadata to a nuget package and read it during a build?

Today we have an assembly attribute and we use Mono.Cecil to scan runtime assemblies from nuget looking for that attribute. This is understandably not performant. Which is why I am considering the marker file approach. I wouldn't block on this change just for our use case, but some metadata system for nupkgs that can be read at build time would be helpful (but obviously that would be an entirely different proposal).

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Just catching up here. 
 We won't need to worry about legacy projects as the changes can be done in a non-breaking way.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: this should be doable.
I'll add it to the spec.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: This has been included in the newest iteration of the spec.

accepted/2022/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: We've learned a lot since this comment :)
We will just use SDKAnalysisLevel to drive the format. If an SDK version aware of the new syntax is used, then NuGet would use V4, otherwise, V3. 

Legacy projects will always use V3.

nkolev92:
@ericstj @ViktorHofer @dsplaisted @baronfel @zivkan @jeffkl 

This is ready for review. 

It's a complicated, but I'd love to get through it async as much as possible because of the length itself :D

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+All of these APIs may require an incremental addition.
+
+### Project to Project references
+
+MSBuild has a [`ProjectReference` protocol](https://github.com/dotnet/msbuild/blob/25fdeb3c8c2608f248faec3d6d37733ae144bbbb/documentation/ProjectReference-Protocol.md), which explains how `ProjectReferences` work.
+MSBuild calls the `GetReferenceNearestTargetFrameworkTask`, which NuGet implements, in order to obtain best TFM matching.
+
+Currently `GetReferenceNearestTargetFrameworkTask` is implemented by running NuGet's "nearest match" algorithm.
+This feature would extend it to first try nearest match, and if there is more than one matching `TargetFramework`, then look for a `TargetFramework` that has an exact name match in both projects.
+If there are more than one "nearest" framework, but no exact name match, then for the first version of this feature, NuGet will report an error.
```
dsplaisted, line 358: How hard would it be to have some sort of extensibility point so that you can add your own logic to choose the "nearest" framework if the NuGet logic results in more than one match?

dsplaisted (dismissed):

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+
+The changes listed are largely what NuGet needs to implement in code we own, but there will be impacts to other components as well.
+In particular the .NET SDK may need some changes to be able to use the assets file correctly.
+Majority of this had fortunately been done in past iterations and the .NET SDK currently depends on are NuGet APIs.
+Other tools, like various Visual Studio components, might have assumptions about TFM uniqueness in the project, which this feature changes.
+Therefore, while the NuGet team works very closely with the .NET SDK team and we will ensure that building projects work, there will increased risk of other features and tools breaking, which would normally not be affected by NuGet changes.
+
+### Restore output (assets file) changes
+
+In PackageReference, NuGet has a contract with the .NET SDK where NuGet writes the assets file and the .NET SDK consumes it.
```
zivkan:
```suggestion
In PackageReference, NuGet has a contract with the .NET SDK where NuGet writes the assets file and the .NET SDK reads it.
```

nitpick: produce-consume, or read-write to use the same metaphore in a single sentence.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+The .NET SDK uses NuGet libraries for the parsing of this file and as such, there'll be no code on the .NET SDK to consume it, but rather all the scenarios can be handled within the different NuGet code.
```
zivkan:
```suggestion
The .NET SDK uses NuGet libraries to parse this file, and as such, as long as NuGet's APIs don't change, the .NET SDK will not need any code changes to read the assets file's new changes.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+
+PackageReference is also supported by non-SDK style projects, which use [dotnet/NuGet.BuildTools](https://github.com/NuGet.BuildTools) which, despite the name, is owned by the non-SDK project system team, not NuGet.
+The changes to the assets file that affect legacy projects will be done in a non-breaking way and as such, changes should not be required there.
+NuGet will utilize the SDKAnalysisLevel property when it writes out an assets file with breaking changes, ensuring that the .NET SDK will be able to read the assets file for the build.
```
zivkan, line 172: Customers are expected to change SDKAnalysisLevel if they have build errors when upgrading to a newer SDK.  I can't remember the name of the property, but there's an SDK version that feels more appropriate for what this is aiming to achieve.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+It is a part of the assets file, but it is also used for the dgspec and the no-op scenarios.
+
+The PackageSpec is not versioned and at this point, I'm not proposing adding a version.
+
+The changes being made here can be divided into 2 parts.
+
+- Additive changes
+- Breaking changes
+
+The additive changes will be an always thing. Every newly serialized PackageSpec will make this change. This should not affect any of the existing scenarios, because at worse the new property is not parsed.
```
zivkan:
```suggestion
The additive changes will be an always thing. Every newly serialized PackageSpec will make this change.
This should not affect any of the existing scenarios, because at worse the new property is not parsed.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+NuGet would add version 3 of the packages lock file, which will [pivot on the target framework alias, just as assets files will](#target-framework-pivots).
+This lock file has a single dependencies section and the key there would be changed to the alias instead.
+
+From:
+
+```json
+{
+  "version": 1,
+  "dependencies": {
+    ".NETCoreApp,Version=v10.0": {
```
zivkan:
```suggestion
    ".NETCoreApp,Version=v8.0": {
```

the example above says the alias is production and the TFM is net8.0

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+}
+```
+
+To:
+
+```json
+{
+  "version": 3,
+  "dependencies": {
+    "net10.0": {
```
zivkan:
```suggestion
    "production": {
```

Line 305 says the alias is going to be used

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+
+- Disabling pack requires setting three properties. Should we aim to reduce this number?
+- Are there any concerns regarding the changes to the [ProjectReference](#project-to-project-references) protocol?
+  - Should the alias matching require the frameworks to be fully equivalent? My instinct is no, since we'd potentially like to allow the test projects to have a newer framework than the libraries.
+
+### TargetFramework alias to block `/` character
+
+As previously shown in [the assets file pivots changes](#target-framework-pivots), NuGet uses `tfm/rid` as the property name in the `$/targets` object.
+Therefore, a `TargetFramework` that includes a `/` will cause parsing challenges.
+NuGet's libraries could attempt to do the split on the last `/` character.
```
zivkan: Considering the assets file has an entry for the ridless target, in addition to the per-rid targets, I think it would be challening to tell the difference between the two.

If the parser keeps raw strings in memory and then parses the alias/rid only once all targets are parsed from json, then you can find what the max number of `/` are, and assume the ridless has n-1.

But I think it's easier if we fail restore if the alias has a `/`. It's easier to block first and later allow, than the other way around.

Since we're already incrementing the version number, I think it'd be ideal if we could make another breaking change to the format, so instead of

```json
"targets": {
   "alias" : {  },
   "alias/rid": {  }
}
```

we could instead have

```json
"targets": {
   "alias": {
     "": { },
     "rid": {  }
  }
}
```

But I suspect this might be a lot harder to implement.


However, as is pointed out just below, the target alias is used as the directory name for the obj and bin directories, and `/` is a directory separator, so I think that's another reason we could just block `/` from the alias

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
ViktorHofer: To add to that question, in the "No match found" case, would it make sense to return that information without raising an error / warning so that we could allow for customization in the msbuild target instead of in the task (which would give us much more flexibility)?

We still want to raise an error if there's no customization (I'm trying to avoid the word "fallback" here as it has a different meaning for NuGet). That would probably need to be moved outside of the task, into the target.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,463 @@
...
+In particular the .NET SDK may need some changes to be able to use the assets file correctly.
+Majority of this had fortunately been done in past iterations and the .NET SDK currently depends on are NuGet APIs.
+Other tools, like various Visual Studio components, might have assumptions about TFM uniqueness in the project, which this feature changes.
+Therefore, while the NuGet team works very closely with the .NET SDK team and we will ensure that building projects work, there will increased risk of other features and tools breaking, which would normally not be affected by NuGet changes.
+
+### Restore output (assets file) changes
+
+In PackageReference, NuGet has a contract with the .NET SDK where NuGet writes the assets file and the .NET SDK reads it.
+The .NET SDK uses NuGet libraries to parse this file, and as such, as long as NuGet's APIs don't change, the .NET SDK will not need any code changes.
+PackageReference is also supported by non-SDK style projects, which use [dotnet/NuGet.BuildTools](https://github.com/NuGet.BuildTools) which, despite the name, is owned by the non-SDK project system team, not NuGet.
```
KalleOlaviNiemitalo: <https://github.com/NuGet.BuildTools> gives me 406 Not Acceptable.  Is that the correct link?

KalleOlaviNiemitalo: <https://github.com/dotnet/NuGet.BuildTools> on the other hand gives 404 Not Found.

KalleOlaviNiemitalo: <https://github.com/dotnet/NuGet.BuildTasks> exists; did you mean that?

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: > How hard would it be to have some sort of extensibility point so that you can add your own logic to choose the "nearest" framework if the NuGet logic results in more than one match?

It's not impossible. I'd love to see any usage of this multiple framework project reference scenario before we want that direction though. 

> To add to that question, in the "No match found" case, would it make sense to return that information without raising an error / warning so that we could allow for customization in the msbuild target instead of in the task (which would give us much more flexibility)?
 
Can you help me understand how this could help?
If there's no nearest match/restore/build should be failing.
I see this as equivalent to a nearest framework not matching at all.

> We still want to raise an error if there's no customization (I'm trying to avoid the word "fallback" here as it has a different meaning for NuGet). That would probably need to be moved outside of the task, into the target.

The target that calls the task does not run during restore. 
In the core restore algorithm, NuGet picks the nearest project ref and that's based on the data provided to NuGet and collected at the beginning of restore. 
Effectively, NuGet does the same thing internally when choosing to flow the package references coming through projects.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Yeah, there's an SDK version property for sure, we raise telemetry on that, but I see SDKAnalysisLevel to mean, behave like the previous SDK version, and to me that'd imply the older restore behavior in general.

I don't feel too strongly though, and can easily be convinced otherwise.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Yeah, I'd be fine with blocking that yeah. 
Honestly, I can't image this to be a super critical scenario that would need to be addressed. 
The more changes we make, the higher the risk of the change anyways.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: > https://github.com/dotnet/NuGet.BuildTasks exists; did you mean that?

That's exactly what I meant. 
I legit don't know how I ended up with that link haha. 

Thanks!

nkolev92: Fixed now.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
zivkan: In my mind, SdkAnalysisLevel is about warnings and errors (diagnostics), not features, and looking at the original spec, I feel my opinion is reinforced: https://github.com/dotnet/designs/blob/main/proposed/sdk-analysis-level.md

The spec doesn't say it, but I always thought SdkAnalysisLevel was inspired by Roslyn's AnalysisLevel, and Roslyn also has LangVersion which I consider to be the feature toggle. The SDK/NuGet doesn't have a "global feature toggle" like Roslyn does with LangVersion.  Using SdkAnalysisLevel is a choice that can be made, but then customers lose the ability to enable new features while disabling new diagnostics. I guess we need to decide whether that's an important distinction to allow customers to have.

nkolev92: I think Chet has a good compass when it comes to SDKAnalysisLevel. 

Curious what you think @baronfel

tldr; 
NuGet will introduce a V4 of the assets file. 
It requires a new .NET SDK reader, ie a specific .NET SDK version. 

Should that "switch" be based off of the SDKAnalysisLevel or NETCoreSdkVersion property.
More context on the pros and cons in the above comments.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
ViktorHofer: I see, fair point. I was basically wondering if we can add an extension point / a callback to provide additional information that would allow us to extend the selection algorithm for the multiple or non found cases.

Runtime today overwrites the entire selection algorithm target (msbuild) and task (nuget) due to not an extension point being available. This isn't great and needs an update every time the P2P protocol is touched and the msbuild target gets updated.

nkolev92: I know you're not focusing on that right now, but what are some practical examples where the runtime needs to overwrite the selection.  
I was hoping alias matching would address a bunch of those needs.

ViktorHofer: RID based selection. I.e. `net10.0-linux` should be able to reference `net10.0-unix` (as unix is a parent of linux in the graph). The current algorithm as you describe in the doc wouldn't be able to handle that but it would support direct matches, i.e. `net10.0-unix` referencing `net10.0-unix`.

We can probably workaround that limitation if necessary or even change the RIDs that we target in TFMs but yeah...

nkolev92: I see. That's the next step then yeah.
Rids are a cross product, but maybe it'll be possible to add first class support for nearest that accounts for the RID as well.

dsplaisted: I probably wouldn't try to include RID matching in NuGet or the SDK for the first version of this, ideally just an extensibility point so that people could add it.

nkolev92: I think in V1, we can just add alias matching support, but I'll create a follow-up for the extensibility point. 
Right now, the current technical implementation does not allow a computational extensibility.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: @baronfel attempted to post the below, but was facing GH problems: 

<hr/>

The intent of SdkAnalysisLevel is to easily apply behaviors to all kinds of SDK-level analysis - not just Roslyn Analyzers, but built-in diagnostics, MSBuild BuildChecks, and similar. Any kind of validation that we do during a build was in scope.  
 
This has evolved slightly to controlling _behaviors_ as well - I am entirely supportive of this. The main reason for this evolution/broadening of meaning is that the user-facing view of this property is often "make the current SDK _behave like_ the SDK band I tell you to". With this lens, applying SdkAnalysisLevel to _behaviors_ in addition to validations is entirely appropriate.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: I added this in future possibilities. 

As I finalize the MVP, I'll think of ways we can expose this that's not overly complex.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+
+- Disabling pack requires setting three properties. Should we aim to reduce this number?
+- Are there any concerns regarding the changes to the [ProjectReference](#project-to-project-references) protocol?
+  - Should the alias matching require the frameworks to be fully equivalent? My instinct is no, since we'd potentially like to allow the test projects to have a newer framework than the libraries.
+- Should target alias based matching be consider when applying AssetTargetFallback as well?
```
donnie-msft:
```suggestion
- Should target alias based matching be considered when applying AssetTargetFallback as well?
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+Example: P1 targets net10.0 with alias apple, P2 targets net472 (apple) and net472 (banana). Should the matching allow net10.0 to match net472 (apple) or should it fail due to ambiguity?
```
donnie-msft: nit
```suggestion
Example: P1 targets net10.0 with alias apple, P2 targets net472 (apple) and net472 (banana).
Should the matching allow net10.0 to match net472 (apple) or should it fail due to ambiguity?
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+
+### TargetFramework alias to block `/` character
+
+As previously shown in [the assets file pivots changes](#target-framework-pivots), NuGet uses `tfm/rid` as the property name in the `$/targets` object.
+Therefore, a `TargetFramework` that includes a `/` will cause parsing challenges.
+NuGet's libraries could attempt to do the split on the last `/` character, but that may lead to ambiguity.
+Since the `TargetFramework` value is used by the .NET SDK as the directory name for build output, should we block invalid characters on Windows file systems? (Linux doesn't prevent any characters, with the possible exception of `/`, since that's used as the directory separator).
+
+Note that in the current iteration, something like `banana/3` is allowed as an alias and it seems to work.
+Adding an error here would be a potential breaking change. It is very unlikely to affect people, but something that would probably need to be documented.
```
donnie-msft: nit
```suggestion
Adding an error here would be a potential breaking change.
It is very unlikely to affect people, but something that would probably need to be documented.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+- `dotnet list package`.
+- Numerous features in Visual Studio.
+  - Test Explorer.
+  - Text editor (left-most dropdown, for TFM specific Intellisense).
+  - Solution Explorer's Dependencies node.
+
+### Lock file changes
+
+NuGet's "repeatable build" feature adds a package lock file to the project directory.
+The restore has a lock file has a similar schema to the assets file, so that schema would also need to be amended.
```
donnie-msft:
```suggestion
The restore lock file has a similar schema to the assets file, so that schema would also need to be amended.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+Fortunately when Central Package Management and its transitive pinning was introduced we introduced the concept of a PackagesLockFile version successfully.
```
donnie-msft: nit
```suggestion
Fortunately, when Central Package Management and its transitive pinning was introduced, we introduced the concept of a PackagesLockFile version successfully.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+
+Only .NET SDK projects will support aliasing, as they are the only ones that support multiple frameworks.
+Aliasing requires all the tooling (.NET SDK, Visual Studio) to be aliasing aware, since there are some breaking changes in the files that these components depend on.
+In .NET 10.0.100, you would get an error if you tried to create a project with 2 of the same framework.
+If someone tries to target duplicate frameworks using a Visual Studio version that's aliasing aware, but pins a .NET SDK version that is not aliasing aware, an error will be raised indicating that the scenario requires a tooling upgrade.
+
+The changes proposed in this document will enable the following hypothetical scenarios.
+
+Note that projects and SDKs are responsible for setting the `TargetFramework*` and `TargetPlatform*` properties, which NuGet uses as the canonical framework to use for compatibility checks.
+Therefore, the exact `TargetFrameworks` value is not important to NuGet, as it is either the SDK author, or the project owner, to set any value that is reasonable to them.
```
donnie-msft:
```suggestion
Therefore, the exact `TargetFrameworks` value is not important to NuGet, as it is the responsibility of either the SDK author, or the project owner, to set any value that is reasonable to them.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+There is no intuitive solution to the pack scenario.
+Given a project with multiple versions of the same framework, pack will fail by default.
+
+NuGet will hinge off off IncludeBuildOutput, IncludeContentInPack and SuppressDependenciesWhenPacking to decide to skip packing for a framework, effectively allowing a framework to be completely excluded.
+
+Pack can only succeed when one of the 2 frameworks doesn't have both of the following `IncludeContentInPack == true or IncludeBuildOutput == true or SuppressDependenciesWhenPackaging == false`.
+
+The pack experience likely requires some learning which may not be possible until we see more scenarios utilizing duplicate frameworks.
+
+> Note that as part of this proposal, there may be some scenarios where alias usage in per framework commands will be validated and potentially fixed. This will discovered during the implementation and won't be contained in detail in the current design.
```
donnie-msft:
```suggestion
> Note that as part of this proposal, there may be some scenarios where alias usage in per framework commands will be validated and potentially fixed. This will be discovered during the implementation and won't be contained in detail in the current design.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+
+### Restore output (assets file) changes
+
+In PackageReference, NuGet has a contract with the .NET SDK where NuGet writes the assets file and the .NET SDK reads it.
+The .NET SDK uses NuGet libraries to parse this file, and as such, as long as NuGet's APIs don't change, the .NET SDK will not need any code changes.
+PackageReference is also supported by non-SDK style projects, which use [dotnet/NuGet.BuildTasks](https://github.com/dotnet/NuGet.BuildTasks) which, despite the name, is owned by the non-SDK project system team, not NuGet.
+The changes to the assets file that affect legacy projects will be done in a non-breaking way and as such, changes should not be required there.
+NuGet will utilize the SDKAnalysisLevel property when it writes out an assets file with breaking changes, ensuring that the .NET SDK will be able to read the assets file for the build.
+
+The assets file will be updated in a bunch of location to use the alias as a key instead of the target framework, effectively allowing duplicate frameworks.
```
donnie-msft:
```suggestion
The assets file will be updated in a bunch of locations to use the alias as a key instead of the target framework, effectively allowing duplicate frameworks.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+```diff
+-"version": 3
++"version": 4
+```
+
+By default for non-SDK style projects, we will always use version 3.
+Non-SDK style project only have a single target, so those project types won't see a change.
+
+For simplicity of implementation, whenever SDKAnalysisLevel is set to 10.0.300 or more, NuGet will use the V4 assets file.
+Whenever the SDKAnalysisLevel is set to a lower version, then NuGet will use the V3 assets file.
```
donnie-msft, line 189: And this ages out after 3 major releases, based on how SDKAnalysisLevel works, right? 
In other words, there's ample time (years?) for a customer to move a solution to version 4.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+The same considerations extend to NuGet.exe restore, which will also be dependent on the SDKAnalysisLevel property.
+
+#### Target Framework Pivots
+
+There are multiple places in the assets file where the "NuGet Framework" is used as a JSON object property key.
+There are 3 top level sections where changes will be needed. The `targets`, `projectFileDependencyGroups` and `project` sections.
+
+Take a project using `<TargetFramework>production</TargetFramework>`, `<TargetFrameworkMoniker>.NETCoreApp,Version=v8.0</TargetFrameworkMoniker>`, and `<RuntimeIdentifiers>linux-x64</RuntimeIdentifiers>`.
+
+Note that the effective framework itself will not be part of the serialized part, but it'll cleaned up by the reader by looking up the target framework based off of the "project" section of the assets file.
```
donnie-msft:
```suggestion
Note that the effective framework itself will not be part of the serialized part, but it'll be cleaned up by the reader by looking up the target framework based off of the "project" section of the assets file.
```

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+
+- Additive changes
+- Breaking changes
+
+The additive changes will be an always thing. Every newly serialized PackageSpec will make this change. 
+This should not affect any of the existing scenarios, because at worse the new property is not parsed.
+
+The following 2 changes will be made on the writer side:
+
+- Add "framework" to the packageSpec, while keeping targetAlias for simplicity sake.
```
donnie-msft: As a reader not as familiar with `targetAlias`, I'm not understanding how this is simple. Why would I use `targetAlias` and `framework`?

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+This is an additive change and will be written out for every new package spec for simplicity.
+- Change the key from framework to alias. In most cases, really almost all non-alias cases, this is likely to be the same and cause no problems.
+The only default cases where the alias and the effective tfm are different are MAUI and platform scenarios.
+
+1. `$/project/restore/frameworks`
+
+    ```diff
+    "frameworks": {
+    -  "net8.0": {
+    -    "targetAlias": "production",
```
donnie-msft, line 252: Is there a scenario where `targetAlias` can be selecting 1 of multiple aliases? Let's say I have "production" and "staging" and "development", can I then set `targetAlias` to any of those three?

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+      +    "framework": "net8.0",
+      +    "targetAlias": "production",
+        }
+      }
+    }
+    ```
+  
+Given that there is no version for the package spec at this point, we can do a simple trick to infer the type of PackageSpec to read.
+
+- When reading a framework, if both targetAlias and framework, then this is equivalent to V4 and a .NET SDK scenario
```
donnie-msft: Could this be simplified from the textual scenarios into a small matrix of scenarios?

donnie-msft (commented):
I had a couple questions, and left a few nits. 
Suggest having Copilot review spelling/etc and also telling it to put each sentence on a new line in the markdown, as I know Andy prefers that.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Yes, it allows plenty of time to move to newer SDKs. 
By that time, all .NET SDKs in support will support v4.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: The simplicity I am referring to is from an implementation perspective. 
It would allow us not to version the "PackageSpec", but just the assets file. 

If we drop the targetAlias, we'll need to somehow version the package spec to tell the writer what to set as the TargetAlias.

Hopefully the implementation PR clarifies this.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Who's the setting them in this case? 
I'm not sure I understand the scenario. 

This is mostly just technical details for us.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: I'll push a follow-up to this.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+But as for some examples, I propose `$/librariesCount` to specify the number of property keys under `$/libraries`.
+`$/targets/tfm1/@count` could be the first property in `$/target/tfm1`, since `@` is an invalid character for a package ID.
+
+### More significant lock file schema changes
+
+I don't want to make the scope of this feature so large that it takes an unreasonably long time to implement.
+However, breaking changes are rarely made in NuGet, so if we have to make a breaking change to enable a feature, it's tempting to take advantage and propose other changes at the same time.
+
+Since the lock file contains the package's content hash, and the package's `nuspec` is included in that content hash, this means that the dependencies in the the lock file are a redundant duplication of information available elsewhere.
+
```
baronfel, line 472: Quick feedback on this spec (which is comprehensive in all the right ways - thank you!) that I wanted to get down:

Bundling more kinds of 'things' into TargetFramework-shaped holes makes sense because the only kind of 'foreach' we've built into our platform so far _is_ `TargetFramework`. An alternative, which would likely be much more work but _may_ be more extensible and/or orthogonal, would be to

> **Consider making the 'pivots' of multitargeting more extensible**
> 
> Meaning, let a project and/or MSBuild SDK define a list of properties that should be cartesian-product-ed together when doing any kind of 'for-each-able' workflow (build, pack, publish, etc).

Now, I don't necessarily think we should _do_ this, bit it comes to mind as an alternative to this proposal, and I don't see a section where alternatives are considered and criteria for rejection are listed.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
```diff
@@ -0,0 +1,467 @@
...
+- When a project reference is made to a project with aliases, the TargetFramework will serve as a secondary pivot in case the standard get nearest target framework returns more than one option.
+The technical explanation has [the gory details](#project-to-project-references)
+- Packages lock file will also be extended to support aliasing. The file format will need to be updated, but it's already version and that work should be straightforward. [Packages Lock File technical details](#lock-file-changes).
+- Visual Studio Package Manager UI does not have great support for multi-targeting and no new work is being proposed here. This is a specific flavor of multi-targeting, and as such no difference from the challenges we have today. [Visual Studio challenges](#visual-studio-challenges)
+- dotnet package add, dotnet package update and dotnet package list will all be updated to support per framework operations in case they do not support it already.
+
+### Pack and duplicate frameworks
+
+There is no intuitive solution to the pack scenario.
```
ericstj: Dependencies only have a single slot for TargetFramework.  I agree that there is no solution to alias dependencies.

Assets/build-output on the other hand is a different story.  For example, folks using alias for runtime it's perfectly reasonable to include all aliased frameworks and allow each target to specify it's package path.  

Then it's up to the project to decide which alias is used for dependencies.

I can imagine us building first-class support for this behavior.  We could designate one alias as being used for dependencies, or we could allow the project to assert that all aliases should have identical dependencies (and validate that).

A couple common cases that I'd like to see us build minimal examples:
1. Project building a .NET package for multiple runtimes with API and dependencies represented in runtime-less output.
2. Project building a roslyn analyzer targeting multiple roslyn versions, omitting all dependencies.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Aliases are secondary pivots for compat. There's no more to them at this point.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Added in this section.

accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
nkolev92: Great points. Added in the spec as future opportunities. We'll very likely come back to this and provide a 1st class experience once we've gotten some adoption.

nkolev92:
Addressed all feedback + review feedback.

@donnie-msft, @zivkan can I get approval so I can merge. 

Thanks!

zivkan (approved):

Issue #5375, opened 2017-06-08
Honor WarningsNotAsErrors


natemcmaster:
We use TreatWarningsAsErrors=true. We want NuGet warnings to appear, but not to fail build. The concept of NoWarn is too strong as it suppresses them completely, but we still want to be aware of the issue. IMO the best property for this (that CSC also uses) is WarningsNotAsErrors, a list of warnings that should not cause errors even if TreatWarningsAsErrors is enabled.

Scenario:
```xml
<Project Sdk="Microsoft.NET.Sdk">
	<ItemGroup>
		<PackageReference Include="Moq" Version="4.7.1" />
	</ItemGroup>
	<PropertyGroup>
		<TargetFramework>netcoreapp2.0</TargetFramework>
		<TreatWarningsAsErrors>true</TreatWarningsAsErrors>
		<WarningsNotAsErrors>$(WarningsNotAsErrors);NU1603</WarningsNotAsErrors>
	</PropertyGroup>
</Project>
```

cc @emgarten

anangaur:
@natemcmaster I am just wondering, why is this not exposed in the UI. IMO this would be the common scenario. Users will start with treating all warnings as errors and then as they understand the warnings from their project, either suppress them completely (possible in UI by "Suppress warnings" OR list them as warnings not treated as errors (WarningsNotAsErrors). May be in the properties UI we should have an option under `Treat warnings as errors`: 
`All except <Textbox>`

Current properties UI for reference:
![image](https://user-images.githubusercontent.com/14800916/26953648-8b8f9a50-4c60-11e7-85f0-b96c2d48fbc6.png)

natemcmaster:
+1 Would be cool if the VS folks could expose this in UI. This property has been a well-known option in the compiler for years. (see [Microsoft.CSharp.Core.targets](https://github.com/dotnet/roslyn/blob/6a98a959cbd296991411e3e9d124aeeb4827ae1f/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L150))

natemcmaster:
Requested: https://github.com/dotnet/project-system/issues/2409

mishra14:
Tagging @rrelyea for triage request.

mishra14:
Closing as there has been no customer traction on this.

mishra14:
Received another request. Reopening this issue for future.

rizi:
Any update on this, it would be great if WarningsNotAsErrors a honored by NuGet.

[...]

PR #12134, opened 2022-10-06
WarningsNotAsErrors proposal

nkolev92:
Proposal for https://github.com/NuGet/Home/issues/5375

[Rendered](https://github.com/NuGet/Home/blob/dev-nkolev92-warningsNotAsErrors/proposed/2022/WarningsNotAsErrors.md)

### DIFF
--- a/proposed/2022/WarningsNotAsErrors.md
+++ b/proposed/2022/WarningsNotAsErrors.md
@@ -34,8 +34,7 @@ For the following table assume `NU1603` is being raised during restore.
 | TreatWarningsAsErrors | NoWarn | WarningsAsErrors | WarningsNotAsErrors | Behavior | Explanation |
 |-----------------------|--------|------------------|---------------------|----------|-------------|
 | true | | | | All warnings are treated as errors | |
-| true | | | | All warnings are treated as errors | |
-| true | | NU1603 |  | Succeeds with no warning. | NU1603 is raised, but because of no warn it won't be shown. |
+| true | NU1603 | |  | Succeeds with no warning. | NU1603 is raised, but because of no warn it won't be shown. |
 | false | | NU1603 | | NU1603 is upgraded to an error. | Only a certain list of warnings are being treated as errors |
 | true | | | NU1603 | Succeeds with a warning. | Only NU1603 is a warning. If any other warning is raised, it'll be upgraded to an error |
 | true | NU1603 | | NU1603 | Succeeds. No warning. | NoWarn takes precedence over both WarningAsErrors and WarningNotAsErrors. |
--- a/proposed/2022/WarningsNotAsErrors.md
+++ b/proposed/2022/WarningsNotAsErrors.md
@@ -1,0 +1,108 @@
+# Honor WarningsNotAsErrors
+
+- Author Name (GitHub username link)
+- Start Date (YYYY-MM-DD)
+- GitHub Issue (GitHub Issue link)
+- GitHub PR (GitHub PR link)
+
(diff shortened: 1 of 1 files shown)


proposed/2022/WarningsNotAsErrors.md
```diff
@@ -0,0 +1,86 @@
...
+The precedence order for the 3 properties by NuGet can summarized as:
+`NoWarn > WarningsAsErrors > WarningsNotAsErrors`.
+
+For the following table assume `NU1603` is being raised during restore.
+
+| TreatWarningsAsErrors | NoWarn | WarningsAsErrors | WarningsNotAsErrors | Behavior | Explanation |
+|-----------------------|--------|------------------|---------------------|----------|-------------|
+| true | | | | All warnings are treated as errors | |
+| true | | | | All warnings are treated as errors | |
+| true | | NU1603 |  | Succeeds with no warning. | NU1603 is raised, but because of no warn it won't be shown. |
```
zivkan: The NoWarn column is empty, but the explanation says that "because of no warn it won't be shown"

proposed/2022/WarningsNotAsErrors.md
```diff
@@ -0,0 +1,86 @@
...
+
+The precedence order for the 3 properties by NuGet can summarized as:
+`NoWarn > WarningsAsErrors > WarningsNotAsErrors`.
+
+For the following table assume `NU1603` is being raised during restore.
+
+| TreatWarningsAsErrors | NoWarn | WarningsAsErrors | WarningsNotAsErrors | Behavior | Explanation |
+|-----------------------|--------|------------------|---------------------|----------|-------------|
+| true | | | | All warnings are treated as errors | |
+| true | | | | All warnings are treated as errors | |
```
zivkan: These two rows are duplicates

proposed/2022/WarningsNotAsErrors.md
nkolev92: Fixed.

proposed/2022/WarningsNotAsErrors.md
```diff
@@ -0,0 +1,85 @@
...
+
+The precedence order for the 3 properties by NuGet can summarized as:
+`NoWarn > WarningsAsErrors > WarningsNotAsErrors`.
+
+For the following table assume `NU1603` is being raised during restore.
+
+| TreatWarningsAsErrors | NoWarn | WarningsAsErrors | WarningsNotAsErrors | Behavior | Explanation |
+|-----------------------|--------|------------------|---------------------|----------|-------------|
+| true | | | | All warnings are treated as errors | |
+| true | NU1603 | |  | Succeeds with no warning. | NU1603 is raised, but because of no warn it won't be shown. |
```
zivkan:
```suggestion
| true | NU1603 | |  | Succeeds with no warning. | NoWarn prevents NU1603 from being raised. |
```

I get that internally, NuGet will still do the NU1603 analysis, create the internal warning, and it's the "output" code that will prevent it from being shown to the user, but I think this implementation detail just doesn't matter to customers. Customers won't see the NU1603, and I think that's all that matters to a design document.

zivkan (approved):

JonDouglas (approved):
Lgtm

proposed/2022/WarningsNotAsErrors.md
nkolev92: Your suggestion makes sense. 
I think it was a late update that led to the sloppy explanation.

Issue #8407, opened 2019-07-30
Issue saving settings in Visual Studio - adding a source throws an exception


nkolev92:
Root issue : https://developercommunity.visualstudio.com/content/problem/640886/failed-to-apply-nuget-package-manager-settings-vs.html

The customer has not provided additional feedback so it's a bit more difficult to investigate.

Root cause is: 

```xml
<entry>
    <record>917</record>
    <time>2019/07/11 05:17:42.112</time>
    <type>Error</type>
    <source>NuGet Package Manager</source>
    <description>System.ArgumentException: An item with the same key has already been added.&#x000D;&#x000A;   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)&#x000D;&#x000A;   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)&#x000D;&#x000A;   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)&#x000D;&#x000A;   at NuGet.Configuration.PackageSourceProvider.SavePackageSources(IEnumerable`1 sources)&#x000D;&#x000A;   at NuGet.Options.PackageSourcesOptionsControl.ApplyChangedSettings()</description>
  </entry>
  <entry>
    <record>918</record>
    <time>2019/07/11 05:17:44.430</time>
    <type>Error</type>
    <source>NuGet Package Manager</source>
    <description>System.ArgumentException: An item with the same key has already been added.&#x000D;&#x000A;   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)&#x000D;&#x000A;   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)&#x000D;&#x000A;   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)&#x000D;&#x000A;   at NuGet.Configuration.PackageSourceProvider.SavePackageSources(IEnumerable`1 sources)&#x000D;&#x000A;   at NuGet.Options.PackageSourcesOptionsControl.ApplyChangedSettings()</description>
  </entry>
  <entry>
    <record>919</record>
    <time>2019/07/11 05:20:39.330</time>
    <type>Error</type>
    <source>NuGet Package Manager</source>
[...]

PR #12172, opened 2022-10-19
Add spec of dotnet nuget config command

### LINKED ISSUE #12297: Write a spec for dotnet nuget config command

heng-liu:
Fixes: https://github.com/NuGet/Home/issues/12297
Add a spec of `dotnet nuget config list` command.

It's a part of epic issue:
 Add a dotnet nuget config command https://github.com/NuGet/Home/issues/12296

### DIFF
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -74,7 +74,7 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.conf
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
-- List all the NuGet configuration file that will be applied, but passing a non-exsiting `WORKING_DIRECTORY`.
+- List all the NuGet configuration file that will be applied, but passing a non-exsiting `--working-directory`.
 
 ```
 dotnet nuget config path  --working-directory C:\Test\NonExistingRepos
@@ -82,7 +82,7 @@ dotnet nuget config path  --working-directory C:\Test\NonExistingRepos
 Error: The path "C:\Test\NonExistingRepos" doesn't exist.
 ```
 
-- List all the NuGet configuration file that will be applied, but passing an inaccessible `WORKING_DIRECTORY`: C:\Test\AccessibleRepos\NotAccessibleSolution. 
+- List all the NuGet configuration file that will be applied, but passing an inaccessible `--working-directory`: C:\Test\AccessibleRepos\NotAccessibleSolution. 
 
 The configuration file under C:\Test\AccessibleRepos\NotAccessibleSolution\NuGet.Config will be ignored without any warning or error.
 
@@ -338,7 +338,10 @@ dotnet nuget config unset defaultPushSource --config-file C:\Users\username\AppD
 ## Future Work
 1. The `dotnet nuget config path/get` is a community ask. We will discuss if adding this command into NuGet.exe CLI, in the future.
 2. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. 
-But the behavior is confusing: the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file. We might need to change this behavior in the future. (Related code: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L229)
+But the behavior is confusing: the `set` command will set the property which appears last when loading(for all writable files), so it's not updating the closest NuGet configuration file, but the user-wide NuGet configuration file.(Related code: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L229)
+We keep the behavior of `dotnet nuget config set` the same with above for now.
+But we might need to change this behavior in the future. 
+Considering this will be a breaking change, we will only consider doing it in main version change, if there are enough votes.
 
 ## Open Questions
 1. To show configuration files locations, is it better to use `--verbosity` option or some option named like `--show-path`? (git command is using `--show-origin` for similar purpose)
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -26,10 +26,9 @@ The following command will be implemented in the `dotnet.exe` CLI.
 
 - List
 
-Lists all the NuGet configuration file locations. 
+Lists all the NuGet configuration settings. 
 
-This command will include all the NuGet configuration file that will be applied, when invoking NuGet command from the current working directory path. The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
-You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
+This command will list merged NuGet configuration settings from one or multiple NuGet configuration files that will be applied, when invoking NuGet command from the current working directory path. 
 
 #### Arguments
 
@@ -46,7 +45,10 @@ If the specified `WORKING_DIRECTORY` doesn't exist, an error is displayed indica
 - -v|--verbosity <LEVEL>
 
 Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. 
-When the verbosity level is detailed or diagnostic, the command will display the merged NuGet configuration.
+When the verbosity level is detailed or diagnostic, the command will display all the NuGet configuration file that will be applied, when invoking NuGet command from the current working directory path.
+
+The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
+You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
 
 - -?|-h|--help
 
@@ -57,7 +59,25 @@ Prints out a description of how to use the command.
 - List all the NuGet configuration file that will be applied, when invoking NuGet command in the current directory.
 
 ```
-dotnet nuget config list
+dotnet nuget config list --verbosity detailed
+
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Microsoft Visual Studio Offline Packages" value="C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\" />
+  </packageSources>
+  <packageRestore>
+    <add key="enabled" value="False" />
+    <add key="automatic" value="False" />
+  </packageRestore>
+  <bindingRedirects>
+    <add key="skip" value="False" />
+  </bindingRedirects>
+  <packageManagement>
+    <add key="format" value="0" />
+    <add key="disabled" value="False" />
+  </packageManagement>
+</configuration>
 
 C:\Test\Repos\Solution\NuGet.Config
 C:\Test\Repos\NuGet.Config
@@ -70,27 +90,39 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 - List all the NuGet configuration file that will be applied, when invoking NuGet command in the specific directory.
 
 ```
+dotnet nuget config list  C:\Test\Repos --verbosity detailed
+
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Microsoft Visual Studio Offline Packages" value="C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\" />
+    <add key="Test Package source" value="C:\work" />
+  </packageSources>
+  <packageRestore>
+    <add key="enabled" value="False" />
+    <add key="automatic" value="False" />
+  </packageRestore>
+  <bindingRedirects>
+    <add key="skip" value="False" />
+  </bindingRedirects>
+  <packageManagement>
+    <add key="format" value="0" />
+    <add key="disabled" value="False" />
+  </packageManagement>
+</configuration>
+
+C:\Test\Repos\NuGet.Config
+C:\Test\NuGet.Config
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+- List only the NuGet configuration settings, when invoking NuGet command in the specific directory.
+
+```
 dotnet nuget config list  C:\Test\Repos
 
-C:\Test\Repos\NuGet.Config
-C:\Test\NuGet.Config
-C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
-C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
-C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
-```
-
-- List all the NuGet configuration file that will be applied, and show the merged NuGet configuration, when invoking NuGet command in the specific directory.
-
-```
-dotnet nuget config list  C:\Test\Repos -v d
-
-C:\Test\Repos\NuGet.Config
-C:\Test\NuGet.Config
-C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
-C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
-C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
-
-The merged NuGet configuration:
 <configuration>
   <packageSources>
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
@@ -125,7 +157,26 @@ Error: The path "C:\Test\NonExistingRepos" doesn't exist.
 The configuration file under C:\Test\AccessibleRepos\NotAccessibleSolution\NuGet.Config will be ignored without any warning or error.
 
 ```
-dotnet nuget config list  C:\Test\AccessibleRepos\NotAccessibleSolution
+dotnet nuget config list  C:\Test\AccessibleRepos\NotAccessibleSolution -v d
+
+<configuration>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Microsoft Visual Studio Offline Packages" value="C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\" />
+    <add key="Test Package source" value="C:\work" />
+  </packageSources>
+  <packageRestore>
+    <add key="enabled" value="False" />
+    <add key="automatic" value="False" />
+  </packageRestore>
+  <bindingRedirects>
+    <add key="skip" value="False" />
+  </bindingRedirects>
+  <packageManagement>
+    <add key="format" value="0" />
+    <add key="disabled" value="False" />
+  </packageManagement>
+</configuration>
 
 C:\Test\AccessibleRepos\NuGet.Config
 C:\Test\NuGet.Config
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -36,6 +36,7 @@ You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nu
 Prints out a description of how to use the command.
 
 - --current-directory
+
 Run this command as if current directory is set to the specified directory.
 
 #### Examples
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -185,6 +185,7 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.conf
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
+#### Commands
 
 - Set
 
@@ -282,8 +283,9 @@ dotnet nuget config unset defaultPushSource --config-file C:\Users\username\AppD
 3. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. But there is no `list` command. And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?
 
 ## Open Questions
-1. What are the recommended verbs in dotnet command when setting and unsetting values?  ('set' and 'unset' work or not?)
-2. `Get` is not mentioned in this doc. It may cause some confusions:
+1. To show configuration files locations, is it better to use `--verbosity` option or some option named like `--show-path`? (git command is using `--show-origin` for similar purpose)
+2. What are the recommended verbs in dotnet command when setting and unsetting values?  ('set' and 'unset' work or not?)
+3. `Get` is not mentioned in this doc. It may cause some confusions:
    Should `get` work for settings not in config section? (It's workable, but then we will miss `set` and `unset` for those as those are not doable)
    Should `get` get the settings from a specific config file(like `set` and `unset`), or get the merged settings from multiple config files?
 
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -358,6 +358,8 @@ Considering this will be a breaking change, we will only consider doing it in ma
 1. Will this command help with diagnosing incorrect setting format?
 <br />No. Incorrect NuGet settings should have seperate error/warning message to tell the customer what's wrong in the setting file. If we have incorrect NuGet settings, all NuGet command, including `dotnet nuget config` command, should display the same error/warning message.
 <br />E.g. if we have an invalid XML problem in one of the NuGet configuration file, running all NuGet command will get an error as following:
+```
 dotnet nuget list source
 error: NuGet.Config is not valid XML. Path: 'C:\Users\username\Source\Repos\NuGet.Config'.
 error:   The 'disabledPackageSources' start tag on line 19 position 4 does not match the end tag of 'configuration'. Line 20, position 3.
+```
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -291,10 +291,8 @@ dotnet nuget config unset defaultPushSource --config-file C:\Users\username\AppD
 
 ## Considerations
 1. Will this command help with diagnosing incorrect setting format?
-No. Incorrect NuGet settings should have seperate error/warning message to tell the customer what's wrong in the setting file. If we have incorrect NuGet settings, all NuGet command, including `dotnet nuget config` command, should display the same error/warning message.
-E.g. if we have an invalid XML problem in one of the NuGet configuration file, running all NuGet command will get an error as following:
-```
+<br />No. Incorrect NuGet settings should have seperate error/warning message to tell the customer what's wrong in the setting file. If we have incorrect NuGet settings, all NuGet command, including `dotnet nuget config` command, should display the same error/warning message.
+<br />E.g. if we have an invalid XML problem in one of the NuGet configuration file, running all NuGet command will get an error as following:
 dotnet nuget list source
 error: NuGet.Config is not valid XML. Path: 'C:\Users\username\Source\Repos\NuGet.Config'.
 error:   The 'disabledPackageSources' start tag on line 19 position 4 does not match the end tag of 'configuration'. Line 20, position 3.
-```
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -305,14 +305,6 @@ If the specified `CONFIG_KEY` is not a key in [config section](https://learn.mic
 
 - --config-file
 
-Specify the config file path to remove the setting key-value pair. If it's not specified, the config file with highest priority will be updated.
-
-- -?|-h|--help
-
-Prints out a description of how to use the command.
-
-- --config-file
-
 Specify the config file path to remove the setting key-value pair. If it's not specified, `%AppData%\NuGet\NuGet.Config` (Windows), or `~/.nuget/NuGet/NuGet.Config` or `~/.config/NuGet/NuGet.Config` (Mac/Linux) is used. See [On Mac/Linux, the user-level config file location varies by tooling.](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#on-maclinux-the-user-level-config-file-location-varies-by-tooling)
 
 - -?|-h|--help
@@ -337,6 +329,7 @@ dotnet nuget config unset defaultPushSource --config-file C:\Users\username\AppD
 
 ## Future Work
 1. The `dotnet nuget config path/get` is a community ask. We will discuss if adding this command into NuGet.exe CLI, in the future.
+
 2. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. 
 But the behavior is confusing: the `set` command will set the property which appears last when loading(for all writable files), so it's not updating the closest NuGet configuration file, but the user-wide NuGet configuration file.(Related code: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L229)
 We keep the behavior of `dotnet nuget config set` the same with above for now.
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -1,21 +1,21 @@
-# dotnet nuget config command
+# NuGet configuration CLI for dotnet.exe
 
 - Author: [Heng Liu](https://github.com/heng-liu)
 - GitHub Issue [8420](https://github.com/NuGet/Home/issues/8420)
 
 ## Problem Background
 
-Currently, there is no command for NuGet users to know about the NuGet configuration file locations. This is inconvienient for both NuGet users and NuGet team. When NuGet users want to figure out where is the merged configuration coming from, or when NuGet team tries to diagnose issues but needs all the configuration files, this command will be helpful.
+Currently, there is no NuGet configuration CLI for dotnet.exe. It's inconvenient for NuGet users to know about the NuGet configuration file locations and figure out where is the merged configuration coming from. 
 
 ## Who are the customers
 
 This feature is for dotnet.exe users.
 
 ## Goals
-Design and implement `dotnet nuget config list` commands.
+Design and implement `dotnet nuget config` command.
 
 ## Non-Goals
-Design and implement other `dotnet nuget config` commands. E.g. add/update/delete
+Design and implement `dotnet nuget config` command with commands other than `list`, E.g. add/update/delete
 
 ## Solution
 The following command will be implemented in the `dotnet.exe` CLI.
@@ -49,10 +49,13 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.conf
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
-### Future Work
+## Future Work
 The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
 
 ## Open Questions
 1. Shall we add `working directory` as an option? So that people could check different NuGet configuration lists without changing current directory.
 2. Do we need to add this into NuGet.exe CLI?
+
+## Considerations
+1. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. But there is no `list` command. And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?
 
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -35,9 +35,12 @@ You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nu
 
 Prints out a description of how to use the command.
 
+- --current-directory
+Run this command as if current directory is set to the specified directory.
+
 #### Examples
 
-- List all the NuGet configuration file that will be applied.
+- List all the NuGet configuration file that will be applied, when invoking NuGet command in the current directory.
 
 ```
 dotnet nuget config list
@@ -49,12 +52,21 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.conf
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
+- List all the NuGet configuration file that will be applied, when invoking NuGet command in the specific directory.
+```
+dotnet nuget config list --current-directory c:\repos\Solution
+
+c:\repos\Solution\NuGet.Config
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
 ## Future Work
 The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
 
 ## Open Questions
-1. Shall we add `working directory` as an option? So that people could check different NuGet configuration lists without changing current directory.
-2. Do we need to add this into NuGet.exe CLI?
+1. Do we need to add this into NuGet.exe CLI?
 
 ## Considerations
 1. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. But there is no `list` command. And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -30,14 +30,15 @@ If no command is specified, the command will default to list.
 Lists all the NuGet configuration file locations. This command will include all the NuGet configuration file that will be applied, when invoking NuGet command from the current working directory path. The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
 You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
 
+#### Arguments
+- CURRENT_DIRECTORY
+
+Run this command as if current directory is set to the specified directory.
+
 #### Options
 - -?|-h|--help
 
 Prints out a description of how to use the command.
-
-- --current-directory
-
-Run this command as if current directory is set to the specified directory.
 
 #### Examples
 
@@ -46,18 +47,21 @@ Run this command as if current directory is set to the specified directory.
 ```
 dotnet nuget config list
 
-c:\repos\Solution\Project\NuGet.Config
-c:\repos\Solution\NuGet.Config
+C:\Test\Repos\Solution\NuGet.Config
+C:\Test\Repos\NuGet.Config
+C:\Test\NuGet.Config
 C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
 - List all the NuGet configuration file that will be applied, when invoking NuGet command in the specific directory.
-```
-dotnet nuget config list --current-directory c:\repos\Solution
 
-c:\repos\Solution\NuGet.Config
+```
+dotnet nuget config list  C:\Test\Repos
+
+C:\Test\Repos\NuGet.Config
+C:\Test\NuGet.Config
 C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
 C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
--- a/proposed/2022/dotnet-nuget-config-spec.md
+++ b/proposed/2022/dotnet-nuget-config-spec.md
@@ -186,12 +186,106 @@ C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
 ```
 
 
+- Set
+
+Set the NuGet configuration settings. 
+
+This command will set the value of a specified NuGet configuration setting.
+
+Please note this command only manages settings in [config section](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section).
+For other settings not in config section, we have/will have other commands. E.g. for [trustedSigners section](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#trustedsigners-section), we have [dotnet nuget trust](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-trust) command.
+
+#### Arguments
+
+- SETTING_KEY
+
+Specify the key of the settings that are to be set.
+
+If the specified `SETTING_KEY` is not a key in [config section](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section), an error is displayed indicating the `SETTING_KEY` could not be set.
+
+- SETTING_VALUE
+
+Set the value of the SETTING_KEY to SETTING_VALUE.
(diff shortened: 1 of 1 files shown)


proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,61 @@
...
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+## Future Work
+The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
+
+## Open Questions
+1. Shall we add `working directory` as an option? So that people could check different NuGet configuration lists without changing current directory.
```
erdembayar: I believe this one should be in current spec. It's not hard to implement and just 1 option. For example, in CI pipeline logging it's useful to just pass working directory rather than switching to different working directories.

heng-liu: Thanks for reviewing! I agree, this should be useful. Added this option :)

erdembayar (approved):

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,74 @@
...
+c:\repos\Solution\Project\NuGet.Config
+c:\repos\Solution\NuGet.Config
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+- List all the NuGet configuration file that will be applied, when invoking NuGet command in the specific directory.
+```
+dotnet nuget config list --current-directory c:\repos\Solution
```
nkolev92: Do we need a `--current-directory` option. Can the location be the argument to the command and be the working directory if not specified.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,74 @@
...
+## Non-Goals
+Design and implement `dotnet nuget config` command with commands other than `list`, E.g. add/update/delete
+
+## Solution
+The following command will be implemented in the `dotnet.exe` CLI.
+
+### `dotnet nuget config`
+
+#### Commands
+If no command is specified, the command will default to list.
```
nkolev92: What happens today with commands such as sources and packages? 
Is there a default command?

nkolev92 (commented):
I'm really excited about this addition :)

proposed/2022/dotnet-nuget-config-spec.md
erdembayar: Yes, that's better. Just pass as argument instead of typing long option keyword.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,74 @@
...
+Run this command as if current directory is set to the specified directory.
+
+#### Examples
+
+- List all the NuGet configuration file that will be applied, when invoking NuGet command in the current directory.
+
+```
+dotnet nuget config list
+
+c:\repos\Solution\Project\NuGet.Config
```
zivkan: We discourage customers trying to nuget.config files in project directories, because VS will never use it, and I'm not sure that command line solution restore will either. Therefore, we shouldn't use discourages practices in our examples.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Good question! There is no default command for sources and packages, not sure if it's because verbs need to go first.
But there is a default command `list` for [dotnet nuget trust command](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-trust#commands).
So, when no sub command is specified, do we want to let people know all the subcommands for `dotnet nuget config`? Or just run dotnet nuget config list command by default?

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks for letting me know! Changed the examples accordingly.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Great suggestion! Changed it into an argument.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Hi @JonDouglas, may I know if you have any suggestions about this? Thanks! :)

nkolev92: Personally, I think we should align to dotnet list package.

heng-liu:
There are two types of implementations:
1. Treat `list` as subcommand.  Example: [dotnet nuget trust](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-trust)
2. Treat `list` as an option. Example: [dotnet nuget locals](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-locals)

I followed the 1st in this spec. But I'm not quite sure if there is any preference. Please advise, thanks!

nkolev92:
I think it has to be a subcommand. 

The subcommand matches `dotnet list package`, `dotnet list reference` etc. 
It matches the direction of the dotnet CLI.

heng-liu:
Hi @baronfel, thanks for your help offline on the guidance!
Could you please take a look and let me know if there is anything I need to change? Thanks! :)

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+## Future Work
+The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
+
+## Open Questions
+1. Do we need to add this into NuGet.exe CLI?
+
+## Considerations
+1. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. But there is no `list` command. And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?
```
dominoFire: I would say yes to this question. Let's create an issue for that.

dominoFire (approved):

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,79 @@
...
+This feature is for dotnet.exe users.
+
+## Goals
+Design and implement `dotnet nuget config` command.
+
+## Non-Goals
+Design and implement `dotnet nuget config` command with commands other than `list`, E.g. add/update/delete
+
+## Solution
+The following command will be implemented in the `dotnet.exe` CLI.
```
zivkan, line 21: I really wish GitHub had threaded/resolvable conversations for comments without a specific file/line number 😔

I'd like to know what the design is for error scenarios, specifically when one or more of the XML files contains syntax errors (from an implementation point of view, this means that `Settings.Load*` apis will fail, so we can't use `loadedConfig.GetConfigFilePaths()`).

Is `dotnet nuget config list` going to report an error and fail to list all the files?  Will it at least tell us the filename of the XML that couldn't be parsed?  Or will all the XML files be listed despite the parsing failure?  Will it tell us which file(s) contain parsing errors?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,79 @@
...
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+## Future Work
+The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
+
+## Open Questions
+1. Do we need to add this into NuGet.exe CLI?
```
nkolev92: IMO we do dotnet.exe first and have nuget.exe be a future consideration.

proposed/2022/dotnet-nuget-config-spec.md
nkolev92: > And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?

I think that's a discussion for a future point. 

Personally, I'd prefer we guide users towards choosing the right config file instead of NuGet making assumptions. 

I'd be hesitant to have VS and NuGet.exe work differently from dotnet.exe.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: That's a good question!
I tested with NuGet.exe and dotnet.exe, and both of them are able to tell us the filename of the XML that couldn't be parsed, as below:
```
nuget list source
NuGet.Config is not valid XML. Path: 'C:\Users\username\Source\Repos\NuGet.Config'.
  The 'disabledPackageSources' start tag on line 19 position 4 does not match the end tag of 'configuration'. Line 20, position 3.
```
```
dotnet nuget list source
error: NuGet.Config is not valid XML. Path: 'C:\Users\username\Source\Repos\NuGet.Config'.
error:   The 'disabledPackageSources' start tag on line 19 position 4 does not match the end tag of 'configuration'. Line 20, position 3.
```
So I think the `dotnet nuget config list` is able to do this as well.

If we'd like to list all config files despite the parsing failure, we'd consider if it worth doing.
Since running all nuget commands will encounter this invalid XML error, users need to fix this invalid XML before running any NuGet commands. So this seems to be a prerequisite and there is no harm if user needs to do this first before knowing the list of config files.
In addition, this error message clearly points out the file path and location,  so it should be easy for users to correct the invalid XML.

The only scenario needs to be considered is, with the invalid XML error, user should able to specifying a specific correct config file to run nuget commands. And if we list all config file locations despite the parsing failure, it might help people to find the path easier. But it seems "listing all config file locations despite the parsing failure" adds very limited value.

Those are my preliminary thoughts. Please let me know if you have any other thoughts :)

The other error scenario I'd like to talk about is, the specified working directory doesn't exist. Shall we go ahead listing user-wide and machine-wide config files? Or show a warning saying the working directory doesn't exist? Or both?

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks for reviewing! Moved it to "Future Work" section as we won't address it in the implementation.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks! Moved it to "Future Work" section as we won't address it in the implementation.

heng-liu:
Hi @baronfel, I think I've addressed all comments for now. 
May I know if you could take a look? Or do you prefer a review meeting late this week or early next week? Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Changed to no default command. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,84 @@
...
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+## Future Work
+1. The `dotnet nuget config list` is a community ask. We will consider adding more commands, like add/update/delete, in the future.
+2. We will discuss if adding this command into NuGet.exe CLI, in the future.
+3. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. But there is no `list` command. And the behavior is confusing (the `set` command will set the property which appears last when loading, so sometimes it's not updating the closest NuGet configuration file). Do we want to implement those subcommand(e.g.`set`) in the future in dotnet.exe differently?
+
+## Open Questions
+1. When the specified `CURRENT_DIRECTORY` doesn't exist, shall we list user-wide and machine-wide config files? Or show a warning saying the working directory doesn't exist? Or do both?
```
erdembayar: I believe it just should error out/warn. If there is error, I'm not sure how much machine wide config matters.

nkolev92: I think we should just error too.

heng-liu: Thanks! Updated with only showing a warning/error.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,84 @@
...
+
+We decided to show a warning/error.
```
zivkan: This means it's no longer an open question, and the outcome can be integrated into the design above.

heng-liu:
I just thought about another scenario, when there is an error when accessing the current directories/it's parent directories.
For now, nuget command just ignore the inaccessible config file and there will be no warning/error. As we check if the config file exists at this [line](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L823), and according to [doc](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.exists?view=net-7.0):
`The Exists method returns false if any error occurs while trying to determine if the specified file exists. This can occur in situations that raise exceptions such as passing a file name with invalid characters or too many characters, a failing or missing disk, or if the caller does not have permission to read the file.`

So, for a folder as below, if I change "Solution" folder to inaccessible, the "NuGet.Config" file under this folder will be ignored and there will be no warning/error.

C:\TEST
│   NuGet.Config
│
└───Repos
    │   NuGet.Config
    │
    └───**Solution**
        │   **NuGet.Config**
        │   Solution.sln
        │
        └───Project
                Class1.cs
                Project.csproj

I'm assuming this is expected (Please let me know if you consider this is a bug). Shall I warn/error for this situation for "dotnet nuget config list" command? Or just ignore it and show the rest of NuGet.Config files without any warning/error(align with other NuGet commands)?

If the current behavior for nuget command is expected(ignore the inaccessible config files without any warning/error), my thoughts are doing this without any warning/error, but add this as a note into the doc of "dotnet nuget config" command. (Applied in the doc). Please let me know if you have any other thoughts. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,84 @@
...
+#### Commands
+
+- List
+
+Lists all the NuGet configuration file locations. This command will include all the NuGet configuration file that will be applied, when invoking NuGet command from the current working directory path. The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
+You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
+
+#### Arguments
+
+- CURRENT_DIRECTORY
```
kartheekp-ms: Incase if we are referring to `CURRENT_DIRECTORY` as working directory, `dotnet store` command has `-w|--working-dir <WORKING_DIRECTORY>` option https://learn.microsoft.com/dotnet/core/tools/dotnet-store#optional-options. I think it is better to use the same option name for this command also.

heng-liu: Thanks for reviewing! Renamed it as `WORKING_DIRECTORY`, but still an argument based on previous discussions. Pls let me know if you have any different thoughts :)

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: You're right :) Integrated into the design and add an example about this scenario.

nkolev92:
> I'm assuming this is expected (Please let me know if you consider this is a bug). Shall I warn/error for this situation for "dotnet nuget config list" command? Or just ignore it and show the rest of NuGet.Config files without any warning/error(align with other NuGet commands)?

I'm guessing we match what commands that use the configs do, for example running restore.

nkolev92 (approved):

heng-liu:
> I'm guessing we match what commands that use the configs do, for example running restore.

You're right. It matched what commands that use the configs.

heng-liu:
Hi @baronfel , may I know if you have some time to take a look at this PR this week? Thanks!

kartheekp-ms:
https://github.com/NuGet/Home/issues/12213 issue may be out of scope for this spec, but the feature request make sense to me.

`When I issue a nuget config command in a repository, I don't get any information. And if I specify -Verbosity detailed, then I only get the version number of the nuget.exe executed.
It would be nice to have nuget config list the current configuration values as being evaluated from the various config files, and if -Verbosity detailed specified also the config file being source for given config value.`

bstordrup:
> #12213 issue may be out of scope for this spec, but the feature request make sense to me.
> 
> `When I issue a nuget config command in a repository, I don't get any information. And if I specify -Verbosity detailed, then I only get the version number of the nuget.exe executed. It would be nice to have nuget config list the current configuration values as being evaluated from the various config files, and if -Verbosity detailed specified also the config file being source for given config value.`

Regarding the #12213, I have a local branch that implements this by adding a `all` dummy item and is used like this:

```
nuget config all -Verbosity detailed
```

To have the -Verbosity work correctly however, the Origin property (and hence SettingsFile class) need to be added to PublicAPI, but these changes are also done in my local branch.

Once this pull request is merged, I can update my branch and make a pull request for the change in #12213

heng-liu:
Thanks @bstordrup and @kartheekp-ms ! I've added the `-verbosity` option into this spec. And the merged NuGet configuration will be displayed when verbosity is detailed.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,146 @@
...
+If the specified `WORKING_DIRECTORY` doesn't exist, an error is displayed indicating the `WORKING_DIRECTORY` doesn't exist.
+
+> [!Note]
+> If `WORKING_DIRECTORY` (or its parent directories) is not accessible, the command will ignore any NuGet configuration files under those directories without any warning/error. This is aligned with other NuGet commands.
+
+#### Options
+- -v|--verbosity <LEVEL>
+
+Sets the verbosity level of the command. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic]. The default is minimal. 
+When the verbosity level is detailed or diagnostic, the command will display the merged NuGet configuration.
```
zivkan: It feels out of place to me to implement "display merged config" as the detailed output for "Lists all the NuGet configuration file locations". 

I feel like the functionality belongs in a `get` verb, not `list`.

I have no idea if anyone will actually want to do it, but also imagine that someone wants to have a script where the output is piped to a file, and then they use an XML processing library/tool to work on the XML.  With this design, it's difficult because of all the non-XML output. Whereas a dedicated command to get the merged config could have only XML output at minimal verbosity, allowing customers to "safely" process the XML output of something like `dotnet nuget config get all -v m`.

You might also want to have a look at `git config` and all its options as prior-art that customers might already be familiar with. Admittedly they use `--list` to list all the config values (what you're calling the "merged config"), rather than listing all the files that contributed. I don't see an option to only list gitconfig files used, but they do have a `--show-origin`, so `git config --list --show-origin` will show all the config key-value pairs, and which

nkolev92: I agree, verbosity should only really be a thing when you want to know more about certain config value. 

IMO here's how it should work longer term: 

- dotnet nuget list config-> Lists all NuGet config file paths.
- dotnet nuget list config globalPackagesFolder -> Gives you the value of the globalPackagesFolder configuration. 
- dotnet nuget add config globalPackagesFolder D:\packages -> Sets the global packages folder in the approrpiate config. (same thing would apply to update)
- dotnet nuget list config globalPackagesFolder --verbosity detailed -> Gives you the value of the global packages folder configuration + the config where it comes from.

I know nuget.exe uses `get` and `set`, but the dotnet CLI uses add/remove/update so we should think in those verbs. 

Note the ordering to match `dotnet add package` and the like commands.

nkolev92: Apologies for not noticing the ordering and being on autopilot. 

dotnet commands are verb driven.

heng-liu: Thanks @zivkan for the great suggestion!
I'm thinking if the following sounds good:
`dotnet nuget config list` to list all config values(merged configurations)
`dotnet nuget config list --show-path` to list all config values(merged configurations) and the config file paths.

//cc @JonDouglas

heng-liu: > Apologies for not noticing the ordering and being on autopilot.
> 
> dotnet commands are verb driven.

Thanks @nkolev92 ! I asked Chet about dotnet command conventions and he helped me confirm that: `At this time we have semi-standardized on dotnet NOUN VERB, so dotnet nuget config list would be the preferred form for this command.` (In email: Re: Please advise on naming of a new dotnet nuget command)

nkolev92: my bad if I'm providing conflicting feedback :D 

Ignore the ordering then. I'm curious what you think about the other behavior.

heng-liu: Thanks @nkolev92 ! Most of them sounds good to me. But seems we still don't have an option to view all config values (what customer mentioned)
So how about the following:

- `dotnet nuget config list`    list all config values (key + value).
- `dotnet nuget config list` --verbosity detailed   list all config values (key + value), and the config file paths at the bottom.
- `dotnet nuget config list <configKey>`  list the value of the specified config key
- `dotnet nuget config list <configKey> --verbosity detailed`    list the value of the specified config key, and the config file path it comes from
-  `dotnet nuget config add <configKey> <configValue> <config file path>`  add the key value into the specified config file
-  `dotnet nuget config update <configKey> <configValue> <config file path>`  update the key value pair in the specified config file
-  `dotnet nuget config delete <configKey> <config file path>` delete the key value pair from the specified config file

But I'm not sure what shall we do if the configuration is something with child attributes, but not a simple key-value pair. E.g. package source, packageSourceMapping etc. Shall we avoid providing add/update/delete command for those configurations?

zivkan: Should we consider `set` and `unset` instead of `add`, `update`, and `delete`?  That's one less verb, and I don't think that `add` vs `update` matters to customers. (for example, would we want `add` to fail if it already exists in the config?   Would we want to fail `update` if the key doesn't exist in the config file?)

> `dotnet nuget config list <configKey>` list the value of the specified config key

nitpick/very opinionated: A single item is technically a special case of a list, but 

> But I'm not sure what shall we do if the configuration is something with child attributes,

nuget.exe config only handles the [`<configuration><config>` part of the config](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section), which are key-value pairs (probably implemented by `Dictionary<string, string>`). The `nuget.exe config` commands could never be used for other parts of the config file, like sources, credentials, package source mapping, etc.  I'm not saying this is good, but be aware that if you want to make `dotnet nuget config`'s scope so much different to `nuget.exe config`'s existing implementation, it's going to make implementing the spec a lot more difficult.  It's definitely not ideal to design features based on how easy it is to implement, but sometimes we need to be pragmatic. As you pointed out, what would `dotnet nuget config add` commands even look like for package sources, credentials, or package source mapping?

nkolev92: > Most of them sounds good to me. But seems we still don't have an option to view all config values (what customer mentioned)

I think we might need a magic word for list all configs such as "all", something we promise to never make a config value for example. 

> Should we consider set and unset instead of add, update, and delete? That's one less verb, and I don't think that add vs update matters to customers. (for example, would we want add to fail if it already exists in the config? Would we want to fail update if the key doesn't exist in the config file?)

I think we should use the same verb as the CLI.

IMO, it's ok with config doesn't handle sources, credentials or source mapping. All of those will have their own individual commands.

erdembayar:
@heng-liu 
Maybe it's off topic. But could this detect nuget.config issues like broken xml tag or duplicate keys etc? It's hard for customers to detect exactly which entry key is causing problem. For example: https://developercommunity.visualstudio.com/t/After-upgrading-Visual-Studio-2022-to-17/10199374#T-ND10201096

heng-liu:
> @heng-liu Maybe it's off topic. But could this detect nuget.config issues like broken xml tag or duplicate keys etc? It's hard for customers to detect exactly which entry key is causing problem. For example: https://developercommunity.visualstudio.com/t/After-upgrading-Visual-Studio-2022-to-17/10199374#T-ND10201096

Thanks for reviewing! I don't think the command will be able to detect broken xml. Please see comment at https://github.com/NuGet/Home/pull/12172#discussion_r1005260706
For duplicate keys, I'm not sure if this is the same with [this one](https://github.com/NuGet/Home/issues/8407#issuecomment-1245890475). If so, it seems an bug not related to this spec. 
So generally speaking, for incorrect settings, I think we should have better error messages.

proposed/2022/dotnet-nuget-config-spec.md
zivkan: FWIW, here's a customer complaint that separate add and update verbs are annoying: 
* https://github.com/dotnet/sdk/issues/29026#issue-1445497084

> I think we should use the same verb as the CLI.

Consistency is good, unless it's consistently user unfriendly 😁

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,298 @@
...
+Prints out a description of how to use the command.
+
+#### Examples
+
+- List all the NuGet configuration file that will be applied, when invoking NuGet command in the current directory.
+
+```
+dotnet nuget config list --verbosity detailed
+
+<configuration>
```
kartheekp-ms: customers may not be interested in the XML content. How about just printing key/value pairs from the merged NuGet.Config file?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,298 @@
...
+#### Arguments
+
+- WORKING_DIRECTORY
+
+Run this command as if working directory is set to the specified directory.
+
+If the specified `WORKING_DIRECTORY` doesn't exist, an error is displayed indicating the `WORKING_DIRECTORY` doesn't exist.
+
+> [!Note]
+> If `WORKING_DIRECTORY` (or its parent directories) is not accessible, the command will ignore any NuGet configuration files under those directories without any warning/error. This is aligned with other NuGet commands.
```
baronfel: This argument feels odd to me - it's not used in any of the other proposed commands (instead explicit `--config-file` options are provided), so why does this concept get first-class treatment here?

Looking at the other commands (set, unset) to me `list` implies listing configuration values (parallel to setting/unsetting them), _not_ locating and showing the entire xml config file.  This command as described feels like a `dotnet nuget config paths` kind of command as a result.

baronfel: The verb `get` is missing here, and it seems like a natural analogue for users that are scripting. Are scripting use cases a motivation here?

baronfel (commented):
Brief comment on list especially.  I have a couple follow up questions on the user experience though - is tab completion a thought here? Specifically it seems like set/unset are prime opportunities to suggest config keys that are valid.

proposed/2022/dotnet-nuget-config-spec.md
kartheekp-ms: `dotnet store` command has `-w|--working-dir <WORKING_DIRECTORY>` option https://learn.microsoft.com/dotnet/core/tools/dotnet-store#optional-options

heng-liu: Thanks for reviewing!
The option is renamed to `-w|--working-dir <WORKING_DIRECTORY>`. Thanks @kartheekp-ms!
The `get` is added. Thanks @baronfel !

heng-liu:
The spec is updated based on our last review meeting.
The main changes are:
dotnet nuget config path : show list of paths of config files
dotnet nuget config get : get all merged configuration settings (XML format)
dotnet nuget config get -v d : get all merged configuration settings with source of config file path (XML format)
dotnet nuget config get <CONFIG_KEY> : get the value of the configuration setting (a string, not XML format)
dotnet nuget config set <CONFIG_KEY> <CONFIG_VALUE>: set the configuration setting
dotnet nuget config unset <CONFIG_KEY> : unset the value of the configuration setting
Please take another look, thank you!

proposed/2022/dotnet-nuget-config-spec.md
nkolev92: > Thanks for reviewing! The option is renamed to `-w|--working-dir <WORKING_DIRECTORY>`. Thanks @kartheekp-ms! The `get` is added. Thanks @baronfel !

I don't think that was the outcome in the meeting. 

When we brought this up for discussion, Daniel pointed out that other commands such as `dotnet build`, `dotnet nuget sources` etc all use a directory or a project file as the first argument. 

I was under the impression that no one disagreed with that? 

I guess there's a recording so we can always figure it out from there :)

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+Design and implement in the same way for NuGet.exe command.
+
+## Solution
+The following command will be implemented in the `dotnet.exe` CLI.
+
+### `dotnet nuget config`
+
+#### Commands
+
+- Path
```
nkolev92: Should the config be `paths` instead? 
I'm not really sure what the guidance for plural vs singular is.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+
+List all the paths of NuGet configuration files that will be applied, when invoking NuGet command from the current working directory path.
```
nkolev92, line 29: Shouldn't the get command be listed here?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+
+The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
+You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
+
+
+#### Options
+
+- -w|--working-dir <WORKING_DIRECTORY>
```
nkolev92: I left the feedback on the original thread. I don't think we want an option. A working directory argument is a natural solution.
`--working-dir` is redundant imo.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+
+Error: The path "C:\Test\NonExistingRepos" doesn't exist.
+```
+
+- List all the NuGet configuration file that will be applied, but passing an inaccessible `--working-dir`: C:\Test\AccessibleRepos\NotAccessibleSolution. 
+
+The configuration file under C:\Test\AccessibleRepos\NotAccessibleSolution\NuGet.Config will be ignored without any warning or error.
+
+```
+dotnet nuget config list  --working-dir C:\Test\AccessibleRepos\NotAccessibleSolution
```
nkolev92: You probably just missed changing the list argument here.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+
+C:\Test\AccessibleRepos\NuGet.Config
+C:\Test\NuGet.Config
+C:\Users\username\AppData\Roaming\NuGet\NuGet.Config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.FallbackLocation.config
+C:\Program Files (x86)\NuGet\Config\Microsoft.VisualStudio.Offline.config
+```
+
+#### Commands
+- Get
```
nkolev92: Ah, now I see it. You just have commands subheadings. I'd recommend not having the same headers in a markdown down as it breaks linking. 

I'd the following: 

`### Commands
#### path
#### set
#### set
```

Hope that helps.

File: proposed/2022/dotnet-nuget-config-spec.md (hunk 12 above)
Comment (nkolev92) [reply proposed/2022/dotnet-nuget-config-spec.md]: Personally I don't think this is valuable in V0, but I won't oppose it as long as others are good with that :)

File: proposed/2022/dotnet-nuget-config-spec.md [hunk 19]
```diff
@@ -0,0 +1,358 @@
...
+```
+
+- Get `http_proxy` from config section, when `http_proxy` is not set in any of the NuGet configuration files.
+
+```
+dotnet nuget config get http_proxy 
+
+Key 'http_proxy' not found.
+
+```
```
nkolev92, line 226: I think we're missing a more natural/more valuable idea. 

`dotnet nuget config get http_proxy --verbosity detailed`  could be: 
http://company-squid:3128@contoso.com" - C:\Users\username\AppData\Roaming\NuGet\NuGet.Config

I think this is is way more valuable than the  `dotnet nuget config get` xml format.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+2. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. 
+But the behavior is confusing: the `set` command will set the property which appears last when loading(for all writable files), so it's not updating the closest NuGet configuration file, but the user-wide NuGet configuration file.(Related code: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L229)
+We keep the behavior of `dotnet nuget config set` the same with above for now.
+But we might need to change this behavior in the future. 
+Considering this will be a breaking change, we will only consider doing it in main version change, if there are enough votes.
+
+## Open Questions
+1. To show configuration files locations, is it better to use `--verbosity` option or some option named like `--show-path`? (git command is using `--show-origin` for similar purpose)
+
+2. `dotnet nuget config get <CONFIG_KEY>` will get the value of the specifc config key (which is aligned with NuGet config command). But
```
nkolev92: I think it's ok if `dotnet nuget config get` shows all the config settings, I just don't think we should show the XML.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,358 @@
...
+`dotnet nuget config get` will get all merged configuration settings in xml format, since many sections are not simple key-value pairs.
+ Do we need to have another verb for getting all merged configuration settings? 
+
+3. `dotnet nuget config get -v d` will display source of each configuration setting key, in a format of comment in xml file. So that people can still redirect the output into a file without breaking any syntax. Does that sound good to you?
+
+3. `--working-dir` is changed from an argument into an option. Because we could not differentiate `dotnet nuget config get <WORKING_DIRECTORY>` and `dotnet nuget config get <CONFIG_KEY>`. Any better ideas?
```
nkolev92: I see the motivation for this change. 

IMO we should add a magic word for getting all the config values, example `all`.

nkolev92 (commented):
This feature is turning out bigger than originally expected, but I think that's for the better! 
A couple of suggestion for tweaks.

Great job @heng-liu

proposed/2022/dotnet-nuget-config-spec.md
nkolev92: Replying to myself again, this isn't redundant but a way to differentiate between getting all and specifying a directory.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks for reviewing! I used `path` in the spec and add this question into open question as #1. 
Hi @baronfel, could you help advise? Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks! Changed the structures accordingly.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Good catch! Thanks! Fixed.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Fixed. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Added this as an example. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks! Changed the format accordingly. We might need to go through all configuration sections for the formats later.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: I added `all` and also add this as #2 in the open questions.
Hi @baronfel , could you pls advise? Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks! Yes, @kartheekp-ms helped find an example of this option:
The Dotnet store command has -w|--working-dir <WORKING_DIRECTORY> option https://learn.microsoft.com/dotnet/core/tools/dotnet-store#optional-options
Since it can be short for `-w`, I think it should be okey to be an option?

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Changed into non-XML format. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
baronfel: Is "all" necessary if there's no longer a 'working dir' argument? In that case, no argument value would mean `all` because the presence of an argument would be a filter.

proposed/2022/dotnet-nuget-config-spec.md
nkolev92: The existence of one instance doesn't make it the best approach imo. 

I'm also one of the few people even commenting on this concern, so I'm not really sure how I feel about it because of that. Whether others don't care, or others prefer the working directory approach.

proposed/2022/dotnet-nuget-config-spec.md
nkolev92: my suggestion is we do the magic word `all` and make the argument the working directory.

`--working-directory`  just seems off. 

I'm also very conscious of the fact that only a few of us have commented on this topic in here, while more did in the meeting :D

nkolev92:
Great summary with the open questions @heng-liu 
I know it's been a lot of iterations, but I really appreciate that the open questions are up to date :)

heng-liu:
Hi @baronfel and @dsplaisted , there are 3 open questions are mainly about dotnet conventions. Could you please advise? Thanks!
Other comments are all addressed. (You may find the details in email I sent on 12/6 ).
Please let me know if you have any other concerns. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,366 @@
...
+
+- CONFIG_VALUE
+
+Set the value of the CONFIG_KEY to CONFIG_VALUE.
+
+#### Options
+
+- --configfile <FILE>
+
+The NuGet configuration file (nuget.config) to use. If specified, only the settings from this file will be used. If it's not specified, `%AppData%\NuGet\NuGet.Config` (Windows), or `~/.nuget/NuGet/NuGet.Config` or `~/.config/NuGet/NuGet.Config` (Mac/Linux) is used. See [On Mac/Linux, the user-level config file location varies by tooling.](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#on-maclinux-the-user-level-config-file-location-varies-by-tooling)
```
zivkan, line 253: So, even if there's a nuget.config in the current directory, if `--configfile` is not specified, the local directory file will not be modified?

While I think it's good that there's an easy way to modify the user-profile nuget.config, I feel like there will be more customers who expect the local directory file to be edited by default, so I think this will be non-intuitive.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,366 @@
...
+1. The `dotnet nuget config path/get` is a community ask. We will discuss if adding this command into NuGet.exe CLI, in the future.
+
+2. NuGet.exe [config command](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config) is implemented. 
+But the behavior is confusing: the `set` command will set the property which appears last when loading(for all writable files), so it's not updating the closest NuGet configuration file, but the user-wide NuGet configuration file.(Related code: https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L229)
+We keep the behavior of `dotnet nuget config set` the same with above for now.
+But we might need to change this behavior in the future. 
+Considering this will be a breaking change, we will only consider doing it in main version change, if there are enough votes.
+
+## Open Questions
+1. Shall we use `dotnet nuget config path` or `dotnet nuget config paths` to get all configuration file paths?
```
dsplaisted: I would go with the plural (`dotnet nuget config paths`) since it returns multiple paths.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,366 @@
...
+
+2. `WORKING_DIRECTORY` is changed from an option to an argument. So we have `dotnet nuget config path <WORKING_DIRECTORY>` and `dotnet nuget config get <ALL|CONFIG_KEY> <WORKING_DIRECTORY>`. Is it okey if we have two arguments in the second command? And if it's key, only the second argument could be optional, right?
```
dsplaisted: This seems OK to me, but I'd like to hear from and defer to @baronfel, as he's the expert.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,366 @@
...
+dotnet nuget config get http_proxy 
+
+http://company-squid:3128@contoso.com"
+
+```
+
+- Get `http_proxy` from config section, when invoking NuGet command in the current directory. Show the source(nuget configuration file path) of this configuration setting.
+
+```
+dotnet nuget config get http_proxy
```
dsplaisted: Is there supposed to be a "show path" or verbosity parameter here?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,366 @@
...
+We keep the behavior of `dotnet nuget config set` the same with above for now.
+But we might need to change this behavior in the future. 
+Considering this will be a breaking change, we will only consider doing it in main version change, if there are enough votes.
+
+## Open Questions
+1. Shall we use `dotnet nuget config path` or `dotnet nuget config paths` to get all configuration file paths?
+
+2. `WORKING_DIRECTORY` is changed from an option to an argument. So we have `dotnet nuget config path <WORKING_DIRECTORY>` and `dotnet nuget config get <ALL|CONFIG_KEY> <WORKING_DIRECTORY>`. Is it okey if we have two arguments in the second command? And if it's key, only the second argument could be optional, right? 
+
+3. To show configuration files locations, is it better to use `--verbosity` option or some option named like `--show-path`? (git command is using `--show-origin` for similar purpose)
```
dsplaisted: `--show-path` is better in my opinion.  I think of verbosity as something to apply to logging, not so much the output of a command like this.  Also in most cases of the .NET CLI, the verbosity doesn't do anything unless it's a command that ends up running MSBuild.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks @dsplaisted for reviewing! Changed into `paths`.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Good catch! Added `--show-path`. Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks for the explanation! Make sense! Changed into `--show-path`.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Hi @baronfel , could you help with this open question? Thanks!

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Yes. This is the behavior in NuGet.exe config command right now. 
You may refer to the [doc ](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-config#options), or the [code](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L208-L230) .
The code indicates if there is no clear tag, default to furthest from the user, which is the user-wide configuration file (machine wide configuration files are excluded as they have `IsReadOnly` flag set to true, so they are not writable).
Unless this is not the expected behavior in NuGet.exe, otherwise we've discussed to keep the behavior consistent between NuGet.exe and dotnet.exe, and we only consider changing the behavior when major version changes, if there are enough upvotes. (Please refer to [Future Work  #2](https://github.com/NuGet/Home/blob/779cfc85bdff9887f47ae36d5705151667a80dcb/proposed/2022/dotnet-nuget-config-spec.md#future-work) in the spec).

proposed/2022/dotnet-nuget-config-spec.md
baronfel: Good eye on the potential for ambiguity of the grammar here. Let me make sure I have the proposed grammar correct:

> dotnet nuget config get <all|CONFIG_KEY> [<WORKING_DIRECTORY>]

meaning 
> the first argument is required, and will either be the string 'all' or the name of an existing config key. the second key is optional and if present represents the working directory to locate the nuget config file from

Is that correct? if so, **I see no problem and am ok with the proposal**. in cases where there is some potential for ambiguity, consider adding a flag to allow for explicit disambiguation (i.e. `-w <path>` like we rejected earlier), but we've already had that conversation so I don't think we need to revisit it.

heng-liu: Yep, that's correct :)
Great! Thanks for the confirmation! I moved this open question to Considerations.

heng-liu:
Thanks all for helping review this spec!
I know there are lots of iterations. Thanks a lot for your patience!

All comments are addressed in this spec. It's ready to review again.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,369 @@
...
+Design and implement in the same way for NuGet.exe command.
+
+## Solution
+The following command will be implemented in the `dotnet.exe` CLI.
+
+### `dotnet nuget config`
+
+### Commands
+
+### Paths
```
erdembayar, line 27: Sorry, I might be missing the context. Why don't default to this command? Why need `dotnet nuget config paths` instead of just `dotnet nuget config`?

heng-liu: No worries! In our review meeting, we discussed and agreed that showing config file paths and showing config file contents should using different words. So we use `dotnet nuget config paths` for showing paths and `dotnet nuget config get` for showing contents.
And `dotnet nuget config` will show the options just like `dotnet nuget config -help`.

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,369 @@
...
+List all the paths of NuGet configuration files that will be applied, when invoking NuGet command from the current working directory path.
+
+The listed NuGet configuration files are in priority order. So the order of loading those configurations is reversed, that is, loading order is from the bottom to the top. So the configuration on the top will apply.
+You may refer to [How settings are applied](https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied) for more details. 
+
+#### Arguments
+
+- WORKING_DIRECTORY
+
+Run this command as if working directory is set to the specified directory. If it's not specified, the current working directory will be used.
```
erdembayar, line 38: If can choose to not pass anything isn't this one is `option` rather than `argument`?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,369 @@
...
+
+If the specified `WORKING_DIRECTORY` doesn't exist, an error is displayed indicating the `WORKING_DIRECTORY` doesn't exist.
+
+> [!Note]
+> If `WORKING_DIRECTORY` (or its parent directories) is not accessible, the command will ignore any NuGet configuration files under those directories without any warning/error. This is aligned with other NuGet commands.
```
erdembayar, line 43: Here not accessible means "path exist, but don't enough privilege/permission" right?

proposed/2022/dotnet-nuget-config-spec.md
```diff
@@ -0,0 +1,369 @@
...
+
+- CONFIG_KEY
+
+Specify the key of the settings that are to be set.
+
+If the specified `CONFIG_KEY` is not a key in [config section](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section), a message is displayed indicating the `CONFIG_KEY` could not be set.
+
+- CONFIG_VALUE
+
+Set the value of the CONFIG_KEY to CONFIG_VALUE.
```
erdembayar, line 247: If value type doesn't match with expected value type, then does it show appropriate message?
If you include 1 unhappy path scenario output for each use case, then that would be great. I learned this from `dotnet list package --format json` spec.

nkolev92 (approved):
Love it. 
Great job.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Not exactly. Here is the definition of [arguments](https://learn.microsoft.com/en-us/dotnet/core/tools/#arguments) and [options](https://learn.microsoft.com/en-us/dotnet/core/tools/#options) in dotnet command world.  
An example of argument is a solution or a project. it can be optional as well.
The folder path is something similar with a solution or a project, so we choose to use it as an argument. We checked this with  .NET sdk team in the review meetings.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Good question! Yes, it means the path exist but don't have enough privilege/permission to access.

proposed/2022/dotnet-nuget-config-spec.md
heng-liu: Thanks @erdembayar ! That's a good suggestion!
However, after checking, NuGet.exe config -Set doesn't check the key at all. So if we run `NuGet.exe config -Set ABC=abc`, the following will be added into NuGet.Config file:
```
  <config>
    <add key="ABC" value="abc" />
  </config>
```
Created a follow-up issue https://github.com/NuGet/Home/issues/12345, added it to the epic https://github.com/NuGet/Home/issues/12296.

heng-liu:
I'll merge the PR since it's approved several weeks ago and there is no new comments. 
Please let me know if there is anything you'd like to discuss and I will create follow-up issues if necessary. Thanks!
FYI @dsplaisted @baronfel

PR #12199, opened 2022-10-27
Update the issue templates

nkolev92:
The issue templates currently have a title that says `bug`, `dcr`, `feature` etc. 

I think this is redundant as we use labels for that. 

Sometimes, the issue title isn't really what the issue is exactly about.
Other times an issue "evolves" as we get clarity.


erdembayar (approved):
I believe it's good change. Because customer doesn't remove `BUG`, `DCR` etc when creating new issue.

jeffkl (approved):

zivkan (approved):

heng-liu (approved):

PR #12225, opened 2022-11-10
Add a bot to Close stale icebox issues, with internal-debugging label

heng-liu:
This is the first step for https://github.com/NuGet/Client.Engineering/issues/1784
The added rulesets are tested in bot-testing repo(but with different waiting time, to shorten the testing period). 
You may check the 1-4 rulesets of bot-testing repo in UI at: https://portal.fabricbot.ms/bot/   and enter "NuGet/bot-testing" and load config.

The testing example is: https://github.com/NuGet/bot-testing/issues/22
1. The searching tasks will check issues with `Status:Inactive` and `internal-debugging`, but without `Status:Excluded from icebox cleanup`, `Triage:NeedsTriageDiscussion`, `Icebox cleanup candidate`, to add label `Icebox cleanup candidate`, and a [comment ](https://github.com/NuGet/bot-testing/issues/22#issuecomment-1300463845)
2. After 14 days, if there is no activity on issues with `Icebox cleanup candidate` and `internal-debugging`, and without `Triage:NeedsTriageDiscussion`, the issue will be closed, and a comment "This issue will now be closed since it had been marked Icebox cleanup candidate, but received no further activity in the past 14 days.".
3. During the 14 days, if there is any comment on the issue, a `Triage:NeedsTriageDiscussion` label will be added. So it won't be closed anymore, until we manually remove the `Triage:NeedsTriageDiscussion` label. And this label will lead to a discussion in triaging meeting.

Note: the `internal-debugging` label is to narrow down the affected issues for the first step, just in case if we find anything wrong.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1575,6 +1575,488 @@
           }
         ]
       }
+    },
+    {
+      "taskType": "scheduled",
+      "capabilityId": "ScheduledSearch",
+      "subCapability": "ScheduledSearch",
+      "version": "1.1",
+      "config": {
+        "frequency": [
+          {
+            "weekDay": 0,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 1,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 2,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 3,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 4,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 5,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          },
+          {
+            "weekDay": 6,
+            "hours": [
+              3,
+              7,
+              11,
+              15,
+              19,
+              23
+            ],
+            "timezoneOffset": -7
+          }
+        ],
+        "searchTerms": [
+          {
+            "name": "isOpen",
+            "parameters": {}
+          },
+          {
+            "name": "isIssue",
+            "parameters": {}
+          },
+          {
+            "name": "hasLabel",
+            "parameters": {
+              "label": "Status:Inactive"
+            }
+          },
+          {
+            "name": "noLabel",
+            "parameters": {
+              "label": "Status:Excluded from icebox cleanup"
+            }
+          },
+          {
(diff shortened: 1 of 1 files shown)


.github/fabricbot.json
```diff
@@ -1575,6 +1575,488 @@
...
+              "label": "internal-debugging"
+            }
+          }
+        ],
+        "taskName": "[Close stale icebox][4-2] Close \"Icebox cleanup candidate\" if no activity within 14 days.",
+        "actions": [
+          {
+            "name": "addReply",
+            "parameters": {
+              "comment": "This issue will now be closed since it had been marked Icebox cleanup candidate, but received no further activity in the past 14 days. It is still possible to reopen or comment on the issue, but please note that the issue will be locked if it remains inactive for another 30 days."
```
nkolev92: Do we need  the long explanation?

I feel like we already add a few verbose messages, so we can just close it :)

nkolev92 (approved):
Should we consider running these bots once a day?

That's what we do for the `WaitingForCustomer` issues.

heng-liu:
> Should we consider running these bots once a day?
> 
> That's what we do for the `WaitingForCustomer` issues.

I think it should be okey if running multiple times a day. First, for the processed results, since we add `Icebox cleanup candidate` label, they will be excluded from next query. Second, since the FabricBot will use the GraphQL API to query and there is a limit of max number of results (I forgot it's 100 or 200). If there are more than 100 or 200 records and we run it multiple times a day, the task will do the searching task incrementally, so it could be done in a day. If there are less than 100 or 200 records, the first query will finish the task and the rest of queries will get 0 record to process.

.github/fabricbot.json
heng-liu: Thanks for bring that out! I asked the question in the teams and will update here later.

heng-liu: Update to not lock the closed issues.

nkolev92 (approved):

zivkan:
> If there are more than 100 or 200 records and we run it multiple times a day, the task will do the searching task incrementally, so it could be done in a day.

Is this documented? 

Because the bot would have to be stateful to remember where it stopped last time in order to know where to resume from the next time.

edit: Ok, I see that the bot's action adds a label which then in theory excludes it from the search results, so for this pattern it will work.

>  If there are less than 100 or 200 records, the first query will finish the task and the rest of queries will get 0 record to process.

We don't get over 100 new issues per day, so after the initial set of stale issues are processed, we will never need more than one execution per day.

.github/fabricbot.json
```diff
@@ -1575,6 +1575,355 @@
...
+            "parameters": {}
+          },
+          {
+            "name": "isIssue",
+            "parameters": {}
+          },
+          {
+            "name": "hasLabel",
+            "parameters": {
+              "label": "Status:Inactive"
```
zivkan, line 1683: What is going to tag issues with `Status:Inactive`, to that it becomes part of this this query's result set?

heng-liu: Thanks for the question! That's a label we've already added, when identifying stale icebox issues. The rulesets is: if there is no activity for more than 2 years and with an icebox label, the `Status:Inactive` will be added. 
Since this is a one-time experiment, this part is done in config portal and not added into configuration files.
If this experiment is good and we want to do this regularly, I'll add a ruleset for this: identifying stale icebox issues.

heng-liu:
> We don't get over 100 new issues per day, so after the initial set of stale issues are processed, we will never need more than one execution per day.

Yes, that's right.
I don't think there is any document about those. I just checked with people who works on the FabricBot team.

jeffkl (approved):

PR #12310, opened 2022-12-07
Auditing projects for package vulnerabilities during restore

JonDouglas:
Introduction to an auditing experience at package restore time where known vulnerabilities will be reported.

This proposal introduces new MSBuild warnings and output to the NuGet restore operation.

[Rendered Spec](https://github.com/NuGet/Home/blob/dev-feature-vinr/proposed/2022/vulnerabilities-in-restore.md)

Thank You 🎉


mattzink:
We use Nexus repo manager and Artifactory as local Nuget.org package mirrors, so it would be great if there was a way in the nuget.config file to specify that Nuget.org is used only for vulnerability lookup, not for package hosting.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+Found 2 vulnerabilities (0 low, 1 moderate, 0 high, 1 critical) in 1 package(s)
+
+warning NU1904: Package 'Contoso.Service.APIs' 1.0.3 has a known critical severity vulnerability, https://github.com/advisories/GHSA-1234-5678-9012
+warning NU1904: Package 'Contoso.Service.APIs' 1.0.3 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-1234-5678-90XY
+```
+
+In cases where no known vulnerabilities are found, restore will provide a normal verbosity message for every project:
+
+```bash
+No known vulnerabilties found for <project path>.
```
loic-sharma:
```suggestion
No known vulnerabilities found for <project path>.

```

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+
+| Warning Code | Severity |
+|--------------|----------|
+| NU1901 | low |
+| NU1902 | moderate |
+| NU1903 | high |
+| NU1904 | critical |
+
+```bash
+Found 2 vulnerabilities (0 low, 1 moderate, 0 high, 1 critical) in 2 package(s)
```
loic-sharma: What's the value in reporting severities with 0 hits? Would it be better to exclude those instead?

```suggestion
Found 2 vulnerabilities (1 moderate, 1 critical) in 2 package(s)

```

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+When NuGet restores, we check whether a source supports vulnerabilities and if so it can go through a protocol of getting this information.
+
+The client needs all the data to make decisions about vulnerability. In order to not have the client redownload a file and limit the total number of HTTP calls we will ever make, the vulnerability resource will be an index that can contain multiple pages.
+
+- The vulnernability resource **must** be an array of objects.
+- Each object **must** contain:
+  - `@name`, a user friendly name for the page, **must** be unique.
+  - `@id`, the url that contains the data.
+  - `@updated`, a UTC timestamp when the content at `@id` was updated last.
+  - `comment`, a user friendly description.
```
loic-sharma, line 153: What's the value for this field? Does this really need to be required? Will this be noisy repetitive values?

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+- The vulnerability resource **must** at minimum contain 1 page.
+- The pages within the vulnerability resource **must** be exclusive.
+- The vulnerability resource may 2 pages.
+  - One page, `base`, represents the data up to a certain point in time. This page **must** be updated periodically, based on a trigger.
+  - The second page, `update`, represents the data from the last update of `base`, to present. This page **should** be updated with new data.
+  - Periodically, for example once a month, the data from `update` **should** be migrated to `base`.
+  - If an entry needs to be removed from `base`, `base` should be updated.
+- Other partioning strategies are workable as well, as long as only 2 pages are receiving all the new data.
```
loic-sharma: Are the `base` and `update` names required if using other partitioning strategies?

Also, what's the value in requiring a maximum of two pages? Does it complicate the client logic to support arbitrary number of pages?

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: Good idea, let's keep this open for discussion!

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+- The vulnernability resource **must** be an array of objects.
+- Each object **must** contain:
+  - `@name`, a user friendly name for the page, **must** be unique.
+  - `@id`, the url that contains the data.
+  - `@updated`, a UTC timestamp when the content at `@id` was updated last.
+  - `comment`, a user friendly description.
+- The vulnerability resource **must** at minimum contain 1 page.
+- The pages within the vulnerability resource **must** be exclusive.
+- The vulnerability resource may 2 pages.
+  - One page, `base`, represents the data up to a certain point in time. This page **must** be updated periodically, based on a trigger.
```
loic-sharma: How big will nuget.org's `base` page be? How quickly will it grow? In the worst case scenario (like an Australian customer), are we concerned about the performance impact of updating this `base` page during a restore?

proposed/2022/vulnerabilities-in-restore.md
nkolev92: Base and update aren't required names, but highly recommended.

The idea with 2 pages updating is that the client will always want to have all the data be up to date. 
It would allow refreshing to be super fast without an overly complicated protocol, as we'd only download the out of date files.

It's balance between CI (where the data is always download) and local, where the data is refreshed.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+  <ItemGroup>
+    <PackageReference Include="PackageId" Version="3.0.0" SuppressedAdvisoryUrls="https://github.com/advisories/GHSA-g3q9-xf95-8hp5" />
+  </ItemGroup>
+```
+
+The project level and package level metadata are not necessarily a package deal, we can choose to only add one.
+
+### Allow sources to upstream vulnerability information
+
+While the NuGet tooling does support multiple sources and a means to control which packages come from which source through [Package Source Mapping](https://learn.microsoft.com/en-us/nuget/consume-packages/package-source-mapping), all configurations are not going to contain nuget.org, and thus won't easily get the data curated from the [GitHub Advisory Database](https://devblogs.microsoft.com/nuget/how-to-scan-nuget-packages-for-security-vulnerabilities/).
```
loic-sharma: Will vulnerability reporting take source maps into account? If two package sources report different vulnerabilities for the same package, will the client report all vulnerabilities?

proposed/2022/vulnerabilities-in-restore.md
nkolev92: These were the numbers a few weeks ago:

vulnerability_all_alternate.json    2.4 Mb
vulnerability_all_alternate.json.gz 78.9 Kb

> are we concerned about the performance impact of updating this base page during a restore?

Definitely a consideration. We'd only redownload when restore actually happens. The whole process takes time, but right now we think those numbers are relatively small and a cost worth paying given the benefit.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+- Pros:
+  - Vulnerabilities are not inherently tied to sources. This is explicit effort undertaken by nuget as shepherds of the eco-system to provide vulnerability information.
+  - This can be additive. We can ship vulnerabilities tied to sources first and then update later.
+
+- Cons:
+  - The client has an established protocol and a means to configure a source.
+  - Experimentation is arguably easier as we're not introducing new syntax, at least not immediately.
+  - The challenge is that we'd need to establish a protocol, and something that could be slower on the uptake.
+
+### Optimizing vulnerability check cadence
```
loic-sharma, line 263: Would it make sense to update vulnerability info in the background in Visual Studio? That way, restore performance in Visual Studio wouldn't be affected by this feature.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,391 @@
...
+While the nuget.org data only contains vulnerability information about packages on nuget.org, often times an `upstream` source is setup by customers, and while they reference their own private feed, they can still download packages from nuget.org. The vulnerability information is valuable in these scenarios.
+
+If this upstream approach is taken, it is very likely that the vulnerability checking has a lot of misses, but this is an understood trade-off.
+
+It is **very important* that if the upstream sources link to the nuget.org vulnerability, the actual nuget.org urls are used wherever possible, as this would allow the client to easily deduplicate the vulnerability information.
+
+### Providing vulnerability information through a json, remotely or locally
+
+Other components may want to provide vulnerability information that's not tied to a source.
+Whatever approach we take there will be a way to represent the vulnerability information through a json file.
```
loic-sharma, line 404: Perhaps there could be a nuget.config for a list of vulnerability info URLs? That'd be useful if I have a single private package source that doesn't point to its upstream vulnerability info URL yet.

proposed/2022/vulnerabilities-in-restore.md
loic-sharma: The 2 page idea makes perfect sense. I'm more curious as to why we *require* a maximum of 2 pages; that seems odd.

nkolev92: It's not really suggesting maximum of 2 pages.
Or at least, it's not supposed to :) 

What that line is trying to communicate is that as few pages as possible should be updated, so that the client is not forced to redownload pages all the time.

proposed/2022/vulnerabilities-in-restore.md
kapsiR: Good idea!  
Additionally this could be another step in CI (cli), so that not every build would be affected by a new download or in other words: 
Have more control over the time when a new file is going to be downloaded

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,392 @@
+# Auditing projects for package vulnerabilities during restore
+
+- [Jon Douglas](https://github.com/JonDouglas/), [Nikolche Kolev](https://github.com/nkolev92), [Peter Yu](https://github.com/ryuyu)
+- Issue: [#8087](https://github.com/NuGet/Home/issues/8087)
+- Related: [#11549](https://github.com/NuGet/Home/pull/11549)
+
+## Summary
+
+Audit your dependencies when restoring your NuGet packages with security vulnerabilities reported to GitHub Advisory Database.
```
kapsiR: I don't know exactly what sources this database is using but are there any plans to support additional databases too? (e.g. Google osv.dev)

proposed/2022/vulnerabilities-in-restore.md
nkolev92: > Would it make sense to update vulnerability info in the background in Visual Studio? That way, restore performance in Visual Studio wouldn't be affected by this feature.

We don't expect this to be taking so long where it's noticeable consistently. 

On the topic of background downloading, it's not always desirable to communicate with sources without user intent. That could for example trigger authentication and we wouldn't want anything to interrupt the user. 

> Additionally this could be another step in CI (cli), so that not every build would be affected by a new download or in other words:

There are a lot of optimizations that happen prior to vulnerability data being downloaded. 
If you're not changing your packages/deleting the obj folder, restore will be snappy and won't redownload sources.

proposed/2022/vulnerabilities-in-restore.md
nkolev92: Yes, it would have to be something along those lines.

proposed/2022/vulnerabilities-in-restore.md
nkolev92: That's a great question. 

Right now, we think no is the answer for that, we "merge" the data, but we can certainly change that based on community input.

proposed/2022/vulnerabilities-in-restore.md
zivkan: The current proposal is that NuGet client gets the list of vulnerabilities from the NuGet package source. How the server implements the vulnerability information resource is up to the server implementation. They certainly could use osv.dev.

Looking at osv.dev, I don't see a way to download the entire database (unless I need to clone, build, and run the github repo?). So, from a NuGet client point of view, osv.dev doesn't fit well with our design.

This could be something added in the future if there's enough demand, but the current plan is that NuGet server implementations are responsible for providing the client with the relevant data.

kapsiR: Thanks to clarify on that, maybe it's worth adding to the docs? 

FYI there are data dumps available:
https://github.com/google/osv.dev#data-dumps

kapsiR: And more important: GH advisory db is a source of osv.dev! 😉

CamiloTerevinto:
This sounds very interesting, thanks @JonDouglas. I have a few comments:

* CLI switches to run/skip auditing 

It's quite common for an entire build to be run twice (feature branch and dev/main branch) for a change in environments with CI/CD. I think that for this to be useful, we should have an easy way of telling `dotnet restore` to run or skip the audit (for example, run the audit on the PR build, skip it in the dev build).

* Fail on X level

It would be useful to have a way to make `dotnet restore` exit as failure when a package with a particular vulnerability level is found (such as treating `NU1904` as error instead of warning).  
However, this might be doable already through something like `<WarningsAsErrors>NU1904</WarningsAsErrors>`.

proposed/2022/vulnerabilities-in-restore.md
zivkan: FWIW, I tested a different file format for the vulnerability data, trading off computation speed vs file size, with the idea that smaller file size will parse more quickly, so the slower per-package computation might still come ahead. Benchmarks suggest yes, overall perf is better, and the current file size for all github reviewed vulnerabilities is 175KB uncompressed, 9KB gzipped. So, we'll be going ahead with the smaller file size format instead now.

nkolev92:
> CLI switches to run/skip auditing

Any CLI switch basically doubles as an MSBuild property. We're proposing MSBuild properties for opting into auditing, so that automatically makes it available as a CLI switch. Example: `dotnet restore -p:NuGetAudit=disable`

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,406 @@
...
+}
+```
+
+While the nuget.org data only contains vulnerability information about packages on nuget.org, often times an `upstream` source is setup by customers, and while they reference their own private feed, they can still download packages from nuget.org. The vulnerability information is valuable in these scenarios.
+
+If this upstream approach is taken, it is very likely that the vulnerability checking has a lot of misses, but this is an understood trade-off.
+
+It is **very important* that if the upstream sources link to the nuget.org vulnerability, the actual nuget.org urls are used wherever possible, as this would allow the client to easily deduplicate the vulnerability information.
+
+### Command Line Restore output
```
zivkan, line 371: @JonDouglas FYI, I moved the vulnerability summary output on command line restores under "Future Possibilities" for two reasons. 1. So we can get a V1 of this feature out sooner. 2. I don't believe this proposal has been specced out well. I have a bunch of questions if we want to implement it.

For example, in the sample output, if there are non-vulerability warnings or errors, where in the output do they go?  Before the vulnerability summary, or below with the sample vulerability warnings?   If before, should vulerability warnings/errors be output twice, once between the summary once after?

What's the expected order of output when restoring a solution with many projects? Wait until all projects finish restoring, and then output all their summaries?  Or output project summaries as each project finishes restoring, even if it means that the vulnerability summary for that project will be somewhere much earlier in the restore log?

When restoring a solution where the same package version is in multiple projects, should we de-duplicate the output?

Anyway, as I said, I think we can take this on in the future if we still think it's useful. Let's get a V1 of the feature done without it, and see how we and customers feel about it.

proposed/2022/vulnerabilities-in-restore.md
zivkan: I updated the spec to try to make it clearer that this is our suggestion, but servers can implement the data partitioning in different ways. The client will just look at the list of files, and their update timestamps to determine whether it can use a previously cached download or not.

proposed/2022/vulnerabilities-in-restore.md
zivkan: I updated the spec to try to make this clearer.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,406 @@
...
+| Warning Code | Severity |
+|--------------|----------|
+| NU1901 | low |
+| NU1902 | moderate |
+| NU1903 | high |
+| NU1904 | critical |
+
+When there are packages used by the project with vulnerabilities, they will be outjust like like any other warning or error:
+
+```text
```
nkolev92, line 63: If you put bash, you get better highlighting :) Not a big deal :D

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,406 @@
...
+  }
+]
+```
+
+- The page within the vulnerability resource **must** have the `packageid` as a primary key.
+- The value of that **must** be an array of objects.
+- The objects within the array of objects **must** contain:
+  - Severity, an `int`, where `0` means `low`, `1` means `medium`, `2` means `high`, `3` means `critical`
+  - Advisory url, a url.
+  - Versions, a version range in [NuGet range syntax](https://learn.microsoft.com/nuget/concepts/package-versioning#version-ranges) of affected package versions.
```
nkolev92: maybe mention that it includes the prerelease versions in the range as well?

proposed/2022/vulnerabilities-in-restore.md
nkolev92: I think some details, like the ones about the handling of multiple vulnerability for a package can still be in the main part as that's likely going to be done.

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,406 @@
...
+Restore output could be modified to make vulnerability information more explicit.
+
+```bash
+Found 2 vulnerabilities (0 low, 1 moderate, 0 high, 1 critical) in 2 package(s)
+
+warning NU1904: Package 'Contoso.Service.APIs' 1.0.3 has a known critical severity vulnerability, https://github.com/advisories/GHSA-1234-5678-9012.
+warning NU1902: Package 'Contoso.Forms' 8.4.1 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-1234-5678-9012.
+```
+
+If a package has more than 1 vulnerability, each vulnerability will have its own warning. These security issues are often likely independent and may require independent fixes.
```
nkolev92: This can probably be in the main part.

nkolev92 (commented):
Great updates @zivkan

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: Let's see what this shapes to be while working to answer the unknown. Thanks for being so thorough with questions to be answered.

proposed/2022/vulnerabilities-in-restore.md
zivkan: The only highlighting I see are the words `in`, `for` and the `<` and `>` characters. Which make sense in bash scripts, but in my opinion it doesn't make sense for these samples.  Why in the sentence "Found 2 vulerabilities in 1 page(s)" should only the word `in`, be highlighted?  🤷

nkolev92: Here's what I am seeing: 

image

The quoted part is being highlighted. It's not a big deal either way, and most definitely a nit :) Your call.

nkolev92 (approved):

JonDouglas:
We have approval from @nkolev92, but @zivkan did you want to approve before merge?

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -0,0 +1,421 @@
...
+
+For example, an Azure Artifacts source could have this in their index.json:
+
+```json
+{
+  "resources" [
+
+  {
+    "@id": "https://api.nuget.org/v3/vulnerability.json",
+    "@type": "VulnerabilityInfo/6.3.0",
```
zivkan:
```suggestion
    "@type": "VulnerabilityInfo/6.6.0",
```

Oops, I already changed this elsewhere in the doc, but I missed it here.

zivkan (approved):

Issue #6437, opened 2018-01-14
Why is there no 'dotnet nuget setApiKey' command?


jamesqo:
I remember that before, I could write `nuget.exe setApiKey XXX`, and then do `nuget push` in a script without specifying my API key so other people couldn't see it in the script. Now, it's necessary to specify the API key as a parameter when you do `dotnet nuget push`, and there is no `dotnet nuget setApiKey` command. Is there a reason for this?

PatoBeltran:
Hey @jamesqo thanks for your feedback. `nuget.exe setApiKey` was a simple command that only added the given api key (encrypted) to your `nuget.config` (which is read by the others nuget commands). There is no specific reason why it hasn't been ported to dotnet.exe. I will tag this issue as a feature request an send it to the backlog.

As a workaround you could use `nuget.exe setApiKey` (or set the values manually in your `nuget.config` and then use `dotnet nuget push`. `dotnet.exe` should read the `nuget.config` and use the api keys found there.

dangershony:
FYI I had to go to `\AppData\Roaming\NuGet\nuget.config` and delete the previous key for the new key to take effect

FlorianHockmann:
>  or set the values manually in your nuget.config

@PatoBeltran Is there any information on how the API key needs to be encrypted? What is used as the key to encrypt the API key?

rrelyea:
I believe this might be similar to encrypting credentials...we don't have cross platform APIs for this available via Dotnet Core.

FlorianHockmann:
So, it is really not possible right now to push NuGet packages to nuget.org on non-Windows platforms without Mono?

lucastheisen:
@PatoBeltran ,

> As a workaround you could use nuget.exe setApiKey (or set the values manually in your nuget.config and then use dotnet nuget push. dotnet.exe should read the nuget.config and use the api keys found there.

This does not appear to work.  I tried:
```
nuget setApiKey <OMITTED> \
    -ConfigFile ~/.nuget/NuGet/Nuget.Config \
    -Source  https://pastdev.com/artifactory/api/nuget/pastdev-local
```
Verified it was added:
```
  <apikeys>
    <add key="https://pastdev.com/artifactory/api/nuget/pastdev-local" value="OMITTED" />
  </apikeys>
```
Then attempted a push:
```
bash-4.2$ dotnet nuget push artifacts/Pastdev.Cli.1.0.0.nupkg
error: Encryption is not supported on non-Windows platforms.
```

[...]

Issue #2352, opened 2016-03-17
NuGet internationalization: switch-off required


jochenwezel:
To support tools based on NuGet, it's required to analyse the output messages of NuGet by these tools (e.g. the CreateNewNuGetPackage). 

The problem is on non-English systems that NuGet automatically logs every message in the supported language - making it very difficult to analyse NuGet output messages (required lot of internationalization work on those depending projects).

For simplifying development for those tools depending on NuGet output messages, an additional command line parameter/switch would be great to disable internationalization of output messages and to just output everything with messages in English language.

Example:
NuGet.exe pack niceApp.csproj -ForceEnglishUI

deadlydog:
Yes, yes, yes please! Great idea :+1:

yishaigalatzer:
@joelverhagen - I think this will take some reflection magic, but I think it is doable by iterating all the loaded assemblies finding the resource managers (flaky I know) and forcing the `resourceCulture` variable to English

joelverhagen:
Fixed with https://github.com/NuGet/NuGet.Client/pull/451.

Per the original request...

> NuGet.exe pack niceApp.csproj -ForceEnglishUI

We chose to make the option name `-ForceEnglishOutput`.

jochenwezel:
absolutely great! :+1: 

One additional question because most depending tools are developed to support older Visual Studio versions, too.

> NuGet feed v2 (VS 2013 and earlier / NuGet 2.x)

If I understand it correctly, NUget 3.5 will be available only for VS2015 or higher because earlier VS versions depend on NuGet 2.x.
Are these changes also visible to VS <= 2013 ?

yishaigalatzer:
These changes are only for nuget.exe, not inside visual studio. We are unable to make these in visual studio proper without a deeper discussion and approvals from the visual studio team.

We are not planning on further investments in 2013 other than critical bug fixes or major blocking issues like new target frameworks.

jochenwezel:
@yishaigalatzer Thank you for this information. I hoped a different result, but I understand your decision, too :-)

alpaix:
Re-opening due to a bug reported by a customer in NuGet/NuGet.Client#451

joelverhagen:
Closed because additional work is tracked here https://github.com/NuGet/Home/issues/3191.

PR #12321, opened 2022-12-15
`dotnet nuget apikey` command spec

zivkan:
spec for https://github.com/NuGet/Home/issues/6437


accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,230 @@
...
+
+Add the `dotnet nuget apikey set` command, to allow customers to set a nuget.org api key in their `nuget.config` file. Also add `dotnet nuget apikey unset`, so customers can clean up secrets easily. Additionally, `nuget.config` will be changed to allow unencrypted (plain text) API keys.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+- `dotnet nuget apikey set`
+
+nuget.org is one of the only package sources that uses API keys (`X-NuGet-ApiKey` HTTP header), rather than source credentials (`Authorization` HTTP header).
```
loic-sharma: Could you expand on what source credentials are? Are these the tokens created by credential providers? Also, don't most private package sources (NuGet.Server, GitHub's NuGet package registry, BaGet) use API keys?

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,230 @@
...
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+- `dotnet nuget apikey` commands `set` and `unset`
+
+The [`dotnet nuget config` spec](https://github.com/NuGet/Home/pull/12172) proposed `set` and `unset`, to mimic `git config` commands, rather than `add` and `remove` or `delete`, which commands like `dotnet add package` and `dotnet nuget add source` use.
```
loic-sharma: FYI there are `dotnet user-secrets set` and `dotnet user-secrets remove` commands (also git also has `git remote remove`). If you haven't already, consider double-checking what's considered idiomatic by the .NET tool team.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: Hey @loic-sharma, it's been a long time!  Good to hear from you.

I went through the [list of NuGet server implementations in our docs for hosting private feeds](https://learn.microsoft.com/en-us/nuget/hosting-packages/overview), and I found about half of them use API key, and the others use package source credentials. (GitHub Package Repository surprised me, because [now they use API key](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry#publishing-a-package), but my memory is correct in that [they used to use package source credentials](http://web.archive.org/web/20210506213928/https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry)).

Anyway, I updated the paragraph, so it's more accurate now, and hopefully more clear as well.

loic-sharma: Gotcha, make sense. Thanks!

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+- Andy Zivkovic ([zivkan](https://github.com/zivkan))
+- Start Date (2022-12-15)
+- GitHub Issue (https://github.com/NuGet/Home/issues/6437)
+- GitHub PR (not yet implemented)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+
+Add the `dotnet nuget apikey set` command, to allow customers to set a nuget.org api key in their `nuget.config` file. Also add `dotnet nuget apikey unset`, so customers can clean up secrets easily. Additionally, `nuget.config` will be changed to allow unencrypted (plain text) API keys.
```
donnie-msft:
```suggestion
Add the `dotnet nuget apikey set` command, to allow customers to set a NuGet server API key in their `nuget.config` file. Also add `dotnet nuget apikey unset`, so customers can clean up secrets easily. Additionally, `nuget.config` will be changed to allow unencrypted (plain text) API keys.
```

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+Therefore, risk can be reduced by providing customers commands to add and remove the API to the `NuGet.config` file, especially if it can be done in a way where the secret is not entered into, or output from, the console.
+
+In `nuget.config`, the API keys are in a different section to source credentials.
+There is a command `nuget.exe setapikey` to set the API key in a config file, however, `nuget.exe` does not ship in Visual Studio or the .NET SDK.
+Also, `nuget.exe` requires mono on Linux and Mac, since `nuget.exe` is built for the .NET Framework.
+Therefore, a command in the .NET SDK will significantly reduce the barrier for customers to set values in the `nuget.config` file.
+
+- `dotnet nuget apikey unset`
+
+Customers may want to remove secrets from their `NuGet.config` files, so an `unset` command should be added.
```
donnie-msft: And this will require `--source`, or will it wipe all keys it can find and let the user know where it searched and what it unset?

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+  -f, --from-file            Treat the <VALUE> argument as a filename, whose contents contain the API key.
+  -s, --source <SOURCE>      The package source which this API should be used with. [Required]
+  --configfile               The NuGet.config file to write the API key to. [Default: c:\users\zivkan\AppData\Roaming\NuGet\NuGet.config]
+  --store-in-clear-text      Don't use .NET's ProtectedData APIs to encrypt API key. Required on platforms that don't support the API.
+  -?, -h, --help             Show command line help.
+```
+
+Note on `--configfile`:
+
+NuGet has been using `--configfile`, rather than `--config-file` for multiple years, so this "incorrect" name is used to remain consistent with other commands.
```
donnie-msft, line 63: Does consistency matter in this case, given it's a new command? I can understand not wanting to break `dotnet restore` by changing it suddenly (but maybe we could create another alias for the same option? Not sure how motivated we and customers are on this)

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+- script: dotnet nuget apikey set -e APIKEY
+  name: Set nuget.org API key
+  variables:
+    APIKEY: $(SECRET_API_KEY_FOR_NUGET_ORG)
+```
+
+Note on `--source`:
+
+The technical feasibility needs to be checked, but `NuGet.config` has a key `defaultPushSource`.
+When this is defined, `--source` should be optional and if `dotnet nuget apikey set` is run without specifying a value for `--source`, the default push source should be used.
```
donnie-msft: > if `dotnet nuget apikey set` is run without specifying a value for `--source`, the default push source should be used.

Really? Isn't this default making it easier to leak API keys to other sources? Would a customer really expect this behavior, and is it a true convenience?

Scenario example... Given:
- SourceA (defaultPushSource)
- SourceB

I want to push a package to **SourceB**:
- `dotnet nuget apikey set <VALUE>` // SourceA's API key is set (intended SourceB)
- `dotnet nuget push` // Intended to push SourceB, but SourceA is default
- SourceA just received my API key for SourceB. I'm not sure if there's a guaranteed error if SourceA doesn't need an API Key, but receives one anyway?

Does the customer know this happened? Is it unlikely that they are always aware when `defaultPushSource` is enabled?

accepted/2023/dotnet-nuget-setapikey.md
dominoFire: Let's use option alias, so we can have both `--configfile` and `--config-file`

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+- `dotnet nuget apikey`
+
+Some care needs to be taken when implementing the default values for options.
+
+For `--configfile`, Linux is case-sensitive, and a customer might be using `nuget.config`, rather than `NuGet.config`.
```
dominoFire: Let's use the default name used in `dotnet new nugetconfig`, which is `nuget.config`

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+The default value factory will need to check the filesystem for which casing actually exists, and only fall back to the default casing if none are found.
```
dominoFire, line 102: I think this add more complications. Let's just use the default nuget.config file name (lowercase) and do not check for file system casing.

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+## Rationale and alternatives
+
+### Which NuGet.config file
+
+At the time this spec is written, all NuGet commands that can write values in `NuGet.config` files have the following unintuitive behavior.
+The docs on [common NuGet configurations](https://learn.microsoft.com/nuget/consume-packages/configuring-nuget-behavior) explains that there's a hierarchy of directories and files checked.
+When making a change, NuGet will look for the first file in that ordered list that contains the relevant section (first child of the XML root element).
+For example, `nuget.exe config` will look for the first file that contains `<config>`, and `dotnet nuget add source` will look for the first file that contains `<packageSources>`.
+While package sources, and certainly other config settings are valuable to be in a solution `NuGet.config` file, and therefore commit into source control, accidentally committing a secret into source control is a security risk.
+Therefore, `dotnet nuget apikey set` will always default to the user scope `NuGet.config` file.
```
dominoFire: How about introducing a new option, named `--use-new-config-lookup` to locate the config file defaulting to the first nuget.config file found?

accepted/2023/dotnet-nuget-setapikey.md
dominoFire: Referring to this comment https://github.com/dotnet/sdk/issues/29480#issuecomment-1347426073 ; on `dotnet nuget` commands, NuGet team can decide which approach is better: noun-first or verb-first. I think as long as we're consistent withing NuGet commands, we're good.

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+- `dotnet nuget apikey unset`
+
+Customers may want to remove secrets from their `NuGet.config` files, so an `unset` command should be added.
+
+- `nuget.config` support for unencrypted `<apikeys>`
+
+Currently `nuget.config` `<apikeys>` section only supports encrypted values, via .NET's [`System.Security.Cryptography.ProtectedData` APIs](https://learn.microsoft.com/dotnet/api/system.security.cryptography.protecteddata).
+As the important message says on the docs page, on Windows it uses Windows' Data Protection API (DPAPI), which isn't supported on Linux or Mac, and [the .NET libraries team have no plans to support it since it will have different security promises than on Windows](https://github.com/dotnet/runtime/issues/22886).
+There is [an issue to consider another way of having less secure "encrypted" passwords](https://github.com/NuGet/Home/issues/1851) on Linux and Mac (like what Mono does for the `ProtectedData` API), but it hasn't been implemented yet.
+Therefore, in order for this feature not to be blocked, `<apikeys>` needs to support unencrypted (plain text) secrets (just as source credentials do).
```
dominoFire: Why not reuse <packageSourceCredentials> with a new entry for api keys, both encrypted and unencrypted?

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+It still uses the `<add key="string" value="string">` syntax, allowing it to use `<add key="cleartextpassword" value="secret" />` and `<add> key="password" value="encrypted" />`.
+But that's because these elements are nested within an element whose name is the package source. To try to do the same with `<apikeys>` would require a breaking schema change, since currently it uses `key` as the package source and `value` as the API key.
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- Need to validate old nuget behavior when `apikeys` contains unexpected `encryption` element. Validate both read and write.
```
dominoFire: On write: nuget.exe just overwrites the `<add/>` entry.
On read: nuget.exe just ignores the bool value.

dominoFire (commented):
Looks good. I suggest focusing this spec on the command line implementation and file issues separately for new encryption and config file lookup.

accepted/2023/dotnet-nuget-setapikey.md
kartheekp-ms: [NuGet.exe setapikey  command docs](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-setapikey#usage) state that `If <source> is omitted, nuget.org is assumed.` If that's not the case we should consider updating the docs.

accepted/2023/dotnet-nuget-setapikey.md
```diff
@@ -0,0 +1,234 @@
...
+Description:
+  Set the API key in a NuGet.config file.
+
+Usage:
+  dotnet nuget apikey set <VALUE> [options]
+
+Arguments:
+  <VALUE> The value of the API key. If neither --from-file or --environment-variable is specified, the value is treated as a string literal.
+
+Options:
```
kartheekp-ms, line 53: nuget.exe setapikey command has `--ForceEnglishOutput` option added to address https://github.com/NuGet/Home/issues/2352 customer ask. I don't see that option in this command. Is that intentional? Please let me know if I am missing something.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: Source is required. In the later section below, where I showed sample CLI help output, the `--source` option has `[Required]`.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: I don't know if aliases were an option with our old CLI parsing library, but it's certainly possible with System.CommandLine.  Good idea.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: @donnie-msft I find the scenario a bit improbable. If SourceA is the default push source, why would the customer use `apikey set` without specifying a source, and expect it to be SourceB?  Kartheek did bring up an interesting point, that `nuget.exe setapikey` appears to default to nuget.org. So, SourceB being nuget.org is the only scenario that makes any sense to me. So with that knowledge, I can accept that security is more important than convenience and that we can make `--source` a required option.

@kartheekp-ms this spec is specifically about a new `dotnet nuget apikey set` command, not the `nuget.exe` command.  I don't know if the docs are correct, but I assume they are. 

@JonDouglas any feedback?  I don't think we need to (or should) carry over nuget.exe defaults in the dotnet CLI (msbuild have changed some defaults, for example). If my proposal goes accepted, then customers publishing packages to nuget.org, when they migrate from `nuget.exe setapikey` to `dotnet nuget apikey set` will need to start specifying nuget.org's service index URL (or source name, assuming they have it in their nuget.config file). While I think it would have been nice to use the default push source for `apikey set` and `apikey unset`, I think `nuget.exe`'s default behaviour adds risk, so perhaps we should make `--source` a mandatory option.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: I disagree. NuGet will create the file, if none of the 3 files already exist, with the default of `NuGet.Config`. Therefore, I expect the vast majority of customers to have `NuGet.Config`, not `nuget.config`. However, it also wouldn't surprise me if we have a lot of `nuget.config` files that were manually created. `NuGet.config` would surprise me, but if we're going to check for 2, we might as well check for all 3.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: I assume that's a typo, you meant` --use-old-config-lookup`?

As I wrote in the spec, I think having the secret written into a file that might be in source control has risk (although adding it to the user-profile nuget.config on a CI agent that keeps state between builds also has risk). But I don't understand why customers would want to have the secret written to a "random" file. I mean, it's technically deterministic, but I don't see using `--config-file` as difficult for customers. An extra option for what I consider to be confusing behaviour doesn't feel worthwhile to me.

accepted/2023/dotnet-nuget-setapikey.md
zivkan: The `dotnet` CLI already has a method to change the output language, [the `DOTNET_CLI_UI_LANGUAGE` environment varliable](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_cli_ui_language), which Fernando recently added support for in our tool commands: https://github.com/NuGet/NuGet.Client/pull/4816

So, I don't want to have an option for one command that does not exist in all the other commands, especially when there's already a supported way to achieve the same thing.

accepted/2023/dotnet-nuget-setapikey.md
donnie-msft: >  I find the scenario a bit improbable. If SourceA is the default push source, why would the customer use apikey set without specifying a source, and expect it to be SourceB? 

@zivkan Are you saying humans never make mistakes?! 😲
What about scripts, can they not make assumptions and rely on a default, but be affected by a context change?

>  I can accept that security is more important than convenience and that we can make --source a required option.

Well, it sounds like you're slowly agreeing to do this anyway, but only because nuget.org is an implied default. Either way, I prefer consistency and security over hidden defaults. Open to hear arguments against it.

accepted/2023/dotnet-nuget-setapikey.md
smarts: I do like set/unset over add/remove. `add` IMHO is ambiguous regarding what happens when you attempt to overwrite. I.e., Will it overwrite the old value? Will it ignore the new value because a value is already present? Will it return an error because a value is already present? I might have to read docs to find out whereas `set` is pretty obvious.

JonDouglas (approved):
Great spec.

My only high level thought on this command is the ordering of the command.

The current proposal is `dotnet nuget apikey set`

Many of the dotnet commands tend to put verb before the noun. I think we should remain consistent with our commands although it is wildly inconsistent across the CLI.

i.e.

`dotnet nuget set apikey`

JonDouglas:
The only other thoughts I have around this command is whether we should have a standalone apikey command and/or to extend the existing source commands that will write the `<apikeys>` section appropriately? It's always nice to have multiple ways of doing the same thing.

i.e. `dotnet nuget add source --apikey`

accepted/2023/dotnet-nuget-setapikey.md
JonDouglas: I don't think we ought to carry over the defaults. This command as designed has a required field that will fail otherwise. I think that is reasonable design instead of assuming a central registry.

accepted/2023/dotnet-nuget-setapikey.md
JonDouglas: All great points.

I think in the scope of this command where we are adding a new apikeys element to a config, you could use the `add/remove` consistency as seen in other commands.

I know it isn't always representative that it can be overwritten (i.e. dotnet add package), but it is the term that was chosen and would be remembered more easily by CLI users who use all other CRUD operations. 

My second vote would be `set / remove`.

I personally see this command as a means to "add a apikey to a source". Similar to if you were to just add the apikey in the `dotnet nuget add source --apikey <blah>` command (if it was supported)

nkolev92 (approved):

accepted/2023/dotnet-nuget-setapikey.md
smarts: @zivkan are you talking about nuget.exe, or `dotnet nuget`?
+1 for using `nuget.config`. It's a simple migration to do for users who need to adapt.
My second vote would be for a simple try one then try the other for `nuget.config` and `NuGet.Config`.

zivkan: The code that checks the files with different casing is shared in both nuget.exe and the dotnet cli, so they both work the same way, in regards to what order the 3 different casings are checked in: https://github.com/NuGet/NuGet.Client/blob/7a06880721feb769b5e6edcc123dacc7b5566414/src/NuGet.Core/NuGet.Configuration/Settings/Settings.cs#L29-L37

This is already a public API in the dll, so there's no need to re-invent anything. The person who implements this spec "just" needs to use the array.

zivkan:
> The only other thoughts I have around this command is whether we should have a standalone apikey command and/or to extend the existing source commands that will write the <apikeys> section appropriately?

A standalone `apikey` command lets customers who don't use nuget.org as a package source (say they're using Azure Artifacts with upstreaming) still publish to nuget.org. At least, in theory this could work. I know NuGet is inconsistent whether `--source` parameters on different commands are source names, source URLs or can be either. I don't know which `push` implements.

Adding an `--apikey` option to `add source` and `update source` are good ideas too. If I were smart, I'd just update the spec now, but I want to do some other things first, so hopefully I'll remember to come back to this and update it later.

jeffkl (approved):

Saibamen:
https://github.com/NuGet/NuGet.Client/pull/7322

Issue #1851, opened 2015-12-18
nuget source password encryption story


yishaigalatzer:
Rob adding text on 2/10/2017:
nuget.exe | dotnet.exe can now use encrypted passwords in nuget.config files -- when running on windows ONLY.
Ideally, we'd support dotnet.exe on other platforms too, so they don't need to store passwords via cleartext.

This would need APIs from .NET Core. See the related https://github.com/dotnet/corefx/issues/6746

yishaigalatzer:
CC @anurse @davidfowl

emgarten:
Is this needed for xplat?

yishaigalatzer:
Yes but not for rc, and debatable for rtm

yishaigalatzer:
@emgarten when RC is out the door, please start a thread with Stephen Toub

mountgellert:
Hi @rrelyea , is there any update on this or on https://github.com/dotnet/corefx/issues/6746 ?

mountgellert:
@mrward , How does your change fix this? Still cannot seem to see a way to use encrypted credentials from a Linux machine.

mrward:
@mountgellert - For some reason merging the latest code into my fork closed this issue when it should not have. Re-opening.

psmulovics:
Any news on this item?

guidobouman:
What's the roadmap on this? Is the NuGet or the dotnet project responsible for this implementation? I can't seem to find anything about this in https://github.com/dotnet/designs/issues/11

psmulovics:
any news on this item?

ghost:
Bump for update.

koliyo:
Please make this a priority, if Microsoft is serious about true cross-platform .NET application development.

We need both MacOS and Linux support!

dquist:
Another serious shortcoming here is that, even on Windows, there's no way to securely enter the password. It can only be provided as plain text through the `-Password` option, in which case it will likely be logged to the shell history.

Come on, this is like CLI authentication 101. There needs to be a way to enter secrets interactively.

See https://github.com/NuGet/Home/issues/8131

Swoorup:
* Upgrades everything to use dotnet core instead of mono. 
* Everything works fine locally. 
* Hopes on to build machine, which repulls packages.
* Notices this damn bug. 
* Cleans packages locally, notices this damn bug.

MaxHorstmann:
> Please make this a priority, if Microsoft is serious about true cross-platform .NET application development.

[...]

PR #12354, opened 2023-01-10
Allow unecrypted API keys in nuget.config

zivkan:
Without this API keys saved in nuget.config files can't work on Linux and Mac, unless https://github.com/NuGet/Home/issues/1851 is implemented.


jeffkl (approved):

JonDouglas (approved):
Great spec. Straight to the point and unblocks other platforms. 🚢

accepted/2023/nuget-config-apikeys-unencrypted.md
```diff
@@ -0,0 +1,115 @@
...
+
+We have got [feedback from customers](https://github.com/NuGet/Home/issues/3245) who manually entered clear text package credentials where NuGet expects encrypted credentials, and NuGet will crash with an obscure error message ("The parameter is incorrect") that is not actionable by customers.
+While we are working on API keys, I propose that we should additionally catch the decryption of encrypted API keys, and if DPAPI throws an exception, report a meaningful error message back to the customer.
+
+The `<packageSources>` section also uses the `<add key="string" value="string">` syntax that most other parts of `nuget.config` use, but it also supports the `protocolVersion="int"` attribute.
+The approach it takes to add `protocolVersion` can be copied to make the `encrypted` attribute work.
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
```
blowdart, line 52: You know why :) So fill this out.

- Users run the risk of checking the file into source control.
- Users run the risk of putting this into a directory not properly ACLed against other users on the machine
- Users run the risk of malware on their machine reading the credentials and sending them elsewhere.

accepted/2023/nuget-config-apikeys-unencrypted.md
```diff
@@ -0,0 +1,115 @@
...
+If a customer runs any command that requires `nuget.config` to be modified, for example `dotnet nuget add source`, NuGet will deserialize the existing config, modify the value as per the command, and then serialize back to disk.
+Since older versions of NuGet don't know about the `encrypted="true/false"` attribute, this attribute will be lost, and then if a newer version of NuGet is run, it will assume that the value is encrypted, and then cause errors.
+
+If NuGet doesn't already handle maintaining sections of the config file that it doesn't know about, then there is no mitigation for this.
+Using a different section name, like `<clearApiKeys>` will still get lost, but at least it won't cause DPAPI to try to decrypt an unencrypted string, and therefore improving the error message will not be as important.
+
+### Unencrypted secrets
+
+Since the API key is a secret, providing a way to reduce security might not be desirable.
+However, at the time that this design spec is being written, there is no alternative on Linux and Mac, making API keys in the `nuget.config` file a Windows-only feature, which is also not desirable.
```
blowdart: Is the Apple Keyring usable from the command line, or without user interaction? If that is possible then on an MacOS device put the credentials in the keyring, and allow the expanded spec to include something like location in the key syntax, and require either a value or a location.

accepted/2023/nuget-config-apikeys-unencrypted.md
agr: I would assume that the last point is not mitigated by having encrypted API keys (if nuget client can decrypt it, so can malware), so it should be irrelevant in this context.

accepted/2023/nuget-config-apikeys-unencrypted.md
zivkan: Oops, I forgot to respond to this earlier. Perhaps I should change the "there are no alternatives" wording, but unfortunately I have zero experience with Mac since I was in high school, which was before OSX ever came out. All this to say I know nothing about Apple Keyring.

As the spec says elsewhere, we have a different issue to add encryption for Mac and Linux, and I hope that whoever researches/implements that will take Apple's Keyring into account: https://github.com/NuGet/Home/issues/1851

accepted/2023/nuget-config-apikeys-unencrypted.md
```diff
@@ -0,0 +1,124 @@
...
+- GitHub Issue (https://github.com/NuGet/Home/issues/6437)
+- GitHub PR (not yet implemented)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+
+Enable unencrypted API keys in `nuget.config` files, so that they can be used on platforms other than Windows.
+
+## Motivation
```
zivkan, line 14: Last week I had an idea which probably means that the benefit of allowing unencrypted API keys in nuget.config becomes close to zero. It would still be useful for customers who use Mac and Linux (until https://github.com/NuGet/Home/issues/1851 is implemented) and push packages with APIs from their dev box.  But most of the feedback we got was from customers pushing in CI scripts and docker containers, so I think this will satisfy most people:

* https://github.com/NuGet/Home/issues/12539

accepted/2023/nuget-config-apikeys-unencrypted.md
zivkan: The "Drawbacks" section is a 2nd level header, and I added two 3rd level headers under. If you look at the rendered spec, it's much more clear. I believe these points are addressed in the "unencrypted secrets" sub-heading, but I did update the spec last week which hopefully made it a little better as well.

ghost:
This PR has been automatically marked as stale because it has no activity for **30 days**. It will be closed if no further activity occurs **within another 15 days** of this comment, unless it has a "Status:Do not auto close" label. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

PR #12408, opened 2023-02-03
FabricBot: fix a condition when switching "waiting for" label

heng-liu:
Fix a condition when switching "waiting for" label.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -237,10 +237,21 @@
               "parameters": {}
             },
             {
-              "name": "activitySenderHasPermissions",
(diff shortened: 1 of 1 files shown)


erdembayar (approved):

nkolev92 (commented):
Should all places where we have activity sender permissions checks contain both admin and write?

jeffkl (approved):

heng-liu:
> Should all places where we have activity sender permissions checks contain both admin and write?

I'll verify if this fixes the issue @zivkan encountered. If yes, then I'll apply this to all places.

PR #12529, opened 2023-04-06
Update "vulnerabilities in restore" spec

zivkan:
* Add restrictions to `@name` element in the vulnerability index's page definition.

This minimizes risk that a malicious payload can harm clients

* Change the client version yo 6.7.0 in the service index.

NuGet didn't implement this in time for 6.6.0.

### DIFF
--- a/proposed/2022/vulnerabilities-in-restore.md
+++ b/proposed/2022/vulnerabilities-in-restore.md
@@ -146,7 +146,10 @@ Periodically, the small file's data can be merged into the large file(s), and th
 
 - The vulnerability resource **must** be an array of objects.
 - Each object **must** contain:
-  - `@name`, a user friendly name for the page, **must** be unique.
+  - `@name`, a short name for the page, used for caching, and has a few restrictions:
+    - **must** be unique.
+    - **must** be between 1 and 32 characters long.
+    - **must** only contain  characters `A` to `Z`, `a` to `z`, `0` to `9`, or be `-` or `_`.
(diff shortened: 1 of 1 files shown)


proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -146,7 +146,10 @@ Periodically, the small file's data can be merged into the large file(s), and th
 
 - The vulnerability resource **must** be an array of objects.
 - Each object **must** contain:
-  - `@name`, a user friendly name for the page, **must** be unique.
+  - `@name`, a short name for the page, used for caching, and has a few restrictions:
+    - **must** be unique.
+    - **must** be between 1 and 32 characters long.
+    - **must** only contain  characters `A` to `Z`, `a` to `z`, `0` to `9`, or be `-` or `_`.
```
zivkan, line 152: I know this is very English-centric, which isn't my intention, but I'm not enough of a Unicode expert to know if using [`char.IsLetterOrDigit`](https://learn.microsoft.com/en-us/dotnet/api/system.char.isletterordigit?view=net-7.0) can be abused, such that if the value is used as a filename, it can somehow escape outside the directory it's intended for.

It's easier to become less restrictive in future versions than become more restrictive. Although in either case, servers that want to support all client versions will always need to use the most restrictive set ever implemented.

jeffkl (approved):

nkolev92 (approved):

Issue #12015, opened 2022-08-10
The http removal will break private NuGet servers (NU1803 warning)


peymanr34:
I completely support the effort to force users to use https, however I'm concerned about not being able to use private NuGet servers (such as BaGet) on my local machine since:

> In November 2024, we will throw an error when a non-HTTPS source is used. You will not be able to opt-out of this behavior. https://devblogs.microsoft.com/nuget/https-everywhere

Please consider keeping the opt-out option permanently or suggest an alternative solution.

nkolev92:
@peymanr34 

NU1803 is a warning. It is likely that your project is elevating it to an error.

erdembayar:
Do we dup to https://github.com/NuGet/Home/issues/12013?

peymanr34:
@nkolev92 
> NU1803 is a warning. It is likely that your project is elevating it to an error.

Yes, I'm well aware that currently it's a warning but it will become an error in the future (Nov 2024) without an option to opt-out.

nkolev92:
> @nkolev92
> 
> > NU1803 is a warning. It is likely that your project is elevating it to an error.
> 
> Yes, I'm well aware that currently it's a warning but it will become an error in the future (Nov 2024) without an option to opt-out.

My bad, I misread it because of the title.

Do you mind tweaking the title to something that indicates that this issue is about the complete removal of non-https feeds?

I can take a stab at it if you'd like me to.

peymanr34:
> My bad, I misread it because of the title.
> 
> Do you mind tweaking the title to something that indicates that this issue about the complete removal of non-https feeds?
> 
> I can take a stab at it if you'd like me to.

Sorry about that, please kindly do so.

timmydo:
Should this throw when the hostname is localhost?

mika76:
Totally agree with this - we have a totally private nuget server which does not even connect to the internet so please don't make me have to secure it for no reason. This should be an operational decision and not something a tool forces on you.

mhsimkin:
I agree with @mika76. We are running ProGet internally.

I need a way to indicate that a nuget source is allowed to use HTTP instead of HTTPS. That can be done via domain suffixes or flagging specific feeds.

To go through the process of either purchasing a cert or getting a cert issued internally, then added to all the windows workstations, and finally, working with my network security team to allow HTTPS traffic to a specific server is a BIG ask.

ITaluone:
+1 from me..

[...]

Issue #4387, opened 2017-01-24
NuGet cannot restore from HTTPS sources that have SSL certificate problems


joelverhagen:
## Idea

We should add a way for users to bypass certain HTTPS certificate errors. If the user has a source that does not have a valid certificate chain, they should still have some way of getting NuGet to interact with this source.

## "Inspirations"

npm and yarn have this:
```
npm config set strict-ssl false
yarn config set strict-ssl false
```

Maven has some extra knobs and switches:
```
-Dmaven.wagon.http.ssl.allowall=true
-Dmaven.wagon.http.ssl.insecure=true
-Dmaven.wagon.http.ssl.ignore.validity.dates=true
```

PHP's Composer
```json
{
    "config": {
        "disable-tls": true,
        "secure-http": false
    }
}
```

curl has it:
```
--no-check-certificate
```

## Specific Scenario

nuget.org server side runs functional tests against the staging slot of DEV, INT, and PROD that use nuget.exe. The DNS for this staging slot is some random blah.cloudapp.net. The certificate used is associated with the production DNS (e.g. *.nuget.org). This means nuget.exe fails SSL checks when running against this staging slot. 

## Proposal

Add a NuGet.config `trustedCertificates` section with standard `<add>` and `<remove>` tags. The value should be the SHA1 fingerprint (a.k.a. thumbprint) of the certificate. This allows developers to specifically opt-in into the tenuous state of SSL validation failures. The risk is mitigated since the developer must opt into it and even then only on a case-by-case basic.

Any certificate added to the `trustedCertificates` set is excluded from SSL validations -- I am thinking ALL validations exposed to use via the .NET API (untrusted root, name mistmatch, etc).

To improve discovery, we can add a log which writes out the fingerprint of the certificate that failed SSL validation at the `error` level.

We should also consider adding an environment variable to set this value. I believe npm has this and this  sibling environment variable would match other NuGet.config keys, such as the location of the user packages directory.

## Implementation

`HttpClientHandler` in `System.Net.Http` 4.3.0 (maybe earlier) has a way to set the certification validation check. This allows an app to configure the certificate validation on a per-request basis, instead of a per-AppDomain basis previously provided by `ServicePointManager` static properties.

With this approach, we should be able to implement the settings for all of our entry points: nuget.exe, .NET CLI, and (most tricky) Visual Studio.

The only possible problem is TFM compatibility with the `System.Net.Http` package. This requires investigation.

## Issues

[...]

PR #12542, opened 2023-04-14
allowInsecureConnections and disableCertificateValidation Proposal

JonDouglas:
This proposal introduces two new properties to control the behavior surrounding insecure connections (http sources) and disabling certificate validations (https sources). This is proposed to address #4387 and #12015 in the .NET 8 timeframe.

[Rendered Proposal](https://github.com/NuGet/Home/blob/insecureconnections/proposed/2023/InsecureConnectionsDisableCertificateValidation.md)

Thank You 🎉

### DIFF
--- a/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
+++ b/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
@@ -34,12 +34,12 @@ There will be multiple ways to opt-out of these two functionalities described ab
 
 #### Package Source NuGet Config
 
-A developer may have anywhere from 2+ sources defined in their projects based on recent survey data. Each package source has its own unique security requirements and a developer should have control on a per package source basis where they can apply the `disableCertificateValidation` property to a HTTPS source. If applied to a HTTP source, nothing will happen. A developer should also be able to apply the `allowInsecureConnections` property to a HTTP source. If applied to a HTTPS source, nothing will happen.
+A developer may have anywhere from 2+ sources defined in their projects based on recent survey data. Each package source has its own unique security requirements and a developer should have control on a per package source basis where they can apply the `disableTLSCertificateValidation` property to a HTTPS source. If applied to a HTTP source, nothing will happen. A developer should also be able to apply the `allowInsecureConnections` property to a HTTP source. If applied to a HTTPS source, nothing will happen.
 
 ```
 <!-- Disables certification validation on a specific https source and allows insecure connections on a specific http source -->
 <packageSources>
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" disableCertificateValidation="true" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" disableTLSCertificateValidation="true" />
     <add key="Contoso" value="http://contoso.com/packages/" allowInsecureConnections="true" />
     <add key="Test Source" value="c:\packages" />
 </packageSources>
@@ -82,7 +82,7 @@ A developer may not want to check in an insecure configuration file and may want
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
 This section will be filled out by the respective person picking up this work in more detail if more needs to be explained. 
 
-`HttpClient` and `HttpClientHandler` should be able to support this functionality easily such as a custom `ServerCertificateCustomValidationCallback` which always returns `true` meaning that any certificate presented by the server will be considered valid for the `disableCertificateValidation` functionality.
+`HttpClient` and `HttpClientHandler` should be able to support this functionality easily such as a custom `ServerCertificateCustomValidationCallback` which always returns `true` meaning that any certificate presented by the server will be considered valid for the `disableTLSCertificateValidation` functionality.
 
 As for `allowInsecureConnections`, this functionality should be fairly easy to revert the initial warning/error messages to take into account this new flag. Simply put, if there's no code flow to this property, a user will continue to see HTTPS warnings/errors encouraging best practice. 
 
--- a/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
+++ b/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
@@ -1,0 +1,124 @@
+# Allow insecure connections and disable certificate validation
+
+- [Jon Douglas](https://github.com/jondouglas)
+- Start Date (2023-04-13)
+- [#4387](https://github.com/NuGet/Home/issues/4387), [#12015](https://github.com/NuGet/Home/issues/12015)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+NuGet does not provide much functionality for controlling the decisions of trust against http and https sources. In recent surveys, we have even seen a general sentiment towards using 2 or more sources for their projects. Many of these sources have different security requirements that should be up to the developer to decide on. 
+
+There are two functionalities that need to be added to NuGet tooling to help companies move over to a "HTTPS Everywhere" ideal world.
+
+1. I have a HTTP server and don't want the HTTPS warning/error.
+2. I have a HTTPS server and I want to ignore SSL/TLS certification validation.
+
+In both of these cases, developers would like more fine-grain control to opt-out of these security best practices such in cases where they have their own local http source or are self-signing with a certificate.
+
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+When we blogged about an initial plan about ["HTTPS Everywhere" for NuGet](https://devblogs.microsoft.com/nuget/https-everywhere/), we were met with widespread feedback on scenarios we aren't always aware of. The primary scenario is that a company may have a HTTP source on a local network that they would like to remain using based on their other company security practices. There's also still a challenge of obtaining a SSL/TLS certificate which may take longer than the timeline we initially proposed (~3 years).
+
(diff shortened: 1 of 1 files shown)


proposed/2023/InsecureConnectionsDisableCertificateValidation.md
```diff
@@ -0,0 +1,124 @@
...
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+There will be multiple ways to opt-out of these two functionalities described above.
+
+#### Global NuGet.config
+
+A developer may choose to want a global setting applied to all sources in which they may allow insecure connections on any defined HTTP source and disable certificate validation on any defined HTTPS source.
```
skofman1, line 52: IMHO we should have the setting per source (as suggested below), and not have a global config. The reasoning is that this config is risky, and the user should consider if it's appropriate for each feed specifically. This will also prevent a situation when a new external feed is added and those settings are applied unintentionally.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
```diff
@@ -0,0 +1,124 @@
...
+
+```
+<!-- Disables certification validation for all HTTPS sources and allows insecure connections from all  HTTP sources -->
+<config>
+    <add key="allowInsecureConnections" value="true" />
+    <add key="disableCertificateValidation" value="true" />
```
joelverhagen, line 58: Feedback from our design review meeting. This is confusing with package signing certificate validation (which would not be impacted). Alternatives:
- `disableSslCertificateValidation`
- `disableTlsCertificateValidation`
- `disableHttpsCertificateValidation`

joelverhagen, line 58: This allows ALL HTTPS problems to be ignored, right? Some other tools allow additional trusted roots so that issues like unexpected CAs, expired certificates, mismatching domain name still cause things to fail. This can be done by providing an alternative or additional cert bundle, as mentioned here https://github.com/NuGet/Home/issues/4387#issuecomment-275274426.

There are several comments around workarounds which leverage modifying the trusted CAs (e.g. `update-ca-certificates`). This is a less drastic solution than just disabling all HTTPS checks.

ngbrown:
How NuGet validates certificates on each platform and how to update the root certificates should be explained in the docs. That should take care of many of the MiM cases for enterprise HTTPS traffic sniffing appliances. (For example NuGet pull from within docker on a enterprise owned computer.)

For my own use, I have a HTTP only NuGet feed, but could self-sign it as long as I am sure I can get NuGet to use the self-signed root.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
blowdart: Very much agree

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
blowdart: Tls please :)

blowdart: I'd be careful of disabling all validation with a single switch. I'd still want to know about name mismatches and expired certs, but that may end up super complicated. 

Might an interactive experience help? I add a package at the command line, there's an TLS error of Untrusted root, or expired cert or subject mismatch, do you want to Ignore (One Time), Ignore (always), stop

Even better would be ignored just for the presented cert. So if, in the case of expiry, you accept the current cert, it then gets updated, and expires again, when its a new cert, error again. 

So maybe a "trust this cert with this sha256"?

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
```diff
@@ -0,0 +1,133 @@
...
+```
+<!-- Allows insecure http source and ignores certificate validation for https source -->
+<insecurePackageSources>
+    <add key="Contoso" value="true" />
+</insecurePackageSources>
+```
+
+#### NuGet Environment Variables
+
+A developer may not want to check in an insecure configuration file and may want to have a local machine or CI/CD override for these scenarios. The following environment variables would behave similarly to the global NuGet config concept above.
```
blowdart, line 75: Again the global here is undesirable. It would ignore cert errors on nuget.org if someone managed to put a proxy in the way, or dns hijack. Global should be avoided.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
JonDouglas: Let's leave this conversation open for more community input. Especially to hear more cases where this might be useful.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
JonDouglas: Let's leave this conversation open for more community feedback on the usefulness of this option or if they'd like something more specific. For now this should bridge parity of other ecosystems and their high level switches.

JonDouglas: These are added in the unresolved questions for now :)

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
robindegen: I agree that it should be per source. I want to keep validating for example nuget.org, but not my internal server.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
robindegen: My main case is being forced to use https internally because http is apparently being dropped even for servers that aren't even on the internet. Interactivity will not help in our ci/cd pipelines. I just want to disabled any and all checks since the security of the internal server is completely pointless.

In conan you simply add a boolean per source if you want to validate it or not. If you do that it will just accept anything. This is simple, to the point and just works. Security should be the users responsibility, not enforced by the tool.

JonDouglas: @robindegen do you have docs to help us understand what you are referring to? Especially on what it is validating? Thank you!

robindegen: Little background for reference: With Conan the remote servers are managed system-wide (not exactly but for the sake of simplicity, you could also run it in a python virtual environment).

When adding a remote server you simply specify verify_ssl=False. What exact things this ignores and allows I don't know but it is enough for my simple use case of simply not having a trusted certificate. I think this is the most common use case when disabling ssl, an internal server that just can be trusted. In my years working in devops i only really see two use cases. Either it's a server that runs https but never had a certificate configured so it is untrusted, or the IT department forgot to renew the certificate and it takes a few days to get that fixed. I think this gets you 99% of the usecases for people that use this feature.

https://docs.conan.io/en/1.53/reference/commands/misc/remote.html (note I linked 1.x documentation, since conan 2 doesn't really have much real world users yet)

For nuget I would expect something similar in nuget.conf. Just an xml attribute per repository.

I think if you really want to verify the identity then just use a proper trust authority. A setting like this i would expect to simply be for the case where everything is internal and there is really no need to verify it. We would use simply http if it wasn't being phased out.

tmds:
One reason users want to disable certificate validation is there's no easy way to trust additional certificates.

I've written a proposal to allow a .NET app to trust additional certificates from the file system through envvars: https://github.com/dotnet/runtime/issues/86243.

This would for example allow in a docker container to obtain certificates via https, and then use them during the nuget restore.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
JonDouglas: I have redacted the other experiences based on feedback.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
JonDouglas: Thank you for your feedback. I have redacted the other experiences to scope a quick solution here.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
JonDouglas: I have redacted the other experiences based on feedback. Thank you.

JonDouglas:
This PR has been updated based on community feedback and should be ready for implementation.

nkolev92 (approved):
Not a fan of the global disabling, but seems like that's something that's been thought through/discussed.

I don't have any other concerns beyond that.

proposed/2023/InsecureConnectionsDisableCertificateValidation.md
```diff
@@ -0,0 +1,135 @@
...
+- maven wagon has a number of settings - https://maven.apache.org/wagon/wagon-providers/wagon-http/
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Are these the best descriptive names for these behaviors?
+    - Could they be simplified more?
+        - Other suggestions are: `DisableHttpsCertificateValidation`, `DisableTLSCertificateValidation`, and `DisableSSLCertificateValidation` (SSL is deprecated).
```
nkolev92, line 127: I missed this, I like `DisableTLSCertificateValidation` best.

JonDouglas: Updated as this came up during team review too.

PR #12573, opened 2023-05-03
Spec for Package Source Mapping Status in PM UI

### LINKED ISSUE #12303: Update Package Source Mapping in VS / PM UI Spec
Some changes and clarifications have been made on the UX for Package Source Mapping in VS.

donnie-msft:
Fixes: https://github.com/NuGet/Home/issues/12303

Package Source Mapping Status in the PM UI ([shortcut to rendered spec](https://github.com/NuGet/Home/blob/dev-donnie-msft-specSourceMappingPMUIEdits/proposed/2023/package-source-mapping-status-PMUI.md))
    - Merged into 17.7 Preview 2 (int preview)
    - Goal is to collect customer feedback on how to proceed with actions in Spec 2 below


proposed/2023/package-source-mapping-actions-PMUI.md
```diff
@@ -0,0 +1,102 @@
...
+### Functional Explanation
+
+When a new package source mapping is being created, the Preview Window will appear showing which packages (the top-level and any transitive dependencies) will be mapped. 
+
+(TBD) It is not yet understood whether there will be a new afforance on the details pane in a new row below the Install/Update buttons, or if source mappings will be handled in the background.
+
+#### A Proposal for Consent to a new Source Mapping
+ A `CheckBox` that allows the customer to choose if they want to add a mapping to the package from the selected source when they `Install`/`Update`. This `CheckBox` will always be shown even if package source mapping is not enabled. Package source mapping can be enabled by the customer checking the `CheckBox` to create the first mapping for the solution's settings. Once package source mapping is enabled, the customer must either have a mapping previously configured, or consent to a new mapping by checking the `CheckBox` to install. Otherwise, Restore will fail as it does today.
+
+ PSM Enabled|Package Mapped|Current Behavior|Proposed Behavior
```
JonDouglas: We should prompt PSM enablement if it is disabled. A button such as "Enable Source Mapping" next to the first case here would be beneficial and onboard people into the following flow.

proposed/2023/package-source-mapping-status-PMUI.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Functional Explanation
+
+On the details pane there will be new information shown below the Install/Update buttons. 
+
+#### Mapping Status
+
+A label will appear under the `Install/Update` button indicating the Package Source Mapping status.
+
+| Project Onboarded | Selected Package is Mapped | Status Icon | Status Text
```
JonDouglas, line 28: Icons might be:

Disabled - Information or a Hyperlink type icon

Onboarded Not Mapped - X or a disconnected link

Onboarded Mapped - Checkmark or a connected link

proposed/2023/package-source-mapping-actions-PMUI.md
donnie-msft: > A button such as "Enable Source Mapping" 

Would the button essentially be the "Configure" VS Options shortcut? Or would it mean "enable it by mapping the next package I install"?

proposed/2023/package-source-mapping-status-PMUI.md
donnie-msft: For not mapped / mapped, some that stick out...
`ConnectionOffline` is a little subtle. Maybe use the `StatusOffline` and `ConnectionWarning`? I also like `StatusOK` for the good to go scenario.

1. x and check 
  image
  image

1.  connection icons
image

1. Reuse nuget icons?
image

proposed/2023/package-source-mapping-actions-PMUI.md
albarry4: I like that idea, Jon. I would picture an option that would mean "enable it by mapping the next package I install", so that the user can have the option to onboard to PSM and utilize these features for future installs, regardless of their actions at the time of opt-in?

proposed/2023/package-source-mapping-status-PMUI.md
albarry4: I think the icons here will end up being pretty small, so I wonder if icon options 2 and 3 would be too busy for this use case? I'd personally opt towards something more simple, like the icons in option 1

proposed/2023/package-source-mapping-actions-PMUI.md
donnie-msft: Re-reading this, it sounds like we just want the Checkbox that is already described. I think this is better than a button because it allows opting in as a toggle, then the existing action buttons. I worry that a prompt could be an nuisance for a customer who can't/doesn't want to opt-in, so then we'd need to have a way to "not show this dialog" as we do for preview window.

proposed/2023/package-source-mapping-status-PMUI.md
donnie-msft: Added icons to the table!

proposed/2023/package-source-mapping-actions-PMUI.md
JonDouglas: I imagine a hyperlink that when clicking will enable the feature in the nuget.config and then prompt for mappings for any future installs.

Or back to the original idea, a toggle that allows people to map.

donnie-msft: We can continue discussing Action related concepts in the separate spec I'm drafting: https://github.com/NuGet/Home/pull/12616

donnie-msft:
Hi @JonDouglas & @albarry4 - thanks for the comments! Is there anything else preventing your approval of this spec? 

We can continue discussing the PM UI Actions and how they relate to source mapping in the spec I'm drafting for that: https://github.com/NuGet/Home/pull/12616. I'd also like to discuss that one with @nkolev92 as well when he's back.

JonDouglas (approved):
lgtm. please get an additional client/PM review if possible.

albarry4 (approved):

donnie-msft:
@NuGet/nuget-client please take a look. I'd like feedback sooner than later so I can make the changes in this release.

jeffkl (approved):

PR #12611, opened 2023-05-22
Update vulnerabilities-in-restore.md

JonDouglas:
Adds a vulnerability mode selection.

### DIFF
--- a/proposed/2022/vulnerabilities-in-restore.md
+++ b/proposed/2022/vulnerabilities-in-restore.md
@@ -305,6 +305,16 @@ However, it is expected that such projects will have a CI build which will perfo
 - Vulnerability scanning can be extended to SBOMs.
 - Support can be added to automatically fix vulnerable dependencies (i.e. a fix experience in CLI / Tooling)
 - Consideration of SDK/Framework scanning for implicit PackageReference that may be vulnernable.
+- Readiness to enable `<NuGetAuditMode>` to `all` for .NET/VS vNext:
+    - Customer feedback from .NET 8.
+    - Satisfaction of direct dependency scanning.
+    - Noise ratio of transitive dependency scanning (i.e. new warnings)
+    - Performance/scalability impact of transitive dependency scanning.
+    - Version resolution to ensure proper vulnerability reporting.
+    - UI/UX considerations for distinguishing direct/transitive vulnerability warnings.
+    - Incremental scanning/caching to avoid redundant scans.
+    - Documentation and education resources for the functionality.
+    - Prioritization and suppression of severity / advisories.
 
 Additionally, most of the [`Rationale and alternatives`](#rationale-and-alternatives) are really future possibilities on their own as they are not always exclusive to the current approach. Here's some further possibilities:
 
--- a/proposed/2022/vulnerabilities-in-restore.md
+++ b/proposed/2022/vulnerabilities-in-restore.md
@@ -82,11 +82,11 @@ To enable the feature, a developer can add `<NuGetAudit>enable</NuGetAudit>` to 
 
 #### Setting Vulnerability Auditing Modes
 
-There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, `transitive`, and `all`.
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, and `all`.
 
-These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, `transitive` will scan for any transitive-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
 
-When a known vulnerability is found that is of the `transitive` level, it will include the path to the project containing the top-level package and including the name and version of the package the vulnerable transitive dependency is coming from. Transitive level known vulnerabilities should not be a warning, but rather a message/informational MSBuild severity as they should not break builds but still be brought up in the Error List as informational.
+When a known vulnerability is found that is of the transitive level, it will include the path to the project containing the top-level package and including the name and version of the package the vulnerable transitive dependency is coming from. Transitive level known vulnerabilities should not be a warning, but rather a message/informational MSBuild severity as they should not break builds but still be brought up in the Error List as informational.
 
 #### Setting an Audit Level
 
@@ -304,6 +304,7 @@ However, it is expected that such projects will have a CI build which will perfo
 
 - Vulnerability scanning can be extended to SBOMs.
 - Support can be added to automatically fix vulnerable dependencies (i.e. a fix experience in CLI / Tooling)
+- Consideration of SDK/Framework scanning for implicit PackageReference that may be vulnernable.
 
 Additionally, most of the [`Rationale and alternatives`](#rationale-and-alternatives) are really future possibilities on their own as they are not always exclusive to the current approach. Here's some further possibilities:
 
--- a/proposed/2022/vulnerabilities-in-restore.md
+++ b/proposed/2022/vulnerabilities-in-restore.md
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
 
 To enable the feature, a developer can add `<NuGetAudit>enable</NuGetAudit>` to their project file as a MSBuild property. To disable the feature, a developer can add `<NuGetAudit>disable</NuGetAudit>` or remove the property from the project file.
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, `transitive`, and `all`.
+
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, `transitive` will scan for any transitive-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
+
+When a known vulnerability is found that is of the `transitive` level, it will include the path to the project containing the top-level package and including the name and version of the package the vulnerable transitive dependency is coming from. Transitive level known vulnerabilities should not be a warning, but rather a message/informational MSBuild severity as they should not break builds but still be brought up in the Error List as informational.
+
 #### Setting an Audit Level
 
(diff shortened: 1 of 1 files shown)


proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
 
 To enable the feature, a developer can add `<NuGetAudit>enable</NuGetAudit>` to their project file as a MSBuild property. To disable the feature, a developer can add `<NuGetAudit>disable</NuGetAudit>` or remove the property from the project file.
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, `transitive`, and `all`.
```
donnie-msft: When do we start discussing how these modes might be brought into the PM UI ? My first impressions are

- Actions (install/update) would audit automatically based on a user setting for AuditMode.
- Allow filtering package lists on AuditMode so you can easily see out of hundreds of packages: are my top-level packages vulnerable? Then, which transitives are vulnerable?

Happy to brainstorm more details on each bullet & other ideas.

donnie-msft (approved):

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
 
 To enable the feature, a developer can add `<NuGetAudit>enable</NuGetAudit>` to their project file as a MSBuild property. To disable the feature, a developer can add `<NuGetAudit>disable</NuGetAudit>` or remove the property from the project file.
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, `transitive`, and `all`.
+
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, `transitive` will scan for any transitive-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
```
zivkan: I don't think we should ever allow `NuGetAudit` to be used without direct package references being checked, so we shouldn't have a `transitive` mode. (I can't imagine a scenario where it would ever be desirable)

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
...
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, `transitive`, and `all`.
+
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, `transitive` will scan for any transitive-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
+
+When a known vulnerability is found that is of the `transitive` level, it will include the path to the project containing the top-level package and including the name and version of the package the vulnerable transitive dependency is coming from. Transitive level known vulnerabilities should not be a warning, but rather a message/informational MSBuild severity as they should not break builds but still be brought up in the Error List as informational.
```
zivkan: What do you think about moving this to the "future possibilities" section? While I definitely want to have package graph path displayed to make it easier, to me it raises questions about if adding NuGetAuditMode without showing the package graph path is acceptable (feature not implemented to spec), and whether showing the path in the audit warning messages is higher priority than other similar features, like `dotnet nuget why`.

anangaur: I tend to agree. However, there should be an easy way for developers to know the full paths (could be multiple?) so that its clear what brings the transitive in. IMO, whether disjoint feature or not, should be part of the vulnerability epic for .NET 8.

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: Yes let's modify this. Do we need a mode for `framework` or `sdk` dependencies? i.e. packages that are apart of a framework/sdk?

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: I have been having various conversations and I think we should try to provide as much provenance in the warning messages as we can so it helps empower the developer to remediation until we can move towards our fixing experiences. `dotnet nuget why` would be a helpful command for further provenance, but we should provide enough in these warnings to give people somewhere to look initially. We have until August for much of this work.

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: I'm going to schedule a larger team meeting to discuss this. It has come to my attention that we could make many experiences audit aware and show appropriately.

proposed/2022/vulnerabilities-in-restore.md
zivkan: Honestly, I don't know.

On one hand, implicit packages (those that MSBuild SDKs & Visual Studio project systems bring in) are intended not to be upgradable by customers. In VS, we prevent those packages from appearing in the "updates" tab, and on the browse & installed tab, up disable the update button. It's the SDK owner's responsibility to upgrade packages, which means customers will need to update to a newer version of the MSBuild SDK.

On the other hand, from a security perspective, the customer's app might use a package with a known vulerability. It might be difficult for them to figure out how to upgrade, and NuGet can't provide generic instructions on how to upgrade. Some SDKs provide their own extensibility (MSBuild property). Other times a `<PackageReference Update` (rather than `Include`) might work, but it depends on MSBuild import order. Other times using a different SDK version is needed, but that assumes that the SDK owner published a new version of their SDK that updated their implicit package versions. But at least by alerting the customer that their project uses a package with a known vulerability they have more knowledge and can investigate.

Honestly, my expectation is that implicit packages should mostly be "build related" packages, not packages with runtime assets, but of course there's nothing technically preventing SDK owners to add packages with runtime assets. So, I think implicit packages should mostly be a non-issue, but technically it could be.

This is potentially something else we can discuss in the meeting you're organizing, if we don't already have more topics than time available.

zivkan (approved):

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
...
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, and `all`.
+
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
+
+When a known vulnerability is found that is of the transitive level, it will include the path to the project containing the top-level package and including the name and version of the package the vulnerable transitive dependency is coming from. Transitive level known vulnerabilities should not be a warning, but rather a message/informational MSBuild severity as they should not break builds but still be brought up in the Error List as informational.
```
skofman1, line 89: The text here can be understood in different ways, so consider clarifying,
Are transitive dependencies shown as message / informational even when NuGetAuditMode is set to "all"?
If NuGetAuditMode is set to "direct", will we still show transitive dependency vulnerabilities as message/informational, or we won't show them at all?

proposed/2022/vulnerabilities-in-restore.md
```diff
@@ -80,6 +80,14 @@ This feature will be opt-in to start and gather feedback from developers.
 
 To enable the feature, a developer can add `<NuGetAudit>enable</NuGetAudit>` to their project file as a MSBuild property. To disable the feature, a developer can add `<NuGetAudit>disable</NuGetAudit>` or remove the property from the project file.
 
+#### Setting Vulnerability Auditing Modes
+
+There will be different modes to audit vulnerabilities based on the developer's or developer's team preference. To do this, a developer will opt-in to a feature called `<NuGetAuditMode>` which will have different modes such as `direct`, and `all`.
+
+These modes should be pretty straight-forward. `direct` will scan for any top-level vulnerabilities, and `all` will scan for both top-level and transitive-level vulnerabilities. The default will be `direct` until the experience is ready to be `all` given that transitive vulnerabilities are the majority of vulnerability notices (90%+).
```
skofman1, line 87: > The default will be `direct` until the experience is **ready** to be `all`

While I agree with this direction, it would be good to specify here what this "readiness' means. What is the decision criteria we would follow to switch the experience to "all" be default?

skofman1, line 87: What are your thoughts on adding a message to the output along the lines of: "Only direct dependencies were scanned for vulnerabilities. To enable scanning transitive dependencies as well, turn on <NuGetAudit>all</NuGetAudit>."

proposed/2022/vulnerabilities-in-restore.md
nkolev92, line 89: If there are more paths, are we gonna include them all? I think we should, as there's many interesting resolution things that can happen.

proposed/2022/vulnerabilities-in-restore.md
nkolev92, line 87: Not sure if it needs to be called out in this spec, but we should make sure we can capture a few metrics regardless of the audit  level.

- Capture number of direct vuln + potentially hashes.
- Capture number of transitive + potentially hashes. This should happen even if direct is the audit level, at least in the initial version. The perf hit is going to be negligible, but the information would be invaluable.

nkolev92 (approved):
LGTM, a couple of suggestions.

proposed/2022/vulnerabilities-in-restore.md
zivkan: Some customers are actively hostile against any non-actionable message that can't also be suprressed. If we has Roslyn analyzer style `.editorconfig` seveity levels on codes, then we could allow customers to customize their builds as they like. But currently we only have NU codes for warnings and errors, informational messages don't have codes and therefore can't be suppressed.

So, it's a tradeoff, especially in large solutions where the message could be shown hundreds of times, once for each solution.

One problem with moving fast, is that we don't have time to carefully plan, or test, or validate, before we have to ship. So, my recommendation for now is no, don't have a message, and we can wait for customer feedback while working on other improvements, and if we don't get feedback, we can do a bit of our own testing to estimate how important it is.

zivkan: I see that a too much information about implementation details for this more customer focused feature spec.

nkolev92: Fair, resolving.

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: Probably best to just skip this for now then until more feedback.

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: direct -> direct dependencies

all -> direct & transitive

We will not show output for transitive dependencies unless the user sets their mode appropriately.

proposed/2022/vulnerabilities-in-restore.md
JonDouglas: added some bullets in future possibilities to help answer

nkolev92:
@JonDouglas can this be merged now?

Issue #12606, opened 2023-05-19
NU1100 is used for both TFM incompatibility and for an unresolved PackageSourceMapping


donnie-msft:
### NuGet Product(s) Affected

NuGet.exe, Visual Studio Package Management UI, Visual Studio Package Manager Console, MSBuild.exe, dotnet.exe, NuGet SDK

### Current Behavior

When an install fails, the message I get is this: "Error    NU1100    Unable to resolve 'Swashbuckle.Core (>= 5.6.0)' for '.NETFramework,Version=v4.7.2/win-x86'. PackageSourceMapping is enabled, the following source(s) were not considered: nuget, nuget-build. "

I don't know if the failure is because of TFM incompatibility, or because I forgot to add source mapping

![image](https://github.com/NuGet/Home/assets/49205731/ddfe8a46-cf89-4f60-aeca-ba499e9981bd)

### Desired Behavior

Specific messages for either scenario. 
Specify whether the TFM not compatible, or is the PackageSourceMapping not configured?

nkolev92:
NU11**  errors are always package unavailable. 

NU12** are compat errors.

donnie-msft:
@nkolev92 - Considering that I'm already creating an NU code in https://github.com/NuGet/Home/issues/12953 
... What if I create another NU111X and move the Package Source Mapping scenario to it?
https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1100#scenario-2

This would sort of make NU111X prefix space a place for 10 source mapping specific error codes.
More importantly, this issue could be closed since it would help the concern of NU1100 representing multiple scenarios.

nkolev92:
This is not really a bug. 
Let's have an offline discussion, the error here says that the package can't be found, nothing to do with compat.

nkolev92:
I removed this from the NuGet.Client project. By adding it to the project, triage never saw this issue.

jeffkl:
Team Review: Closing as not a bug.  The docs for NU1100 state that if you're using package source mapping and the package isn't found but exists on your feeds you need to add a mapping.

https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1100

PR #12616, opened 2023-05-26
[Draft] Spec for Package Source Mapping Actions in PM UI

donnie-msft:
_This Draft PR has been replaced by PR # 12810_

[Draft] Package Source Mapping Package Management Actions in the PM UI ([shortcut to rendered spec](https://github.com/NuGet/Home/blob/dev-donnie-msft-specSourceMappingActionSpec/proposed/2023/package-source-mapping-actions-PMUI.md))

Spec for: Issue # 11366
  - Marked Draft as there are questions about how we should proceed 
  - Various edge cases are unclear


proposed/2023/package-source-mapping-actions-PMUI.md
```diff
@@ -0,0 +1,102 @@
...
+ A proposal for the changes to the preview window:
+![PM UI - a proposal for the changes to the preview window](../../meta/resources/PackageSourceMapping/PMUI_preview.png)
+
+#### Restore Errors
+
+There are a few errors the customer could make while trying to install or update a package with a mapping that would make `Restore` fail. First, there could be issues with the transitive packages. Maybe there are transitive packages that are already mapped to a different source or the source that the customer is trying to map a package to does not support some of the transitive packages. 
+
+The customer could also try to update a package that is already mapped to a different source, but the source it is mapped to does not support the version they are trying to update to. In these scenarios, `Restore` will fail in the same way it does today. The customer can see why `Restore` failed in the output window. 
+
+Currently, `NU1100` is used for both an incompatible TFM, and for a missing Package Source Mapping.
```
nkolev92: I think I mentioned in the issue, but NU1100 is always a missing package. 

Missing package for source mapping vs not shouldn't matter. This was a deliberate decision at that point. 

Ofc it can be a subject to change, but I think this particular statement is not correct.

proposed/2023/package-source-mapping-actions-PMUI.md
```diff
@@ -0,0 +1,102 @@
...
+A proposal to make more specific errors for restore errors related to package source mapping has been made:
+- NU1100 is used for both TFM incompatibility and for an unresolved PackageSourceMapping [#12606](https://github.com/NuGet/Home/issues/12606)
+
+### Technical Explanation
+
+Package Management action logic will pass down a new mapping package ID and source name to Restore. Preview Restore Utilities will look for this new mapping and use it in the preview restore: By appending 2 package patterns, a "*" and the new package ID, along with the new mapping source name to Settings in memory.
+
+Once preview restore determines the newly installed package IDs (top-level and transitive), those IDs will be checked against existing mappings. If an existing mapping is found, no new mapping is written for that package.
+
+If a package is found in the Global Packages Folder (GPF), that package will be mapped to the selected source.
```
nkolev92: Can you more specific, what is the selected source in this? The previously downloaded one?

proposed/2023/package-source-mapping-actions-PMUI.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Unresolved Questions
+
+None
+
+## Future Possibilities 
+
+1. Currently, if an `Install`/`Update` cannot find transitive packages on the mapped source being mapped, the restore will fail. Reaching out to other configured sources and allowing the customer to select a different source for these packages may resolve this issue. However, care must be taken to avoid unintentionally leaking requested package IDs to these secondary sources when the customer hasn't agreed to such a query.
+
+1. If a package is found in the Global Packages Folder (GPF), the PM UI could look at the source in the `nupkg.metadata` file. A new mapping could either be created for this source auotmatically, or it could be presented to the customer in the Preview Window or another affordance.
```
nkolev92, line 172: An idea would be:

* Use the mapping from the GPF if the source exists.
* Warning if the source doesn't. 

I think https://github.com/NuGet/Client.Engineering/issues/1166 getting implemented will complement this nicely.

proposed/2023/package-source-mapping-actions-PMUI.md
```diff
@@ -0,0 +1,102 @@
...
+## Explanation
+
+### Functional Explanation
+
+When a new package source mapping is being created, the Preview Window will appear showing which packages (the top-level and any transitive dependencies) will be mapped. 
+
+(TBD) It is not yet understood whether there will be a new afforance on the details pane in a new row below the Install/Update buttons, or if source mappings will be handled in the background.
+
+#### A Proposal for Consent to a new Source Mapping
+ A `CheckBox` that allows the customer to choose if they want to add a mapping to the package from the selected source when they `Install`/`Update`. This `CheckBox` will always be shown even if package source mapping is not enabled. Package source mapping can be enabled by the customer checking the `CheckBox` to create the first mapping for the solution's settings. Once package source mapping is enabled, the customer must either have a mapping previously configured, or consent to a new mapping by checking the `CheckBox` to install. Otherwise, Restore will fail as it does today.
```
nkolev92: Is there a screenshot that we can see?

TBH, I feel like *if* you have source mapping enabled, and a package is not mapped, the mapping additon should be the default by whatever means you opt into it.

nkolev92 (commented):
Thanks for this write up. I'm not sure I follow how https://github.com/NuGet/NuGet.Client/pull/5238 is a consequence of this proposal.

proposed/2023/package-source-mapping-actions-PMUI.md
donnie-msft: Selected source always refer to the upper right drop-down selection in the PM UI. I can add the terms used at the beginning of the spec for clarity.

proposed/2023/package-source-mapping-actions-PMUI.md
donnie-msft: Yes, it's missing, but the suggestion was to be more specific that it's missing *because* package source mapping didn't allow restore to retrieve it.

Here, the NU1100 is shown only because I deleted the mapping for the package I tried to install:
image

nkolev92: So NU1100 says, `PackageSourceMapping is enabled, the following source(s) were not considered: X, Y`. 

How would you make that clearer?

proposed/2023/package-source-mapping-actions-PMUI.md
nkolev92: I think terms is a great idea. 

That distinction is important in that scenario since there's a *chance* that the package version on the mapped sources and global packages folder aren't the same.

proposed/2023/package-source-mapping-actions-PMUI.md
donnie-msft: The proposal was for me to create something distinct. Copying from the Desired behavior in https://github.com/NuGet/Home/issues/12606:

> Specific messages for either scenario.
> Specify whether the TFM not compatible, or is the PackageSourceMapping not configured?

nkolev92: My point is that we already have different text for source mapping vs no source mapping scenarios. Are you proposing a new error code? This was something we considered and *specifically opted against* since when a package is not resolved, the issue could be, package doesn't exist, package exists but version doesn't, no suitable version can be selected etc. All of those get the same "due to source mapping, these sources were not considered". You can look at that in https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Commands/RestoreCommand/Diagnostics/UnresolvedMessages.cs by searching for `isPackageSourceMappingEnabled `.
PR regarding this https://github.com/NuGet/NuGet.Client/pull/4374.

For the 2nd line, compatibility has different warning/error codes, NU1100s vs NU1200s. In fact they are generated at different points, since in the first case there is no package, but in the 2nd case, there's a package that *was* analyzed.

donnie-msft: Hm, when Svetlana and I were walking through the changes to PM UI, I'm sure we both saw a case where the package *was* available on a source, but we got a NU1100.  I'll have to see if I can recreate that. 

According to the docs, TargetFramework is called out as well:

> Unable to resolve 'Dependency (>= 1.0.0)' for 'TargetFramework'

https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1100

nkolev92: > According to the docs, TargetFramework is called out as well:

It's calling that out not because it is a compatibility problem, but because projects can multi target.

PR #12705, opened 2023-06-26
NuGet Proposal Process simplification

donnie-msft:
Fixes: https://github.com/NuGet/Client.Engineering/issues/2357

Change the README to just have 2 folders: `Proposed` and `Accepted`.

I've also moved/edited all the old specs to meet this simplified process.


JonDouglas:
I'd recommend we don't include status or feature labels in PRs nor similar sections in the proposals. I would recommend we just simplify the existing process to remove `implemented` and `withdrawn` while pointing people to just include proposals to `accepted` by default.

We can just simplify to be like https://github.com/dotnet/designs if people agree that's easiest.

donnie-msft:
> I'd recommend we don't include status or feature labels in PRs nor similar sections in the proposals. I would recommend we just simplify the existing process to remove `implemented` and `withdrawn` while pointing people to just include proposals to `accepted` by default.
> 
> We can just simplify to be like https://github.com/dotnet/designs if people agree that's easiest.

Ok, I have removed the verbiage about "status:" in the document's header. I probably won't delete all the existing doc headers to comply, since at the time those were written, status was applicable. Going forward, we can just rely on the state of the linked GitHub Issue to indicate whether the feature was merged or closed.

meta/README.md
```diff
@@ -16,22 +16,20 @@ All discussions surrounding a proposal are covered by the [.NET Foundation code
...
+Once accepted, it will be scheduled by the NuGet contributors to be implemented or put up for grabs for anyone to implement by making a PR in the appropriate repository.
 
-## What happens when a proposal is withdrawn?
+## What happens when a proposal is implemented?
 
-When a proposal is reviewed for its accuracy and relevance and no longer fits the goals of the project or is no longer a candidate for implementation, a proposal will be moved from `accepted` to `withdrawn` with a clear reason as to why it was moved. 
+When the changes described in the accepted proposal have been implemented and merged into the relevant repository and due for release the corresponding proposal will remain in the `accepted` folder. The linked GitHub issue will be closed upon merging to the relevant branch. 
 
-## What happens when a proposal is implemented?
+If you'd like to implement an `accepted` proposal, please make a PR in the appropriate repository and mention the proposal in the PR.
```
nkolev92: might be a good idea to check that no one else is working on it by commenting on the issue related to the proposal.

donnie-msft: Ok, added "and Issue".

jeffkl (approved):

PR #12709, opened 2023-06-27
Rendering Markdown readmes in the PMUI package details page

albarry4:
[Rendered Spec](https://github.com/NuGet/Home/blob/albarry4-patch-1-1/proposed/2023/PMUI-Readme-rendering.md)

### DIFF
--- a/proposed/2023/PMUI-Readme-rendering.md
+++ b/proposed/2023/PMUI-Readme-rendering.md
@@ -1,0 +1,45 @@
+# Markdown Readme Rendering
+* Start Date: 6/27/23
+* Authors: Allie Barry ([albarry4](https://github.com/albarry4))
+* Issue: https://github.com/NuGet/Home/issues/12583
+
+## Summary
+
+Support rendering markdown readmes in the package details page for NuGet in Visual Studio to make package affordances more accesible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem. 
+
+## Motivation
+
+In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as READMEs. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed README file and implementing features to facilitate README viewing and inclusion would positively impact this number.   
+
+## Explanation
+
+### Functional Explanation
+
+The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the README in general, and make it feel like an important and crucial element of a package.  
+
+The desired experience would somewhat mirror the user interface on NuGet.org, which already supports rendered Markdown viewing. When a customer views the package detail page of a package with an included markdown readme, they will see this proposed UI mock ![Mock](https://github.com/NuGet/Home/assets/89422562/0017703e-2572-4c3c-8f1c-7cb72a01f25f)
+
+
+### Technical Explanation
+
+To render markdown in Visual Studio, we can use Markdig to convert the MD to HTML, and then use either Webview2 or Daytona to render the HTML and stylize. 
+
+Currently, the Markdown editor team within Visual studio uses webview 2 to render tehir HTML for general MD rendering within Visual Studio. Using webview2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently recieving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over webview2 would avoid accesibility bugs, and cut out the effort required to write out own CSS styling. The markdown editor team is currently in the process of transitioning from using webview2 to Daytona, and reccomended that, if possible, our team should use Daytona off the bat.  
+
+## Drawbacks
+
+None
+
+## Rationale and Alternatives
+
+An alternative here would be to include a link within the package detail page which would open up the rendered readme externally. Should we find complications with directly embedding the rendered readme into the packag detail page, this would be a proposed alternative.
+
+
+## Unresolved Questions
+
+None
+
+
+## Future Possibilities 
+
+TBD
--- a/proposed/2023/PMUI-Readme-rendering.md
+++ b/proposed/2023/PMUI-Readme-rendering.md
@@ -9,13 +9,13 @@ Support rendering markdown Readmes in the package details page for NuGet in Visu
 
 ## Motivation
 
-In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as READMEs. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed README file and implementing features to facilitate README viewing and inclusion would positively impact this number.   
+In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as Readmes. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed Readme file and implementing features to facilitate Readme viewing and inclusion would positively impact this number.   
 
 ## Explanation
 
 ### Functional Explanation
 
-The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the README in general, and make it feel like an important and crucial element of a package.  
+The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the Readme in general, and make it feel like an important and crucial element of a package.  
 
 The desired experience would somewhat mirror the user interface on NuGet.org, which already supports rendered Markdown viewing. When a customer views the package detail page of a package with an included markdown readme, they will see this proposed UI mock ![Mock](https://github.com/NuGet/Home/assets/89422562/89b0295c-64d5-42a4-a52c-83dea2807edc)
 
--- a/proposed/2023/PMUI-Readme-rendering.md
+++ b/proposed/2023/PMUI-Readme-rendering.md
@@ -25,7 +25,7 @@ The desired experience would somewhat mirror the user interface on NuGet.org, wh
 To render markdown in Visual Studio, we can use Markdig to convert the MD to HTML, and then use either Webview2 or Daytona to render the HTML and stylize. 
 
 
-Currently, the Markdown editor team within Visual studio uses webview 2 to render tehir HTML for general MD rendering within Visual Studio. Using webview2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently recieving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over webview2 would avoid accesibility bugs, and cut out the effort required to write out own CSS styling. The markdown editor team is currently in the process of transitioning from using webview2 to Daytona, and reccomended that, if possible, our team should use Daytona off the bat.  
+Currently, the Markdown editor team within Visual studio uses webview 2 to render their HTML for general MD rendering within Visual Studio. Using WebView2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently receiving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over WebView2 would avoid accessibility bugs, and cut out the effort required to write our own CSS styling. The markdown editor team is currently in the process of transitioning from using WebView2 to Daytona, and recommended that, if possible, our team should use Daytona from the beginning.
 
 ## Drawbacks
 
--- a/proposed/2023/PMUI-Readme-rendering.md
+++ b/proposed/2023/PMUI-Readme-rendering.md
@@ -5,7 +5,7 @@
 
 ## Summary
 
-Support rendering markdown readmes in the package details page for NuGet in Visual Studio to make package affordances more accesible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem. 
+Support rendering markdown Readmes in the package details page for NuGet in Visual Studio to make package affordances more accessible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem.
 
 ## Motivation
 
--- a/proposed/2023/PMUI-Readme-rendering.md
+++ b/proposed/2023/PMUI-Readme-rendering.md
@@ -17,7 +17,11 @@ In the latest NuGet in Visual Studio 17.5 customer survey, our team found that o
 
 The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the Readme in general, and make it feel like an important and crucial element of a package.  
 
-The desired experience would somewhat mirror the user interface on NuGet.org, which already supports rendered Markdown viewing. When a customer views the package detail page of a package with an included markdown readme, they will see this proposed UI mock ![Mock](https://github.com/NuGet/Home/assets/89422562/89b0295c-64d5-42a4-a52c-83dea2807edc)
+The desired experience would be similar to the current design of the package details page in NuGet in Visual Studio, but we would add a tab at the top of the view where the user could toggle back and forth to view the regular package details that exist currently, and a rendered Readme file (if a Readme exists in the package contents). The only change in the current package details page would be the removal of the "Readme: View Readme" list item. Below shows a mock of the desired view of both the Readme and the Package Details tabs: 
+
+![package details](https://github.com/NuGet/Home/assets/89422562/c106ca79-e26e-4b1d-b5ea-273651f79857)
+
+![Readme](https://github.com/NuGet/Home/assets/89422562/81b24877-f12f-4783-905c-4a155d3c7693)
 
 
 ### Technical Explanation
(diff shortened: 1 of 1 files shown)


mckennabarlow (approved):
This is great, supporting rendering markdown in the PMUI will help in increasing awareness and motivation for high quality READMEs.

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Explanation
+
+### Functional Explanation
+
+The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the README in general, and make it feel like an important and crucial element of a package.  
+
+The desired experience would somewhat mirror the user interface on NuGet.org, which already supports rendered Markdown viewing. When a customer views the package detail page of a package with an included markdown readme, they will see this proposed UI mock ![Mock](https://github.com/NuGet/Home/assets/89422562/0017703e-2572-4c3c-8f1c-7cb72a01f25f)
```
nkolev92: I'm guessing the proposed changes are just the details tab correct?

donnie-msft: That mock is dense with features that don't exist yet. How much of it is included in this readme spec?

albarry4: Hi! The proposed changes here are just the details tab and the ability to render markdown READMEs directly in the package details pane. Any features in the mock that dont exist yet can be ignored for the time being -- but the general layout in the mock showing the rendered README in the details tab is what is being proposed here!

nkolev92: If that's the case, it wouldn't be worst idea to crop the image to include only that part. 
The overall screenshot might send some mixed signals, and suggest something we're not committing to/planning right now.

donnie-msft: I'd ideally want more than just cropping the image, because that leaves the question of how will we render *just* the Readme in the UI we have today? 

1. Will we split the details pane into 2 columns (left is Readme, right is Package metadata/Details) 
2. Will it be the 1st new tab in the details pane? This implies that existing details are in a new tab (would produce 2 total tabs: the existing content in tab1, and the readme in tab2)
3. Will we just render Readme content below the package description?
4. other?

albarry4: @nkolev92 @donnie-msft, I just created and updated the spec with a full new mock that doesn't have any features that don't exist. This new mock should address Donnie's comment above!

donnie-msft: There's a new screenshot with tabs, so that aligns with my option # 2, but it shows a total of 5 tabs. Just wanted confirmation that this spec should result in all 5 tabs? 
image

I think I'm clear on how these tabs would work: 
✅ README
✅ Frameworks
✅ Dependencies

But I think these require work outside the scope of the spec:
⚠ Used By
⚠ Versions

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Motivation
+
+In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as READMEs. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed README file and implementing features to facilitate README viewing and inclusion would positively impact this number.   
+
+## Explanation
+
+### Functional Explanation
+
+The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the README in general, and make it feel like an important and crucial element of a package.
```
donnie-msft: Should the README be shown by default when viewing a package? In other words, the README tab will be shown first, pushing the package details to a separate tab (and requiring a separate click). 

Seems to work for nuget.org, so I'd guess we're happy with that. It's a performance consideration for the VS side, so that's why I want to be clear on our goal.

albarry4: Part of my goal with this proposal was to contribute to the general push for README inclusion and README awareness within NuGet. I feel that making the README shown by default would both 1) motivate more package authors to include a high-quality readme that demonstrates package affordances and usage information, and in turn 2) make this information more accessible and prevalent for package consumers (I included some findings from the recent survey in the "Motivation" section above that explains why I think this is important). I'm open to hearing any other ideas for a potential layout here as well!

proposed/2023/PMUI-Readme-rendering.md
albarry4: I've updated the mock again to just show those three tabs! Apologies, I missed editing out that piece as I was creating the new mock

anangaur:
> Regarding the Options row in the PMUI per package: image

I see that it currently lists an option to preview window  during an operation:

image

 But it seems to take a lot of real estate moving important info like README further down. Is this option important to show always? Or can it move to some NuGet settings under "Options"? Or can we find a better place for it?

proposed/2023/PMUI-Readme-rendering.md
zivkan: With the updated screenshot, it's not clear where the existing package details will be. For example, license, version, authors, data published, project URL, report abuse link.

proposed/2023/PMUI-Readme-rendering.md
zivkan: Have we asked customers to see what they want to see most prominently? If there are more customers that want to see dependencies & project URL & license, then always showing the readme will provide those customers a worse experience, even if improving readme awareness is a good goal.

JonDouglas:
Here's a couple of considerations:

1. While a long term solution would be multiple tabs, there could be a short term solution of two tabs:
    - README
    - Details / Information (What exists today)
2. If tab parity is part of the scope of this proposal, we should create mockups for each of the unique tabs and show the existing UI elements that would be placed inside. From the current mockup that might be:
    - README
    - Details / Information (Most package metadata)
    - Frameworks (May not apply here yet)
    - Dependencies (package dependencies metadata)

In my opinion, this proposal can be scoped for the minimal amount of work to add the value of the embedded README. The current mockup is pretty close to what I imagine this would be, but i'd consider 1) above as VS PM UI doesn't have the NuGet.org tab parity quite yet and that would warrant another proposal in my opinion. It is fully possible to extract the "Framework" / "Dependencies" / "Details" from the current PM UI to be in line with the current mockups, but we would need to consider 2).

JonDouglas:
> > Regarding the Options row in the PMUI per package: image
> 
> I see that it currently lists an option to preview window during an operation:
> 
> image
> 
> But it seems to take a lot of real estate moving important info like README further down. Is this option important to show always? Or can it move to some NuGet settings under "Options"? Or can we find a better place for it?

I think it could be fine where it is or moved up near the install operation rather than its own section(i.e. remove the horizontal rule). The main concern would be to keep it nearby for certain legacy project types. Or even detect new project types and even hide it or move these options to VS options to be more global. My vote personally would be to make NuGet install options global and not even in the PM UI details panel.

https://learn.microsoft.com/en-us/nuget/consume-packages/install-use-packages-visual-studio#nuget-package-manager-options-control

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+### Technical Explanation
+
+To render markdown in Visual Studio, we can use Markdig to convert the MD to HTML, and then use either Webview2 or Daytona to render the HTML and stylize. 
+
+
+Currently, the Markdown editor team within Visual studio uses webview 2 to render tehir HTML for general MD rendering within Visual Studio. Using webview2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently recieving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over webview2 would avoid accesibility bugs, and cut out the effort required to write out own CSS styling. The markdown editor team is currently in the process of transitioning from using webview2 to Daytona, and reccomended that, if possible, our team should use Daytona off the bat.  
+
+## Drawbacks
```
JonDouglas, line 34: Some potential drawbacks:

- We would be changing a core experience in VS and that can move people's cheese.
- We do not know the current preference of people wanting to see a README or package information.
- There is limited space to render a README in VS today compared to say NuGet.org or other package registry websites.
- There might? be a small performance hit to render the README or fetch it.
- etc

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Unresolved Questions
+
+None
+
+
+## Future Possibilities 
+
+TBD
```
JonDouglas: Maybe consider the future possibilities of NuGet.org parity, the frameworks tab, etc and how those would build on-top of this current proposal. For example, the VS UI doesn't have a concept of `Frameworks` yet, but could leverage NuGet.org to provide that data and experience to VS.

JonDouglas (commented):
Looking good! Consider providing as much visualizations or further explanation of the functionality so it is easier to get a "big picture" idea.

For example consider writing about what happens when a user clicks a tab and what they might see. Accompany that with some visualizations or descriptions of the relevant information.

anangaur:
> Or even detect new project types and even hide it or move these options to VS options to be more global. My vote personally would be to make NuGet install options global and not even in the PM UI details panel.

+1 to remove it from this view altogether and move it elsewhere globally.

anangaur:
@albarry4 , Can you also put the proposed screenshot for the Solution view since that has additional complexity with Project selection UI?

proposed/2023/PMUI-Readme-rendering.md
nkolev92: In Visual Studio, I also we should be mindful adding a lot of new tabs. 
Today most of that information is within 1 pane (license, author, url, dependencies etc), we should keep that.

IMO it should be `README` and `Package Details` unless we have a good reason to believe we need the extra tabs.

nkolev92: Also, the `frameworks` tab is a whole major feature itself, so unless we want that now (we don't afaik), we should not include it.

proposed/2023/PMUI-Readme-rendering.md
donnie-msft: > Have we asked customers to see what they want to see most prominently? If there are more customers that want to see dependencies & project URL & license, then always showing the readme will provide those customers a worse experience, even if improving readme awareness is a good goal.

Maybe the default tab should be configurable by customers? By default: Readme tab. Could be in VS Options, or a pin on the tabs, or whatever UX likes for this concept.

donnie-msft (commented):
Thanks for writing this up, and addressing feedback so quickly! I have some questions that may need customer development regarding the README tabs and surrounding UI:

1. "Can we move the tab(s) up and get rid of the "Options" section?"
    1.  We don't have telemetry on the usage. Do you want that to be able to analyze usage first, or do you have enough customer development to proceed with changes?
    1. `packages.config` is where most options come into play. Perhaps we keep showing the options pane only for `packages.config` going forward?
    1. In Figure1, I show a new Options "gear icon" button. We *could* move the entire details pane options panel functionality into the VS Options dialog exclusively, and use the Gear icon as a shortcut. 

1. What's the default tab? Some customers may not want README to be the default. Making the first tab customizable could be done in a few ways:
    1. Setting in VS Options (use gear icon to shortcut, see Figure1).
    1. Pin icons on the tabs (or similar, need to chat with UX board about it first. See Figure 1).
    1. Remember the last tab. 
      1. When switching packages, the details could remember the tab previously selected
      1. When closing/reopening PM UI, should details pane remember the tab that was open last time? (Similar to how PM UI reopens to Browse/Updates/Installed tabs)

_Figure 1_
image

proposed/2023/PMUI-Readme-rendering.md
albarry4: @zivkan not explicitly, but during our interviews on package quality, package documentation (including READMEs and other sources of getting started/bugs/use case information) was mentioned by all interviewees, and was most commonly indicated as the single most important factor (outside of package affordances and legality/security concerns) in determining package quality, and in turn determining which packages to install. This, coupled with the results of the VS 17.5 survey, seemed to me to be sound evidence that customers would like to see README information prominently displayed when viewing package details.

aortiz-msft:
@albarry4: We removed [draft] from the title based on active discussion and PR status.

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
+# Markdown Readme Rendering
+* Start Date: 6/27/23
+* Authors: Allie Barry ([albarry4](https://github.com/albarry4))
+* Issue: https://github.com/NuGet/Home/issues/12583
+
+## Summary
+
+Support rendering markdown readmes in the package details page for NuGet in Visual Studio to make package affordances more accesible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem.
```
jebriede: Capitalizing 'Readme' in all cases for consistency and fixed a typo.


```suggestion
Support rendering markdown Readmes in the package details page for NuGet in Visual Studio to make package affordances more accessible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem.
```

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+* Issue: https://github.com/NuGet/Home/issues/12583
+
+## Summary
+
+Support rendering markdown readmes in the package details page for NuGet in Visual Studio to make package affordances more accesible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem. 
+
+## Motivation
+
+In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as READMEs. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed README file and implementing features to facilitate README viewing and inclusion would positively impact this number.
```
jebriede: 'READMEs' is written in all caps in this paragraph. Perhaps we should standardize on this way of writing it throughout the document.

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+### Technical Explanation
+
+To render markdown in Visual Studio, we can use Markdig to convert the MD to HTML, and then use either Webview2 or Daytona to render the HTML and stylize. 
+
+
+Currently, the Markdown editor team within Visual studio uses webview 2 to render tehir HTML for general MD rendering within Visual Studio. Using webview2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently recieving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over webview2 would avoid accesibility bugs, and cut out the effort required to write out own CSS styling. The markdown editor team is currently in the process of transitioning from using webview2 to Daytona, and reccomended that, if possible, our team should use Daytona off the bat.
```
jebriede: Fixing typos and replacing idiom at the end for clarity.

```suggestion
Currently, the Markdown editor team within Visual studio uses webview 2 to render their HTML for general MD rendering within Visual Studio. Using WebView2 would require us to use CSS to create our own styling for the rendered HTML. A drawback to their solution is that the markdown editor team mentioned that they are currently receiving a lot of accessibility bugs caused by their CSS, so this solution is not ideal. Daytona offers inbuilt styling, so using Daytona over WebView2 would avoid accessibility bugs, and cut out the effort required to write our own CSS styling. The markdown editor team is currently in the process of transitioning from using WebView2 to Daytona, and recommended that, if possible, our team should use Daytona from the beginning.
```

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+### Functional Explanation
+
+The goal of this new experience would be to integrate the experience of viewing a rendered Readme directly into NuGet in Visual Studio. Additionally, we want to draw more attention to the README in general, and make it feel like an important and crucial element of a package.  
+
+The desired experience would somewhat mirror the user interface on NuGet.org, which already supports rendered Markdown viewing. When a customer views the package detail page of a package with an included markdown readme, they will see this proposed UI mock ![Mock](https://github.com/NuGet/Home/assets/89422562/89b0295c-64d5-42a4-a52c-83dea2807edc)
+
+
+### Technical Explanation
+
+To render markdown in Visual Studio, we can use Markdig to convert the MD to HTML, and then use either Webview2 or Daytona to render the HTML and stylize.
```
jebriede, line 29: We should introduce the abbreviation of MD at the first use of the abbreviation.

E.g.: `markdown (MD)` then we can refer to it as MD in the rest of the document.

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+
+Support rendering markdown readmes in the package details page for NuGet in Visual Studio to make package affordances more accesible for customers browsing packages within the PMUI, and to increase motivation for high quality Readme inclusion for package authors. This will be part of a larger effort to enhance the experience around Readmes and increase their inclusion rate across packages in the NuGet ecosystem. 
+
+## Motivation
+
+In the latest NuGet in Visual Studio 17.5 customer survey, our team found that one of the largest problems that our customers face while using NuGet in the Visual Studio ecosystem is around understanding package affordances – specifically calling out the usage of package metadata such as READMEs. Only 35.35% of survey respondents feel that they can adequately understand the affordances of a package (does it do what I need to to do when I need it?) while browsing and installing packages. Including a detailed README file and implementing features to facilitate README viewing and inclusion would positively impact this number.   
+
+## Explanation
+
+### Functional Explanation
```
jebriede, line 16: This section should outline details about how the experience will work. Currently, the document is missing details that are relevant to the UI in Visual Studio, such as consideration for installed packages, multiple package sources, and packages with README.txt files.

A few questions that come to mind that this section should answer as part of the functional spec (probably not an exhaustive list, happy to collaborate on ideas, but these are the ones that come to mind immediately):

1. Will README files only be rendered for packages in the Browse tab?
2. Will we show a README for installed packages in the Installed tab? If so, it would be worth calling out where we will be getting this README from. I would imagine we'd be pulling it in from disk rather than from the remote package source.
3. Will we show README for packages in the Updates tab?
5. Will we show README markdown for the Solution-level package manager or just Project-level package manager?
6. What will we do in cases where there's a README.txt file for an installed package? Will we display that as well in the details pane or is this experience only for README if it's in markdown format?
7. What will we show in the details pane if there is no README for the selected package?

proposed/2023/PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,47 @@
...
+
+None
+
+
+## Rationale and Alternatives
+
+An alternative here would be to include a link within the package detail page which would open up the rendered readme externally. Should we find complications with directly embedding the rendered readme into the packag detail page, this would be a proposed alternative.
+
+
+## Unresolved Questions
```
jebriede, line 50: I'm not sure where to leave this comment so I chose this section. :-)

What are the measures of success of this feature? What will we be using to determine if this experience of showing the README in the details pane has had the desired effect and meets the goal outlined in the section above?

jebriede (dismissed):
Thanks for writing up this spec! This is an exciting addition to consider for the PM UI. I've left some comments. Please reach out if I can be of help defining the experience or if you have any questions!

JonDouglas (approved):
I think this proposal is ready to merge. Implementation details can come when we start work on it.

PR #12733, opened 2023-07-09
Cluster

paulnsac:
-P

### DIFF
--- a/.github/workflows/ibm.yml
+++ b/.github/workflows/ibm.yml
@@ -1,0 +1,75 @@
+# This workflow will build a docker container, publish it to IBM Container Registry, and deploy it to IKS when there is a push to the "dev" branch.
+#
+# To configure this workflow:
+#
+# 1. Ensure that your repository contains a Dockerfile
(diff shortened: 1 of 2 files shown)


zivkan:
Considering the contents of this repo, this PR makes absolutely no sense, so it doesn't look like an honest mistake of something someone intended for their own fork. At the same time, it's not obviously malicious, as it would require this repo to have certain secrets with the specific names. So, I have no idea what to make of this, but in any case it's just not appropriate.

paulnsac:
How may we handle this

paulnsac:
Friend since we are both aggravated  come to an agreement on solving important issues

zivkan:
Handle what? The PR title doesn't make sense, the opening comment "-P", makes no sense.

I'm replying just in case you're a beginner learning, but your two comments since I closed the issue are generic like some chat bot. And neither of your two comments following my first comment address the point I made saying that the PR doesn't make sense in the context of this repo.

So, if you're a human, please explain what you're trying to achieve. Otherwise, I assume this is a bot account.

PR #12735, opened 2023-07-11
Remove projects from list package command

Greybird:
Hello,

This proposal aims to remove projects from `list package` command output

[Rendered Proposal](https://github.com/NuGet/Home/blob/dc1912eceba3914536c276600e36dbcc928f6801/proposed/2023/remove-projects-from-list-package-command.md)

### DIFF
--- a/proposed/2023/project-exclusion-for-transitive-dependencies.md
+++ b/proposed/2023/project-exclusion-for-transitive-dependencies.md
@@ -11,12 +11,15 @@ Currently, when listing transitive packages using `dotnet list package --include
 
 ## Motivation 
 
-Packages and projects are two vastly different types of references.
-Packages are published to a repository, allowing consumption by multiple pieces of software and repositories. Metadata are attached to packages or package versions, either included in the package, or added as a decoration - for deprecation or vulnerabilities for example -, which can be consumed to take decisions. On the contrary, projects are _local_ entities, living in a single repository, and do not have an existence outside of it.
+While packages and projects are very close - packages being packed projects -, packages differ from project because they can be published to a package repository. This package repository allows metadata not included in the package initially, to be added as a decoration - for deprecation or vulnerabilities for example -. In turn, these additional metadata can be consumed to take decisions.
 
 The `list package` command already supports filtering the results to retrieve only deprecated (`--deprecated`), vulnerable (`--vulnerable`), or not up-to-date (`--outdated`) packages. These commands only return actual packages, which is understandable as the underlying information used to choose what to report is managed by the _package_ manager, aka nuget.
 
-Allowing to remove projects from the output when none of these flags is provided would allow users to get an accurate view of their package dependencies. This would ensure that the output of the command can be processed in an uniform way by either a user, or a piece of software - through the machine readable output - to support any report, or process associated with packages, even those with no issues related to deprecation, vulnerabilities, or up-to-dateness.
+When `list package` is used without any of these flags,
+* using `--include-transitive` lists `direct` project dependencies as transitive.
+* not using `--include-transitive` does not list project dependencies at all.
+
+Allowing to remove projects from the output would allow users to get an accurate view of their package dependencies. This would ensure that the output of the command can be processed in an uniform way by either a user, or a piece of software - through the machine readable output - to support any report, or process associated with packages, even those with no issues related to deprecation, vulnerabilities, or up-to-dateness.
 
 ## Explanation
 
@@ -112,6 +115,8 @@ It adds a new option to a command, which could result in it being more difficult
 This design is retro compatible with the previous behavior of the command. It also limits the output of the command to the expected results.
 
 Other approaches could have been taken:
+* Considering that including projects into the output is a bug, and fixint it.
+* Considering that projects are part of the output, and ensuring that direct (resp. transitive) project references are reported as direct (resp. transitive). In this case, adding a flag to be able to retrieve only package is probably worth it, like mentioned below.
 * Adding a `--include-project` option and modifying the default behavior of the list command: this would have made the whole command consistent, whether a `--deprecated`, `--vulnerable`, or `--outdated` option is passed or not. However, this would create a breaking change with previous behavior
 * Adding the type of dependency to the output: this would have made the output filterable afterwards, without adding flags. However, this could break console output parsing - which could be acceptable -, and could force an `output-version` increment as the machine-readable format would change. It would also require an active filtering process after the command is run. Whether this proposal is implemented, this could be an interesting option if consequences are accepted.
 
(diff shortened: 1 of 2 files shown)


JonDouglas:
@Greybird Thank you so much for the proposal. Let me put together a small team review in the next couple of weeks and you'll start to see some comments on this PR. Thanks for your patience while we do this!

proposed/2023/project-exclusion-for-transitive-dependencies.md
```diff
@@ -0,0 +1,130 @@
...
+- GitHub PR https://github.com/NuGet/NuGet.Client/pull/5302
+
+## Summary
+
+Currently, when listing transitive packages using `dotnet list package --include-transitive`, both project and packages dependencies are reported. The goal of this proposal is to provide a way to filter the projects out of the report.
+
+## Motivation 
+
+Packages and projects are two vastly different types of references.
+Packages are published to a repository, allowing consumption by multiple pieces of software and repositories. Metadata are attached to packages or package versions, either included in the package, or added as a decoration - for deprecation or vulnerabilities for example -, which can be consumed to take decisions. On the contrary, projects are _local_ entities, living in a single repository, and do not have an existence outside of it.
```
nkolev92: This paragraph is probably too strong of a statement. 
Projects and Packages are incredibly close. When you use PackageReference, and you pack a project, by default, all the project references *become* package references. 

The NuGet resolver treats these equivalently. Packages can be preferred over projects if the situation calls for it. 

[ResolvingProjectAsPackageReference.zip](https://github.com/NuGet/Home/files/12059545/ResolvingProjectAsPackageReference.zip)

proposed/2023/project-exclusion-for-transitive-dependencies.md
```diff
@@ -0,0 +1,130 @@
...
+
+The `list package` command already supports filtering the results to retrieve only deprecated (`--deprecated`), vulnerable (`--vulnerable`), or not up-to-date (`--outdated`) packages. These commands only return actual packages, which is understandable as the underlying information used to choose what to report is managed by the _package_ manager, aka nuget.
+
+Allowing to remove projects from the output when none of these flags is provided would allow users to get an accurate view of their package dependencies. This would ensure that the output of the command can be processed in an uniform way by either a user, or a piece of software - through the machine readable output - to support any report, or process associated with packages, even those with no issues related to deprecation, vulnerabilities, or up-to-dateness.
```
nkolev92: Testing on the latest .NET SDK, projects *are* not included in the default output at all. 
With `--include-transitive`, they are included in the transitive packages (which is incorrect). 

See my example from above: 

Project Root -> Project INtermediate (1.0.0) -> Project Leaf (1.0.0)
Project Root -> Package Leaf (2.0.0)

In this case the output is:

```xml
Project 'Root' has the following package references
   [net8.0]:
   Top-level Package      Requested   Resolved
   > Leaf                 2.0.0       2.0.0

   Transitive Package      Resolved
   > Intermediate          1.0.0
```

Seems like there's some intent behind that and we migt just be dealing with a bug.

proposed/2023/project-exclusion-for-transitive-dependencies.md
```diff
@@ -0,0 +1,130 @@
...
+   [net7.0]:
+   Top-level Package      Requested   Resolved
+   > package3             30.0.1      30.0.1
+
+   Transitive Package       Resolved
+   > package4               1.0.5
+   > projectA               1.0.0
+```
+
+Adding `--exclude-project` to the command would make `dotnet list solution.sln package --include-transitive --exclude-project` produce the following output:
```
nkolev92: I think these 2 examples actually do summarize that we're having an inconsistent behavior and understanding that may be priority #1. 

If the command truly did list projects, `projectA` wouldn't be listed as transitive to projectB, it's a direct reference. I'm guessing it just couldn't have a root (?) and it got left there.

proposed/2023/project-exclusion-for-transitive-dependencies.md
```diff
@@ -0,0 +1,130 @@
...
+
+If we do not implement this option, then users of the command will have to be able to filter projects out of the output. However, this is not an easy task, and it is subject to errors if for example a project and a package have the same name.
+
+## Prior Art
+
+* [Machine readable JSON output for dotnet list package](https://github.com/NuGet/Home/blob/a6fea29b47359df9a1a03c381459e2b72c6062e9/proposed/2022/DotnetListPackageMachineReadableJsonOutput.md)
+
+## Unresolved Questions
+
+* Should the nature of the dependency (project or package) be included in the command output (or at least in the machine-readable one) ?
```
nkolev92: IMO, I think that might be a miss in the command in general. 

The project graph is most valuable when you know which packages are brought through your project.

nkolev92 (commented):
Digging deeper, we might just be dealing with a bug, rather than something that requires a new option. 

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.CommandLine.XPlat/Utility/MSBuildAPIUtility.cs#L781-L783

proposed/2023/project-exclusion-for-transitive-dependencies.md
Greybird: I added details on the behavior of the command, and mentioned that packages are reported as transitive dependencies hile they are direct, for the sake of clarity.

Greybird:
Thanks for your comments.
I proposed some changes related to the inaccuracies your pointed in the proposal.
Of course, depending on the direction the discussion will take, its content might change vastly, or be void.

Greybird:
@nkolev92 I tried running the code from the initial introduction of the feature (in https://github.com/NuGet/NuGet.Client/pull/2358,) and it seems to me that it returned the project at that time (had to play a bit with a few things to make this work, so cannot be 100% assertive).
The requirements at the time of development are unclear from the pull request threads, and the code (no unit tests for example), but I suppose it is possible that projects might not have been considered at all ? As such, subsequent code (and especially the one which introduced the comments about excluding projects) might have been written to maintain an existing behavior with no underlying reason ?

martinrrm:
@Greybird Thanks for the contribution! We are going to have a meeting to review this design, we should be able to get back to you next week.

Greybird:
@martinrrm : Thanks, I just filled another PR, https://github.com/NuGet/NuGet.Client/pull/5335, to remove projects from being reported if this is considered the way to go, as per @nkolev92 comments. Of course they are mutually exclusive.

martinrrm:
@Greybird Thanks for filling the other PR, it helped us a lot to decide what decision to make. We decided that removing project is the way to go (https://github.com/NuGet/NuGet.Client/pull/5335). Could it be possible to update this spec to reflect the PR instead? Please let me know if there is anything I can help you with. Thanks again for helping us fix this bug!

Greybird:
Hi @martinrrm, I submitted a new proposal and updated the PR accordingly.
Please tell me if you need other changes, I'll gladly implement them!

nkolev92 (approved):

nkolev92:
Thanks for updating the spec @Greybird!

I'll merge this given the client PR is already merged as well.

PR #12780, opened 2023-07-28
[Draft][Proposal] PMUI Tab Consolidation

albarry4:
This proposal presents a concept for a consolidated NuGet UI in Visual Studio. 

[Rendered Spec](https://github.com/NuGet/Home/blob/42bc88abc32b77bd1cc42b583c39b0f213e8840f/proposed/2023/pmui-tab-consolidation.md)

Thank You 🎉

### DIFF
--- a/proposed/2023/pmui-tab-consolidation.md
+++ b/proposed/2023/pmui-tab-consolidation.md
@@ -1,0 +1,71 @@
+# NuGet PMUI Tab Consolidation
+
+- Allie Barry @albarry4
+- Start Date 7/27/2023
+
+## Summary
+
+This proposal suggests a solution to reduce the 3 tab format of the NuGet PMUI in Visual Studio as a first step to simplify and modernize the UI, and to address recent customer feedback.  
+
+## Motivation 
+
+We have recently recieved feedback from custmers that the NuGet PMUI design is "clunky" and "requires too much clicking over all the tabs, checkboxes... takes too much clicking and surfing to navigate to varius changelogs". Based on this feedback, I wanted to propose an initial solution to specifcially target and reduce the amount of clicking, and to simplify and consolidate the controls within the PMUI by reducting the number of tabs available. Specifically, this proposal aims to consolidate the "Installed" and "Updates" tabs into a single window titled "My Packages".
+
+Note: I want to include data here about tab click-through rates, but am still working on getting access to that data!
+
+## Explanation
+
+### Functional explanation
+
+Put simply, the end goal of this proposal is to reduce the number of tabs in the NuGet PMUI, and to simplify the format of the User interface. 
+
+In this new experience, instead of seeing three tabs ("Browse", "Installed", "Update"), the user will see two tabs at the top of the screen, "Browse", and "My Packages". Below, I will describe how these two tabs will behave. 
+
+The "Browse" tab will remain unchanged from the current design. 
+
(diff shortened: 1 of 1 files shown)


proposed/2023/pmui-tab-consolidation.md
```diff
@@ -0,0 +1,71 @@
...
+
+In this new experience, instead of seeing three tabs ("Browse", "Installed", "Update"), the user will see two tabs at the top of the screen, "Browse", and "My Packages". Below, I will describe how these two tabs will behave. 
+
+The "Browse" tab will remain unchanged from the current design. 
+
+Within the "My Packages" tab, the user will see a list of the packages that they have installed in the current project. The layout of the tab will remain mostly unchanged from the general layout of the seperate "Installed" and "Updates" Tabs that we see today, but with a few tweaks to consolidate all of this information into one tab. 
+
+If no installed backages have an update available, this is what the user will see:
+
+![image](https://github.com/albarry4/Home/assets/89422562/ce769bc6-ae6d-48b0-b093-730881d7ef2d)
```
nkolev92, line 30: In general, prefer to reference resources within the repo.

proposed/2023/pmui-tab-consolidation.md
```diff
@@ -0,0 +1,71 @@
...
+
+
+If the user has a package installed with an update available, this is what the user will see:
```
nkolev92, line 33: I like how this looks if my goal is to look at the packages that have updates only.
Should this be the default though? Or do we want a filter/checkbox something that does this separation. 

Is there a world where the packages with updates/without updates are within a single list?

proposed/2023/pmui-tab-consolidation.md
```diff
@@ -0,0 +1,71 @@
...
+
+If the user has a package installed with an update available, this is what the user will see: 
+
+![image](https://github.com/albarry4/Home/assets/89422562/f13f6a26-e023-4aca-bb47-826581d61952)
+
+
+
+### Technical explanation
+
+All packages with an update available will have a checkbox next to it, with the option at the top of the screen to select all, and bulk update all packages with an available update, or handle them on a one-at-a-time basis. This behaves similarly to the current functionality of the "Updates" tab
```
nkolev92, line 41: This sounds awesome.

proposed/2023/pmui-tab-consolidation.md
```diff
@@ -0,0 +1,71 @@
...
+
+
+## Prior Art
+
+The basis of this design comes from other similar and more modern virtual software marketplace designs we see today such as the Microsoft Store app for Windows. The current experience in the Microsoft Store allows for the user to browse apps of different types (Apps, Games, Entertainment), and then there is one single tab view titled "Library" which displays all of the different apps that a user has installed, and when an update is available in one of these apps, it is bubbled up to the top and the user can install the update. Once the update has completed, the app resturns the the list with the rest of the up to date apps. The end-goal experience for NuGet would be similar to this. 
+
+Below shows a screenshot of how this experience looks when an app on the Microsoft Store has an update available.
```
nkolev92, line 58: Drawing inspiration from this is reasonable. 
To only thing for me is that in the MS store, you nearly *always* want to be on latest, so updates being first makes a lot of sense. With packages that's not always the case as there are more things that come with updating a package (tests/runtime validaiton and sometimes just ignoring certain updates for many different reasons.).

nkolev92 (commented):
Looks great. Some suggestions.

proposed/2023/pmui-tab-consolidation.md
erdembayar: Is this view same for both project and solution?
What happens to `Consolidate` tab?

PR #12853, opened 2023-08-28
Update the spec template

nkolev92:
- Added a special note to set the correct expectations that the title should be amended to better reflect the proposal.
- Removed Start Date - Not sure what value this carries
- Removed GitHub PR - Which PR? Product PR? Usually that comes after the design. Spec repo PR? A bit of chicken and egg problem here.

### DIFF
--- a/meta/template.md
+++ b/meta/template.md
@@ -1,4 +1,4 @@
-# Title
+# ***Untitled***
 <!-- Replace `Title` with an appropriate title for your design -->
 
 - Author Name <!-- GitHub username link -->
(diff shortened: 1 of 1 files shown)


donnie-msft (approved):

meta/template.md
```diff
@@ -1,9 +1,8 @@
 # Title
```
donnie-msft:
```suggestion
# ***Untitled***
```

nkolev92: Sure, I don't mind it. 

We can even go aggressive and say 
***Title goes here***

heng-liu (approved):

zivkan (approved):

Issue #5138, opened 2017-04-28
exact match and version search for nuget.exe list command


jjanuszkiewicz:
(See https://nuget.codeplex.com/workitem/2130 for a related discussion)

I'd like to be able to find a specific package (exact match on package ID) using the nuget.exe command line client. I'd also like to be able to find a specific version of the package (to check if that version exists in the feed).

Something like:

    # list the most recent version of MyPackage, matching by exact package ID
    nuget.exe list MyPackage -ExactMatch

    # list all versions of MyPackage, matching by exact package ID
    nuget.exe list MyPackage -ExactMatch -AllVersions

    # list the provided version of MyPackage, matching by exact package ID
    nuget.exe list MyPackage -ExactMatch -Version 1.2.3.4

So in short, the -ExactMatch option would cause the search term to be an exact match on the package ID. The -Version option (which probably only makes sense with -ExactMatch) would cause only the specified version to be listed (if it exists).

I would use this in scripts (outside of Visual Studio) instead of having to rely on hacks (e.g. trying to install a specific version of the package and checking if it succeeds or not).

DannyMeister:
I have the additional use cases for the -Version option:
```
# list the available compatible versions of MyPackage via version wildcard
nuget.exe list MyPackage -ExactMatch -Version 1.2.*

# find the available beta releases for a specific version of MyPackage
nuget.exe list MyPackage -ExactMatch -Version 2.0.0-beta*
```
I would think probably anything supported by the Version attribute of `<PackageReference>` would probably make sense, though I'm not sure whether that currently supports my beta requirement.

Alchemy86:
Is this possible yet at all

nigurr:
Hi,

We have similar case where we want to get all versions of specific package. 
We don't need to even to search for package as we know exactly what we are looking for.

medianick:
Here's my workaround:

`nuget list SomePackage -AllVersions | findstr /E " 1.2.3"`

This will set ERRORLEVEL to 0 if SomePackage 1.2.3 exists in the source, but 1 if it is not. The combination of the `/E` switch (match end of line only) and the space before the version causes an exact match of the full version number rather than erroneously matching a substring of it.

jstutson:
This is such a common need in script publishing, especially for CI builds.  I'm extremely surprised there is no method for asking a remote repo whether it has a package at a given version.

nigurr:
Surprised to see there's not any traction nor feedback from the owners. It's being there for more than an year

[...]

PR #12870, opened 2023-09-06
Spec: Improve NuGet Package Search functionality

Nigusu-Allehu:
Spec for improving NugGet package search across nuget.exe and dotnet. 
Issue: https://github.com/NuGet/Client.Engineering/issues/1803

[Rendered Spec](https://github.com/NuGet/Home/blob/fdf89db34cf2d710ba9c710f61cfcb48f90db729/proposed/2023/nuget-package-search.md)
### Update
Thanks to all of you for your reviews! Based on your feedback, as @nkolev92 suggested, it's clear that splitting the specification into two or more distinct documents: one for transitioning from nuget.exe list to nuget search, and another for introducing dotnet package search will help me address your concerns more comprehensively. In the meantime, please feel free to make further reviews and suggestions. 

Here is the spec for dotnet package search : https://github.com/NuGet/Home/pull/12886

### DIFF
--- a/proposed/2023/nuget-package-search.md
+++ b/proposed/2023/nuget-package-search.md
@@ -11,7 +11,7 @@ This specification will discuss three key improvements to the NuGet package sear
 ## Motivation 
 
 <!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
-Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a not confusing search functionality on both the Dotnet CLI and nuget.exe.
+Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a more straight-forward search functionality on both the Dotnet CLI and nuget.exe.
 ## Explanation
 
 ### Functional explanation
@@ -22,20 +22,22 @@ Improving the nuget search functionality across all platforms is highly requeste
 A new flag, `-ExactMatch`, will be added to the `search` command. When this flag is used, it will only return packages whose names exactly match the query.
 - **Version not specified** : In this case, the latest version of the package specified is listed .\
 *Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch` \
-*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch`
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match`
 - **Version specified** : In this case, the package with the specific version number is listed.\
 *Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -Version 1.2.3.4` \
-*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch -Version 1.2.3.4`
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match --version 1.2.3.4`
 - **All Versions** : In this case, all versions of the package with the exact match are listed.\
 *Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -AllVersions` \
-*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch -AllVersions`
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match --all-versions`
+
+If no exact match is found, display an error: `Error: An exact match for package <PackageName> is not found.`
 #### 2. Deprecating list Command: 
 The list command is planned to be deprecated, given its functionalities can be covered by the search command. A warning message will be displayed to inform users of this change. `Warning: The 'list' command is deprecated. Use 'search' instead.` \
 The functionalities of the `list` command can be replicated as follows
 * In nuget.exe `nuget list` will be `nuget search`.
 * In Dotnet, `dotnet nuget search` will be added and it will be the same as `nuget search`.
 #### 3. Introducing nuget search to Dotnet:
-A new command dotnet nuget search will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
+A new command `dotnet nuget search` will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
 ### Technical explanation
 
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
--- a/proposed/2023/nuget-package-search.md
+++ b/proposed/2023/nuget-package-search.md
@@ -1,0 +1,72 @@
+# NuGet Package Search
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- [Nigusu](https://github.com/Nigusu-Allehu) <!-- GitHub username link -->
+- [GitHub Issue](https://github.com/NuGet/Home/issues/6060) <!-- GitHub Issue link -->
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+This specification will discuss three key improvements to the NuGet package search functionality. These improvements are: adding an `-ExactMatch` option to the `search` command, deprecating the `list` command in favor of `search`, and introducing a `dotnet nuget search` command to the Dotnet CLI.
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a not confusing search functionality on both the Dotnet CLI and nuget.exe.
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option: 
+A new flag, `-ExactMatch`, will be added to the `search` command. When this flag is used, it will only return packages whose names exactly match the query.
+- **Version not specified** : In this case, the latest version of the package specified is listed .\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch`
+- **Version specified** : In this case, the package with the specific version number is listed.\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -Version 1.2.3.4` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch -Version 1.2.3.4`
+- **All Versions** : In this case, all versions of the package with the exact match are listed.\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -AllVersions` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch -AllVersions`
+#### 2. Deprecating list Command: 
+The list command is planned to be deprecated, given its functionalities can be covered by the search command. A warning message will be displayed to inform users of this change. `Warning: The 'list' command is deprecated. Use 'search' instead.` \
+The functionalities of the `list` command can be replicated as follows
+* In nuget.exe `nuget list` will be `nuget search`.
+* In Dotnet, `dotnet nuget search` will be added and it will be the same as `nuget search`.
+#### 3. Introducing nuget search to Dotnet:
+A new command dotnet nuget search will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+The `-ExactMatch` flag will require modifying the query parser for the `search` command in NuGet.
+
+For deprecating the list command, the current implementation needs to be flagged as obsolete and redirect users to the `search` command with a warning message.
+
+Lastly, the introduction of nuget `search` to Dotnet will require adding a new command.
+## Drawbacks
+
+<!-- Why should we not do this? -->
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+NuGet.exe already supports package search with nuget search but lacks an exact match feature. Dotnet CLI also has a list command but serves a different purpose. Aligning these functionalities can simplify the user experience.
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+What would be the transition plan for users currently relying on the list command?
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->


proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,72 @@
...
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a not confusing search functionality on both the Dotnet CLI and nuget.exe.
```
donnie-msft: nit: 
Instead of: `with a not confusing`
suggestion: `with a more straight-forward `

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,72 @@
...
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option:
```
donnie-msft, line 21: The positive cases are clear, but what if the exact match isn't found in each scenario? If they're all the same, listing it once is fine.

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,72 @@
...
+- **All Versions** : In this case, all versions of the package with the exact match are listed.\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -AllVersions` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch -AllVersions`
+#### 2. Deprecating list Command: 
+The list command is planned to be deprecated, given its functionalities can be covered by the search command. A warning message will be displayed to inform users of this change. `Warning: The 'list' command is deprecated. Use 'search' instead.` \
+The functionalities of the `list` command can be replicated as follows
+* In nuget.exe `nuget list` will be `nuget search`.
+* In Dotnet, `dotnet nuget search` will be added and it will be the same as `nuget search`.
+#### 3. Introducing nuget search to Dotnet:
+A new command dotnet nuget search will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
```
donnie-msft:
```suggestion
A new command `dotnet nuget search` will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
```

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,72 @@
...
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option: 
+A new flag, `-ExactMatch`, will be added to the `search` command. When this flag is used, it will only return packages whose names exactly match the query.
+- **Version not specified** : In this case, the latest version of the package specified is listed .\
```
donnie-msft, line 23: Are we sure about defaulting to latest version with `ExactMatch`? I wouldn't intuitively think that is what would happen by specifying `-ExactMatch` without a parameter. 

Our existing `list` behavior breakdown is:
1. `(no version options specified)`: list latest version
2. `-AllVersions`: list all versions
3. Exact isn't possible

So why not introduce number 3 above as `-ExactMatch <version required>` ?

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,72 @@
...
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> -ExactMatch`
```
donnie-msft: For .NET CLI, the format would change to `--exact-match`, I believe. Examples: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-nuget-push

donnie-msft (commented):
The motivation section is strong! I like the explanation as well, but have a few points raise...

proposed/2023/nuget-package-search.md
Nigusu-Allehu: Good catch. I listed it once, as they all should output the same error.

proposed/2023/nuget-package-search.md
Nigusu-Allehu: I see, it could be confusing to users. However, the current implementation of `list` does not achieve an exact match. For example, if I were to do `nuget list packageA` and a package named `PackageA`, and `PackageAmazing`  exist, both packages will be printed out. However, the user wants the exact match only. With `nuget search PackageA -ExactMatch` only the latest version of `PackageA` will be printed out as the user wants the exact match only. if they specify the version, that specific version will be printed out., or if they do `-AllVersions`, all of them will be displayed. If we were to include `-ExactVersion` as the third behaviour, users would not be able to achieve: 
1. Exact matching a package name without a version number
2. Listing all versions of an exact match, not just all matching packages.

JonDouglas (changes requested):
Well put together. Here's my main comments.

We should only backport functionality to `nuget CLI` if we get it for relatively no cost while implementing `dotnet CLI`. I believe that's the case for many our our shared commands today.

For the deprecation, I would suggest we don't "deprecate", but rather we just alias the new term called search. I don't see strong evidence to go through further deprecation steps here, they seem wildly disruptive for little value.

`dotnet nuget search` could also be `dotnet search package` or something related around the package noun and command group. I think it is more ideal there than under the `nuget` group from a UX perspective. Perhaps we need more data there and can do a twitter survey for a fast response?

As for the actual `dotnet search package` or `dotnet nuget search` (or even `dotnet search`) functionality, I believe a new proposal should be created to outline the goals of the design and how it would differ from `nuget CLI`'s current experience. In my opinion, this requires a bit more dedication and thought rather than to just re-implement what the `nuget CLI` provides today. If we want to start with a pure port of `nuget CLI` to `dotnet CLI`, that has merit. But we should keep in mind that there is lots of potential to make this a great core experience with the help of @baronfel and team :)

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match --all-versions`
+
+If no exact match is found, display an error: `Error: An exact match for package <PackageName> is not found.`
+#### 2. Deprecating list Command: 
+The list command is planned to be deprecated, given its functionalities can be covered by the search command. A warning message will be displayed to inform users of this change. `Warning: The 'list' command is deprecated. Use 'search' instead.` \
+The functionalities of the `list` command can be replicated as follows
+* In nuget.exe `nuget list` will be `nuget search`.
+* In Dotnet, `dotnet nuget search` will be added and it will be the same as `nuget search`.
+#### 3. Introducing nuget search to Dotnet:
+A new command `dotnet nuget search` will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
```
JonDouglas, line 40:
```suggestion
A new command `dotnet nuget search` will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `nuget list` functionality.
```

martinrrm:
Looking at the current [documentation of nuget search command ](https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-search), I don't see the option `-AllVersions` are we planning in adding that feature in this work too?

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option:
```
kartheekp-ms, line 21: I am curious to understand, what are the advantages or practical use cases for the `-ExactMatch` option in `dotnet nuget search` command? Specifically, when a user already knows the precise package name they need, how does this option enhance their experience or benefit them?

One use case, I can think of is to discover available versions. Customers may want to use the `-ExactMatch` option to discover all available versions of a specific package. For example, they might want to see a list of all versions for a package before deciding which exact version to install.

Are there any other use cases?

kartheekp-ms, line 21: Does the `dotnet nuget search` command output include information about `deprecated and/or vulnerable` package versions? If we consider this command as equivalent to the NuGet.org search or Visual Studio Package Manager UI search, then I think it would be beneficial to customers to display such information.

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+The list command is planned to be deprecated, given its functionalities can be covered by the search command. A warning message will be displayed to inform users of this change. `Warning: The 'list' command is deprecated. Use 'search' instead.` \
+The functionalities of the `list` command can be replicated as follows
+* In nuget.exe `nuget list` will be `nuget search`.
+* In Dotnet, `dotnet nuget search` will be added and it will be the same as `nuget search`.
+#### 3. Introducing nuget search to Dotnet:
+A new command `dotnet nuget search` will be introduced to bring package search functionality to the Dotnet CLI. This will help in resolving the highly requested `dotnet nuget list` functionality.
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+The `-ExactMatch` flag will require modifying the query parser for the `search` command in NuGet.
```
joelverhagen, line 44: Which HTTP resource will `-ExactMatch` use? For the [search resource](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource), nuget.org supports `packageid:My.Exact.Id` to return a single result in the search results. But this is not guaranteed to work across various package feeds. You can perhaps consider using the [package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource) (aka registration) resource, you can get exact ID metadata but there is no guarantee the metadata available on search vs. package metadata is going to be the same.

For the client-side filtering option, the unfortunate reality is that a poorly implemented, although spec compliant search resource may not return an exact match in the first page of results. In fact, there were periods of time when exact match didn't come first (or on the first page of results) on NuGet.org! So I don't see client side filtering as ideal either.

Generally, I'm curious about the specific details of "require modifying the query parser".

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option: 
+A new flag, `-ExactMatch`, will be added to the `search` command. When this flag is used, it will only return packages whose names exactly match the query.
+- **Version not specified** : In this case, the latest version of the package specified is listed .\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match`
+- **Version specified** : In this case, the package with the specific version number is listed.\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -Version 1.2.3.4` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match --version 1.2.3.4`
```
joelverhagen, line 28: The search endpoint does not return metadata associated with previous versions. It only returns the version number and the download count. I see the `nuget.exe search knapcode.torsharp` shows the description as well. The search index does not contain descriptions for anything but the latest version (given the prerelease and semverlevel flags).

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+- **All Versions** : In this case, all versions of the package with the exact match are listed.\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch -AllVersions` \
+*Dotnet Usage* : `dotnet nuget search <PackageName> --exact-match --all-versions`
+
+If no exact match is found, display an error: `Error: An exact match for package <PackageName> is not found.`
```
joelverhagen, line 33: Does this kind of error get an `NU` code?

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+#### 2. Deprecating list Command:
```
joelverhagen, line 34: The `list` command appears to support `-IncludeDelisted` and `-AllVersions`. I see `-AllVersions` getting pulled forward into `search` but not `delisted`. Is that by design?

By the way, the `list -AllVersions` option has not worked for years on NuGet.org, it's tracked here: https://github.com/NuGet/NuGetGallery/issues/3274. This may mean you have an opportunity to redefine the behavior with less back-compat impact. Or it means we're finally fixing a longstanding bug :)

proposed/2023/nuget-package-search.md
joelverhagen: +1 on this question. NuGet.org search endpoint now contains both vulnerable and deprecated information so you could consider leveraging that.

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a more straight-forward search functionality on both the Dotnet CLI and nuget.exe.
+## Explanation
+
+### Functional explanation
```
joelverhagen, line 17: I'm curious if we should anticipate the ask that we had on `dotnet list package` for `--format`, but for `dotnet search`.

nkolev92: I don't know if we need to commit to a format just yet, but I think the implementation should definitely separate the data collection + visualization. 

Reading the motivation behind `ExactMatch`, people basically want to use that for scripting.

proposed/2023/nuget-package-search.md
nkolev92: The motivation for this is linked in https://github.com/NuGet/Home/issues/5138. 

It's something that'd be used for scripting.

proposed/2023/nuget-package-search.md
nkolev92: `-AllVersions` doesn't really work and hasn't in a while (server bugs)

IMO we should start with a more minimal version of the command.

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+#### 1. Adding `-ExactMatch` Option: 
+A new flag, `-ExactMatch`, will be added to the `search` command. When this flag is used, it will only return packages whose names exactly match the query.
```
nkolev92, line 22: Will this be a client or server side implementation?

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+- **Version not specified** : In this case, the latest version of the package specified is listed .\
+*Nuget.exe Usage* : `nuget search <PackageName> -ExactMatch` \
```
nkolev92, line 24: I think it's ok for dotnet package search to be the only command with the new functionality such as exact match.

proposed/2023/nuget-package-search.md
nkolev92: For the most part, we've kept `NU` codes for pack/restore functionalities where these things are configurable via build.

proposed/2023/nuget-package-search.md
nkolev92: I think we should try to go with a slim version of search, at least in first proposal. 
We can add more options based on community feedback while getting this design reviewed or after we ship it.

proposed/2023/nuget-package-search.md
nkolev92: Do we have a pragmatic solution that works in every case? 

I'm wondering if we need to commit to a trade-off that's likely to cause the least amount of problems. 
IMO, using metadata resource and ignoring inconsistencies is likely to work the best. 

I think we should choose 1 and document others options though! Great call out!

nkolev92 (commented):
I think it would be great to have separate specs for the deprecation of nuget.exe list to nuget search and the addition of `dotnet package search`.

proposed/2023/nuget-package-search.md
joelverhagen: > IMO, using metadata resource and ignoring inconsistencies is likely to work the best.

Yes this is the way. You already have a unified model so strategically selecting output/display properties based on the intersection between search and registration seems like the right move.

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Improving the nuget search functionality across all platforms is highly requested by customers. As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command. In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality. Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source. The only difference between the two is that the search command allows for filters. In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates. Addressing these issues will provide users with a more straight-forward search functionality on both the Dotnet CLI and nuget.exe.
```
zivkan, line 14: If each sentence on a new line, then it's easier for people to comment on something specific. I think it's also much, much easier to read, although markdown is intended to be read rendered, not as code. Anyway, the rendered markdown doesn't start a new paragraph unless you have an empty line, so putting each sentence on a new line doesn't change the rendered result.

```suggestion
Improving the nuget search functionality across all platforms is highly requested by customers.
As shown in this [issue](https://github.com/NuGet/Home/issues/5138), customers would like to have `-ExactMatch` option added to the search command.
In this other [issue](https://github.com/NuGet/Home/issues/6060), we can see that it has been repeatedly requested by customers for the Dotnet CLI to adopt a search functionality.
Lastly, as discussed in this [issue](https://github.com/NuGet/Home/issues/7912) the `nuget list` command is redundant with `nuget search` as both commands basically display a list of packages from a source.
The only difference between the two is that the search command allows for filters.
In addition, the `nuget list` command can be confusing as there is a `nuget list` command in the Dotnet CLI with a different functionality: lists configured sources and client certificates.
Addressing these issues will provide users with a more straight-forward search functionality on both the Dotnet CLI and nuget.exe.
```

proposed/2023/nuget-package-search.md
```diff
@@ -0,0 +1,74 @@
...
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- [Nigusu](https://github.com/Nigusu-Allehu) <!-- GitHub username link -->
+- [GitHub Issue](https://github.com/NuGet/Home/issues/6060) <!-- GitHub Issue link -->
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+This specification will discuss three key improvements to the NuGet package search functionality. These improvements are: adding an `-ExactMatch` option to the `search` command, deprecating the `list` command in favor of `search`, and introducing a `dotnet nuget search` command to the Dotnet CLI.
+## Motivation
```
zivkan, line 11: formatting: keep an empty line around each heading/paragraph/etc.

If you're editing the markdown in VSCode, consider installing the markdownlint extension. It'll give you generic advice on how formatting and syntax to help improve readability and sometimes syntax errors. Useful when writing docs with images, to make sure the link to the image is correct.

proposed/2023/nuget-package-search.md
zivkan: Strong +1 from me on this. I'm sure there will be some customers who will be upset if it's not added to nuget.exe, but I believe our focus should very much be on the `dotnet` CLI, which works across Linux, Mac, and Windows. NuGet.exe techncally can as well , using Mono Project, but that's been in maintenance mode for a long time. Time spent on improving nuget.exe is less valuable than bringing new things to the ecosystem. I'd only add it to nuget.exe if it's almost zero effort.

proposed/2023/nuget-package-search.md
zivkan: NuGet.Client's source code is full of bad patterns we don't want to reproduce in new code. To me the question should be does an NU code provide customer benefit?

If a customer reports a problem to us, and they copy a message with an NU code, does that help us diagnose the problem more quickly?

Might a feature be confusing for inexperienced customers, and a docs page, easily found via an NU code, help them more quickly find relevant documentation?

Personally, I think some kind of "server doesn't contain package 'X'" is sufficiently clear, so I don't see a benefit in an NU code.

proposed/2023/nuget-package-search.md
Nigusu-Allehu: Thank you @kartheekp-ms , here is a separate [spec ](https://github.com/NuGet/Home/pull/12886) for dotnet package search. As you recommended, I have added `deprecated and/or vulnerable` as part of the output.

Issue #12766, opened 2023-07-20
dotnet list package --outdated, --vulnerable, --deprecated do not respect package source mapping


zivkan:
### NuGet Product Used

dotnet.exe

### Product Version

8.0.100-preview.5.23303.2

### Impact

It bothers me. A fix would be nice

### Repro Steps & Context

1) The NuGet.Client repo uses package source mapping. Add a mapping for package `Microsoft.SourceLink.GitHub` to the `dotnet-public` feed.
2) Run `dotnet list package --outdated` on any project in the repo to get the list of installed packages for which there is a newer version available.

**Expected:**
The output should show `Microsoft.SourceLink.GitHub` at version 1.1.1.

**Actual:**
The output shows `Microsoft.SourceLink.GitHub` at a version higher than 1.1.1, which indicates it is coming from a different package source and it's reporting a package version that cannot be restored.

Issue #12567, opened 2023-04-27
Support for credential providers deployed via .NET tools


JohnSchmeichel:
### NuGet Product(s) Affected

NuGet.exe, MSBuild.exe, dotnet.exe, NuGet SDK

### Current Behavior

Currently, credential providers support \*.exe for .NET Framework or \*.dll for .NET Core, and are usually split between two folders under the NuGet credential provider base path. For .NET Framework the providers are discovered by searching credentialprovider*.exe and executing that program, and for .NET Core they are discovered by searching credentialprovider*.dll and executing ```dotnet <credentialprovider>.dll```. Prior to .NET Core 2.0 this split was required as .NET Core only supported platform agnostic dll's (with the side affect of excluding platform dependent credential providers not based on .NET). With the latest versions of .NET however, this split is no longer necessary and requires multiple versions to be supported and deployed. This is also reinforced by the two plugin path environment variables ```NUGET_NETFX_PLUGIN_PATHS``` and ```NUGET_NETCORE_PLUGIN_PATHS``` which need to be set per framework version.

A deployment solution for dotnet is [.NET tools](https://learn.microsoft.com/en-us/dotnet/core/tools/global-tools) which provide a seamless .NET install and management experience for NuGet packages. Using .NET tools as a deployment mechanism has been a recurring feature ask for users of the .NET ecosystem which need to authenticate with private repositories. The ideal workflow would be (for private repositories like Azure DevOps):
1. Customers have dotnet CLI tools installed
2. Run ```dotnet tool install -g Microsoft.CredentialProviders```
3. Run ```dotnet restore``` with a private endpoint and it 'just works' (e.g. credential providers from step 2 are used during credential acquisition and used to authenticate against private endpoints)

This almost works today, but for Windows .NET Framework only. The goal is to support this cross-platform for all supported .NET runtimes.

### Problems
This work for Windows and .NET Framework today with the following caveats:
1. You must use the "tool path" option when installing to get the tool in a path known by NuGet. By default .NET tools are installed to %USERPROFILE%\\.dotnet\\tools which is a location not examined by NuGet today. This can be specified explicitly, but can be error prone and doesn't take into account any environment customization:
    ```
    dotnet tool install -g Microsoft.CredentialProviders --tool-path %USERPROFILE%\.nuget\plugins\netfx\CredentialProvider.Microsoft
    ```
[...]

Issue #4855, opened 2017-03-20
Expand `dotnet nuget` to include more of the commands that nuget.exe has such as dotnet nuget config 


automatonic:
`dotnet msbuild` is clever about its project/solution context and knows how to find the right version of msbuild to support the task at hand. This is especially useful for a build agent doing CI/CD work where there are potentially many projects with potentially many tool/sdk/runtime version profiles.

I think this is equally valuable in `dotnet nuget` commands, as evidenced by the existence of:

 - `dotnet nuget delete`
 - `dotnet nuget locals`
 - `dotnet nuget push`
 - `dotnet pack` (i assume)

But when working through automation tasks, I immediately wanted:

 - `dotnet nuget sources`

 > I was trying to dynamically build a `nuget.config` file at the solution level that had the correct PAT/credentials (secret build variables) to access a private nuget feed on a multi-tenant build agent without checking in secrets.

I could also see an argument for:

 - `dotnet nuget config`
 - `dotnet nuget delete`
 - `dotnet nuget setapikey`

I mentioned this to Nate McMaster [here](http://disq.us/p/1gzxs4z), and he encouraged me to take it up in this repository. 

Thanks for your time and consideration. Please let me know if there is a better/preferred channel in which to request this sort of feature set.

rrelyea:
Thanks for your asks. Helps us to prioritize future work here.
/cc @anangaur

mishra14:
Another request: https://github.com/dotnet/cli/issues/7339

natemcmaster:
+1 for `dotnet nuget sources` for adding credentials. This is important to us as we've been asked to switch to VSTS private feeds.

nkolev92:
I took a look at *all* the NuGet.exe commands and generated a table for the behaviors


| Command Name | Description | NuGet.exe | dotnet.exe | msbuild.exe | Notes |
|--------------|-------------|-----------|------------|-------------|-------|
| add | add package to hierarchical source | Yes | [No](https://github.com/NuGet/Home/issues/12004) | N/A | Parity not reached. No community asks I could find. |
| client-certs | Client certs management | Yes | Yes | N/A | Full parity. |
| config | Individual config items management | Yes | [Yes](https://github.com/NuGet/Home/issues/10294)  | N/A |  Full parity. |
| delete | Unlist package | Yes | Yes | N/A | Full Parity |
[...]

PR #12886, opened 2023-09-19
Spec for `dotnet package search` : implements a `search` functionality.

Nigusu-Allehu:
This PR 
- provides a spec for implementing `search` functionality in DotNet.
- Here is the implementation PR: https://github.com/NuGet/NuGet.Client/pull/5466
I previously made a spec that included this as a sub project, however, I noticed that making its own spec for it would be a better idea. Here is a link to the other PR : https://github.com/NuGet/Home/pull/12870 .


nkolev92:
This spec doesn't contain ideas such as `exactmatch`. I think the separation should be: 

`nuget.exe list` is deprecated in favor of `nuget.exe search` and `dotnet package search` as a completely new command with some new options.

Nigusu-Allehu:
> This spec doesn't contain ideas such as `exactmatch`. I think the separation should be:
> 
> `nuget.exe list` is deprecated in favor of `nuget.exe search` and `dotnet package search` as a completely new command with some new options.

I have updated it to include the Exact Match option.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- [Nigusu](https://github.com/Nigusu-Allehu) <!-- GitHub username link -->
+- [GitHub Issue](https://github.com/NuGet/Home/issues/6060) <!-- GitHub Issue link -->
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add `nuget list` functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.  
+## Motivation
```
zivkan: nitpick: empty line between headings and previous content. This happens multiple places throughout the file.

If you edit the markdown in VSCode, if you install the "MarkdownLint" extension, I believe it'll tell you hints like this.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+        NuGet.Bootstrapper | 2.6.0
+        CommandLineParser20 | 2.0.0
+        NuGet.VerifyMicrosoftPackage | 1.0.0
+        NuGet.for.MSBuild | 2.1.0 ...
+* Using ``dotnet package search NuGet.CommandLine --ExactMatch --Verbosity Quiet`` on the other side will have the following output
+
+         NuGet.CommandLine | 6.7.0
+## Drawbacks
+
+<!-- Why should we not do this? -->
```
zivkan, line 70: It hammers the server, which is a reason not to do it. Consider nuget.org, which claims to have 371,611 unique package IDs. What's the customer scenario where they "need" to list over 370,000 packages on the server?

Additionally, the request to add a `nuget list` equivalent to the `dotnet` CLI was created before a `dotnet nuget search` command existed. Therefore, it's entirely possible that a `dotnet nuget search` makes `nuget list` unnecessary.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+In nuget.exe there is `nuget list` command which does the same thing. However, customers would like this functionality to be available in dotnet.
```
zivkan: Similar to my other comment, do we have any signals that a search command would not meet their need?

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+This is a customer requested functionality as discussed in this issue: https://github.com/NuGet/Home/issues/6060. Adding the functionality will aid our customers to be able to use the command when writing scripts.
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+Imagine you have your dotnet CLI open. You would like to get the list of all packages from a source `<MySource>`. No worries, you can just use the following command : `dotnet package search --source <MySource>`. It will provide you with the list of all the packages in the source `<MySource>`.
```
zivkan: Why is listing all packages on a source more compelling than searching for a keyword?  I don't understand the use case.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+The `package search [search terms] [options]` command will have the following options 
+
+| Option | Function |
+|---------|:----------|
+| `--Source` | A list of packages to search |
+| `--ExactMatch` | Return exact matches only as a search result |
+| `--Verbose` | Displays a detailed ist of information for each package |
+| `--Prerelease` | Allow prerelease packages to be shown. |
+| `--IncludeDelisted` | Allow unlisted packages to be shown |
```
zivkan: If the implementation of this command is going to use the search endpoint, I don't think that `--IncludeDelisted` is going to work.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+| `--NonInteractive` | Do not prompt for user input or confirmations.|
+| `--Help` | Show command help and usage information |
+| `--Verbosity` | Display the amount of details in the output: normal, quiet, detailed. |
+| `--ConfigFile` | The NuGet configuration file. If not specified, the hierarchy of configuration files from the current directory will be used. |
```
zivkan: NuGet.exe follows PowerShell conventions, which uses a single `-` and Pascal Case for options. For example `-SomeArg`.

The `dotnet` CLI uses GNU conventions, which is a single `-` for single letter args, and `--` with snake case for multi-letter args. For example, `-s` or `--some-arg`. The single character, single `-` is because GNU's getopt library lets you use `-123` as equivalent to `-1 -2 -3`. I don't know if System.CommandLine, and therefore the `dotnet` CLI supports this.

These option names should be updated to meet the appropriate naming convention.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+|||
+#### **Algorithm**
+1. Read and parse arguments from user
+2. Use the nuget.exe Search API to look for the specific term in the provided source. The search filter will be composed of the `--Source`, `Search term`, `--Prerelease`, `--IncludeDelisted` values.
```
zivkan: I don't know what "the nuget.exe Search API" is.

I believe that `nuget.exe list` uses a NuGet Server V2 API (not documented), which means that it doesn't work with V3 servers, such as https://api.nuget.org/v3/index.json.  I don't think there's value in introducing a new command for a protocol we consider deprecated.

If you mean the [V3 protocol search resource](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource), then it might be worthwhile for this spec to acknowledge that search APIs don't gauruntee "correct" enumeration of all packages. Due to complexities in how search indexes work, plus a search index might be updated in the middle of a "list", some packages might get skipped, or some pacakges must get listed in multiple responses.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,100 @@
...
+                <Package Name> | <Latest Package Version> | <Amount of Downloads>
+                <Description of the Package(short form)>
+    - **Detailed** :
+
+                <Package Name> | <Latest Package Version> | <Amount of Downloads>
+                Deprecated : YES/NO | Vulnerable : YES/NO
+                <Description of the Package(Readme file)>
+                License URL : <URL>
+
+#### **New Option `ExactMatch`**
```
zivkan: ExactMatch doesn't seem relevant to a spec on proving a `nuget.exe list` equivalent in the `dotnet` CLI.

zivkan (commented):
sorry, my brain appears to be in pessimistic mode today. Generally, if we have a search command, I don't see the value in allowing customers to list/enumerate every package on a server.

proposed/2023/Add-dotnet-search.md
baronfel: This GNU behavior of combining single-letter flags is called 'bundling' and System.CommandLine (along with the dotnet CLI) do support this.

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: Thank you, the extension helped.

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: I think you make a good point. However, according to https://github.com/NuGet/Home/issues/6060 it looks like, the users want the functionality in dotnet.

zivkan: I had a look through that issue. I don't see anything saying that list is specifically needed. In fact, I feel like that issue reinforces the idea that the request was made before any `search` command was available. I still think that `search` without any kind of list should be sufficient.

Nigusu-Allehu: Yeah, that sounds good to me. @nkolev92 , what do you think?

nkolev92: I agree with @zivkan.

list is implemented based on the V2 protocol, and based on what was available back then. 
Listing all packages is a super heavy operation with the feeds as they are and we don't need it at this point.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
+# ***Add `nuget search` Functionality to DotNet***
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- [Nigusu](https://github.com/Nigusu-Allehu) <!-- GitHub username link -->
+- [GitHub Issue](https://github.com/NuGet/Home/issues/6060) <!-- GitHub Issue link -->
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add `nuget search` functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.
```
nkolev92: The specification seeks to add a search functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`. 

`nuget search` might be imply some ambiguity with nuget.exe search. We really want a search functionality and the design on it is as you're proposing it in this spec.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add `nuget search` functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.  
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+As outlined in the customer request on this GitHub [issue](https://github.com/NuGet/Home/issues/6060), there is a call for the incorporation of `nuget list` functionality into dotnet. The customers are requesting for this feature, but the `search` command can replicate the same functionality, with the only difference being the quantity of package results it outputs. The search command mitigates unnecessary server use by limiting the number of results through the `--take` option. By integrating this search functionality, we aim to accommodate our customers' needs, especially for scripting purposes.
```
nkolev92: I'd just say `add a search` functionality. Think of it as list++ instead of moving the nuget list functionality in there. 

Basically people want to be able to search for packages through dotnet.exe the same way they can in Visual Studio or nuget.exe.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+Imagine you have your dotnet CLI open. You would like to look up a NuGet package named `MyPackage` from a source `<MySource>`. No worries, you can just use the following command : `dotnet package search MyPackage --source <MySource>`. It will provide you with the list of the packages in the source `<MySource>` that match with the search criteria.
```
nkolev92: The functional explanation should probably contain the outputs, the options, arguments etc. 
It's about how the customer would use it. 

So I tihnk most of the technical explanation can just be moved into the functional one.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+The `package search [search terms] [options]` command will have the following options
+
+| Option | Function |
+|---------|:----------|
+| `--source` | A list of packages to search |
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--non-interactive` | Do not prompt for user input or confirmations.|
```
nkolev92: if you check `dotnet restore --help` and compare it to `nuget.exe restore /?`, you'll see they have different interactivity behaviors. 

`dotnet` is by default noninteractive, while nuget.exe is by default interactive. So in here, you need an `--interactive` switch and something that matches the same description as the one from restore.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+Imagine you have your dotnet CLI open. You would like to look up a NuGet package named `MyPackage` from a source `<MySource>`. No worries, you can just use the following command : `dotnet package search MyPackage --source <MySource>`. It will provide you with the list of the packages in the source `<MySource>` that match with the search criteria.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+The `package search [search terms] [options]` command will have the following options
+
+| Option | Function |
+|---------|:----------|
+| `--source` | A list of packages to search |
```
nkolev92:
```suggestion
| `--source` | A source to search |
```

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--non-interactive` | Do not prompt for user input or confirmations.|
+| `--take` | The number of results to return. The default value is 20.|
+| `--help` | Show command help and usage information |
+| `--verbosity` | Display the amount of details in the output: normal, quiet, detailed. |
```
nkolev92: As part of the functional spec, we should define how an output under the different verbosities looks like.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+|||
+
+#### **Algorithm**
+
+1. Read and parse arguments from user
```
nkolev92: this is probably too much detail for a spec.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+2. Use the nuget.exe Search API to look for the specific term in the provided source. The search filter will be composed of the `--source`, `[Search term]`, `--prerelease` values.
```
nkolev92: Can you pass more than 1 source? What happens if you don't pass a source?

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+3. Based on `--verbosity` value, provide with a list of packages with various verbosity
```
nkolev92: This is great info!

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+                >[Package Name] | [Latest Package Version] | [Amount of Downloads]
+                <Description of the Package(short form)>
+  - **Detailed** :
+
+                >[Package Name] | [Latest Package Version] | [Amount of Downloads]
+                Deprecated : True/False | Vulnerable : True/False
+                <Description of the Package(Readme file)>
+                License URL : [URL]
+
+#### **New Option `-exact-match`**
```
nkolev92: This is a new command, so technically all options are new. I'd just list out the arguments/options and then have a section for each one to provide how they work.

proposed/2023/Add-dotnet-search.md
nkolev92: I'm guessing this is meant to be resolved.

nkolev92 (commented):
Sorry for the slow turnaround. 
Nearly all of my feedback is on the spec body, more so than the final command. 

Let's work offline together.

proposed/2023/Add-dotnet-search.md
baronfel: We in the SDK definitely want to move to interactive by default when we can detect it's safe to do so: https://github.com/dotnet/sdk/issues/35165

Hoping for .NET 9!

nkolev92: @baronfel Would the recommendation be to stick with an `--interactive` flag in the meantime or should we remove it altogether?

proposed/2023/Add-dotnet-search.md
zivkan: Some tools, like [`git log`, allow you to pass a format string](https://www.git-scm.com/docs/git-log#_commit_formatting), so you can format it in any way you want. In theory, you can probably get it to generate (almost) JSON.

I don't want to over complicate things, and I know this will feel hard because it's a pattern we don't already have anywhere else in NuGet, but should we consider removing `--verbosity` and do some kind of `--format "\i | \v | \o\n\d"` ? 

We'd need to come up with a format code for every metadata item, but then customers could possibly do something like ``$json = '[' + `dotnet package search sql --format "{\"id\": \"\\i\", \"version\": \"\\v\", \"description\": \"\\d\" }"` + ']'`` and then process it as json. We'd also need to think a bit harder about what other escaping we'd need, this was just a 5 second example. For example, we'd probably need two copies of all properties that have free text, just in case the metadata contains a double quotes character that would otherwise break json. For example, `\d` could output the package description as-is, and `\D` outputs the package description with json encoding.

Since we're a .NET project, we could consider more `string.Format` style syntax, like ``--format "{id} | {version} | {downloads}\n{description}"`` and ``--format "\{ \"id\": \"{id}\", \"version\": \"{version}\", \"description\": \"{description:j}\" \}"``, which is clearer to people not familar with C's `printf` style formatting, but much, much more verbose.

My concern is that if we start off with `--verbosity`, but then later get requests to add something like `--format`, then there might be complicated questions or implementation details around what happens when `--format` and `--verbosity` are provided at the same time. Although a simple answer would be to error out when both are provided at the same time, commands that can't be used in a wrong way are generally better than commands that can. Hence, why I think it's worthwhile to think about whether this is something we would want to do in a V1.  It would also be a great feature to encourage customers to get off `nuget.exe search` and onto `dotnet package search`

I'm really not sure how complex this might be to implement. Might be worth doing a 1 day spike to see how close we can get a string format parser working, to figure out if this suggestion is feasible?

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,107 @@
...
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+Make the search command bigger and vast
```
zivkan: IMO this is too vague. If you don't have specific examples of what feature could be added, I'd leave this section empty

proposed/2023/Add-dotnet-search.md
nkolev92: Maybe we propose no `--verbosity` in v1?

TBH, I don't haste just having 1 verbosity, especially since adding more things will be easier than taking away things

Nigusu-Allehu: I like this idea. I think `--format` would be more useful and practical to users. I will have just one verbosity, `--Normal`, when the `--format` is not specified. I am thinking of, for this PR, removing `--verbosity`. And then adding `--format` in its own spec PR and we can discuss its syntax in that PR.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add search functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.  
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+As outlined in the customer request on this GitHub [issue](https://github.com/NuGet/Home/issues/6060), there is a call for the incorporation of `nuget.exe list` functionality into dotnet. The customers are requesting for this feature, but a `search` command can replicate the same functionality, with the only difference being the quantity of package results it outputs. A search command mitigates unnecessary server use by limiting the number of results through the `--take` option. By integrating this search functionality, we aim to accommodate our customers' needs, especially for scripting purposes.
```
zivkan: Putting each sentence on a separate line makes it easier to comment on a specific line. Also, this is being completely pedantic, I feel like the focus should not be on the `nuget list` command, but brining the dotnet cli closer to feature parity of nuget.exe. For better or worse, NuGet/Home#6060 is being used to track both the list and search commands (see https://github.com/NuGet/Home/issues/4855#issuecomment-1205936210)

```suggestion
The `dotnet` CLI is missing a way for customers to search for packages from the command line, as is possible with `nuget.exe search`.
This brings NuGet experiences in the `dotnet` CLI one step closer towards feature parity.

While [NuGet/Home#6060](https://github.com/NuGet/Home/issues/6060) (the linked issue) also discusses the `nuget.exe list` command, the issue was created before `nuget.exe search` was added.
The [NuGet Server API Search Resource specification requires all packages to be returned when no `q` parameter is provided](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#request-parameters).
Therefore, this proposal does not add a `list` equivalent to the `dotnet` CLI, as it is effectively redundant.
```

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,91 @@
...
+
+The `package search [search terms] [options]` command will have the following options
+
+| Option | Function |
+|---------|:----------|
+| `--source` | A source to search |
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--interactive` | Allows the command to stop and wait for user input or action (for example to complete authentication).|
+| `--take` | The number of results to return. The default value is 20.|
```
zivkan, line 33: There also needs to be a `--skip` option, so customers can search for a second page, when the first doesn't contain what they're looking for.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,91 @@
...
+        >NuGet.for.MSBuild | 2.1.0 ...
+
+- Using ``dotnet package search NuGet.CommandLine --exact-match`` on the other side will have the following output
+
+         >NuGet.CommandLine | 6.7.0 | Downloads: N/A
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+This command will prepare a search parameter using the inputs of  `Search Term`, `--take`, and `--prerelease`. This parameter is then used prepare an API query to the specified source. In the dotnet/sdk repo, there is already a command, `dotnet tool search`, which does a similar thing. However, it specifies the package type to be only `dotnettool`. Instead of adding a new set of classes and methods, I will modify the methods and classes used by `dotnet tool search` to do a general search when needed. This will allow both `dotnet package search` and `dotnet tool search` to use the same API request class. The result of the query is then parsed to provide output to users.
```
zivkan: one sentence per line makes it easier to provide comments and suggestions on individual lines. It renders the same as long as there are no empty lines

```suggestion
This command will prepare a search parameter using the inputs of  `Search Term`, `--take`, and `--prerelease`.
This parameter is then used prepare an API query to the specified source.
In the dotnet/sdk repo, there is already a command, `dotnet tool search`, which does a similar thing.
However, it specifies the package type to be only `dotnettool`.
Instead of adding a new set of classes and methods, I will modify the methods and classes used by `dotnet tool search` to do a general search when needed.
This will allow both `dotnet package search` and `dotnet tool search` to use the same API request class.
The result of the query is then parsed to provide output to users.
```

zivkan: > Instead of adding a new set of classes and methods, I will modify the methods and classes used by `dotnet tool search` to do a general search when needed. This will allow both `dotnet package search` and `dotnet tool search` to use the same API request class. The result of the query is then parsed to provide output to users.

I don't think this is the right approach, if you mean what I think this is saying. The `dotnet tool search` code was added such a long time ago, I don't remember if they asked us, the NuGet Client team, for an API change to support the new query syntax that nuget.org added. Maybe they didn't ask, because what we add into NuGet.Client is generalised to support any NuGet server implementation, not just nuget.org, and this package type filtering is specific to nuget.org.  Maybe the SDK/CLI team did ask us, but we didn't commit to doing it quickly enough for them, so they decided to re-implement parts of NuGet.Protocol themselves to unblock their scenario.

Anyway, I feel like we should be encouraging re-use of our NuGet.Protocol package, not custom implementations of the NuGet server API.

Our docs on the NuGet Client SDK have an almost complete example of how `dotnet package search` should be implemented: https://learn.microsoft.com/en-us/nuget/reference/nuget-client-sdk#search-packages. It's probably more or less the same code that `nuget.exe search` has. No need to extend dotnet/sdk's existing code to support what our code already supports.

It would be great for us, the NuGet.Client team, to add relevant APIs into NuGet.Protocol's search to do package type filtering (it will be needed if I get my way with https://github.com/NuGet/Home/issues/12567), at which time the dotnet/sdk repo can remove their custom NuGet protocol API code. However, adding package type filtering into NuGet.Protocol is out of scope of this spec on adding `dotnet package search`.

Nigusu-Allehu: Thanks Andy, I have changed my implementation to use Nuget.Protocol and left dotnet tool search as it is.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,98 @@
...
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+I believe `dotnet tool search` is a subset to `dotnet package search`, as a result we could deprecate it. If it is necessary we could add an option in `dotnet package search` to allow users to specify that they want to only search dotnet tool.
```
zivkan: `dotnet tool search` is not a subset of `dotnet package search`. `tool search` specifically searches only for packages of type `DotnetTool`, whereas `dotnet package search` and `nuget.exe search` does not send a `packageType` query filter. Even if a `--package-type` filter was added to `dotnet package search`, `dotnet tool search` would still be valuable to avoid needing to memorize the package type identifier, though that could be partially mitigated by putting a list of well-known package types in the command help message. `dotnet tool search` also avoid customers needing to know that tools are shippable as NuGet packages and customers running `dotnet tool -h` can easily discover there's a search command.  I think there's value in `dotnet tool search` existing, rather than being deprecated if there's an alternative.

zivkan (approved):

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,98 @@
...
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add search functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.  
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+The `dotnet` CLI is missing a way for customers to search for packages from the command line, as is possible with `nuget.exe search`.
+This brings NuGet experiences in the `dotnet` CLI one step closer towards feature parity.
+
+While [NuGet/Home#6060](https://github.com/NuGet/Home/issues/6060) (the linked issue) also discusses the `nuget.exe list` command, the issue was created before `nuget.exe search` was added.
```
nkolev92: nit: this seems like more of an explanation about the scope rather than motivation specific.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,98 @@
...
+The `package search [search terms] [options]` command will have the following options
+
+| Option | Function |
+|---------|:----------|
+| `--source` | A source to search |
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--interactive` | Allows the command to stop and wait for user input or action (for example to complete authentication).|
+| `--take` | The number of results to return. The default value is 20.|
+| `--skip` | The number of results to skip, for pagination |
```
nkolev92: Default 0.

proposed/2023/Add-dotnet-search.md
nkolev92: I agree. we shouldn't deprecate, but there's maybe a world where they're implemented by the same code.

nkolev92 (approved):

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,99 @@
...
+
+| Package ID                                  | Latest Version | Authors | Downloads       |
+|---------------------------------------------|----------------|---------|-----------------|
+| Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
+| Newtonsoft.Json.Bson                        | 1.0.2          |         | 554,641,545     |
+| Newtonsoft.Json.Schema                      | 3.0.15         |         | 39,648,430      |
+| Microsoft.AspNetCore.Mvc.NewtonsoftJson     | 7.0.12         |         | 317,067,823     |
+...
+
+- Using ``dotnet package search NuGet.CommandLine --exact-match`` on the other side will have the following output
```
kartheekp-ms:
``` suggestion
- Using ``dotnet package search Newtonsoft.Json --exact-match`` on the other side will have the following output
```

The ouput table below has `Newtonsoft.Json` as package id.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,99 @@
...
+The `package search [search terms] [options]` command will have the following options
+
+| Option | Function |
+|---------|:----------|
+| `--source` | A source to search |
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--interactive` | Allows the command to stop and wait for user input or action (for example to complete authentication).|
+| `--take` | The number of results to return. The default value is 20.|
+| `--skip` | The number of results to skip, for pagination. The default value is 0. |
```
kartheekp-ms, line 34: I am reviewing the spec and have a question regarding the pagination aspect, especially when both --skip and --take options are used together.

From my understanding, introducing these options implies that users might view the search results in a paginated manner. If a user passes both the `--skip` and `--take` options, would the command be designed to wait for user input after displaying the first page of results? This could offer users an opportunity to view subsequent pages without having to re-run the command, enhancing user experience.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,99 @@
...
+The `dotnet` CLI is missing a way for customers to search for packages from the command line, as is possible with `nuget.exe search`.
+This brings NuGet experiences in the `dotnet` CLI one step closer towards feature parity.
+
+While [NuGet/Home#6060](https://github.com/NuGet/Home/issues/6060) (the linked issue) also discusses the `nuget.exe list` command, the issue was created before `nuget.exe search` was added.
+The [NuGet Server API Search Resource specification requires all packages to be returned when no `q` parameter is provided](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#request-parameters).
+Therefore, this proposal does not add a `list` equivalent to the `dotnet` CLI, as it is effectively redundant.
+
+## Explanation
+
+### Functional explanation
```
kartheekp-ms, line 19: Does the search command consider package source mapping settings as well? I'm curious to know how the new search functionality will interact with existing configurations and settings, ensuring that it provides accurate and relevant results based on the user's preferences and restrictions.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The specification seeks to add search functionality to dotnet. The functionality will be added to dotnet as a search command: `dotnet package search`.  
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Customers have expressed a desire for the functionality of the nuget.exe list command to be integrated into the dotnet CLI. Here is a discussion on the request: [NuGet/Home#6060](https://github.com/NuGet/Home/issues/6060). By implementing the search feature, we can effectively address this need, as it will allow users to retrieve and display a list of packages, thereby achieving the primary objective of the list command in a more streamlined manner within the dotnet environment.The [NuGet Server API Search Resource specification requires all packages to be returned when no `q` parameter is provided](https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#request-parameters). Therefore, this proposal does not add a `list` equivalent to the `dotnet` CLI, as it is effectively redundant.
```
kartheekp-ms, line 15: I think we should also mention somewhere that `.NET SDK` has `list package` command already. https://learn.microsoft.com/dotnet/core/tools/dotnet-list-package

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Explanation
+
+### Functional explanation
```
kartheekp-ms, line 19: If the latest version of the package has been deprecated or is vulnerable, does the command output that information? It would be helpful so that customers can make informed decisions in case they would like to install that package.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,94 @@
...
+| Option | Function |
+|---------|:----------|
+| `--source` | A source to search |
+| `--exact-match` | Return exact matches only as a search result |
+| `--prerelease` | Allow prerelease packages to be shown. |
+| `--interactive` | Allows the command to stop and wait for user input or action (for example to complete authentication).|
+| `--take` | The number of results to return. The default value is 20.|
+| `--skip` | The number of results to skip, for pagination. The default value is 0. |
+| `--help` | Show command help and usage information |
+|||
```
kartheekp-ms: nit: Maybe we don't need the additional row.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### **Option `--source`**
+
+This option will specify a list of sources to search from. If a source is not specified using this option, the sources in the `nuget.config` file will be used.
```
kartheekp-ms, line 39: Perhaps it is worth clarifying that if no `source` option is passed to the command, the effective NuGet settings will be applied to obtain the results, as detailed here: https://learn.microsoft.com/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied.

proposed/2023/Add-dotnet-search.md
```diff
@@ -0,0 +1,94 @@
...
+
+#### **Option `-exact-match`**
+
+- This option will allow for users to be able to search and have only exact matches as an output.
+- For example if a user uses `dotnet package search Newtonsoft.Json`
+
+| Package ID                                  | Latest Version | Authors | Downloads       |
```
kartheekp-ms, line 47: I was wondering if we could display the source name in the output when multiple sources are configured, and no `source` option is passed to the command.

Nigusu-Allehu: Yes the source will be displayed and then a table of the search result follows for each source. I will update the spec to indicate this

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: The `dotnet list package`  "Lists the package references for a project or solution". As a result I was not sure how to mention it in this spec

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: The `--skip` and `--take` options are passed to the search api as parameters. By using `--skip x`, users will be able to skip the first x packages from being returned in the search result. By using `--take x`, users would be able to limit the maximum number of packages outputted at once. If they are passed together as `--take x --skip y`, x number of packages following the first y packages are printed out. https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: I am not sure about that. Currently, what I do is go through each of the package sources specified in the config file and do a search API request to each of the sources.

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: That is a good point raised Kartheek. Currently, no. The reason is that I plan to implement a `--format` option that would allow users to print out the search results in the format they want. That would allow users to be able to use the output in a script or pipeline easily.

proposed/2023/Add-dotnet-search.md
donnie-msft: Kartheek raises a great question. I believe most CLI search functionality that we have today doesn't honor source mapping. I'm not clear what customers expect. What's the context around using this command? With `dotnet list package`, the context is around a solution, so we *do* want to honor source mapping for that command (https://github.com/NuGet/Home/issues/12766).

Reconciling some of the following thoughts may help reach a decision:
1. It's optional to provide a `--source`. What do customers expect when it's not specified? What if they specify a source and the package isn't found (do we remind them with a warning that source mapping may affect results)?
2. There's no `--ConfigFile` option. If there were, it'd be more support for honoring source mapping.
3. Running a search with no options, do customers expect the context to matter? If my current directory is a solution, should the search's context be the solution by default? Or is search more about user/machine-wide configuration?
4. When the user-level config is applying to the CLI command, ideally there wouldn't be any source mapping in that `nuget.config`.

nkolev92: Re: Source mapping 

I'd say for the purposes of this implementation, search is just that, search. I don't think we want source mapping to affect search results, but that's a separate topic that we shouldn't worry about in this context.

Source Mapping is about installing packages. People should be able to discover packages on different feeds in case they want to change mappings or something like that.

PR #12921, opened 2023-10-05
Feature spec for suppressing specific vulnerabilities

zivkan:
[view rendered spec](https://github.com/NuGet/Home/blob/9d98231d88d70b18e61b37079f725e410ee0e6a0/accepted/2023/NuGetAudit-supress-advisory.md)

Note: I don't expect this to get implemented particularly soon, and I'm not going to have a lot of capacity to make frequent updates to this spec, in case it gets a lot of feedback. But I'm putting it out there now, so it has time to get sufficient feedback before it gets accepted.

### DIFF
--- a/accepted/2023/NuGetAudit-supress-advisory.md
+++ b/accepted/2023/NuGetAudit-supress-advisory.md
@@ -61,7 +61,7 @@ Unfortunately, restore has many entry points, each of which read MSBuild propert
 Therefore, just like adding any other property or item, it will need modifications in all of them.
 The list is: MSBuild task (so msbuild and dotnet restore), static graph restore, nuget.exe, VS CPS projects (nominations), VS non-CPS projects (LegacyPackageReferenceProject).
 
-##### CollectNuGetAuditSuppressions target
+##### `CollectNuGetAuditSuppressions` target
 
 We already have `CollectPackageReferences`, `CollectPackageDownloads`, `CollectFrameworkReferences`, and `CollectCentralPackageVersions`.
 The VS restore path needs these targets in order to collect any items we want to pass as inputs to the restore process. This has a side-effect that it provides customers, and other MSBuild SDK authors, a way to customize their build and dynamically add or modify items as needed.
@@ -72,7 +72,11 @@ Additionally, we'll need to check if Visual Studio's project systems need any ch
 If NuGet's restore always ran as an MSBuild target, the customers and MSBuild SDK authors could probably just have used `BeforeTarget="Restore"`.
 However, in Visual Studio that's not the case, so it won't work.
 
-This new target needs to be added to our list of 'targets to run' in multiple places. In order to avoid adding these new targets in multiple places every time, we will also take this opportunity to unify all the 'Collect' targets (5, after we add this one) into a single `CollectRestoreInputs` target, which we can re-use for everything going forward.
+##### `_CollectRestoreInputs` target
+
+The new `CollectNuGetAuditSuppressions` target needs to be added to our list of 'targets to run' in multiple places. In order to avoid adding these new targets in multiple places every time, we will also take this opportunity to unify all the 'Collect' targets (5, after we add this one) into a single `_CollectRestoreInputs` target, which we can re-use for everything going forward.
+
+This target is strictly an internal implementation detail, meant to reduce the amount of work needed to add new targets in the future. This is not intended as a target for customers to hook onto.
 
 #### `dotnet list package --vulnerable`
 
--- a/accepted/2023/NuGetAudit-supress-advisory.md
+++ b/accepted/2023/NuGetAudit-supress-advisory.md
@@ -58,8 +58,21 @@ Here is a complete project example:
 #### NuGetAudit (restore)
 
 Unfortunately, restore has many entry points, each of which read MSBuild properties and items in a different way.
-Therefore, just like adding any other property, it will need modifications in all of them.
+Therefore, just like adding any other property or item, it will need modifications in all of them.
 The list is: MSBuild task (so msbuild and dotnet restore), static graph restore, nuget.exe, VS CPS projects (nominations), VS non-CPS projects (LegacyPackageReferenceProject).
+
+##### CollectNuGetAuditSuppressions target
+
+We already have `CollectPackageReferences`, `CollectPackageDownloads`, `CollectFrameworkReferences`, and `CollectCentralPackageVersions`.
+I believe the original design of these targets was specifically for `PackageDownload`, to allow the .NET SDK to determine instruct NuGet to download runtime packs and BCL reference assemblies only when they're not already available locally.
+The VS restore path needs these targets in order to collect any items we want to pass as inputs to the restore process. This has a side-effect that it provides customers, and other MSBuild SDK authors, a way to customize their build and dynamically add or modify items as needed.
+
+We will add a `CollectNuGetAuditSuppressions` target with the new `NuGetAuditSuppress` item. We will need to not only modify our own *NuGet.targets* to add it to the appropriate `DependsOnTargets`, but we also need to investigate MSBuild's static restore or .NET SDK to see if it needs to be added to the list of targets it "hardcodes" it runs.
+Additionally, we'll need to check if Visual Studio's project systems need any changes, specifically for design time builds.
+
+If NuGet's restore always ran as an MSBuild target, the customers and MSBuild SDK authors could probably just have used `BeforeTarget="Restore"`.
+However, in Visual Studio that's not the case, so it won't work.
+Could we consider a single `CollectRestoreInputs` target, which we can re-use for everything going forward, rather than adding a new target for every item type we add?
 
 #### `dotnet list package --vulnerable`
 
@@ -72,14 +85,16 @@ On the other hand, we don't know of any package source that provide any package 
 Finally, `dotnet list package` already reads some properties out of the project, but doesn't currently retrieve any MSBuild items.
 Therefore, some new abstractions or helper methods will probably be needed, and there might not be existing examples in our code to copy-paste, as restore uses different APIs to get the same data.
 
+
 ## Drawbacks
 
 <!-- Why should we not do this? -->
-- Priority
+
+### Priority
 
 The linked issue has few up-votes, so we don't yet have evidence that it's a high impact feature.
 
-- Alternatives
+### Alternatives
 
 Customers can already achieve a similar outcome, but it will require a lot more effort on their end.
 Firstly, they wouldn't use NuGetAudit, at least on CI.
@@ -101,11 +116,13 @@ Alternatively, if this feature is implemented, then it's much easier for custome
 Some NuGet features are only available via *nuget.config*, and not MSBuild items or properties.
 However, we had [one customer provide feedback that they'd like the suppression to be per-project](https://github.com/NuGet/Home/issues/11926#issuecomment-1683384471), rather than per repo.
 This allows a project already using a package with a known vulnerability to prevent CI pipeline errors, while preventing other projects in the solution from starting to use it.
-This seems like an important use-case that should be enabled.
+This seems like an important use-case that should be enabled, and MSBuild items allow these settings to be applied at the solution-level (*directory.packages.props*) or project-level.
 
 ### MSBuild item vs property
 
-The current proposal is to use an MSBuild item. NuGet already uses MSBuild properties for various settings, such as additional package sources, and we can learn from experiences using it.
+The current proposal is to use an MSBuild item.
+
+NuGet currently uses MSBuild properties for various settings, such as additional package sources, and we can learn from experiences using it.
 
 Firstly, adding a per-project item into an MSBuild property has worse syntax (at least in my opinion): `<NuGetAuditSuppress>$(NuGetAuditSuppress);https://sample.test/CVE</NuGetAuditSuppress>`.
 
@@ -120,10 +137,19 @@ Secondly, when there are a long list of values, it similarly ends up with a poor
 
 While creating a new MSBuild item for each advisory URL uses more characters I feel it "fits" better with XML & MSBuild generally.
 
-Finally, MSBuild properties don't allow additional metadata.
-By using an item, we can add metadata, such as `Packages`: `<NuGetAuditSuppress Include="https://example.com/CVE1" Packages="ExamplePackage1;ExamplePackage2" />`.
+Further, MSBuild properties don't allow additional metadata.
+By using an item, we can add metadata, such as `Packages`:
+  
+  `<NuGetAuditSuppress Include="https://example.com/CVE1" Packages="ExamplePackage1;ExamplePackage2" />`,
+
+  or `Justification`:
+  
+  `<NuGetAuditSuppress Include="https://example.com/CVE1" Justification="mitigated risks" />`.
+
 For package sources this would have been helpful to enable credentials to authenticated package sources.
-For `NuGetAuditSuppress`, it could enable new future features more easily.
+For `NuGetAuditSuppress`, it could enable new features more easily.
+
+On the other hand, we do not read any items for packages.config projects, so relying on a completely MSBuild item based solution would add some technical complexity. Properties are also better for performance and memory concerns, and will involve less overhead than items. Further, if we decided to reverse our item vs property decision later, moving from properties to items is easier than the other way around. 
 
 ### Metadata on PackageReference items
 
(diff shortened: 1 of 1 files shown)


proposed/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,231 @@
...
+
+Additionally, by using MSBuild items and properties, it forces MSBuild to do these allocations on every project load, even when restore (or `dotnet list package`) is not run.
+Putting it in *nuget.config* would not only make it much easier to avoid duplicate strings, it would also only be read when NuGet is actually used vs every MSBuild project evaluation.
+
+However, basic tests on MSBuild evaluation performance doesn't show measurable regressions when adding 100 additional items into a project, and I don't believe it's likely that customers will suppress 100 different advisories.
+Therefore, even though using MSBuild items is theoretically the highest allocation design choice, given how much other work MSBuild does, it doesn't appear to result in a measurable performance impact.
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
```
skofman1: When thinking about prior art we can draw inspiration from, consider GitHub dependabot suppressions, CG suppressions, NPM. It will be interesting to see the granularity that each solution provides (per package, per project, per advisory, etc.). Assuming that NuGet users already have experience using those other vuln suppression mechanism, they likely expect a similar customer experience from NuGet.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <OutputType>exe</OutputType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <NuGetAuditSuppress Include="https://github.com/advisories/GHSA-5crp-9r3c-p9vr" />
```
kartheekp-ms, line 50: In general, when developers suppress code analysis warnings, there is an option to provide justification for future reference in the `GlobalSuppressions.cs` file. For example, in Component Governance, there is an option to provide a justification for dismissing a security vulnerability alert. 

Would it make sense to also provide users with an option to add justification for suppressing these warnings?

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+If we use this approach for direct packages, there's no obvious alternative for transitive packages.
+Using a new MSBuild item for transitive packages, but metadata on a PackageReference item for direct packages is confusing because developers need more context in order to choose the correct syntax.
+It also means duplication when a package is direct in one project, but transitive in others.
+Having the suppress metadata flowed down to transitive packages, as Include/Exclude/Private Assets does is another option, because it's behavior that already exists.
+However, this makes it difficult to suppress centrally, like in a *Directory.Packages.props* file, as a transitive package will have different "entry points" in different projects in a solution.
+
+### CollectNuGetAuditSuppress target
+
+We already have `CollectPackageReferences`, `CollectPackageDownloads`, `CollectFrameworkReferences`, and `CollectCentralPackageVersions`.
+I believe the original design of these targets was specifically for `PackageDownload`, to allow the .NET SDK to determine instruct NuGet to download runtime packs and BCL reference assemblies only when they're not already available locally.
```
nkolev92: Collect targets came before that. It allowed a hook for tweaking items.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+When a single advisory is relevant to multiple packages, as is common for .NET, ASP.NET, and NuGet advisories, it will require duplication in multiple packages.
+
+Additionally, this will only work for direct packages, not transitive packages.
+If we use this approach for direct packages, there's no obvious alternative for transitive packages.
+Using a new MSBuild item for transitive packages, but metadata on a PackageReference item for direct packages is confusing because developers need more context in order to choose the correct syntax.
+It also means duplication when a package is direct in one project, but transitive in others.
+Having the suppress metadata flowed down to transitive packages, as Include/Exclude/Private Assets does is another option, because it's behavior that already exists.
+However, this makes it difficult to suppress centrally, like in a *Directory.Packages.props* file, as a transitive package will have different "entry points" in different projects in a solution.
+
+### CollectNuGetAuditSuppress target
```
nkolev92: Do we absolutely need a collect target? 
Is this a technical requirement for CPS?

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+
+`cargo-audit`'s config can contain a list of advisory IDs to ignore: https://docs.rs/cargo-audit/latest/cargo_audit/config/struct.AdvisoryConfig.html
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+### Per-package suppressions
```
nkolev92, line 219: I feel like this should be a very, very future thing, unless we get a very loud feedback from customers this is absolutely necessary and the feature is not usable without it.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+However, this makes it difficult to suppress centrally, like in a *Directory.Packages.props* file, as a transitive package will have different "entry points" in different projects in a solution.
+
+### CollectNuGetAuditSuppress target
+
+We already have `CollectPackageReferences`, `CollectPackageDownloads`, `CollectFrameworkReferences`, and `CollectCentralPackageVersions`.
+I believe the original design of these targets was specifically for `PackageDownload`, to allow the .NET SDK to determine instruct NuGet to download runtime packs and BCL reference assemblies only when they're not already available locally.
+I'm not sure why the other `Collect*` targets were added.
+This has a side-effect that it provides customers, and other MSBuild SDK authors, a way to customize their build and dynamically add or modify items as needed.
+
+We will add a `CollectNuGetAuditSuppress` target with the new `NuGetAuditSuppress` item. We will need to not only modify our own *NuGet.targets* to add it to the appropriate `DependsOnTargets`, but we also need to investigate MSBuild's static restore or .NET SDK to see if it needs to be added to the list of targets it "hardcodes" it runs.
```
nkolev92: Do we absolutely need this target? 
Can we avoid adding a target for this? Something we should check with the CPS/project-system teams, whether there's a technical need for this target (guessing no). 

At this point, I don't see why we need an extensibility over this.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+If we add per-package support in the first version of `NuGetAuditSuppress`, then it reduces risk for this future change.
+Additionally, given the need to modify all of NuGet's restore entry points, it reduces effort.
+
+On the other hand, this feature has not yet been requested by any customers, so might be a form of scope creep.
+The more simple `NuGetAuditSuppress` without per-package support will be quicker to implement, and significantly easier to write automated tests for.
+
+### NuGetAuditSuppress name
+
+Is there a better name than "NuGetAuditSuppress"?
+Taking cargo's lead, perhaps `NugetAuditIgnore`?
```
nkolev92: Consider NoWarn like syntax.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Tooling support
+
+Should we add any `dotnet` CLI commands to add/remove `NuGetAuditSuppress` items?
```
nkolev92: I think no. 
This should be easy to manually edit.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+
+Should we work with dotnet/project-system to display suppressions in Solution Explorer? Should project properties provide a GUI to edit suppressions?
```
nkolev92: Suppressions are not within the context of a package, so it'd be unusual.

nkolev92: Maybe project properties?

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+<NuGetAuditSuppress>
+  https://example.com/CVE1;
+  https://example.com/CVE2;
+<NuGetAuditSuppress>
+```
+
+While creating a new MSBuild item for each advisory URL uses more characters I feel it "fits" better with XML & MSBuild generally.
+
+Finally, MSBuild properties don't allow additional metadata.
+By using an item, we can add metadata, such as `Packages`: `<NuGetAuditSuppress Include="https://example.com/CVE1" Packages="ExamplePackage1;ExamplePackage2" />`.
```
nkolev92: I'm not sure this is something that's desirable. It introduces a syntax not available for any other item. 

If we want suppression on a per package level, maybe a package item is more appropriate? I understand the transitive dependency limitations, but this is arguably something we're not going to add immediately anyways.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+<!-- What is the impact of not doing this? -->
+
+### MSBuild vs nuget.config
+
+Some NuGet features are only available via *nuget.config*, and not MSBuild items or properties.
+However, we had [one customer provide feedback that they'd like the suppression to be per-project](https://github.com/NuGet/Home/issues/11926#issuecomment-1683384471), rather than per repo.
+This allows a project already using a package with a known vulnerability to prevent CI pipeline errors, while preventing other projects in the solution from starting to use it.
+This seems like an important use-case that should be enabled.
+
+### MSBuild item vs property
```
nkolev92, line 125: For perf/agility reasons my preference is a property, but I think we need customer validation for that. 

Going from property to item is doable, but going from item to property would cause more issues for customers.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,230 @@
...
+
+The current proposal is to use an MSBuild item. NuGet already uses MSBuild properties for various settings, such as additional package sources, and we can learn from experiences using it.
```
nkolev92: One thing I forgot to mention is that for PC we never read items, so relying on a completely msbuild/item centric approach for packages.config adds technical complexity

accepted/2023/NuGetAudit-supress-advisory.md
advay26: With items, users will be able to add their own justification attribute if they wish. We're not looking to provide any tooling around a "Justification" attribute, but users can still add it themselves.

accepted/2023/NuGetAudit-supress-advisory.md
advay26: I spoke to Tom and a few others on the project systems side, and they confirmed that we do need a target if we want to collect any items and pass them onto the VS restore path.

accepted/2023/NuGetAudit-supress-advisory.md
advay26: > I spoke to Tom Meschter, and he confirmed that we do need a target if we want to collect any items and pass them onto the VS restore path.

https://github.com/NuGet/Home/pull/12921#discussion_r1508432638

accepted/2023/NuGetAudit-supress-advisory.md
advay26: We're not going to be adding per-package suppressions right now, so I've moved this to the 'Future Possibilities' section.

accepted/2023/NuGetAudit-supress-advisory.md
advay26: We eventually decided on items over properties, primarily because they allow for metadata, and will make it easier to add features later on, but I did make a note of the perf concerns in the spec.

accepted/2023/NuGetAudit-supress-advisory.md
advay26: Makes sense. I've made note of the added technical complexity here.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,242 @@
...
+The VS restore path needs these targets in order to collect any items we want to pass as inputs to the restore process. This has a side-effect that it provides customers, and other MSBuild SDK authors, a way to customize their build and dynamically add or modify items as needed.
+
+We will add a `CollectNuGetAuditSuppressions` target with the new `NuGetAuditSuppress` item. We will need to not only modify our own *NuGet.targets* to add it to the appropriate `DependsOnTargets`, but we also need to investigate MSBuild's static restore or .NET SDK to see if it needs to be added to the list of targets it "hardcodes" it runs.
+Additionally, we'll need to check if Visual Studio's project systems need any changes, specifically for design time builds.
+
+If NuGet's restore always ran as an MSBuild target, the customers and MSBuild SDK authors could probably just have used `BeforeTarget="Restore"`.
+However, in Visual Studio that's not the case, so it won't work.
+Could we consider a single `CollectRestoreInputs` target, which we can re-use for everything going forward, rather than adding a new target for every item type we add?
```
zivkan: This question should be resolved before the spec is merged. From memory after I wrote the initial draft of this spec, I found that static graph resote does have a list of targets to run, which means that we can replace it with just one, and then change NuGet.targets to have the one target that in turn calls all the others. So, I think it's technically feasible.

Or we can take the easy option and just append the new target to the list of targets that need to run in both places.

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,242 @@
...
+
+Additionally, by using MSBuild items and properties, it forces MSBuild to do these allocations on every project load, even when restore (or `dotnet list package`) is not run.
+Putting it in *nuget.config* would not only make it much easier to avoid duplicate strings, it would also only be read when NuGet is actually used vs every MSBuild project evaluation.
+
+However, basic tests on MSBuild evaluation performance doesn't show measurable regressions when adding 100 additional items into a project, and I don't believe it's likely that customers will suppress 100 different advisories.
+Therefore, even though using MSBuild items is theoretically the highest allocation design choice, given how much other work MSBuild does, it doesn't appear to result in a measurable performance impact.
+
+### Tooling support
+
+We considered adding `dotnet` CLI commands to add/remove `NuGetAuditSuppress` items, but it should be fairly easy to edit these items manually, so we don't believe it's necessary at this time.
```
zivkan: This seems like it should be under Future Possibilities, not rejected designs.

zivkan (commented):
only two minor things, otherwise lgtm!

accepted/2023/NuGetAudit-supress-advisory.md
advay26: It would be helpful to have a central location where we can add new 'Collect' targets, rather than having to add it in multiple places every time, so we can make that change with the suppression item. I've updated the spec to say we'll add a new `CollectRestoreInputs` target that runs all the individual 'Collect' targets in one place.

zivkan (commented):
Since I created this PR, GitHub won't let me approve it, but it LGTM now.

jeffkl (approved):

accepted/2023/NuGetAudit-supress-advisory.md
```diff
@@ -0,0 +1,243 @@
...
+
+We will add a `CollectNuGetAuditSuppressions` target with the new `NuGetAuditSuppress` item. We will need to not only modify our own *NuGet.targets* to add it to the appropriate `DependsOnTargets`, but we also need to investigate MSBuild's static restore or .NET SDK to see if it needs to be added to the list of targets it "hardcodes" it runs.
+Additionally, we'll need to check if Visual Studio's project systems need any changes, specifically for design time builds.
+
+If NuGet's restore always ran as an MSBuild target, the customers and MSBuild SDK authors could probably just have used `BeforeTarget="Restore"`.
+However, in Visual Studio that's not the case, so it won't work.
+
+This new target needs to be added to our list of 'targets to run' in multiple places. In order to avoid adding these new targets in multiple places every time, we will also take this opportunity to unify all the 'Collect' targets (5, after we add this one) into a single `CollectRestoreInputs` target, which we can re-use for everything going forward.
```
nkolev92: I'd change the name to `_CollectRestoreInputs`. 
Then we're effectively communicating that this is not a target to hook onto. 

I'd also add this is explicitly implementation detail, not a new target customers can depend on.

nkolev92 (approved):

accepted/2023/NuGetAudit-supress-advisory.md
advay26: That makes sense. I've added that in now

anangaur:
It's easy to forget about suppressions done in a config file that's not in your face. Are we thinking about some messaging (info) to state that? Like as part of restore, something like: "Note: Some warnings due to packages having vulnerabilities have been suppressed. See <TBD>, for a list of those suppressions."

PR #12970, opened 2023-10-26
Add `--configFile`, `--format` and `--verbosity` options to `dotnet package search`

Nigusu-Allehu:
Update `dotnet package search` options
- Add `--configFile`, `--format` and `--verbosity` 
- Make `--exact-match` list all versions rather than just the latest version: This will enable users get more information about a package and its versions rather than just confirmation that an exact match exists.


proposed/2023/Add-dotnet-search.md
```diff
@@ -43,21 +44,78 @@ This option will specify a list of sources to search from. If a source is not sp
...
+  - Description
+  - Vulnerabilities
+  - Deprecation
+  - Project URL
+
+  The output will take the following format
+
+        [
+            {
+                "source": "source1",
```
donnie-msft: When reading this PR, it came to my mind that it implies "source name", not "source URL". That's fine. We do utilize source name with our CLI.

Customers shouldn't be getting themselves into this situation, but since it came to my mind.
This is the scenario that I didn't follow the best practice and <clear/> my package sources in my solution's config. My machine-wide config also has nuget.org with a different source name. 

For example, I configured nuget.org under the name "nuget" in solution config and "nugetzz" in machine config.
running `nuget.exe search newtonsoft.json`
I got 2 resultsets: one for nuget and one for nugetzz.

I suppose your command will also do this. Otherwise, you'd have to choose a source name.

I don't have more context to advise you on whether you should group them like
`Source: {{URI}} ({{list of source name(s)}})`
eg, `Source: "https://api.nuget.org/v3/index.json (nuget.org, nugetzz)"`

proposed/2023/Add-dotnet-search.md
```diff
@@ -28,6 +28,7 @@ The `package search [search terms] [options]` command will have the following op
 |---------|:----------|
 | `--source` | A source to search |
 | `--exact-match` | Return exact matches only as a search result |
+| `--format` | Format the output accordingly. |
```
donnie-msft: What's the default?

Nigusu-Allehu: The default would be Table

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: Nice catch Donnie. Currently, my implementation does not group sources. So, it would be 
`"source" : "nuget"`...
`"source" : "nugetzz"`...
However, grouping them by URI is definitely something to think about. I am not sure what the best practice would be for the customers.

donnie-msft (approved):

proposed/2023/Add-dotnet-search.md
```diff
@@ -43,21 +44,78 @@ This option will specify a list of sources to search from. If a source is not sp
...
+        |---------------------------------------------|----------------|---------|-----------------|
+        | Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
+- `--format List` : each package will be listed line by line
+
+        >NuGet.CommandLine | 6.7.0 | Downloads: N/A
+        >NuGet.CommandLine.XPlat | 6.7.0 | Downloads: N/A
+        >NuGet.Commands | 6.7.0 | Downloads: N/A
+        >NuGet.exe | 3.4.3 | Downloads: N/A
+
+- `--format Json` : A json output will be outputted. It will contain the following metdata.
```
nkolev92: Do we want the --format option to change the output?

Do we need something that controls the extra verbosity first?

Nigusu-Allehu: By change the output do you mean changing the amount of information being outputted(list has less information than json)? If that is the case, my thinking was to include more information in the json output and allow users take what they want from it.

Nigusu-Allehu: We could definitely have a `--verbosity` option and allow all the formats have the same amount of information that is determined by the `--verbosity` option.

Nigusu-Allehu: I added `--verbosity` option to allow similar output throughout the `--format` options

danmoseley:
It was suggested in https://github.com/NuGet/NuGet.Client/pull/5466#issuecomment-1791260443 to add --configFile in this PR, I think

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,91 @@ This option will specify a list of sources to search from. If a source is not sp
 #### **Option `-exact-match`**
 
 - This option will allow for users to be able to search and have only exact matches as an output.
+- The outputted list will contain all versions of the matching package ID
```
danmoseley:
```suggestion
- The results will contain all versions of the matching package ID
```
TIL outputted is a word -- but it's an ugly one..

danmoseley: emitted is a more everyday alternative

proposed/2023/Add-dotnet-search.md
```diff
@@ -24,15 +24,18 @@ Imagine you have your dotnet CLI open. You would like to look up a NuGet package
...
-| `--source` | A source to search |
-| `--exact-match` | Return exact matches only as a search result |
-| `--prerelease` | Allow prerelease packages to be shown. |
-| `--interactive` | Allows the command to stop and wait for user input or action (for example to complete authentication).|
-| `--take` | The number of results to return. The default value is 20.|
-| `--skip` | The number of results to skip, for pagination. The default value is 0. |
-| `--help` | Show command help and usage information |
+| Option           | Function |
+|------------------|--------------|
+| `--configFile`   | The path to the NuGet config file to use. |
```
danmoseley: here or in a configFile section below can it explain how it finds the config file if you do not pass this flag ? presumably it should also say it only uses exactly one config file.

danmoseley: oh, doh, I thought this md was going to be the doc...

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,91 @@ This option will specify a list of sources to search from. If a source is not sp
...
-|---------------------------------------------|----------------|---------|-----------------|
-| Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
+        Source: nuget.org
+        | Package ID                                  | Latest Version | Authors | Downloads       |
+        |---------------------------------------------|----------------|---------|-----------------|
+        | Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
+
+#### **Option `--format`**
+
+This option will allow the specification of the output format. The option will take one of these three arguments: Table, List, or Json. The outputs will depend on verbosity. The following examples are for normal verbosity.
```
nkolev92: I don't think we use capitalization option values.

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,91 @@ This option will specify a list of sources to search from. If a source is not sp
...
+
+- `--format Table` the output will be in a tabular form as shown below
```
nkolev92: Do we want to have both list and table? 

I'm wondering because `dotnet list package` itself has 2 format options, json or console.

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,91 @@ This option will specify a list of sources to search from. If a source is not sp
...
+            }
+        ]
+
+#### Option `--verbosity`
+
+  controls the amount of information included in the output of a command. This allows you to tailor the output to your needs, providing either a concise summary or a detailed breakdown of the information returned by the command. 
+
+Here are the details of each verbosity level:
+
+- **`--verbosity quiet`** : The output will contain minimal information. For each package, it will display only the following metadata:
```
nkolev92: The description says the output will contain minimal information. Shouldn't the log level be minimal?

I think `quiet` is a bit of an odd option for an operation whose primary result is what gets written for the user. 
restore on the other hand, writes an assets file, that's the primary purpose, not the logs themselves.

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,91 @@ This option will specify a list of sources to search from. If a source is not sp
...
+- **`--verbosity normal`** : The output will provide a moderate amount of information. For each package, it will display the following metadata:
+  - Package ID
+  - Latest Version
+  - Authors
+  - Downloads
+  
+- **`--verbosity detailed`** : The output will be highly detailed. For each package, it will display the following metadata:
+  - Package ID
+  - Latest Version
+  - Authors
```
nkolev92: This was a miss in the original write-up, but we do have a discussion of owners vs authors that should probably factor in here. 

cc @albarry4 @donnie-msft

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: Here is the PR I made for the public documentation: https://github.com/dotnet/docs/pull/37747.  I will be continuously updating it and will only be merged once the command is released. I will make sure to explain it in the doc.

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: fixed: `table` and `json` now

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: fixed

proposed/2023/Add-dotnet-search.md
nkolev92: @joelverhagen what are your thoughts? 

fyi @aortiz-msft @mariaghiondea

joelverhagen: IMO owners should be shown if available. It is a credible value compared to authors and aligns with current PM UI work. CC @donnie-msft 

Client should detect the presence of `owners` property in search results. Note that packages can have zero owners so you need to detect presence of the property not whether the owners array is empty.
Problem is registration API does not contain owners. It's only on the search API. Meaning `--exact-match` will show authors. You may opt for `packageid:<ID>` for `--exact-match` on NuGet.org but that's leaky.

There is inconsistency in .NET CLI already.
`dotnet new search cake` -> owners
`dotnet tool search cake` -> authors

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
+  - LatestVersion
+  - Authors
+  - Downloads
+
+  The output will take the following format
+
+        [
+            {
+                "source": "source Name1",
+                "packagesList": [
```
joelverhagen: nit: `packages` instead of `packageList`

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
+
+        Source: nuget.org
+        | Package ID                                  | Latest Version | Authors | Downloads       |
+        |---------------------------------------------|----------------|---------|-----------------|
+        | Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
+
+- `--format json` : A json output will be outputted. It will contain the following metadata.
+  - Source
+  - Package ID
+  - LatestVersion
```
joelverhagen: Will latest version be normalized and/or full version?
If `1.0.0.0+metadata` is the latest version what will show?

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
+
+  The output will take the following format
+
+        [
+            {
+                "source": "source Name1",
+                "packagesList": [
+                    {
+                        "author": "",
+                        "downloads": "",
```
joelverhagen: nit: `totalDownloads` -- we regularly get questions on nuget.org about how download counts are calculated. this is the total for all versions.

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
-Source: nuget.org
-| Package ID                                  | Latest Version | Authors | Downloads       |
-|---------------------------------------------|----------------|---------|-----------------|
-| Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
-| Newtonsoft.Json.Bson                        | 1.0.2          |         | 554,641,545     |
-| Newtonsoft.Json.Schema                      | 3.0.15         |         | 39,648,430      |
-| Microsoft.AspNetCore.Mvc.NewtonsoftJson     | 7.0.12         |         | 317,067,823     |
-...
+        Source: nuget.org
+        | Package ID                                  | Latest Version | Authors | Downloads       |
```
joelverhagen: Downloads are not available on registration API. How does this work for `--exact-match`?

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
+  - Downloads
+
+  The output will take the following format
+
+        [
+            {
+                "source": "source Name1",
+                "packagesList": [
+                    {
+                        "author": "",
```
joelverhagen: `authors` not `author` since the nuspec and protocol treat it as plural

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
-|---------------------------------------------|----------------|---------|-----------------|
-| Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
-| Newtonsoft.Json.Bson                        | 1.0.2          |         | 554,641,545     |
-| Newtonsoft.Json.Schema                      | 3.0.15         |         | 39,648,430      |
-| Microsoft.AspNetCore.Mvc.NewtonsoftJson     | 7.0.12         |         | 317,067,823     |
-...
+        Source: nuget.org
+        | Package ID                                  | Latest Version | Authors | Downloads       |
+        |---------------------------------------------|----------------|---------|-----------------|
+        | Newtonsoft.Json                             | 13.0.3         |         | 3,829,822,911   |
```
joelverhagen: Curious how the digits are rendered per localization. Comma separated thousands/millions is not used in all locales.
https://en.wikipedia.org/wiki/Decimal_separator#Conventions_worldwide

proposed/2023/Add-dotnet-search.md
```diff
@@ -41,23 +44,87 @@ This option will specify a list of sources to search from. If a source is not sp
...
+  
+- **`--verbosity detailed`** : The output will be highly detailed. For each package, it will display the following metadata:
+  - Package ID
+  - Latest Version
+  - Authors
+  - Downloads
+  - Description
+  - Deprecation
+  - Project URL
+  - Vulnerable
```
joelverhagen: How much vulnerability info will be shown? Seems like it could be inspired by dotnet list package and how it renders vulnerability info.

Nigusu-Allehu: I was thinking a binary information(True/False)

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: That is a good point. Currently, for --exact-match, downloads are N/A

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: I have update it to show N/A if there is no vulnerability info otherwise it will show True.

proposed/2023/Add-dotnet-search.md
donnie-msft: In PM UI, the approach I'm proposing is `<package id> "by" <owner if available, or author if not>`.
If the CLI output (eg, for table format) had a column header **"By"**, and it did the same logic, that would make for consistency between CLI and VS.

donnie-msft: In PM UI, something will indicate that it's connected to a nuget.org owner. Not sure what that might be in CLI output. 

#### Author idea for CLI:
_source: private feed (Author)_
|Package ID|By|
|--|--|
|Newtonsoft.Json|James Newton-King|
...

Customers need to understand that absense of an indicator means it's not an "owner". Could add a footnote to explain?

#### Owner idea 1 for CLI:
_source: nuget.org (Owner)_
|Package ID|By|
|--|--|
|Newtonsoft.Json|dotnetfoundation (nuget.org), jamesnk (nuget.org), newtonsoft (nuget.org)|
...

#### Owner idea 2 for CLI:
_source: nuget.org (Owner)_
|Package ID|By|
|--|--|
|Newtonsoft.Json|dotnetfoundation*, jamesnk*, newtonsoft*|
...

*Owner according to package source

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: The normalized version is shown.

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: total downloads are `long` values in the metadata. When converting them to a string, I use the current culture.

accepted/2023/Add-dotnet-search.md
```diff
@@ -48,7 +48,7 @@ This option will specify a list of sources to search from. If a source is not sp
 - For example if a user uses `dotnet package search Newtonsoft.Json`
 
         Source: nuget.org
-        | Package ID                                  | Latest Version | Authors | Total Downloads |
+        | Package ID                                  | Latest Version | Owners | Total Downloads |
```
donnie-msft, line 50: Will these be comma-delimited list of owners?

accepted/2023/Add-dotnet-search.md
```diff
@@ -113,13 +113,13 @@ Here are the details of each verbosity level:
...
   - Latest Version
-  - Authors
+  - Owners
   - Total Downloads
   
 - **`--verbosity detailed`** : The output will be highly detailed. For each package, it will display the following metadata:
   - Package ID
   - Latest Version
-  - Authors
+  - Owners
```
donnie-msft, line 121: Sounds good to show Owner for this first iteration, and if we learn that customers want Author, we can discuss a second column for that in the next iteration.

donnie-msft (approved):

proposed/2023/Add-dotnet-search.md
Nigusu-Allehu: Updated the spec. After yesterday's discussion, it was not favored to show owners and Authors both together. However, it was clear that we want to show `Owners` property. As a result, we will be showing `Owners` in this iteration, and we can decide on whether and how we can include authors information later if there is a request for it.

accepted/2023/Add-dotnet-search.md
Nigusu-Allehu: yes

nkolev92 (approved):

PR #13024, opened 2023-11-17
Technical explanation for `disableTLSCertificateValidation`

Nigusu-Allehu:
Expands technical explanation for implementing a `disableTLSCertificateValidation` option which allows disabling TLS certificate validation.

### DIFF
--- a/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
+++ b/proposed/2023/InsecureConnectionsDisableCertificateValidation.md
@@ -80,9 +80,8 @@ A developer may not want to check in an insecure configuration file and may want
 ### Technical explanation
 
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
-This section will be filled out by the respective person picking up this work in more detail if more needs to be explained. 
 
-`HttpClient` and `HttpClientHandler` should be able to support this functionality easily such as a custom `ServerCertificateCustomValidationCallback` which always returns `true` meaning that any certificate presented by the server will be considered valid for the `disableTLSCertificateValidation` functionality.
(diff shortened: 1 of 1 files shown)


proposed/2023/InsecureConnectionsDisableCertificateValidation.md
```diff
@@ -80,9 +80,8 @@ A developer may not want to check in an insecure configuration file and may want
...
-This section will be filled out by the respective person picking up this work in more detail if more needs to be explained. 
 
-`HttpClient` and `HttpClientHandler` should be able to support this functionality easily such as a custom `ServerCertificateCustomValidationCallback` which always returns `true` meaning that any certificate presented by the server will be considered valid for the `disableTLSCertificateValidation` functionality.
+This proposal introduces a feature to manage the `HttpClientHandler` for conditionally disabling TLS certificate validation. By setting a boolean flag, `disableTLSCertificateValidation`, to true, the HttpClientHandler is configured to bypass the standard certificate checks via a custom `ServerCertificateCustomValidationCallback` that unconditionally returns `true`. This means that any SSL/TLS certificate, regardless of its source, expiration status, or domain name mismatch, will be accepted as valid, effectively disabling the usual security checks performed for HTTPS connections. **Note** This is done per source level.
```
donnie-msft: If you add a new line after every sentence, you'll build a habit that will make Andy super happy :) and also makes future edits smaller as text is spread across multiple lines.

Other nit edits:

```suggestion
This proposal introduces a feature to manage the `HttpClientHandler` for conditionally disabling TLS certificate validation. By setting a boolean flag, `disableTLSCertificateValidation`, to true, the `HttpClientHandler` is configured to bypass the standard certificate checks via a custom `ServerCertificateCustomValidationCallback` that unconditionally returns `true`. This means that any SSL/TLS certificate, regardless of its source, expiration status, or domain name mismatch, will be accepted as valid, effectively disabling the usual security checks performed for HTTPS connections. 

> [!Note]
> `disableTLSCertificateValidation` applies at the package source level.
```

donnie-msft (approved):
I'm not personally a fan of the way the original proposal disables all certificate errors. However, I understand that customers want an escape hatch. I worry that customers will forget they've enabled this and miss some other certificate issue (besides TLS, which was on their mind at the time they enabled the setting).

Issue #12501, opened 2023-03-21
Swap authors for owners in Visual Studio Package Manager UI


JonDouglas:
### NuGet Product(s) Involved

Visual Studio Package Management UI

### The Elevator Pitch

Currently in the Visual Studio Package Manager UI, we show the `authors` field which is a comma delimited list that is commonly used to help a developer know who "authored" the package.

This concept is a bit outdated as NuGet.org provides a `owners` field in the search service that provides an **up-to-date maintainer list that can be used as a primary and if a server doesn't provide it, can fallback on `authors`**. This is not to be confused with the local [`owners` nuspec field](https://learn.microsoft.com/en-us/nuget/reference/nuspec#owners).

https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#search-result

Some further details as pointed out by @zivkan might be that we would only display owners for sources that support it and are fetched from the remote and ignore some nuspec metadata in the global packages folder. 

This will help reduce confusion of a field that is commonly used to impersonate notable entities or have changed in ownership in the years since it was published where the original author is still listed.

![MicrosoftTeams-image (3)](https://user-images.githubusercontent.com/1288848/226624059-925af6f6-1422-4b4e-a070-7bb8283163fe.png)

Related: https://github.com/NuGet/Home/issues/10260

zivkan:
We also need to remember that nuget.org is not the only package feed out there, and we need to take into account how other package sources work. In most cases, they blindly copy the package's nuspec owners field. Many nuget server implementations don't have a concept of different owners for different packages.

What behaviour do we want when a package on a non-nuget.org source has an owners field?  What behaviour do we want when the package doesn't have the owner field?

I 100% agree that the proposal is correct for nuget.org. My concern is what the ideal behaviour is for all non-nuget.org sources.

baronfel:
For what it's worth, on the Template Engine we are also evaluating when and where to surface Owners instead of Authors, since as @JonDouglas pointed out from a NuGet.org perspective Owners are the unit of trust - they define the package prefix reservations.  Our discussion is [here](https://github.com/dotnet/templating/issues/3456), but since our `search` functionality for Templates is very strongly bound to NuGet.org we decided that showing the Owner and highlighting it in the search UI in some fashion was preferable.

joelverhagen:
> I 100% agree that the proposal is correct for nuget.org. My concern is what the ideal behaviour is for all non-nuget.org sources.

[...]

PR #13030, opened 2023-11-20
Spec: Display Known Owners on the PM UI Browse Tab

donnie-msft:
Design for: https://github.com/NuGet/Home/issues/12501

[Rendered proposal](https://github.com/NuGet/Home/blob/dev-donnie-msft-ownerAuthorSpec/accepted/2023/owner-author-pmui.md)


accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,78 @@
...
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+Currently, `Owners` JSON parsed from the Search API is immediately converted into a comma-delimited string of owners.
+To avoid breaking any clients depending on this, `Owners` will continue being a pre-processed string with comma-delimited owners.
+A new property, `OwnersEnumerable` (naming TBD) will be added to [IPackageSearchMetadata](https://github.com/NuGet/NuGet.Client/blob/dc2fadacd311646754ce6b062489b403338972aa/src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs) and [PackageSearchMetadataContextInfo](https://github.com/NuGet/NuGet.Client/blob/dc2fadacd311646754ce6b062489b403338972aa/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs).
+The enumerable owners will allow displaying custom UI for each owner.
+
+Building a profile URL will be done by applying the pattern for each owner: `www.nuget.org/profiles/<owner name>`.
+
+TBD: For non-nuget.org sources...
```
joelverhagen: We had a similar problem for linking to the package details page on a source. I added a `PackageDetailsUriTemplate` resource. Docs: https://learn.microsoft.com/en-us/nuget/api/package-details-template-resource 

You could do something similar with a `OwnerDetailsUriTemplate` resource. If not present, you can simply leave the owner name without a link (plain text).

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,78 @@
...
+The "author" metadata is also misleading as it's a free-form text field which doesn't imply any credentials.
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+![Owners from NuGet.org are shown beside each package ID with a hyperlink and a NuGet Package Icon](../../meta/resources/OwnerAuthor/ownerNuGetOrgLinks.png)
```
joelverhagen: I find the many little nuget icons a little noisy. What do you think about only having the last one or exclude them entirely and just having the links? I take it you mimicked the icon on the package details link already present.

Might be worth a quick pass from the UX team

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,78 @@
...
+
+<!-- Why should we not do this? -->
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+Using `by:` in the same manner as `www.nuget.org`, we could start letting this imply Owner, and introduce a new term for Author.
```
joelverhagen: Personally I think what you have in the main design is fine. We never stated "authors" before so I don't know if many users have a strong expectation of what this value is (owners vs authors) meaning distinguishing the two might not add a lot of value. Just my opinion.

accepted/2023/owner-author-pmui.md
donnie-msft: That's exactly what I wanted to find, thanks!

accepted/2023/owner-author-pmui.md
donnie-msft: Yeah, unfortunately UX team cancelled their office hours last week when I was going to get their input.

Absolutely it does feel noisy. I'd say that noise is good when it reiterates trust. However, now that I'm moving towards profile links (and you seem to agree with the direction), that's probably sufficient on its own.

I think customers would pick up that links are owners, and anything else is questionable/ needs more investigation.

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,78 @@
...
+![Owners from NuGet.org are shown beside each package ID with a hyperlink and a NuGet Package Icon](../../meta/resources/OwnerAuthor/ownerNuGetOrgLinks.png)
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+Currently, `Owners` JSON parsed from the Search API is immediately converted into a comma-delimited string of owners.
+To avoid breaking any clients depending on this, `Owners` will continue being a pre-processed string with comma-delimited owners.
+A new property, `OwnersEnumerable` (naming TBD) will be added to [IPackageSearchMetadata](https://github.com/NuGet/NuGet.Client/blob/dc2fadacd311646754ce6b062489b403338972aa/src/NuGet.Core/NuGet.Protocol/Model/IPackageSearchMetadata.cs) and [PackageSearchMetadataContextInfo](https://github.com/NuGet/NuGet.Client/blob/dc2fadacd311646754ce6b062489b403338972aa/src/NuGet.Clients/NuGet.VisualStudio.Internal.Contracts/ContextInfos/PackageSearchMetadataContextInfo.cs).
+The enumerable owners will allow displaying custom UI for each owner.
```
joelverhagen, line 107: What happens if there are many, many owners? Or the cumulative comma separated list is very long?

```kusto
NiPackageOwners
| extend ['Comma\'d'] = array_strcat(Owners, ", ")
| summarize
    (['Max comma separated len'], ['Example ID 1']) = arg_max(strlen(["Comma'd"]), Id),
    (['Max owner count'], ['Example ID 2']) = arg_max(array_length(Owners), Id)
```

|Max comma separated len|Example ID 1|Max owner count|Example ID 2|
|---|---|---|---|
|192|KenticoCommunity.ConnectionStringModule|12|Cake|

joelverhagen, line 107: What happens if there are no owners? What is displayed?

```kusto
NiPackageOwners
| where array_length(Owners) == 0
| join kind=inner NiPackageDownloads on LowerId
| join kind=inner NiPackageVersions on Identity
| where IsListed == true
| distinct ['Package ID'] = Id, TotalDownloads
| order by TotalDownloads desc
| take 5
```

|Package ID|TotalDownloads|
|---|---|
|NDesk.Options|3158883|
|Core.Common|404489|
|CommonServiceLocator.WindsorAdapter|374352|
|Mono.Posix-4.5|281608|
|Core.Calendar|264348|

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,78 @@
...
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+Customers need a clear understanding of who created a NuGet package in order to trust that package.
+The term "owner" has become overloaded as NuGet has evolved, which has caused confusion. 
+The "author" metadata is also misleading as it's a free-form text field which doesn't imply any credentials.
+
+## Explanation
+
+### Functional explanation
```
joelverhagen, line 28: Will the details pane (right side of the Browse tab, after clicking a result) show owners and or authors still?

I know most of the stuff on the right comes from registration which doesn't return owners but you could plumb the owners data from the search result into the right pane to display purposes.

accepted/2023/owner-author-pmui.md
donnie-msft: Author will be shown as the fallback

accepted/2023/owner-author-pmui.md
donnie-msft: Was thinking to add "Owner(s)" above "Author(s)", probably with the same profile links as packages list.

accepted/2023/owner-author-pmui.md
joelverhagen: This may be fine, if it's not linked as an owner. But this could confuse or even deceive our users. Consider these packages that have no owners but have authors that match real NuGet.org usernames:

```kusto
let LatestVersionWithNoOwners = NiPackageOwners
| where array_length(Owners) == 0
| join kind=inner NiPackageDownloads on LowerId
| join kind=inner NiPackageVersions on Identity
| where IsLatestSemVer2 == true
| join kind=inner NiPackageManifests on Identity
| distinct ['Package ID'] = Id, LatestVersion = Version, TotalDownloads, Authors
| order by TotalDownloads desc; 
let AllOwners = NiPackageOwners
| mv-expand Owner = Owners to typeof(string)
| distinct Owner, LowerOwner = tolower(Owner);
LatestVersionWithNoOwners
| extend LowerAuthors = tolower(trim("\\s", Authors))
| join kind=leftouter AllOwners on $left.LowerAuthors == $right.LowerOwner
| project ['Package ID'], TotalDownloads, Authors, ['Matching owner'] = Owner
| where isnotempty(['Matching owner'])
| order by TotalDownloads
| take 10
```

|Package ID|TotalDownloads|Authors|Matching owner|
|---|---|---|---|
|Mvc3CodeTemplatesCSharp|31125|Microsoft|Microsoft|
|OData.Helper|26299|Microsoft|Microsoft|
|WindowsAzureStorage.Helper|20959|Microsoft|Microsoft|
|SharedCode.Common.Security|9604|e-Construmarket|e-Construmarket|
|EFCTP4|7070|Microsoft|Microsoft|
|Mvc3CodeTemplatesVB|5691|Microsoft|Microsoft|
|FsLib.Common|5577|ifish|iFish|
|EFCTP4.SampleBlogModel|4199|Microsoft|Microsoft|
|Moosend.Wrappers.CSharpWrapper|4059|moosend|moosend|
|Lykke.Service.LykkeDevelopers.Client|3197|Lykke|lykke|

accepted/2023/owner-author-pmui.md
joelverhagen: Might be good to have a ❔ icon next to the authors or even owners to describe the difference on hover. Don't know if this is part of the VS UI language but there could be prior are of how to describe nuanced topics inline of UI.

Seems good otherwise

accepted/2023/owner-author-pmui.md
donnie-msft: I'd only render owner hyperlinks when we have owner metadata from the package source. If the author looks like an owner, it won't have a hyperlink (or indicative icon, if we keep that). This is the reason my alternative introduced a new term, 'contributors', to start differentiating these with a descriptor.

joelverhagen: I think showing either owners or authors but not both (per source) is a better design, but I don't feel very strongly. I think I am erring on the side of caution (authors are not secured like owners) and you are erring on the side on usability (provide the user with what information we have). Pros and cons. Perhaps we can raise the point to others for additional feedback? I'll leave the decision to you.

donnie-msft: > I think showing either owners or authors but not both (per source) is a better design

👍and that's what I'm proposing as well. There is never both owner/author shown. If we have owner, show it with links, if we don't, show author which will never have links.

joelverhagen: I think my previous message is unclear or I am misunderstanding you.

Consider these cases, assuming we know NuGet.org supports returning owners on packages (and this means it will return owners property on ALL packages, even if the property is an empty array)

1. NuGet.org selected as only source, package is returned with 0 owners.
2. All source selected, package is returned with 0 owners, package exists only on NuGet.org 
3. All source selected, package version is returned with 0 owners, package exists on non-NuGet.org feeds 
4. All source selected, package version is returned with 0 owners, package exists on both NuGet.org and other feeds

For case 1 and 2, I propose that authors are NOT shown since we know the originating feed has a validated owners property and we think showing "no owners" or literally nothing after the package ID is better than some non-validated authors field.

For case 3, it seems totally reasonable to show (unlinked) authors field

For case 4, it's tricky. I'm not sure what to do but without package source mapping we can't guarantee the download will even come from NuGet.org. So perhaps having undefined/non-deterministic behavior for owners vs. authors display is okay too.

Anyhow, if nuget.org is the only selected feed I would expect authors to appear on no results since displaying them for ownerless packages right next to results with owners displayed might accidentally lend trustworthiness to the authors value that really should be reserved for the owners value.

Said another way, do you want "Microsoft" (unlinked, the authors value) to show when you search for OData.Helper on nuget.org?

accepted/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
```diff
@@ -1,99 +1,99 @@
...
+
+When someone attempts to run pack on an SDK-based project, NuGet will fail with the following error:
+
+```console
+Error NU5049: The `pack` command for SDK-style projects is not supported, use `dotnet pack` to pack this project instead. You may set the ENABLE_LEGACY_NUGETEXE_CSPROJ_PACK environment variable to revert to the previous packing behavior.
+```
+
+This *breaking change* will be announced via blog well before the NuGet.exe carrying the change ships.
+
+### Technical explanation
```
joelverhagen: Will there be a new `SearchQueryService` resource version so that server can explicitly state they support owners in the response JSON?
https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource#versioning

Alternatively, I guess you are checking for the presence of the `owners` property?

We didn't do this for deprecation or vulnerabilities but depending on whether you want to make owners vs. authors a feed level decision or an individual search result level decision, it might help to have it declared in the service index. Something to consider.

accepted/2023/owner-author-pmui.md
donnie-msft: Today, the UI truncates a single line of text at the width of the packages list. The reasoning is that the packages list doesn't intend to show the full context of metadata, but instead provides a "metadata preview" to give customers a little more than "package ID" when searching. Rich details are always available in the details pane. 

This maximizes real estate in the packages list. Alternate designs may come up as we rethink the PM UI layout and architecture.

I believe I'll propose to keep that philosophy even with owners and profile hyperlinks. It's true that some owners may not be visible (depending on the actual width of the packages list), but customers can:
1. View the tooltip to find the complete owner list & package description.
2. Select the package to find a complete Owners listing in the details pane.

donnie-msft: I didn't realize you were writing about the scenario of merging metadata when the `All` source is selected. 

For Case 1 & 2, not showing anything (even author) is an interesting idea. My instinct is that some customers rely on author to quickly spot which internal team created a package. Removing it entirely is enticing, but I think I'll avoid that for this spec.

My plan was to show Trusted Owner provided from a source, if available, otherwise continue showing author. 

The notion of an icon paired with each Owner plays an important role in the UI. We need a fixture that always means Trusted Owner, even if there isn't a profile hyperlink that can be attached to the owner.

> Said another way, do you want "Microsoft" (unlinked, the authors value) to show when you search for OData.Helper on nuget.org?

I'm leaning yes, since it already does so. The new affordances won't be there, so we can start helping customers learn that this rendering means something lesser than a Trusted Owner.

However, I like how nuget.org has an explanation when there's no owners, which I could implement in the Tooltip of the packages list (and maybe another icon/affordance):  `This package has no owners and is not being actively maintained.`.

Perhaps these can be done in iterations (1. keep author, introduce new UI affordances. 2. hide author in certain cases, 3. hide author from packages list in all cases)....

accepted/2023/owner-author-pmui.md
donnie-msft: UX board agreed - the icons are too noisy - so we'll plan to just show links. Thanks!

accepted/2023/owner-author-pmui.md
donnie-msft: Added a screenshot of this.

accepted/2022/NuGetExeWarnWhenSDKProjectsGetPacked.md
donnie-msft: Added to Future Possibilities:
> Stop falling back to `author` when a package source supports Trusted Owners.

accepted/2023/owner-author-pmui.md
donnie-msft: Will look into how to add a help link in development, but imagining the ℹ️ icon will do the trick.

accepted/2023/owner-author-pmui.md
donnie-msft: I've scoped this spec to single-source only, and mentioned this in Future Possibilities. Leaving the comment open for now so others are more likely to see it.

Wonderwhothen:
Not a problem

Wonderwhothen:
This open also blocked current owner is not the only one

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,214 @@
+# Display Trusted Owners on the PM UI Browse Tab
```
nkolev92: Is `Trusted` too strong here? Who trusts these owners? How can I declare that I trust someone?

Should we just call them owners?

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,214 @@
...
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+Instead of showing only `author` as the PM UI does today, the packages list will now show Owners when available from a package source, and only fallback to author when Owners are not available.
+For this iteration, a single package source must be selected (ie, not `All`) and the `Browse` tab must be selected.
+
+The Details Pane will show both Owner and Author for a selected package.
+
+The Owners in both of these panes will be shown as a hyperlink.
```
nkolev92: What if owners is part of the the search response, but there's no OwnerDetailsTemplate?

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+Instead of showing only `author` as the PM UI does today, the packages list will now show Owners when available from a package source, and only fallback to author when Owners are not available.
+For this iteration, a single package source must be selected (ie, not `All`) and the `Browse` tab must be selected.
```
nkolev92, line 35: What do we show on all? Just the authors?

nkolev92 (commented):
A few small questions.

Wonderwhothen:
Good questions!

Wonderwhothen:
I have a bit of  concern over decision my is  2f authentication,also being
spammed, contributors.

accepted/2023/owner-author-pmui.md
donnie-msft: Another word can work. The problem with calling everything "owners" is in the motivation; the word has multiple meanings and doesn't convey whether it's from a source of trust (package source) or a piece of metadata in the package.

accepted/2023/owner-author-pmui.md
donnie-msft: There are 2 possibilities, and the 2nd one is `PackageDetailsUriTemplate`.
https://github.com/NuGet/Home/blob/dev-donnie-msft-ownerAuthorSpec/accepted/2023/owner-author-pmui.md#packages-list-ui-for-ownerauthor

accepted/2023/owner-author-pmui.md
donnie-msft: Correct. Nothing will change, for now.

accepted/2023/owner-author-pmui.md
nkolev92: > from a source of trust

Similar as earlier, `trust` raises too many questions :D 

`Package source owners`, `Current owners`? 

Words are difficult :D

donnie-msft: `Package source owners` - I like except that it's a little confusing as it sounds like the admins of the servers.
This isn't too time critical to this spec as it's just the nomenclature we're using in code and eventually docs. Getting a definition/word nailed down sooner than later would be good.

nkolev92: > This isn't too time critical to this spec

100%, this is spec/docs feedback, but doesn't change the overall intent of the spec.

accepted/2023/owner-author-pmui.md
skofman1: @donnie-msft , I was about to start a similar thread when I saw this conversation. First of all, thank you and @joelverhagen for the in-depth discussion of this edge case.

> Anyhow, if nuget.org is the only selected feed I would expect authors to appear on no results since displaying them for ownerless packages right next to results with owners displayed might accidentally lend trustworthiness to the authors value that really should be reserved for the owners value.

I share this concern. The 'author' metadata can be easily gamed by attackers, so we shouldn't display it for nuget.org packages. Even if it looks differently than 'Trusted owners', I don't know if users will be able to easily tell the difference and make the right trust decision. 
Please don't consider my comment as blocking. Happy to have an offline chat if it helps. :)

donnie-msft: I've redacted the fallback to author when the package source provides Trusted Owners.
https://github.com/NuGet/Home/blob/dev-donnie-msft-ownerAuthorSpec/accepted/2023/owner-author-pmui.md#packages-without-a-trusted-owner

accepted/2023/owner-author-pmui.md
```diff
@@ -0,0 +1,214 @@
...
+
+N/A
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+1. Edge case: The package source responds with a Trusted Owner, but doesn't provide URL templates for **Owner Details** nor **Package Details**.
```
nkolev92: Maybe show something like (incomplete metadata from source).

JonDouglas (approved):
Great spec! I think it is really straight forward and important for our security efforts long term.

I personally think the NuGet.org icon is a nice touch that helps separate the central registry and private ones. But I think how it is right now is what we're after here.

accepted/2023/owner-author-pmui.md
donnie-msft: Definitely worth considering something like this, especially once more package sources support Owner.
For this first iteration, how do you feel about just making these prerequisites for showing Owner? So, if Search API returns owner, and at least one URL template is available, show Owner, otherwise keep current behavior.

nkolev92: I don't feel too strongly. 
I just suggested this because I thought it was the more transparent behavior.

nkolev92 (approved):

Nigusu-Allehu (approved):

PR #13133, opened 2024-01-05
Add `problems` value to `dotnet package search` JSON output

Nigusu-Allehu:
Add `problems` section to the JSON output. This section will be used to store and report errors.

### DIFF
--- a/accepted/2023/Add-dotnet-search.md
+++ b/accepted/2023/Add-dotnet-search.md
@@ -80,11 +80,12 @@ This option will allow the specification of the output format. The option will t
   - Owners
   - TotalDownloads
 
-  The output will take the following format
+  The output will take the following format. Any problems with the source will be reported in the `errors` value.
 
         [
             {
                 "source": "source Name1",
+                "errors": [],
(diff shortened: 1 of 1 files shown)


accepted/2023/Add-dotnet-search.md
```diff
@@ -80,11 +80,12 @@ This option will allow the specification of the output format. The option will t
   - Owners
   - TotalDownloads
 
-  The output will take the following format
+  The output will take the following format. Any problems with the source will be reported in the `errors` value.
```
nkolev92: I'd take a look at https://github.com/NuGet/Home/blob/dev/accepted/2022/DotnetListPackageMachineReadableJsonOutput.md#errorwarning-handling as an example.

We probably need to add at least a level and add a problems section instead of just errors. 

Another thing that we missed was the format version. The dotnet list package --format spec is very complete, so it's worth getting some inspiration off of that.

nkolev92 (approved):

jeffkl (approved):

Issue #12307, opened 2022-12-07
Warn when vulnerabilities are detected during packages.config restore


nkolev92:
- [x] https://github.com/NuGet/NuGet.Client/pull/5383
- [x] Design configuration knobs (if any) for packages.config restore - https://github.com/NuGet/Home/pull/13154
- [x] https://github.com/NuGet/Home/issues/13253
- [x] Add telemetry for package vulnerability checking at packages.config restore time
- [x] Enable vulnerability checking by default 
- [ ] https://github.com/NuGet/docs.microsoft.com-nuget/issues/3247

nkolev92:
Similar work in https://github.com/NuGet/NuGet.Client/pull/4140.

Issue #13118, opened 2023-12-31
NuGet not respecting local cache


Nefcanto:
### NuGet Product(s) Affected

MSBuild.exe, dotnet.exe

### Current Behavior

NuGet does not honor the locale cache. We have restored a project with many packages in a docker image to cache images so that we can run an app without an internet connection, but we get this error:

```
ApiContainer | MSBuild version 17.8.3+195e7f5a3 for .NET
ApiContainer |   Determining projects to restore...
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [/Project/AdminApi/Api.csproj]
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error :   Resource temporarily unavailable (api.nuget.org:443) [/Project/AdminApi/Api.csproj]
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error :   Resource temporarily unavailable [/Project/AdminApi/Api.csproj]
ApiContainer | 
ApiContainer | Build FAILED.
ApiContainer | 
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error : Unable to load the service index for source https://api.nuget.org/v3/index.json. [/Project/AdminApi/Api.csproj]
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error :   Resource temporarily unavailable (api.nuget.org:443) [/Project/AdminApi/Api.csproj]
ApiContainer | /usr/share/dotnet/sdk/8.0.100/NuGet.targets(156,5): error :   Resource temporarily unavailable [/Project/AdminApi/Api.csproj]

```

### Desired Behavior

When packages are cached once, NuGet should use them if the internet connection is not available. Preferably it should use them and not connect to the internet at all, even if it's available. Because packages are already there on the disk.

### Additional Context

This is our `Dockerfile`:

```
# https://mcr.microsoft.com/v2/vscode/devcontainers/dotnet/tags/list
# https://mcr.microsoft.com/v2/devcontainers/dotnet/tags/list
# https://github.com/devcontainers/images/blob/main/src/dotnet/.devcontainer/Dockerfile
# https://hub.docker.com/_/microsoft-dotnet-sdk
ARG VARIANT="8.0-bookworm-slim"
[...]

PR #13154, opened 2024-01-13
Vulnerability checking in PC restore

nkolev92:
Design for https://github.com/NuGet/Home/issues/12307


accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+See [vulnerabilities in PackageReference restore motivation](../2022/vulnerabilities-in-restore.md#motivation).
+
+## Explanation
+
+### Functional explanation
+
+Packages.config restore combines all packages regardless of project. Instead of raising warnings and errors on a per project level, at restore time, they're raised on a per package level.
+
+The configuration knobs for packages.config restore auditing functionality will be NuGet.config based.
+In particular, there will be 2 new configuration keys:
```
zivkan: I disagree with this.

Firstly, as you already listed in the unanswered questions section, should PackageReference projects use this config?

Secondly, from a customer point of view, it means they'll have to duplicate the config (when not using defaults), and will have (at least) two places to update when changing values. This will have even higher impact when [suppressing specific advisories](https://github.com/NuGet/Home/pull/12921) is implemented, as that feature will have to design a packages.config specific alternative, and customers may need to duplicate the suppressions.

packages.config already reads MSBuild properties for lock files, so configuring and using lock files is the same, regardless if the project uses packages.config or PackageReference.  I think NuGetAudit for packages.config should be the same.

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+- SourcesWithVulnerabilityData
+
+## Drawbacks
+
+- Vulnerability checking comes with a performance hit.
+
+## Rationale and alternatives
+
+- Enable vulnerability checking on demand.
+- Enable vulnerability checking at installation time only.
```
zivkan: Checking at installation time only is not a realistic alternative. The primary benefit of audit is to get notifications when a package that previously did not have known vulnerabilities now does. Therefore, in my opinion, checking only at installation time misses the intent for why the feature is useful.

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+    </config>
+</configuration>
+```
+
+Whenever a vulnerability for a package is discovered, a warning will be raised indicating the severity and advisory url.
+Note that warnings as errors and no warn are not support in packages.config projects, so these warnings will not fail the build.
+
+> Package 'Contoso.Service.APIs' 1.0.3 has a known critical severity vulnerability, https://github.com/advisories/GHSA-1234-5678-9012.
+
+For performance considerations, vulnerability checks are only going to be performed when the packages.config restore downloads a package.
```
zivkan: I don't agree with this.

Firstly, it makes packages.config package audit on dev machines of very limited usefulness. Therefore, my interpretation is that it appears that this is being designed as a CI-only feature. But since packages.config doesn't support warning as errors, it obligates customers to check their CI logs frequently.

Secondly, consider just Azure Pipelines for a moment. Microsoft hosted agents are stateless (wiped clean after every job). However, self-hosted agents are not. Azure Virtual Machine Scale Set agents are configurable stateful or stateless. On stateful machines, the `checkout` task has a parameter to clean, or not, the repo. Therefore, customers who set this to skip cleaning will have populated solution packages folders before restore runs. Furthermore, whther you're on a stateful or stateless CI agent, some customers use [the cache task](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/cache-v2?view=azure-pipelines) to populate the global packages, or solution packages directory, more quickly than NuGet can download and extract packages.

So, even on CI, there are many scenarios where NuGet won't be downloading or extracting a package on restore.

IMO there are too many scenarios where "audit only when downloading packages" will prevent recent advisories from being presented to customers, so I don't think it's a good design.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: Removed.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: Usage of packages.config lock is minimal, so I'm not sure it's necessarily a well established pattern. 
Maybe we make the nuget.config apply to everything? 

I'm also wondering if the technical overhead of adding individual project support in packages.config is worth it in V1 of the feature.

nkolev92: fwiw, I'll schedule a meeting to discuss this.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: We need a better approach than every time for packages.config. 

We've already gotten some feedback where people have ended up disabling audit for PR because it goes online even when all the packages are on disk. I'm struggling to find that issue now, but i'll figure it out. 

Another alternative is when the audit data changes, but that's a lot more complicated *without* the fast up to date checks for audit data. 

A different idea is only when a package is downloaded *and on first* restore. 
Either way, I think every single time might lead to some noise.

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+
+Packages.config restore combines all packages regardless of project. Instead of raising warnings and errors on a per project level, at restore time, they're raised on a per package level.
+
+The configuration knobs for packages.config restore auditing functionality will be NuGet.config based.
+In particular, there will be 2 new configuration keys:
+
+| Key | Acceptable values | Description | Default |
+|-----|-------------------|-------------|---------|
+| auditForPackagesConfig | enable, disable | Enables or disables NuGet Audit for packages config projects | If not specified, the default will be `enable` |
+| auditLevelForPackagesConfig | Critical, high, moderate, low | Configurations the default audit level for NuGet audit for packages config projects |  If not specified, the default will be `disable` |
```
donnie-msft: 2 typos?

```suggestion
| auditLevelForPackagesConfig | Critical, high, moderate, low | Configures the default audit level for NuGet audit for packages config projects |  If not specified, the default will be `low` |
```

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+- CheckPackagesDurationSeconds
+- SourcesWithVulnerabilityData
+
+## Drawbacks
+
+- Vulnerability checking comes with a performance hit.
+
+## Rationale and alternatives
+
+- Enable vulnerability checking on demand.
```
donnie-msft: How would this work? Is this for CLI or VS?

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Prior Art
+
+- [Vulnerability reporting in PackageReference restore](../2022/vulnerabilities-in-restore.md)
+
+## Unresolved Questions
+
+- Should the NuGet.config configuration *affect* the PackageReference defaults as well?
+- Should vulnerability checks be performed at every packages.config restore?
+- Should we add the vulnerability metrics in the vs/nuget/restoreinformation event or a dedicated one?
```
donnie-msft: What reasons are there to not just create a dedicated one?

donnie-msft (commented):
In VS, we have the Vulnerability InfoBar that appears directing customers to the PM UI.

Wondering if we need to ensure customers are aware this feature has become enabled for their packages.config projects. Is VS' "What's New" page enough?

Scenario: a legacy project has been used internally for years with a vulnerable package. Suddenly, VS vNext begins showing errors/infobars popping up, despite no change in vulnerabilities. As a customer, would it be obvious this is a new feature lighting up, or would it appear as a new vulnerability in my project and cost me time verifying and potentially raising false alarms in my org?

accepted/2024/vulnerabilities-in-packages.config-restore.md
zivkan: there are 20+ common properties that get added to every telemetry event, and the telemetry service throttles when too much telemetry is being sent. Therefore, many events where the component (NuGet, in this case) adds only a few properties is much more likely to get throttled than one event where many properties are added.

donnie-msft: Good point. My understanding is that it's based on the quantity of events, not the payload of events. So, this question is whether to add 1 event versus adding to the existing properties. Either way, seems reasonable to reuse the event we have.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: This could be CLI primarily, VS technically has PM UI. 

It's more of throwaway idea. I think on demand doesn't move the problem enough to the forefront of the user.

accepted/2024/vulnerabilities-in-packages.config-restore.md
donnie-msft: Here's one I know of: https://github.com/NuGet/Home/issues/13118#issuecomment-1882409588

nkolev92:
Thanks for the comments @donnie-msft 

> In VS, we have the Vulnerability InfoBar that appears directing customers to the PM UI.
> Wondering if we need to ensure customers are aware this feature has become enabled for their packages.config projects. Is VS' "What's New" page enough?
> Scenario: a legacy project has been used internally for years with a vulnerable package. Suddenly, VS vNext begins showing errors/infobars popping up, despite no change in vulnerabilities. As a customer, would it be obvious this is a new feature lighting up, or would it appear as a new vulnerability in my project and cost me time verifying and potentially raising false alarms in my org?

I think this is something we should figure out in general for new features. I think we can do that separately from this particular feature and not block on the spec. 
Do you know if anyone started a discussion on this topic? If not, I can do it.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: I changed it to default, and added details in unresolved questions to figure out if we need to be more selective in order not to regress the overall experience.

donnie-msft:
> Thanks for the comments @donnie-msft
> 
> > In VS, we have the Vulnerability InfoBar that appears directing customers to the PM UI.
> > Wondering if we need to ensure customers are aware this feature has become enabled for their packages.config projects. Is VS' "What's New" page enough?
> > Scenario: a legacy project has been used internally for years with a vulnerable package. Suddenly, VS vNext begins showing errors/infobars popping up, despite no change in vulnerabilities. As a customer, would it be obvious this is a new feature lighting up, or would it appear as a new vulnerability in my project and cost me time verifying and potentially raising false alarms in my org?
> 
> I think this is something we should figure out in general for new features. I think we can do that separately from this particular feature and not block on the spec. Do you know if anyone started a discussion on this topic? If not, I can do it.

Sure, feel free. Not aware of anyone else asking this question. It is just a thought I had reading the spec.

FYI, I'm not blocking on this spec (hence I didn't use the Changes Requested state). It's probably fine to call out that we expect customers to see What's New or we're just not concerned with any jarring effects caused by lighting it up at this time.

EDIT: this comment highlights the "no new warnings for minor releases" idea. Wonder how that applies to this spec.
https://github.com/dotnet/sdk/pull/38037#issuecomment-1897215896

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: I was using this as a sanity check. I'll just add it to the main event.

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+
+- Enable vulnerability checking on demand.
+
+## Prior Art
+
+- [Vulnerability reporting in PackageReference restore](../2022/vulnerabilities-in-restore.md)
+
+## Unresolved Questions
+
+- Should we use MSBuild properties instead?
```
JonDouglas: 👍

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+- Should the NuGet.config configuration *affect* the PackageReference defaults as well?
```
JonDouglas: Not sure I like the idea of nuget.config config being introduced here nor affecting PR.

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+- What should the vulnerability check frequency be? Every time might lead to a lot of overhead on packages.config restores, since no-ops are currently very fast.
```
JonDouglas: What makes the most sense from a gut instinct?

accepted/2024/vulnerabilities-in-packages.config-restore.md
```diff
@@ -0,0 +1,93 @@
...
+The configuration knobs for packages.config restore auditing functionality will be NuGet.config based.
+In particular, there will be 2 new configuration keys:
+
+| Key | Acceptable values | Description | Default |
+|-----|-------------------|-------------|---------|
+| auditForPackagesConfig | enable, disable | Enables or disables NuGet Audit for packages config projects | If not specified, the default will be `enable` |
+| auditLevelForPackagesConfig | Critical, high, moderate, low | Configures the default audit level for NuGet audit for packages config projects |  If not specified, the default will be `low` |
+
+The audit functionality for packages.config restore will be enabled by default.
+To disable it, one can specify a property in the config section of the configuration file.
```
JonDouglas: This might get confusing between having this and also a MSBuild property.

Is the rationale here that this would be used for legacy project system?

accepted/2024/vulnerabilities-in-packages.config-restore.md
JonDouglas: is there any technical or feasibility issues with using the same pattern we've established earlier w/ MSBuild properties? 

I like re-using the same pattern if possible.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: In the future push, I explained on a heuristic that's specifically focusing on the Visual Studio experience, where every restore checking vulnerabilities might be too costly. 

The heuristic is: 
* Refresh vulnerabilities anytime a package is installed in the packages folder. This includes downloading, copying from the gpf.
* Refresh vulnerabilities during the first restore in Visual Studio.
* Replay known vulnerability warnings in Visual Studio in every subsequent when there's no new packages downloaded.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: I recommend reading: https://github.com/NuGet/Home/blob/dev-nkolev92-vulnerabilityCheckingInPCRestore/accepted/2024/vulnerabilities-in-packages.config-restore.md#msbuild-properties-vs-nugetconfig-for-packagesconfig-restore-configuration. 

It covers the in depth thought process. 
Let me know what you think.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: I detail that in https://github.com/NuGet/Home/blob/dev-nkolev92-vulnerabilityCheckingInPCRestore/accepted/2024/vulnerabilities-in-packages.config-restore.md#msbuild-properties-vs-nugetconfig-for-packagesconfig-restore-configuration . 

We have places where we have duplicate configurations, one via nuget.config and another one via property. The property tends to override the nuget.config value. 

The reasoning here is that in packages.config world, for those customers, nuget.config makes more sense.

accepted/2024/vulnerabilities-in-packages.config-restore.md
nkolev92: My argument is that it's not a pattern we've adopted for packages.config projects. 
Technically, they're both feasible, but for packages.config projects using MSBuild property is not really a thing like it is in PackageReference.

donnie-msft (approved):

nkolev92:
TLDR from the meeting: 

* We'll go with properties for now, but the concerns about nuget.config vs msbuild are greater for the suppressions work, so if the suppressions work leans nuget.config, we might do the same for the vulnerability checking.

Issue #12583, opened 2023-05-11
Support viewing README files in Visual Studio


JonDouglas:
### NuGet Product(s) Involved

Visual Studio Package Management UI

### The Elevator Pitch

Last December, it was announced that you can edit and render markdown files now in Visual Studio. 

https://devblogs.microsoft.com/visualstudio/write-markdown-without-leaving-visual-studio/

It is a regular ask from the developer community to be able to read READMEs of packages while they are inside the PM UI whether browsing or getting started with a NuGet package. 

Developers should see a README in a tab similar to the [NuGet.org re-designed tabs](https://devblogs.microsoft.com/nuget/introducing-the-new-nuget-org-package-details-page/):

![image](https://github.com/NuGet/Home/assets/1288848/e9f08ccb-2f34-4b25-9d8f-a27e439cc304)

Inside of Visual Studio:

![image](https://github.com/NuGet/Home/assets/1288848/bb080e1d-5a8f-4be7-b116-660fc193a2ff)

Please 👍 or 👎 this comment to help us with the direction of this feature & leave as much feedback/questions/concerns as you'd like on this issue itself and we will get back to you shortly.

Thank You 🎉

jgonz120:
.

PR #13200, opened 2024-01-26
Add Tech Spec for rendering README in PM UI

jgonz120:
Tech spec for [#12583](https://github.com/NuGet/Home/issues/12583)


accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,77 @@
...
+
+It's also currently marked as obsolete since the interface has not been finalized. So when an upgrade is made we may have to change how we use the control.
+
+## Rationale and alternatives
+By using an existing control we maintain consistency throughout the IDE and can rely on the owner to fix any bugs with the control.
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
```
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Is there any other use of the IMarkdownPreview control that we can mention as Prior Art or are we the first ones taking a dependency on it?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,77 @@
+# Markdown Readme Rendering in PM UI
+
+- Jonatan Gonzalez ([jgonz120](https://github.com/jgonz120)) 
+- Issue [#12583](https://github.com/NuGet/Home/issues/12583) <!-- GitHub Issue link -->
+- [Feature Spec](https://github.com/NuGet/Home/blob/7943122dffa435f4daeee600efcc5b744cd2e97e/accepted/2023/PMUI-Readme-rendering.md)
+
+## Summary
+
+We want to add update the PM UI to render the ReadMe.md file while browing packages.
```
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Consider adding some specificity to the summary about where and when the readme will be rendered in the PM UI. Here's a suggestion, but feel free to reword as you see fit. This suggestion also removes wording about what we intend to do, instead focusing on describing the feature:

```suggestion
Adds Readme.md file rendering to the package details pane for the selected package in the PM UI.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,77 @@
...
+- Issue [#12583](https://github.com/NuGet/Home/issues/12583) <!-- GitHub Issue link -->
+- [Feature Spec](https://github.com/NuGet/Home/blob/7943122dffa435f4daeee600efcc5b744cd2e97e/accepted/2023/PMUI-Readme-rendering.md)
+
+## Summary
+
+We want to add update the PM UI to render the ReadMe.md file while browing packages. 
+
+## Motivation 
+
+We want help our customers understand what a package does and if it will be helpful to them. This could also increase ReadMe adoption if developers know their users will see the ReadMe.
```
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: The functional spec covers this well and we don't want to repeat what the functional spec says, but maybe add a brief description of how the Readme would help a user understand what a package does. Here's a suggestion, but feel free to reword it:

```suggestion
The Readme file is an essential source of information that can help customers understand what a package does. By rendering the Readme file directly in the PM UI, customers can easily access the information they need to determine if a package will be helpful to them while searching, browsing, and managing their projects’ packages. This enhancement streamlines the user experience and may encourage package authors to create comprehensive Readme documentation when customers will be more likely to rely on it for useful information about the package.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,77 @@
...
+
+//We update the current markdown being rendered by calling "UpdateContentAsync"
+markdownPreview.UpdateContentAsync(markDown ?? string.Empty, ScrollHint.None)
+
+//IMarkdownPreview.VisualElement contains the FrameworkElement to be passed to the view
+MarkdownPreviewControl = markdownPreview.VisualElement
+```
+The ReadMe file will be selected from disk if available otherwise pulled from package source. 
+
+The ReadMe will be displayed for both solution level and project level package manager.
```
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: I think these can be moved to the Function Explanation section which will help answer where we are going to be fetching the Readme from and in which PM UI levels it will be supported.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Functional explanation
+When a package is selected we will determine if a ReadMe file exists and if it does we'll render it in the PM UI. 
+
+The PM UI will be updated to have tabs for the Package Details and the the ReadMe. This UX will be displayed for both the Browse and installed tabs.
+![Alt text](https://github.com/NuGet/Home/assets/89422562/81b24877-f12f-4783-905c-4a155d3c7693)
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+### Technical explanation
```
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Can we add a brief and high-level detail of how we are designing the fetching / reading of the Readme markdown from disk versus from fetching it remotely from the package source? In the latter case, are there any additional considerations for packages in a local package source (e.g. a folder on disk with NuGet packages)?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (Nigusu-Allehu) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]:
```suggestion
We want to update the PM UI to render the ReadMe.md file while browsing packages. 
```

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jebriede) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: How will we handle any errors while reading the Readme file? Will we show them place of the rendered Readme, or in a gold bar, or in the Output pane, etc.?

jebriede (commented):
Looks good! Left a few suggestions.

JonDouglas:
Looks good left a few comments to help

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Added a screenshot from the PR in VS experience

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I added some questions regarding this. I found that in my NuGet folder there were ReadMe.md files in the root of the package and in the _content folder for the package. 

As for getting the ReadMe from the API. Is the API we use well documented? I imagine it would be since it would need to be implemented by anyone who wants to implement a source right?

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I'm trying to understand what an error could mean for the renderer. If that means that the MD is malformed then we should see this happening on NuGet.org right? 

Either way I would think the error should be visible in the PMUI since putting it in the Output pane would be confusing especially if it's not already open. What's a gold bar?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,88 @@
...
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
+    - [ ] Display the tab with a message saying there is no ReadMe defined.
+1. Where do we get the ReadMe from when it's not on the disk?
+    - Does the current API return the ReadMe information when searching for a package? 
+    - If not is there an API to retrieve the ReadMe information?
+1. Where are the ReadMe files saved in a package? 
+    - There are examples of them being in root folder as well as a _content folder. 
+1. What do we want the UX to be when an exception or error occurs while reading a ReadMe file? 
+1. Do we want the ReadMe to update whenever a new version is selected for the current package?
```
Comment (jgonz120) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: @albarry4, do you have thoughts on this? I'm guessing it's possible that the ReadMe may change between versions, so I think it makes sense to render the ReadMe for the current version.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
+    * Current suggestion is to focus on ReadMe.md. Once that's complete see how ReadMe.txt would appear when rendered. 
+    * How many packages have a ReadMe.txt vs ReadMe.md?
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
+    - [ ] Display the tab with a message saying there is no ReadMe defined.
+1. Where do we get the ReadMe from when it's not on the disk?
```
Comment (jgonz120) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: @NuGet/nuget-client from looking through the API docs it doesn't look like there's a specific end point for downloading the ReadMe files. Is there any concern with downloading the nupkg to extract the readme from it when trying to render the ReadMe? This would mean that every time a user looks at a package we would download the file even though they haven't hit install.

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Does the server "host" these files somewhere similar to the icons, license files etc?

That's the approach we take for license/icons.

There are concerns with download packages we aren't going to use, in particular due to https://learn.microsoft.com/en-us/nuget/concepts/package-installation-process, it'd basically mess with package source mapping.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,88 @@
...
+    - Does the current API return the ReadMe information when searching for a package? 
+    - If not is there an API to retrieve the ReadMe information?
+1. Where are the ReadMe files saved in a package? 
+    - There are examples of them being in root folder as well as a _content folder. 
+1. What do we want the UX to be when an exception or error occurs while reading a ReadMe file?
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: If this is a transient problem, my dream would be a message and reload button in the middle. Perhaps too a link to, or just a reminder, to use the PackageDetailsURL to open the URL with README and everything else the feed has for the package.

If the package doesn't have a README, I'd think we would state that (without a retry button)

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I think my (possibly unrealistic) _ideal_ is README should match whatever version is selected. That's how nuget.org works as well when selecting older package versions (you also get the older README). 

If that's too costly, we could instead _always_ show Latest README by default, with some declaration that "we're showing latest README" and a way to load the older, version-specific one.

The reason I'm suggesting we provide options for older and Latest is that I see a use case for both. A newer README may have relevant changes that affect how to use even older versions of the package. At the same time, the older README (historical, yet point-in-time with the package itself) may be more specific to the older APIs and expectations of that older package. I'm not sure if there's a "standard"/"best-practice" for editing your package's README as you update your packages. That knowledge would also help answer this one way or another, but I vote for "latest by default, with flexibility" as a compromise.
 
A reminder of our behavior across tabs:
- Browse tab defaults to Latest Version. Therefore, latest README will be most common.
- Installed tab defaults to Installed Version.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,88 @@
...
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
+    * Current suggestion is to focus on ReadMe.md. Once that's complete see how ReadMe.txt would appear when rendered. 
+    * How many packages have a ReadMe.txt vs ReadMe.md?
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
+    - [ ] Display the tab with a message saying there is no ReadMe defined.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: 👍

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: > Either way I would think the error should be visible in the PMUI since putting it in the Output pane would be confusing especially if it's not already open.

👍 Agreed. In another comment, I mention an error message/retry button.

donnie-msft (commented):
Great progress - there's a lot of cool stuff coming with this work!

btw, I'd probably edit the title of the PR to mention README.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I just posted to the server channel to see if they have an API for the ReadMe. I didn't see anything documented though. The [ReadMeUrl] (https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#catalog-entry) in the package metadata is a link to ReadMe section in nuget.org. 

Also for private feeds it's possible that these endpoints aren't implemented fully, for example you search for newtonsoft.json the ReadMe link is missing in our private feeds, but it's there if you use nuget.org as a source. So I guess another implementation detail would be if should we rely on the private feed to source the readme?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+    - Can use Nuget.Packaging to get ReadMe location from nuspec. 
+1. What do we want the UX to be when an exception or error occurs while reading a ReadMe file? 
+1. Do we want the ReadMe to update whenever a new version is selected for the current package?
+    Yes.
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+Currently there is no way to access the Readme.md without downloading the nupgk. The server team has the ability to return us the MD but this isn't documented. We could make a change to the NuGet API to include RawReadmeUrl which a link to download the MarkDown.
```
Comment (kartheekp-ms) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]:
```suggestion
Currently there is no way to access the Readme.md without downloading the nupkg. The server team has the ability to return us the MD but this isn't documented. We could make a change to the NuGet API to include RawReadmeUrl which a link to download the MarkDown.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
+    * Current suggestion is to focus on ReadMe.md. Once that's complete see how ReadMe.txt would appear when rendered. 
+    * How many packages have a ReadMe.txt vs ReadMe.md?
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: I'm a fan of hiding it. Keeping it there to say there's no readme is unusual.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+![Alt text](../../meta/resources/ReadMePMUI/PullRequestExperience.png) 
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
+    * Current suggestion is to focus on ReadMe.md. Once that's complete see how ReadMe.txt would appear when rendered.
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: I think we show the text file if it's there and not render.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
+    * Current suggestion is to focus on ReadMe.md. Once that's complete see how ReadMe.txt would appear when rendered. 
+    * How many packages have a ReadMe.txt vs ReadMe.md?
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
+    - [ ] Display the tab with a message saying there is no ReadMe defined.
+1. Where do we get the ReadMe from when it's not on the disk?
+    - Does the current API return the ReadMe information when searching for a package?
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Putting my feedback from offline here: 

IMO, we should honor a couple of design principles: 

- No explorative downloads. So no guessing whether there's a readme. The server should tell us if it's there. 
- Some way to create get or create a url to get the readme file. Matches license/icon and avoids attempting partial downloads which raise a bunch of questions.
- 
We need this to be a documented behavior so that other servers can implement it.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+1. What do we show if there is no ReadMe defined?
+    - [ ] Hide the tab from the the details pane
+    - [ ] Display the tab with a message saying there is no ReadMe defined.
+1. Where do we get the ReadMe from when it's not on the disk?
+    - Does the current API return the ReadMe information when searching for a package? 
+      - Yes, but not documented
+    https://api.nuget.org/v3-flatcontainer/{PACKAGE ID}/{PACKAGE VERSION}/readme
+    
+        Example: https://api.nuget.org/v3-flatcontainer/newtonsoft.json/13.0.2/readme
+1. Are we okay with pulling ReadMe exclusively from NuGet.org?
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: I'd pull readme from wherever it's advertised by the server. 
We don't special case in our code right and we shouldn't add anything.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+    - Yes
+1. Where are the ReadMe files saved in a package?
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: I'm guessing this is no longer an open equestion.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: If an exception or error occurs, we should say an exception or error occured. A way to refresh it would be nice, but not critical for MVP.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Readme's are per id/version, so let's just follow that.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (kartheekp-ms) [inline accepted/2024/[TS]PMUI-Readme-rendering.md:1]: FYI. I like Andy's suggestion from another spec, which is to have one sentence per line in the markdown source. https://github.com/NuGet/Home/pull/13242/files#r1488885933

The following examples render as a paragraph, but it would be hard to add a comment in example 1 because the line has 2 sentences, whereas example 2 is easy to review.

**Example 1:**
```
Great spec, Jonatan. I felt few comments to being with.
```

**Example 2**
```
Great spec, Jonatan.
I felt few comments to being with.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+
+Currently WebView2 controls always render ontop of other controls in the view. [Secnario 25254665](https://microsoft.visualstudio.com/Edge/_workitems/edit/25254665). PM UI needs to be updated to ensure items don't scroll off screen.
+
+## Rationale and alternatives
+By using an existing control we maintain consistency throughout the IDE and can rely on the owner to fix any bugs with the control.
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
```
Comment (joelverhagen) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: NuGet.org readme rendering is prior art, for our NuGet users. They may expect an experience somewhat familiar on NuGet.org.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (joelverhagen) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Different readers may interpret this differently. The search endpoint (https://learn.microsoft.com/en-us/nuget/api/search-query-service-resource) which populates the right pane in the Browse tab does not return any indication of readme.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+The IMarkdownPreview is currently being used when creating a new pull request inside of Visual Studio.  
+![Alt text](../../meta/resources/ReadMePMUI/PullRequestExperience.png) 
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+1. Will we show readme in the updates tab?
+1. What do we show if the package has a ReadMe.txt instead of md?
```
Comment (joelverhagen) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: IMO this is not a question of txt vs. md file extension. It's just about whether the package author put `<readme>` in the nuspec (or associated csproj property resulting in nuspec property at pack time)

https://learn.microsoft.com/en-us/nuget/reference/nuspec#readme

If they happen to have a md in their package but no readme nuspec property, it should not be shown. If the nuspec readme property points to a txt file, it should be treated as markdown since we spec it as only supporting markdown.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Prior Art
+The IMarkdownPreview is currently being used when creating a new pull request inside of Visual Studio.  
+![Alt text](../../meta/resources/ReadMePMUI/PullRequestExperience.png) 
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
```
Comment (joelverhagen) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Will non-embedded readmes be supported? As mentioned on the teams threads, a package author can provide a readme outside of the .nupkg and this is visible on NuGet.org. It's also mutable. I've seen indication from customers in the past that this is a desirable feature as an alternative to embedded readmes since it allows fixups later and, for these users, readme authoring is a step sometime after pack is done in their desired workflow. I think @clairernovotny may have thoughts+data on this. Despite my feeling, a large number of readmes are non-embedded:

Counts on nuget.org.
```kusto
NiPackageReadmes
| where ResultType !in ("Deleted", "None")
| summarize count() by ResultType
```

|ResultType|count_|
|---|---|
|Embedded|748714|
|Legacy|248580|

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: If we're focusing on the files that are currently downloaded in the MVP it's likely that when they switch versions the new version won't have be downloaded and the ReadMe will disappear.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Where is no endpoint for getting the ReadMe so we're going to focus on rendering the ReadMe for downloaded packages only.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Yea, I think the plan would be to update the search endpoint to include the url to download the md directly similar to LicenseUrl and IconUrl.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: That matches what I'm doing in code so I like it 😊

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Is there only one version of the non-embedded files ReadMe for the package? 

As it stands this plan was only for Embedded ReadMe, seeing as we're going to focus on packages that are already downloaded. We would prefer to have a URL provided by server allowing us to download the MD for packages which aren't downloaded. We don't want to download the package temporarily due to perf concerns. That URL could return the non-embedded ReadMe and Client wouldn't know the difference.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: If a package isn't downloaded, wouldn't we just go fetch the data from the servers anyways? If that's the case, then we'd be able to display it correct?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: image
How does this look?

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Light Mode
image

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: And High Contrast bc I'm already here
image

Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: That could work. I'd lean towards my suggestion of a control with icon/text and possibly reload button. As discussed offline, get input from the UX Board on it.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (joelverhagen) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Non-embedded readme and embedded readme are both per-version. 

Server could theoretically supply the non-embedded README in the URL, and yes client wouldn't know the difference. The only problem is that off of the Browse tab, like on the installed tab there is not necessarily server metadata available to get the non-embedded readme. But this could maybe be resolved by fetching it from the package source in a new metadata fetching flow. Not sure. There are options.

Comment (joelverhagen) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Take an analogy of embedded license and embedded icon. The legacy counterparts have the content available externally but it still sort of works. The data is available somehow.

For old icon URL (not embedded icons), nuget.org caches the icon and surfaces them in storage and API endpoints. They look just like embedded icons if you look and JSON returned from our APIs.

For licenses, you can still click through to the license URL for such cases. In other words, graceful fallback.

It feels a little strange to not have the legacy readmes available. Maybe a link to the readme on NuGet.org (not downloading the markdown) as the graceful fallback is enough.

nkolev92:
Review Meeting:
@jgonz120 
Has all the feedback here been addressed? 
Ready for another review?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,117 @@
...
+markdownPreview.UpdateContentAsync(markDown ?? string.Empty, ScrollHint.None)
+
+//IMarkdownPreview.VisualElement contains the FrameworkElement to be passed to the view
+MarkdownPreviewControl = markdownPreview.VisualElement
+```
+##### Locating the ReadMe
+We will focus the MVP on loading the ReadMe only for packages which are already downloaded.
+
+Will use ZipArchive to locate the nuspec within the Nupkg.
+Then use NuspecReader to locate the ReadMe within the NupKg, and extract it using ZipArchive.
```
Comment (kartheekp-ms) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: README file is already extracted to the global packages folder, correct? I am trying to understand the reasoning behind using NuspecReader to locate the file within the nupkg. Please let me know if I am missing something.

image

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,116 @@
...
+    Updating the package metadata to include a link to the readme, and a direct download similar to the way icon is implemented.
+1. ~~Where are the README files saved in a package?~~
+    - Can use Nuget.Packaging to get README location from nuspec.
+1. What do we want the UX to be when an exception or error occurs while reading a README file? 
+1. ~~Do we want the README to update whenever a new version is selected for the current package?~~
+   - Yes.
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Is that tab selection of readme vs details going to be sticky?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,116 @@
...
+nuget.org currently renders the README and our users will expect them to look the same.
+Ex. https://www.nuget.org/packages/Newtonsoft.Json#README-body-tab
+
+## Unresolved Questions
+1. Will we show README in the updates tab?
+1. ~~What do we show if the package has a README.txt instead of md?~~
+    * The README is spec is written as only accepting MD.
+    So we will use the nuspec to determine where the README is and treat it as md, even if the file is actually txt. 
+1. ~~What do we show if there is no README defined?~~
+    * Show the README tab with a message saying there is no README for the selected pacakge/version.
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: When you first open the package details view and a package has no readme, what would be the default? I'm assuming details since it's more usable?

nkolev92 (commented):
LGTM 👏 

A couple of experience questions about the details and readme tabs.

nkolev92:
Can we update the title of this PR to something more representative?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Yup, right now it keeps the state between tabs and opening/closing the PM UI

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Right now the default is to whatever was used last. If a tab has never been selected it goes to the README tab.

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Should we go to the more useful tab for the package instead?

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,207 @@
...
+##### PackageBaseAddress/6.12.0
+
+A new version of the [package content](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource) resource type which will include a url definition for downloading the README.
+
+```json
+{
+    "version": "3.0.0",
+    "resources": [
+        ...,
+        {
```
Comment (nkolev92) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Why isn't this a template like `ReportAbuseUriTemplate`?
it seems to fit that better. 

Example:
```json
        {
            "@id": "https://www.nuget.org/packages/{id}/{version}/ReportAbuse",
            "@type": "ReportAbuseUriTemplate/3.0.0-rc",
            "comment": "URI template used by NuGet Client to construct Report Abuse URL for packages used by RC clients"
        }
```

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: The [package content](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource) resource describes itself an endpoint for fetching the package's contents. So, I think it makes sense to include it there. While I don't think we implement it in the code this endpoint can be used to download the nuspec and the nupkg.

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: This is a server feedback, but probably a good idea to update the resource comment to indicate that readme can be downloaded as such as well. 

Today, only the nupkg is suggested within the index.json. 

cc @joelverhagen

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -30,19 +30,104 @@ When no README is available we will display a messsage in the README section.
...
+            ],
+            ...
+        }
+    ],
+    ...
+}
+```
+##### PackageBaseAddress/6.12.0
 
 A new version of the [package content](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource) resource type which will include a url definition for downloading the README.
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Pedantic: This example below is not showing the URL to download the readme, it's showing a base URL, for which a URL has to be generated.  It also doesn't tell us what the generated URL will be.  See [the docs](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource) for examples how they explain to both service and client implementors what the expected URL will be.

Guessing from the registration example's rawReadMeUrl, I assume the proposed filename will be `readme`.  Consider the two documented resources both have file extensions, and embedded readmes only support markdown, should it be `readme.md`?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (joelverhagen) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: The comment is in NuGet.Services.Index. Whoever updates the service index with this protocol can update the comment.
I updated https://github.com/NuGet/NuGetGallery/issues/10116 with the suggestion.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I can update the doc, as for the URI. Users can pick whatever name they want for the README. If we set the api {@id}/{LOWER_ID}/{LOWER_VERSION}/readme.md it implies like are specifically looking for that file. As opposed to /readme which leaves it open to whatever readme is available for that package.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: @zivkan brought up a good point about this. Is the expectation that newer versions of an API are backwards compatible with all versions?

PackageBaseAddress/6.12.0 (We can download README)
PackageBaseAddress/6.15.0 (We can download Icons)

From the client perspective should we expect the endpoint provided by PackageBaseAddress/6.15.0 to also allow downloading the README?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (zivkan) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: yes, the doc needs to explain how server implementors tell client what filename to use (what pattern client will use to generate the URl that will be requested)

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (zivkan) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: pedantic: icon isn't a good example, because both search results and registration blobs both contain icon URL. So, there's no need for client to ever probe a generated endpoint based on resource version. Exactly how I wish raw readme URL was implement  😁 

But yes, the point is that if there's some feature in the future, will this design force servers to opt into readme probing if they want to implement the new feature?

I don't know enough about how `SourceRepository.GetResourceAsync` works

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I don't know if it's specified explicitly anywhere, but everything we have done so far, resources are only additive.

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Only additive meaning that if we got back a service index with only the `PackageBaseAddress/6.15.0` resource type we would expect it to comply with all the previous versions as well? @joelverhagen  is that how NuGet.org has been operating as well?

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Yes, I think we'd expect it to comply to previous versions. 

Doesn't mean we can't change it, but if we did, we'd need to better define the protocol and live with the chances that there might be bugs. 

Another thing to consider is that even if we did try to load a readme, we'd fail and it'd be something we'd handle, since readmes aren't mandatory, so in practice it might not even matter that  much.

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Yea but servers may not want to be exposing themselves to additional authenticated calls to an endpoint that they don't support. I think implementing something similar to the ReportAbuseTemplate would actually make more sense given the implications. 

```json
        {
            "@id": "https://apidev.nugettest.org/v3/flatcontainer/{LOWER_ID}/{LOWER_VERSION}/readme",
            "@type": "ReadmeUriTemplate/16.13",
            "comment": "URI template used by NuGet Client to construct a URL for downloading a package's README"
        }
```

@joelverhagen do you have thoughts on this from a NuGet.org perspective?

Comment (nkolev92) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: We should only support lowercase id/version, the same way flat container, and registration do.
We normalize to lowercase everywhere in product, for example the global packages folder is fully lowercase. 
We don't really want to add ambiguity in the protocol.

Comment (joelverhagen) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: **tldr;** `ReadmeUriTemplate` seems better at least from a tactical perspective, great idea @nkolev92!

Readme URI template seems just fine too. Regarding the conversation about whether resources are addition, when to do breaking changes, etc:

There is no rule that says resource `Foo/Y` has all of the capabilities of `Foo/X` and more (given Y > X). Remember, resource types are opaque strings and the version part of the resource type string is for display purposes. No version parsing is done. In fact any such rule would be problematic for when we DO want to make breaking changes. 

Expanding this concept out a bit. In a parallel universe we could call have called `PackageBaseAddress/3.0.0` as `PackageBaseAddressWithNuspec` and then `PackageBaseAddress/6.12.0` as `PackageBaseAddressWithReadme`. Then later when we add icons we could call it `PackageBaseAddressWithIcons`. (setting aside the valid point that @zivkan made about icons already being solved -- let's use it for illustrative purposes)

Generally speaking, each resource version should describe what it does and what relationship it has with the other versions. It's certainly most simple to always be additive, but it's not required. Take an example `RegistrationsBaseUrl/3.4.0`: this had a breaking change of introducing "gzipped responses". Does it have all of the capabilities of the previous (lower) version `RegistrationsBaseUrl/3.0.0-rc`? No, the capability of returning responses without transfer encoding was lost (non-additive). Note that `RegistrationsBaseUrl/3.6.0` explicitly states that it also has `gzip` responses: https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#registrationsbaseurl360. Certainly docs could be improved to explicitly state SemVer 1.0.0 versions are included also but docs are easy to improve over time.

If you don't like the nuance of transfer encoding and additive/breaking, then yes you could say all changes have been additive. But that's because we haven't removed scenarios from V3 before. And there is nothing in the docs or protocol that says it has to be this way.

Would `PackageBaseAddressWithIcons` imply there are readmes? Only if we want it to and document it as such. If we didn't want it to do that, we would say NuGet.org needs to keep `PackageBaseAddressWithReadme` in the service index when it adds `PackageBaseAddressWithIcons` to declare its continued support of readmes. But service indexes need to do this anyways! Because older versions are looking for `PackageBaseAddressWithReadme` and not `PackageBaseAddressWithIcons` and they should continue to work. For newer clients that know about icons and readmes, they would look for `PackageBaseAddressWithReadme` to conditionally fetch READMEs and `PackageBaseAddressWithIcons` to conditionally fetch icons. 

So when we want to release a new `PackageBaseAddress` after README, we would just say `PackageBaseAddress/6.15.0` does not declare support for READMEs. We could even say it is a breaking change but that perhaps too strong of language because `PackageBaseAddress/6.15.0` and `PackageBaseAddress/6.12.0` and co-exist with the same URL.

Anyways, that's the theory/design around service index.

The `ReadmeUriTemplate` resource seems great. I don't feel strongly one way or another from a protocol perspective. But from a tactical perspective it has a very good benefit: this could even provide China customers immediate support for readmes via the `globalcdn` DNS name (or a new one we make to allow Microsoft CDN to remain as primary), e.g.
`https://globalcdn.nuget.org/v3-flatcontainer/{LOWER_ID}/{LOWER_VERSION}/readme`. Perhaps an okay short-term solution until we get readmes replicated into China storage.

Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I like it, updated.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,208 @@
...
+When no README is available we will display a messsage in the README section.
+![alt text](../../meta/resources/ReadMePMUI/NoReadMeFound.png)
+
+##### README File Sources
+
+* RawReadmeFileUrl in the [package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource).
+* README direct download specified in the [package content](https://learn.microsoft.com/en-us/nuget/api/package-base-address-resource).
+* Downloaded nupkg.
+
+#### Nuget API
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: nit: I'd move this API stuff to the Technical section.

```suggestion
#### NuGet API
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Motivation 
+
+The README file is an essential source of information that can help customers understand what a package does.
+By rendering the README file directly in the PM UI, customers can easily access the information they need to determine if a package will be helpful to them while searching, browsing, and managing their projects’ packages.
+This enhancement streamlines the user experience and may encourage package authors to create comprehensive README documentation when customers will be more likely to rely on it for useful information about the package.
+
+## Explanation
+### Functional explanation
+#### PM UI
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]:
```suggestion
#### Tabs introduced to the Details pane

(I know from your feature branch that there's new tabs introduced to the details pane... consider highlighting that major UI change here independent of the README discussion.)

#### README tab behavior
(In addition to what you already wrote, maybe you could comment on how you'll dynamically add the README tab. I think that's an interesting functionality that you're introducing.)
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,208 @@
...
+This UX will be displayed for both the Browse and installed tabs.
+It will also be displayed for both the solution level and project level package managers.
+![Alt text](https://github.com/NuGet/Home/assets/89422562/81b24877-f12f-4783-905c-4a155d3c7693)
+
+When no README is available we will display a messsage in the README section.
+![alt text](../../meta/resources/ReadMePMUI/NoReadMeFound.png)
+
+##### README File Sources
+
+* RawReadmeFileUrl in the [package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource).
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: In your feature branch PR,[ you mentioned](https://github.com/NuGet/NuGet.Client/pull/6029#discussion_r1779153245) that reading the README should be coming from the remote always. I'm curious why the GPF won't work. That's a great thing to specify in this spec.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,208 @@
...
+//IMarkdownPreview.VisualElement contains the FrameworkElement to be passed to the view
+MarkdownPreviewControl = markdownPreview.VisualElement
+```
+#### Locating the README
+
+Create a new implementation of the `INuGetResource` interface, `ReadMeDownloadResource`.
+This will only be available for sources which have implemented the new `PackageBaseAddress` resource type.
+
+Update `IPackageSearchMetadata` to include the `RawReadmeUrl` field. 
+`LocalPackageSearchMetadata` will put the location of the readme on the disk in this field.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: This is where I'm wondering if the README.md file exists on disk, should that just be used rather than always downloading the ReadmeURL?

Will there be any caching so at least during a session, the same URL isn't downloaded multiple times?

donnie-msft (commented):
I forgot to submit my review last week - sorry about the delay!

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I added additional details lemme know what you think

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I think a cache is a good idea, I added it as a future possibility for now.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I updated it to better reflect what the functionality should be. I think using the GPF is fine for the browse tab, as long as there is a remote source available. That way the user doesn't experience confusion as to why the readme tab is conditionally available while exploring the browse tab.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+
+##### README Tab
+This tab is rendered if a remote source available to download the readme or the Installed tab is selected and the selected version is found in the global packages folder.
+This is to ensure that users using the Browse tab will not see the README tab if they do not have a remote source that allows README downloads.
+
+If the README tab is rendered and we do not find a README then a message is rendered encouraging the user to contact the package author to upload a README.
+![alt text](../../meta/resources/ReadMePMUI/NoReadMeFound.png)
+
+##### Package Details
+This tab is always rendered and contains the package details information along with the Vulnerabilty and Depreciation information.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Should the vulnerability warning area be moved above the tabs so that it's always visible? 

I'm thinking that it's not good to "hide" the warning down in the tab, so that customers who use the README tab (especially if they use it primarily, or exclusively) will never see this vulnerability info:

image

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+It will also be displayed for both the solution level and project level package managers.
+
+##### README Tab
+This tab is rendered if a remote source available to download the readme or the Installed tab is selected and the selected version is found in the global packages folder.
+This is to ensure that users using the Browse tab will not see the README tab if they do not have a remote source that allows README downloads.
+
+If the README tab is rendered and we do not find a README then a message is rendered encouraging the user to contact the package author to upload a README.
+![alt text](../../meta/resources/ReadMePMUI/NoReadMeFound.png)
+
+##### Package Details
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Please add a screenshot of how this scenario will look with the single tab. It would be valuable since there's currently a bug about how when just 1 tab is visible, it's aligned differently than when there's 2 tabs visible.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Prior Art
+The IMarkdownPreview is currently being used when creating a new pull request inside of Visual Studio.
+![Alt text](../../meta/resources/READMEPMUI/PullRequestExperience.png) 
+
+nuget.org currently renders the README and our users will expect them to look the same.
+Ex. https://www.nuget.org/packages/Newtonsoft.Json#README-body-tab
+
+## Unresolved Questions
+1. ~~Will we show README in the updates tab?~~
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: nit: Not sure if crossing out questions is the best way to say it's answered

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+    Yes
+1. ~~What do we show if the package has a README.txt instead of md?~~
+    * The README is spec is written as only accepting MD.
+    So we will use the nuspec to determine where the README is and treat it as md, even if the file is actually txt.
+1. ~~What do we show if there is no README defined?~~
+    * Show the README tab with a message saying there is no README for the selected pacakge/version.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]:
```suggestion
    * Show the README tab with a message saying there is no README for the selected package/version.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+1. ~~Where do we get the README from when it's not on the disk?~~
+    - Will document two ways for servers to provide a the readme.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]:
```suggestion
    - Will document two ways for servers to provide the readme.
```

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+Investigate ways for encouraging package owners to publish READMEs through this experience. 
+
+Implement the option for users to opt out of rendering all images from README, similar to outlook with external images.
+
+Update `INuGetPackageFileService` to allow for authenticated feeds.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: This will help with package icons as well!

meta/resources/ReadMePMUI/NoReadMeFound.png
donnie-msft, line 1: low priority: I changed the text a bit so could update the screenshot. Not too important though

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+    - Can use Nuget.Packaging to get README location from nuspec.
+1. ~~What do we want the UX to be when an exception or error occurs while reading a README file?~~
+    - Inform the users that an exception occurred. If time permits allow for a retry.
+1. ~~Do we want the README to update whenever a new version is selected for the current package?~~
+   - Yes.
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: These sound great! Is there a roadmap/epic/etc tracking these?

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md:1]: What does [TS] prefix mean on this file? It affects how filenames are sorted.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Moving it outside would make the readme difficult to read. Some options that come to mind

1. Add the warning icon to the Package Details Tab when there is a Vulnerability/Deprecation. 
2. Move both of these outside and make them collapsable. 
3. Default to the Package Details tab if there are any vulnerabilities or deprecations.
4. Move Vulnerabilities and Deprecations to their own tabs.

I think the 3rd option would be the easiest option.

Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I prefer 1&2, maybe even 4. 
3 Is my least preference as it essentially stomps on the user's choice for default tab (or last used tab).

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,216 @@
...
+                },
+                ...
+            ],
+            ...
+        }
+    ],
+    ...
+}
+```
+##### ReadmeUriTemplate Resource Type
```
Comment (loic-sharma) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Did we consider instead introducing a new version of the `PackageBaseAddress` resource that has this well known URI? I don't feel strongly about this, just a thought

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md:1]: I added it to mean TechSpec 😅

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: We originally planned for a new version of `PackageBaseAddress` but instead opted for `ReadmeUriTemplateResource`. 

There were questions about version, if we create PackageBaseAddress/6.13.0, then in the future implement a new one, PackageBaseAddress/6.20.0 that added another file, what that mean that we could use 6.20.0 to download the readme as well? Would a private feed have to implement a readme endpoint even if they don't support readme? 

Additionally, by setting using a different resource type feeds can specify a URI schema that better suites their implementation.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jebriede) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: This is a great point about keeping vulnerability info visible.
@jgonz120 those are some good options to handle the scenario. If a vulnerability is present, I could see that as justifiable reason for showing the details tab with the vulnerability info at the top. If we were to consider option 4, it should also default to the tab w/ vulnerability info if there is a vulnerability present. For good measure, let's run these options by the UX board.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+            "@type": "RegistrationsBaseUrl/6.13.0"
+        },
+        ...
+    ]
+}
+```
+###### Example Response
+```json
+{
+    "@id": "https://apidev.nugettest.org/v3/registration5-gz-semver2/newtonsoft.json/index.json",
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: response of what? The section immediately before this talks about the service index.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+
+* ReadmeFileUrl in the [package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource).
+* README direct download specified in a new resource `ReadmeUriTemplate`.
+* Downloaded nupkg.
+
+#### NuGet API
+##### New RegistrationsBaseUrl version
+
+A new version of the [package metadata](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource) resource type will be documented which will include the field **ReadmeFileUrl**.
+This will be a link to download the README and will only be filled if a readme is available to download.
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: Oops, I had a brain fart

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+                        "dependencyGroups": [ ... ],
+                        "description": "Json.NET is a popular high-performance JSON framework for .NET",
+                        "iconUrl": "https://apidev.nugettest.org/v3-flatcontainer/newtonsoft.json/13.0.3/icon",
+                        "id": "Newtonsoft.Json",
+                        "language": "",
+                        "licenseExpression": "MIT",
+                        "licenseUrl": "https://dev.nugettest.org/packages/Newtonsoft.Json/13.0.3/license",
+                        "readmeUrl": "https://dev.nugettest.org/packages/Newtonsoft.Json/13.0.3#show-readme-container",
+                        //New field
+                        "readmeFileUrl": "https://apidev.nugettest.org/v3/flatcontainer/newtonsoft.json/13.0.3/readme",
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: This new field is described/introduced later in the document than this example. I think that's confusing, as someone who has been reading this document top to bottom.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+        ...
+    ]
+}
+```
+In code we would take the id provided and replace the {lower_id} and {lower_version} in the strings with the package we are trying to get the README for.
+
+### Technical explanation
+#### Rendering Markdown
+
+We will use the [Microsoft.VisualStudio.Markdown.Platform](https://dev.azure.com/azure-public/vside/_artifacts/feed/vs-impl/NuGet/Microsoft.VisualStudio.Markdown.Platform/overview/18.0.39-preview-g659b28ccd8) package to render the README in the IDE. This will allow us to leverage a centralized tool for rendering markdown in the IDE.
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: one sentence per line would be great to keep lines shorter

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+}
+```
+In code we would take the id provided and replace the {lower_id} and {lower_version} in the strings with the package we are trying to get the README for.
+
+### Technical explanation
+#### Rendering Markdown
+
+We will use the [Microsoft.VisualStudio.Markdown.Platform](https://dev.azure.com/azure-public/vside/_artifacts/feed/vs-impl/NuGet/Microsoft.VisualStudio.Markdown.Platform/overview/18.0.39-preview-g659b28ccd8) package to render the README in the IDE. This will allow us to leverage a centralized tool for rendering markdown in the IDE. 
+
+A new instance of the [IMarkdownPreview](https://devdiv.visualstudio.com/DevDiv/_git/VS-Platform?path=%2Fsrc%2FProductivity%2FMarkdownLanguageService%2FImpl%2FMarkdown.Platform%2FPreview%2FIMarkdownPreview.cs) can be created using [PreviewBuilder](https://devdiv.visualstudio.com/DevDiv/_git/VS-Platform?path=/src/Productivity/MarkdownLanguageService/Impl/Markdown.Platform/Preview/PreviewBuilder.cs).
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: This feels like implementation detail that doesn't need to be a specification document. Plus our specs are publicly available, so links to closed source doesn't feel appropriate to me.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+So when an upgrade is made we may have to change how we use the control.
+
+WebView2 controls always render ontop of other controls in the view.
+[Secnario 25254665](https://microsoft.visualstudio.com/Edge/_workitems/edit/25254665).
+PM UI needs to be updated to ensure items don't scroll off screen.
+
+`INuGetPackageFileService` in it's current implementation cannot download files from feeds that require authentication.
+
+## Rationale and alternatives
+By using an existing control we maintain consistency throughout the IDE and can rely on the owner to fix any bugs with the control.
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: maintain consistency with what?

VS extension readmes might allow different markdown to what nuget.org does for package readmes, so it seems to me that consistency with nuget.org's readme rendering is more important than consistency with readmes of some unrelated component.

I feel like the advantage of using the existing component is quicker development, and hopefully less maintainance.

The drawbacks are inconsistencies if the markdown dialect is different to what nuget.org already supports. And if the team who own the component aren't open to changing the markdown dialect for nuget, then it reduces NuGet's ability to make improvements to the supported markdown dialect.

I feel like this description of "alternatives and alternates" isn't giving a fair or accurate pro-con list.  In fact, reading this spec, it's not clear to me if the drawbacks were even considered.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -0,0 +1,213 @@
...
+    - Show the README tab with a message saying there is no README for the selected package/version.
+1. Where do we get the README from when it's not on the disk?
+    - Will document two ways for servers to provide the readme.
+    Updating the package metadata to include a link to the readme, and a direct download similar to the way nuspec and nupkg are implemented.
+1. Where are the README files saved in a package?
+    - The nuspec defines where to locate the README.
+1. What do we want the UX to be when an exception or error occurs while reading a README file?
+    - Inform the users that an exception occurred. If time permits allow for a retry.
+1. Do we want the README to update whenever a new version is selected for the current package?
+   - Yes.
```
Comment (zivkan) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: This might be a personal preference, but since these "unresolved questions" have answers now, I'd prefer the answers to be integrated throughout the rest of the document and removed from here.  Honestly, I'm not a fan of this "unresolved questions" section.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: This section is all still under the "New RegistrationsBaseUrl version" section, though I think all these subheadings are getting confusing.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: We do if we want to give our partners the flexibility to choose a solution that works best for them.  We don't have a commitment from anyone to implement the new RegistrationsBaseUrl  version.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I filled it out a bit. 

When I said consistency, I meant how the Markdown is rendered, a markdown heading should appear the same throughout the IDE regardless of which experience you're in. You're right about the dialect though, I added that bit. I don't think it's that big of a concern though, unless NuGet.org decides to move to a completely new dialect... which would be a bunch of work for us even if we owned the control outright.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Agreed.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I removed the links and just skimmed it down to say we're using the control

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I moved it around a bit, hopefully its less confusing? Let me know.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: No, should I create those now? 😅

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (jgonz120) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: Is it okay to resolve this as I follow up with the UI board? This feels more like an implementation detail, I can change the screenshot to not include the vulnerability information if that helps.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md:1]: Could you please remove it. It will always float to the top of the folder when sorting a directory by filenames.

accepted/2024/[TS]PMUI-Readme-rendering.md
Comment (donnie-msft) [reply accepted/2024/[TS]PMUI-Readme-rendering.md]: I feel it's also a design detail. When we show informative text around vulnerabilities, it's now moving from a permanent display to a tabular display. That said, yes, I believe it's acceptable to address it in a follow-up PR based on UX board feedback.

accepted/2024/[TS]PMUI-Readme-rendering.md
```diff
@@ -171,6 +171,8 @@ WebView2 controls always render ontop of other controls in the view.
 [Secnario 25254665](https://microsoft.visualstudio.com/Edge/_workitems/edit/25254665).
 PM UI needs to be updated to ensure items don't scroll off screen.
 
+`INuGetPackageFileService` in it's current implementation cannot download files from feeds that require authentication.
```
Comment (donnie-msft) [inline accepted/2024/[TS]PMUI-Readme-rendering.md]: nit: `its`

donnie-msft (approved):

Issue #12409, opened 2023-02-06
Solution-Level or Repo-Level Lock Files


yaakov-h:
### NuGet Product(s) Involved

Other/NA

### The Elevator Pitch

Managing multiple project lock files for a huge solution or a repo can be daunting. This was even called out in 2018 when lock files were introduced:

https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/#solution-or-repo-lock-file

Has there been any further thought or development on solution-level or repository-level lock files?

The link in those paragraphs doesn't contain any further information on centralised lock files, only on centralised versioning, and I can't find any related issues here on GitHub either.

### Additional Context and Details

This is particularly important to me as I don't want to have to go and update approx. 1000 csproj files when updating a core package such as a Roslyn Analyzers package, or Newtonsoft.Json, or similar.

This also gets much more difficult when multiple developers are trying to make changes (e.g. adding a ProjectReference which adds new transitive NuGet dependencies, which I believe requires updating the lock file?), and looks like it will quite quickly lead to merge conflict hell.

In many of our projects already use [Paket](http://github.com/fsprojects/Paket), which provides a single top-level lock file for all dependencies defined in a `paket.dependencies` file (analogous to `nuget.config` + `Directory.Packages.props`). This makes batch operations such as upgrading or downgrading a package across an entire repo a fairly easy and straightforward process. Changes to individual projects require no changes outside of their own project files (`.csproj` + `paket.references`) and as such enable many type of concurrent changes to be made without file merge conflicts.

Please 👍 or 👎 this comment to help us with the direction of this feature & leave as much feedback/questions/concerns as you'd like on this issue itself and we will get back to you shortly.

Thank You 🎉

heng-liu:
Hi @yaakov-h , we have introduced CPM(Central Package Management). Is this the feature you're looking for?  
Please refer to the following for more details:
https://devblogs.microsoft.com/nuget/introducing-central-package-management/
https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

yaakov-h:
@heng-liu Central Package Management seems to be completely orthogonal to lock files. I can use CPM with lock files, I can use CPM without lock files, I can use lock files without CPM, and I can use neither.

[...]

PR #13288, opened 2024-03-04
Nuget CPM Lock file Proposal (#12409)

CEbbinghaus:
After some Internal deliberation I have come up with a fairly straightforward design as to how a Central lock file could be implemented in Nuget. This addresses #12409 which is currently the 3rd highest rated issue.

[Rendered Proposal](https://github.com/CEbbinghaus/Nuget/blob/repo-level-lock-files/accepted/2024/repo-level-lock-file.md)

Appreciate any feedback <3


CEbbinghaus:
@dotnet-policy-service agree company="WiseTech Global"

DrNeil:
This would be **_super_** useful.

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
+﻿# ***Repository Level Lock File***
+
+- Author: [@CEbbinghaus](https://github.com/CEbbinghaus) 
+- Issue: [Nuget/Home#12409](https://github.com/NuGet/Home/issues/12409) 
+
+## Summary
+
+Applications consisting of multiple solutions (using assembly references between projects) Cannot control their transitive dependencies by using Nuget. The current lock file is on a per-project basis rather than an application/repository basis. Given a `Directory.packages.props` file, each project could be resolved with different versions of transitive dependencies, which a single central lock file would solve, by allowing consistency between package versions across multiple solutions.
```
yaakov-h:
```suggestion
Applications consisting of multiple solutions (using assembly references between projects) cannot control their transitive dependencies by using Nuget. The current lock file is on a per-project basis rather than an application/repository basis. Given a `Directory.packages.props` file, each project could resolve different versions of transitive dependencies, which a single central lock file would solve by allowing consistency between package versions across multiple solutions.
```

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Motivation 
+
+Monorepositories with multiple solutions utilizing assembly references that deploy all their code together as a single application are unable to use Nuget (with or without CPM) due to transitive version conflicts. This is due to nuget being able to resolve different versions of transitive dependencies for each individual build, as they occur in isolation from one another. A single central lock file allows all transitive dependencies to have locked versions defined at the root, thereby eliminating third-party dependency management software such as Paket.
```
yaakov-h:
```suggestion
Monorepositories with multiple solutions utilizing assembly references that deploy all their code together as a single application are unable to use Nuget (with or without CPM) due to transitive version conflicts. This is due to NuGet being able to resolve different versions of transitive dependencies for each individual build, as they occur in isolation from one another. A single central lock file allows all transitive dependencies to have locked versions defined at the root, thereby eliminating third-party dependency management software such as Paket.
```

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+We can draw many parallels between Paket & Nuget (with a central lock file) function. For one, we define all the primary packages & versions at the root. Projects define which specific packages they depend on within the relative projects. Where paket would use groups to include dependencies at different versions, NuGet uses the `VersionOverride` property within the csproj. This is the primary difference in how locks are handled for version differences since Paket utilizes groups and writes the locked package versions into the `paket.lock` file. Nuget (with central locking) writes all the pinned versions into the local csproj lock file.
```
yaakov-h:
```suggestion
We can draw many parallels between Paket & NuGet (with a central lock file) function. For one, we define all the primary packages & versions at the root. Projects define which specific packages they depend on within the relative projects. Where paket would use groups to include dependencies at different versions, NuGet uses the `VersionOverride` property within the csproj. This is the primary difference in how locks are handled for version differences since Paket utilizes groups and writes the locked package versions into the `paket.lock` file. Nuget (with central locking) writes all the pinned versions into the local csproj lock file.
```

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+NodeJS has the concept of [Workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces#skip-to-content) which act more like Solutions since NodeJs only has the concept of Projects. However a Workspace does allow for all packages to be resolved to the same version, ensuring no version conflicts between different projects within the same workspace.
```
yaakov-h:
```suggestion
NodeJS has the concept of [Workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces#skip-to-content) which act more like Solutions since NodeJS only has the concept of Projects. However a Workspace does allow for all packages to be resolved to the same version, ensuring no version conflicts between different projects within the same workspace. 
```

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+NodeJS has the concept of [Workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces#skip-to-content) which act more like Solutions since NodeJs only has the concept of Projects. However a Workspace does allow for all packages to be resolved to the same version, ensuring no version conflicts between different projects within the same workspace. 
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+* How would the `Directory.packages.props` dependencies be separated out from the project dependencies to generate the lock file? 
+* How does the publish command handle this as it doesn't have full context of all packages. (Could be solved by the restore always running with --locked-mode)
```
yaakov-h, line 78: Why doesn't `dotnet publish` get the full context in a universe where `dotnet restore` / `dotnet build` would?

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+* How will a project deleted from the repository get deleted from the lock file.
+* How will Lock files be generated with nested `Directoy.packages.props` files.
```
yaakov-h:
```suggestion
* How will Lock files be generated with nested `Directory.packages.props` files?
```

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Future Possibilities
+<!-- What future possibilities can you think of that this proposal would help with? -->
+Using the `TargetFrameworks` property within the `Directory.packages.props` file could narrow down the frameworks resolved for the provided packages. This can simplify the lock file and reduce some unnecessary dependency resolution.
```
yaakov-h: This would also reduce noise in the file which could reduce unnecessary merge conflicts.

accepted/2024/repo-level-lock-file.md
CEbbinghaus: [Source](https://github.com/NuGet/Home/issues/12409#issuecomment-1849543218)
> Something else to consider is that dotnet publish -f tfm -r some_rid also does an implicit restore, setting MSBuild global properties, preventing NuGet from getting the "raw" project information. Although I think this already causes problems with the existing lock file feature. So it's best if a new design can either learn from the previous experience, or at least have the same mitigations (always use --no-restore with publish? it also means that customers must explicitly set their RIDs in the project file) before restore.

I am honestly not too sure what that is about as I haven't had much of a problem with publish & Lock files, But there seems to be an issue

accepted/2024/repo-level-lock-file.md
CEbbinghaus: Amended the section pointing out this additional benefit

CEbbinghaus:
What can we do to further progress this proposal?

JonDouglas:
@CEbbinghaus Thank you kindly for starting a discussion on this proposal! Give me some time to look through it and get the team together to review it. I didn't see the notification earlier as I've been terribly busy with some other things. I apologize in advance for the ~2 month delay.

CEbbinghaus:
I appreciate you taking a look at it now. Please do let us know if anything needs addressing

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+<!-- Explain the proposal as if it were already implemented, and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real-life examples, and low-fidelity mockups or pseudocode to show how this proposal would look. -->
+
+The `Directory.packages.props` now uses the`<NuGetLockFilePath>[path]</NuGetLockFilePath>` property within a `<PropertyGroup>` to specify its lock file path. It becomes the source of truth for all builds. Project level lock files are no longer required unless a package version is overridden.
+
+```xml
+<Project>
+    <PropertyGroup>
+        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+        <NuGetLockFilePath>lockfile.json</NuGetLockFilePath>
```
kasperk81, line 29: i love JSON as much as the next guy. but if Directory.Packages.props, NuGet.props and most of the files are XML based, wouldn't it be better to not mix stacks? historically lock files use .json format in various programming platforms, so that might be one defense for it (but not a very compelling one).

yaakov-h: The existing related files are already JSON - `packages.lock.json` and `project.assets.json`.

CEbbinghaus: Personally my preference is also xml but the nuget/dotnet team have already settled on a json based lock file and this change is supposed to be as small and simple as possible.

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented, and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real-life examples, and low-fidelity mockups or pseudocode to show how this proposal would look. -->
+
+The `Directory.packages.props` now uses the`<NuGetLockFilePath>[path]</NuGetLockFilePath>` property within a `<PropertyGroup>` to specify its lock file path. It becomes the source of truth for all builds. Project level lock files are no longer required unless a package version is overridden.
```
purepani, line 23: Is there a reason to make the lock file path configurable in the project? AFAIK, most package managers leave it unconfigured, and that way, the package manager wouldn't have to first read the configuration file first in order to install packages, only the lock file(which makes it easier to separate into steps).

CEbbinghaus: The package manager would need to read the Directory.packages.props file anyways as it is what enables CentralPackageManagement in the first place? ([relevant docs](https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#enabling-central-package-management))

We could get away with setting a property like `<RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>` to enable CPM Lock files and implicitly assume the `<NuGetLockFilePath>` to be `packages.lock.json` but we can also achieve the same with a single property and a check `$(NuGetLockFilePath) = ''`. At the end it will be up to the maintainers to give feedback which way they prefer but the lock file path will need to be configurable in any case.

purepani: > The package manager would need to read the Directory.packages.props
Yeah, but you can separate that into steps right? I'm imagining a use case of being able to separate it into layers:
You read the configuration file and can generate the lock file in one step, and then you can read the lock file and get the packages in a different step. The reason this could be useful is that you can write a completely separate program to read the lock file if you wanted to(one that assumes that you're using CPM), and you wouldn't have to read the configuration file to figure out where the lock file is. 

I personally can't think of a great reason to not have the lock file at the top level of a repo(or at some fixed path relative to the top level), and I can think of more reasons to avoid being able to configure it(Easier to make external tools to consume the lock files and process them separately, avoiding conflicts between, say, a git submodule with it's own lock file that's configured to generate this lock file, etc.).

Given that, as far as I'm aware, most package managers(mostly python and node ones) that I've used have a fixed path for their global lock files, I thought I'd at least bring it up.

yaakov-h: You would need a way to define the "top level" location. This allows for that.

Keep in mind that not every repo-project relationship is 1:1, particularly in a monorepo situation where you might want a small handful of "top levels" rather than only one.

purepani: It should just be in the same folder as the Directory.Packages.props file, the same way that the "top level" of a nodejs project is where the package.json is.

CEbbinghaus: Sorry i feel you might have misunderstood. This is exactly the case @purepani the `<NuGetLockFilePath>` specifies the lock file for the CPM rather than each individual project. This will be located relative to the `Directory.Packages.props` at the root of the project.

Now that you mention it we will likely need a property like `RestorePackagesWithLockFile` but specifically to enable CPM lock files but the lockfile will be located at the root NOT under individual projects

purepani: I see so this property is meant for each individual project rather than just the top level of the projects; that makes way more sense. 
The way pnpm solves this is by symlinking the dependencies into each workspace after doing a `pnpm install` at the top level. If you only wanted some dependencies, you could just install it for that group. In this case, it is just the dependencies and not the lock file, but you could certainly symlink the lock file if you wanted(though the only advantage of that approach would be that you wouldn't have to move to the top level every time you wanted to install a package). 

(note: I only mentioned pnpm specifically because I know how it works a bit better. Presumably other node package managers work similarly)

JonDouglas:
> I appreciate you taking a look at it now. Please do let us know if anything needs addressing

Just an update, we have various folks out and were busy with our recent conference. I will try to get this scheduled once people are back from summer vacations.

eclairevoyant:
bad bot, still important

CEbbinghaus:
can we add a nostale tag?

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+- Issue: [Nuget/Home#12409](https://github.com/NuGet/Home/issues/12409) 
+
+## Summary
+
+Applications consisting of multiple solutions (using assembly references between projects) cannot control their transitive dependencies by using Nuget. The current lock file is on a per-project basis rather than an application/repository basis. Given a `Directory.packages.props` file, each project could resolve different versions of transitive dependencies, which a single central lock file would solve by allowing consistency between package versions across multiple solutions.
+
+
+## Motivation 
+
+Monorepositories with multiple solutions utilizing assembly references that deploy all their code together as a single application are unable to use Nuget (with or without CPM) due to transitive version conflicts. This is due to NuGet being able to resolve different versions of transitive dependencies for each individual build, as they occur in isolation from one another. A single central lock file allows all transitive dependencies to have locked versions defined at the root, thereby eliminating third-party dependency management software such as Paket.
```
nkolev92, line 13: I think something worth mentioning here is the concept of transitive pinning: https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#transitive-pinning, which allows you to control the transitive dependencies by effectively elevating them to top level.

accepted/2024/repo-level-lock-file.md
nkolev92, line 23: There's a technical complexity here since a property value doesn't get path resolved, so knowing where it comes would be required. 
The simpler approach might be the already mentioned just making the file live next to the directory.packages.props.

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+        <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+        <NuGetLockFilePath>lockfile.json</NuGetLockFilePath>
+    </PropertyGroup>
+    <ItemGroup>
+        <PackageVersion Include="Stuff" Version="1.0.0" />
+    </ItemGroup>
+</Project>
+```
+
+This lock file is generated during any restore operation during the evaluation of the `Directory.packages.props` file. This allows the central lock file to be generated by any project as all of the dependencies are isolated.
```
nkolev92, line 37: The biggest challenges delivering this feature are the preexisting msbuild paradigms and we're not really tackling that here. 

We need to dive deeper into the technical complexities to understand the feasibility of simply having a lock file.

How would the scope of the msbuild based restore operation be defined? Currently restore is either project based or solution based. Projects don't know about all other projects. 

Here's a practical example: 

You have 2 projects and 3 packages.

Project 1:
```xml
<PackageReference Include="A" />
```

Project 2:
```
<PackageReference Include="B" />
```

DPP
```xml
<PackageVersion Include="A" Version="1.0.0" />
<PackageVersion Include="B" Version="1.0.0" />
```

A 1.0.0 -> C 1.0.0
B 1.0.0 -> C 2.0.0

Without pinning:

- Project 1: A 1.0.0 and C 1.0.0
- Project 2: B 1.0.0 and C 2.0.0

These projects aren't necessarily correlated so the only way for them to know that they're conflicting with C is to either: 
- Consider what each project references each restore
- Enforce version declaration for every package version encountered. 

Today, a project only knows about itself and it's closure, so this would be a departure in how restore/build work, one that'd come with some perf considerations. 

The 2nd idea could theoretically be solved with transitive pinning. 

You can declare `C 2.0.0` in DPP, enable transitive pinning and then the versions would be resolved would be equivalent. 
You'd basically need to exhaustively declare *all* versions.

Within the current toolset, The central lock file as proposed is really a more algorithmically advanced transitive pinning that gets applied across *all* projects. 

Another challenge is package creation for any projects that have a central lock file. 

NuGet works really hard to make projects and the package versions of them behave as closely as possible. 
What this means is that every PackageReference you have becomes a dependency in a generated package.
To avoid runtime issues, transitive pinning elevates those transitively pinned packages to top level dependencies. The practical example is Project 1 from above, where it could unwittingly take a dependency on C 2.0.0, and then if it gets published as a package with only its direct dependencies lead to runtime failures later. 
I understand that's not really a relevant scenario here, but it is a concept that the design needs to account for.

accepted/2024/repo-level-lock-file.md
```diff
@@ -0,0 +1,84 @@
...
+
+NodeJS has the concept of [Workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces#skip-to-content) which act more like Solutions since NodeJS only has the concept of Projects. However a Workspace does allow for all packages to be resolved to the same version, ensuring no version conflicts between different projects within the same workspace. 
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
```
nkolev92, line 75: We need to add the mechanics of version resolution in the unresolved questions.

nkolev92 (commented):
Thanks for starting the write-up. 
I've left some comments about some technical/functional challenges we'd need to figure out before taking this further.

accepted/2024/repo-level-lock-file.md
yaakov-h: Transitive pinning only helps when you recognise that you've gotten into this situation and want to do something about it. It doesn't prevent you from getting into a conflict in the first place unless you're very eager and go and manually specify every individual transitive dependency and keep track of them when updating packages.

accepted/2024/repo-level-lock-file.md
yaakov-h: > the central lock file as proposed is really a more algorithmically advanced transitive pinning that gets applied across all projects

Isn't that exactly what a lock file does, by specifying the entire dependency graph and the versions used of each dependency?

> The practical example is Project 1 from above, where it could unwittingly take a dependency on C 2.0.0, and then if it gets published as a package with only its direct dependencies lead to runtime failures later.

I assume we would need to keep track of which dependencies are direct and which are transitive, and `dotnet pack` would only look at direct dependencies when building a package. A transitive dependency can always be elevated to a direct dependency.

accepted/2024/repo-level-lock-file.md
nkolev92: That's correct. It's basically exhausitve transitive pinning where all packages are pinned. 
The lock file you're asking for would effectively be that, but potentially in a more automated fashion.

accepted/2024/repo-level-lock-file.md
nkolev92: > Isn't that exactly what a lock file does, by specifying the entire dependency graph and the versions used of each dependency?

There's a key distinction. All version resolution is project based. 
Lock files don't change the resolution. They skip it. 

The idea for central lock files makes the version resolution more involved and requires to account for transitive packages declared from all projects, not just the current one. 

You're not just asking for a file, you're asking for a completely new resolver. Beyond scoping, the resolver part is the most challenging technical part. 

> I assume we would need to keep track of which dependencies are direct and which are transitive, and dotnet pack would only look at direct dependencies when building a package. A transitive dependency can always be elevated to a direct dependency.

"dotnet pack would only look at direct dependencies when building a package" -> This doesn't work in all scenarios. 
Here's an example: 

Project 1 -> Package A 1.0.0 -> Package B 1.0.0
Regular resolution leads to A 1.0.0, B 1.0.0. 
When the project is packaged, it only needs A 1.0.0 as a dependency, since at minimum B 1.0.0 will be resolved due to A's dependency. 

Central lock file example could bump package B to 2.0.0. 
Project 1 code ends up coding against B 2.0.0. 
If the Project 1 package ends up declaring A 1.0.0, then when anyone installs that package generated from project 1, they'd get B 1.0.0, because no one knows about 2.0.0 in the Project 1 graph, it's there due to how other projects work. If we don't elevate B 2.0.0 to a direct dependency, then the consumer of Project 1's generated package would end up with runtime failures.

nkolev92:
Some general food for thought for the concept of centralized lock files. 

In general, you probably only need to consolidate the versions being used across your runnable projects. Your services, your tools etc, not across your libraries as well (and never downgrade ofc).
Just because you make your projects reference a specific version, that doesn't mean they'll all work. You could be referencing dependencies that reference different versions of a particular package, so consolidation is naturally going to happen in that case. 

There's of course the theory for handling breaking changes in dependencies, but that'd only work if you control all libraries that reference a package with a breaking change, in which case regular CPM would work just fine.

accepted/2024/repo-level-lock-file.md
nkolev92: Some general food for thought for the concept of centralized lock files. 

In general, you probably only need to consolidate the versions being used across your runnable projects. Your services, your tools etc, not across your libraries as well (and never downgrade ofc).
Just because you make your projects reference a specific version, that doesn't mean they'll all work. You could be referencing dependencies that reference different versions of a particular package, so consolidation is naturally going to happen in that case. 

There's of course the theory for handling breaking changes in dependencies, but that'd only work if you control all libraries that reference a package with a breaking change, in which case regular CPM would work just fine.

JonDouglas:
Hey all, adding a quick high-level comment on this proposal as we did a quick team review yesterday.

First, I just want to say, thank you so much for this proposal! The goal of introducing a central lock file concept seems well received based on the number of reactions in the short period of time since it was introduced here and between the linked issue.

Here are some of the challenges we're looking through:

- Feasibility of implementing central lock files, challenges of regenerating lock files, and a lack of a mechanism that represents the entire project tree.
- Merge conflicts, challenges of updating package versions w/ the central lock file concept.
- Significant changes in the tooling and workflows to accommodate central lock files.

Upsides:

- Ensuring consistency in dependency versions across multiple solutions / repository.
- Recognizing the community's evolving needs for better dependency management in monorepos and multi-solution projects.
- Evolving beyond a project and solution level

We encourage further discussion and exploring this topic further while we continue to assess feasibility with others teams (i.e. MSBuild, etc) and how this would be implied throughout the .NET ecosystem.

Please continue to comment, raise your support/disapproval, and provide any prior art to help us investigate further!

yaakov-h:
&lt;generic activity here&gt;

accepted/2024/repo-level-lock-file.md
CEbbinghaus: For flexibility I prefer the idea of having the lock file path configurable which also tracks consistently with lock files on a project. The better solution is probably setting the default property to be next to the `Directory.packages.props` and instead offering (if it doesn't already exist) a `$(ThisCurrentFilePath)` MSBuild property which can be used to set the file path relative to the Directory.packages.props rather than something relative to the project.

nkolev92: Yeah, it's all doable, we'd just need to be specific to avoid confusion.

accepted/2024/repo-level-lock-file.md
CEbbinghaus: I prefer the idea of promoting any transitive dependency at a version outside the already resolved range into a primary dependency so that any downstream library consumers don't need to think about it

ajeckmans:
Bad bot! :)

CEbbinghaus:
Can we just give this Item a No-Stale flag so we don't have activity for the sake of activity. People will keep sending a message whenever the bot tries to stale it

rkm:
Boop

> Can we just give this Item a No-Stale flag so we don't have activity for the sake of activity. People will keep sending a message whenever the bot tries to stale it

+1

yaakov-h:
that sounds like a you problem, microsoft-github-policy-service.

TheAngryByrd:
_insert recent activity_

ajeckmans:
Here we go again...

jeffkl:
@CEbbinghaus apologies for the long delay, I'm looking into this feature and was wondering if we could do a call to go over some of the scenarios with you to get your input on how they should be handled.  In particular, I'd like to discuss:

1. The central lock file cannot be updated when restoring a single project or set of projects, only when restoring the project.  Since these smaller restores don't restore everything, there's no way to know all of the dependencies. Is that acceptable?  Should NuGet log a warning when central lock files are in use and a user is only restoring a subset of projects telling them that the lock file wasn't regenerated?
2. Since lock file regeneration isn't as easy for "all projects in the repo", should there be a new experience to regenerate the lock file.  Like a `dotnet nuget lock` that you point to a solution and it gets all of the projects and their flattended dependencies to generate the lock file?
3. Would a new workflow like this fit in well with a users' existing pattern?  Or is there a better way that we haven't thought of?

The main problem I'm trying to solve is that when NuGet restores, it never really knows the full transitive closure of every project in an entire repository.  System like Paket use proprietary files and it enumerates folders looking for them to sort of accomplish this. But in the .NET ecosystem, we use MSBuild projects as the lowest common denominator which is why lock files are currently tied to that.  Associating a lock file with a solution sounds possible, but there are scenarios where restores aren't executed for a solution and there are repositories with more than one solution file in them.  A new paradigm of a "repo-wide" restore is probably not possible at the NuGet level, we'd need all of .NET to adopt the concept of a repository root.  Even with `Directory.Packages.props` you can have multiple files in your repository using MSBuild functionality.  

So, I'm looking into a good way to get users the functionality they want and deliver a good experience.  Since you've taken the time to send this design proposal, I wanted to have a discussion with you to brainstorm ideas, if you're open to it.  I'm also looking for any other area experts who are interested in talking about it with me.

CEbbinghaus:
@jeffkl happy to set up a call. It would be good to discuss some possibilities around those unanswered questions, especially given that these are things paket supports and people need.

yaakov-h:
> we'd need all of .NET to adopt the concept of a repository root

https://github.com/dotnet/designs/pull/328 ?

ajeckmans:
...

TheAngryByrd:
*the sound of crickets*

Frulfump:
The sound of silence

yaakov-h:
that is once again a you problem, bot.

Frulfump:
Ah bot we meet again, still an issue...

IanKemp:
@JonDouglas please review this.

TheAngryByrd:
meow (activity for the bot)

IanKemp:
NuGet maintainers, do your jobs.

yaakov-h:
shoo, bot.

rkm:
Boop

IanKemp:
2026 and the Microsoft tumbleweed rolls...

IanKemp:
> This PR has been automatically marked as stale because it has no activity for **30 days**. It will be closed if no further activity occurs **within another 330 days** of this comment. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

No.

CEbbinghaus:
Still no No-Stale feed Flag? Cmon...

yaakov-h:
time for some more activity then

Frulfump:
> @jeffkl happy to set up a call. It would be good to discuss some possibilities around those unanswered questions, especially given that these are things paket supports and people need.

Did you have a call and if so what came out of it?

CEbbinghaus:
@Frulfump Yes we had a call, The key takeaways are the following:

Microsoft is unwilling to implement our current design as it stands, as that necessitates a breaking change in dependencies and therefore forces everybody into a very narrow box. From microsoft's point of view the primary usage of this feature is not to lock all versions of every transitive dependency across a monorepo to the same version, but rather ensure consistent versioning (between builds) across many different services / libraries / packages / applications.

To this end the crux of the discussion was, that to make this feature viable we would likely require multiple "levels" of strictness for the lockfile. With on one end being our solution which forces **every single** package within the `Directory.Packages.props` to be compatible with every other package (very hard given the flexibility of MSBuild) and locked together into one "master" lockfile of which every csproj would be referencing a part of.

And on the other end being Microsoft/Jeff's plan of having one big Json file with every individual lock file for every project that is part of the solution / multi project build merged into a single object. This allows for each project to resolve its own transitive dependency versions and keep them locked, but does not guarantee consistent versioning amongst projects.

There may be a middle ground solution but given the two approaches outlined above were the main targets of our goals & microsofts, It wasn't in our interest to investigate other potential "modes".

yaakov-h:
can u not, bot?

TheAngryByrd:
bump for bot

Frulfump:
> This PR has been automatically marked as stale because it has no activity for **30 days**. It will be closed if no further activity occurs **within another 330 days** of this comment. If it is closed, you may reopen it anytime when you're ready again, as long as you don't delete the branch.

But this PR is tagged with "[Status:Do not auto close](https://github.com/NuGet/Home/issues?q=state%3Aopen%20label%3A%22Status%3ADo%20not%20auto%20close%22)"?!

accepted/2024/repo-level-lock-file.md
Nigusu-Allehu: Maybe we could look into a command that generates an exhaustive transitive pinning. The command could be ran once and it would pin all transitive packages for the repo in the DPP file. It would probably be an inconvenience to run the command every time a direct dependency changes

accepted/2024/repo-level-lock-file.md
Nigusu-Allehu: Agreed. Just to make sure I understood, is this about when one package that appears as a transitive dependency in two different projects has different versions?

TheAngryByrd:
bump for bot

fifty-six:
This PR has been automatically marked as not stale because I don't want it to be stale.

pomianowski:
Working with several hundred repositories, each with between 10 and 1400 projects, equals hundreds of lock files.
I think a lot of people want to achieve consistent CPM (which is almost mandatory for large repos) along with a central lock file that allows me to enforce it. Currently, the ONLY alternative I know of is setting `CentralPackageTransitivePinningEnabled` to `true` and then adding every known package to `Directory.Packages.props`. This amounts to a huge mess.

I would like one file in the root repository with a header like this:

```xml
<Project>
  <PropertyGroup>
    <NuGetLockFilePath>$(MSBuildThisFileDirectory)/packages.lock.json</NuGetLockFilePath>
    <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
    <CentralPackageTransitivePinningEnabled>true</CentralPackageTransitivePinningEnabled>
    <EnablePackageVersionOverride>false</EnablePackageVersionOverride>
    <NuGetAudit>true</NuGetAudit>
    <NuGetAuditLevel>moderate</NuGetAuditLevel>
  </PropertyGroup>
  <ItemGroup>
    <PackageVersion Include="Aspire.Hosting.PostgreSQL" Version="13.*" />
    <PackageVersion Include="OpenTelemetry.Exporter.OpenTelemetryProtocol" Version="1.*" />
...
  </ItemGroup>
</Project>
```

Please disable the __no recent activity__. Centralized package management and locking of transitive packages is crucial for applications that are meant to actually work, not be POCs on GitHub.

PR #13376, opened 2024-04-05
load package source attributes in PMUI.

Nigusu-Allehu:
### **This is targeting 17.12**
This proposal aims to enable the PMUI to load package source attributes. Consequently, users will have the capability to view and modify these values.
Rendered: https://github.com/NuGet/Home/blob/697beb7ea1c8e320004591bfceb0989c45cdb060/accepted/2024/PMUI-packagesource-option.md


donnie-msft (commented):
Thanks for taking the initiative to write up this idea! This is probably something we should do for Unified Settings, and not necessarily in the existing VS options. Would you like to drive this change as part of https://github.com/NuGet/Client.Engineering/issues/2712 ?

Gathering some customer input would be great. I personally don't use these features, so finding customers who do more regularly would be best to weigh in here. In general, I think a UI for these would be useful.

It'd be great if you can you find some data to go along with this spec. Either customer signals from telemetry (if we have any for these), or work with PM to survey customers about these properties and learn how a UI would help them.

Nigusu-Allehu:
> Thanks for taking the initiative to write up this idea! This is probably something we should do for Unified Settings, and not necessarily in the existing VS options. Would you like to drive this change as part of [NuGet/Client.Engineering#2712](https://github.com/NuGet/Client.Engineering/issues/2712) ?
> 
> Gathering some customer input would be great. I personally don't use these features, so finding customers who do more regularly would be best to weigh in here. In general, I think a UI for these would be useful.
> 
> It'd be great if you can you find some data to go along with this spec. Either customer signals from telemetry (if we have any for these), or work with PM to survey customers about these properties and learn how a UI would help them.

As we discussed offline, the ability for users to add HTTP sources is essential. Implementing the `allowInsecureConnections` checkbox will enable users to consent to adding an HTTP source, which otherwise would be blocked in PMUI. Therefore, I believe collecting additional data is unnecessary at this point

Nigusu-Allehu:
@NuGet/nuget-client please feel free to leave a comment.

PR #13379, opened 2024-04-06
Update 'dotnet nuget why' output format

advay26:
[Rendered spec](https://github.com/NuGet/Home/blob/e83e772d7372288fc4475ba8b511dca97768a006/accepted/2022/dotnet-nuget-why-proposal.md)

The output of `dotnet nuget why` would be easier to read if we printed it as a tree, rather than as individual paths.

We will also deduplicate graphs when multiple frameworks have the same dependency graph.

SolutionWithMultipleProjects

The spec contains more examples.


nkolev92 (approved):

JonDouglas:
Looks beautiful. Should we evangelize this further?

JonDouglas (approved):

accepted/2022/dotnet-nuget-why-proposal.md
```diff
@@ -118,6 +118,11 @@ dotnet nuget why [<PROJECT>|<SOLUTION>] <PACKAGE_NAME>
 
 dotnet nuget why -h|--help
 ```
+
+We will show the resolved versions of packages in the graph, rather than the versions requested by the top-level package.
+
+We will also deduplicate graphs in certain cases. If we have already shown a package's dependencies in the graph in another place, we will truncate the graph at that node and show a `(*)` symbol to indicate that the package's dependencies have already been shown. Further, if multiple target frameworks in a project have the same dependency graph, we will show a combined graph for them. (See [examples](#examples))
```
zivkan:
```suggestion
We will also deduplicate graphs in certain cases.
If we have already shown a package's dependencies in the graph in another place, we will truncate the graph at that node and show a `(*)` symbol to indicate that the package's dependencies have already been shown.
Further, if multiple target frameworks in a project have the same dependency graph, we will show a combined graph for them.
(See [examples](#examples))
```

I don't know if it's widely considered best practice, but I certainly consider it a markdown best practice to put 1 sentence per line. Markdown renders it the same way, but in the future if it gets modified, the diff will be clearer. GitHib in particular has troubles with diffs on very long lines. Plus it allows people commenting in PRs to comment on a single sentence, rather than the entire pargraph.

zivkan: I'm not really a fan of this `(*)` proposal. I looked at the screenshots before reading the text, and I couldn't guess what it meant. So, if we add it, I think it's very important that the program output tell customers what it means.

However, even if the subtree was already output in another part of the graph, I don't think that trimming it provides a big value. It will make the output shorter, sure, but it actually doesn't in the examples in your screenshots, becuase the subtree it trimmed was just one package. And even if it did make the output shorter, there's also the risk of making it harder to understand, since the customer will now need to look for other instances of the parent package, to see what the children packages actually are.

zivkan (approved):
I love the improved outpt.  I'm not a fan of the `(*)` to remove duplicate subtrees, but overall I think it's a huge improvement to readability.

Mrxx99:
Habe you considered dotnet nuget explain {package}? Maybe as an alias to why? Because that is already common for npm and many people using both would only need to remember one. (I had tried dotnet nuget explain several times)

JonDouglas:
I'd recommend a few UX improvements if possible.

1. Use spectre.console's tree - https://spectreconsole.net/widgets/tree
2. Use spectre.console's markup - https://spectreconsole.net/markup a color like yellow or red to show the "why" package ID. In the example above, `System.Text.Json` could be colored RED so the user can easily know where it is coming from and how deep it is in the graph. Similar to the work we did with the search experience and showing red for matching terms.

(These can also be accomplished with what we have today, see the --vulnerable or search experience code for examples) - Drew or Nigusu may have experience/pointers.

advay26:
> Habe you considered dotnet nuget explain {package}? Maybe as an alias to why? Because that is already common for npm and many people using both would only need to remember one. (I had tried dotnet nuget explain several times)

@Mrxx99 There was a discussion around naming at the time the spec was originally merged: https://github.com/NuGet/Home/pull/11875#discussion_r916377800, and there was some interest even back then in having both `why` and `explain` as aliases. At the time, we'd opted for `why` as a starting point, and left it open for us to add an alias in the future. For now, I've added this into the 'Future work' section of the spec. We'll look to start with `dotnet nuget why`, and can then consider improvements as we receive feedback on the feature.

accepted/2022/dotnet-nuget-why-proposal.md
advay26: Thanks! I'll start doing that

advay26: This seems fair. Having to look through the rest of the graph to see the subtree isn't ideal, especially as default behavior. We can consider adding this as a CLI option in the future, but we'll show the complete graph for now.

I've moved the duplicate subtree trimming into the 'Future work' section now, as a potential CLI option (a boolean option allowing users to de-duplicate subtrees, or a verbosity option, etc.)

zivkan (approved):
:shipit:

Issue #10714, opened 2021-04-01
Download counts do not appear for packages suggested by Package Recommender


joelverhagen:
## Details about Problem

NuGet product used: Visual Studio
Product version: 16.10.0, preview 2
Worked before? Version prior to package recommender

## Repro steps and/or sample project

Open the Browse tab on the PM UI. 
Ensure you have Package Recommender enabled (my understanding is that it's an A/B test, but I am not sure)

![image](https://user-images.githubusercontent.com/94054/113339476-5bc60b00-92df-11eb-983a-bd83f0fdc0a8.png)

joelverhagen:
Strangely, the download counts work for the V2 feed:
https://www.nuget.org/api/v2
![image](https://user-images.githubusercontent.com/94054/113339667-aba4d200-92df-11eb-8722-e5758e89a76a.png)

dominoFire:
@birgithi FYI

zkat:
@jebriede can you please follow up with the package recommender team to see if this is even possible?

nkolev92:
@chgill-MSFT @jebriede 

Can you please define the expected behavior. 
Is this still something we care about?

stamminator:
Looks like the "over one billion" issue again: https://github.com/NuGet/Home/issues/10864. Maybe a regression caused by https://github.com/NuGet/NuGet.Client/pull/4255.

![image](https://user-images.githubusercontent.com/10633170/151434171-12767a68-e907-4971-ba47-266ab4bb5f88.png)

nkolev92:
@stamminator 

Shouldn't be related to that. 
The gist is that the download counts are only available through the `search` server side API and NuGet does not call that right now, as it'd require it to call it for every individual package.

stamminator:
@nkolev92  Seems like an awful coincidence though that in both cases, it seems to be related to hitting the 1 billion figure.

nkolev92:
@stamminator 

In this the download counts are not getting display for any of the recommended packages. 
If you look at your screenshot, you'll notice that none of the packages with a 'star' have a download count.

PR #13514, opened 2024-06-05
Update Owner Author spec to indicate author is always shown for IntelliCode Suggestions

donnie-msft:
The implementation as of now will be that Recommended packages will never shown Known Owner, so reflect this in the spec.

[Rendered](https://github.com/NuGet/Home/blob/3b1cecf65e08da6e199a7d8896d5a7385d3aba7d/accepted/2023/owner-author-pmui.md)


donnie-msft:
@skofman1 & @joelverhagen , referring to your thoughts in https://github.com/NuGet/Home/pull/13030#discussion_r1446405442, I wanted to see if we can align on the IntelliCode Suggestions / recommended packages maintaining author. Otherwise, these will have to be blank. I think they're separated enough that it's not as confusing or misleading as your original concern.

skofman1:
@donnie-msft , agree. The IntelliCode recommendations will be popular packages, so I don't think that showing the Author value has any trust risk. 
In an ideal world it would be wonderful to be able to show Owner links for those as well, but I understand that it would be significantly more work and out of scope for this change. Can we add this suggestion to our backlog? Perhaps the community will be interested in contributing it :)

donnie-msft:
Sure, we had https://github.com/NuGet/Home/issues/10714 which is the same root cause of that behavior, but I went ahead and filed one calling out package owners to see if the community notices that with the new Known Owner feature.
https://github.com/NuGet/Home/issues/13516

nkolev92 (approved):

Issue #12619, opened 2023-05-30
PM UI:  should not recommend deprecated packages


dtivel:
### NuGet Product Used

Visual Studio Package Management UI

### Product Version

VS 17.7.0 Preview 2.0 [33729.141.main]

### Impact

It bothers me. A fix would be nice

### Repro Steps & Context

I'm not sure how to provide repro steps for NuGet package recommender.

That said, I opened PM UI, and the Browse tab displayed a deprecated package as recommended.

![image](https://github.com/NuGet/Home/assets/12734758/93e29211-6dfc-4456-85f1-ed13514a7db2)

Not sure if the recommendation should not be made at all or if it should be behind a "Show deprecated packages" option, but it seems like a bad idea to _recommend_ deprecated packages.

### Verbose Logs

```shell
Microsoft Visual Studio Enterprise 2022
Version 17.7.0 Preview 2.0 [33729.141.main]
VisualStudio.17.IntPreview/17.7.0-pre.2.0+33729.141.main
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Enterprise

Visual C++ 2022   00476-80000-00000-AA281
Microsoft Visual C++ 2022

ASP.NET and Web Tools   17.7.137.65386
ASP.NET and Web Tools

Azure App Service Tools v3.0.0   17.7.137.65386
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   17.7.137.65386
Azure Functions and Web Jobs Tools

C# Tools   4.7.0-2.23274.2+2b33f51dabed9c603211996922e05fd762f72b00
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.4.32 (main@0bf34d7)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Microsoft Azure Tools for Visual Studio   2.9
Support for Azure Cloud Services projects

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Mono Debugging for Visual Studio   17.7.16 (be7117c)
Support for debugging Mono processes with Visual Studio.

NuGet Package Manager   6.7.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

[...]

PR #13530, opened 2024-06-10
Spec: Include Metadata with IntelliCode Suggested Packages

donnie-msft:
Spec for https://github.com/NuGet/Home/issues/10714


donnie-msft:
Related: https://github.com/NuGet/Home/issues/12619

Issue #13191, opened 2024-01-25
[Bug Bash] NuGet restore doesn’t work after checking the “Allow NuGet to download missing packages” in Options dialog for C# Console App (.NET 8.0) 


v-luzh:
### NuGet Product Used

Visual Studio Package Management UI

### Product Version

Dev\6.10.0.13

### Worked before?

It is not a regression since it reproduced on D16.11\34301.259.

### Impact

It bothers me. A fix would be nice

#### Repro Steps: 
1.	Create a C# Console App (.NET 8.0) project in VS and open the PM UI window. 
2.	Install several packages (e.g. Newtonsoft.Json and log4net), go to the path “C:\Users\[username]\.nuget\packages” and delete an installed package (e.g. log4net) from the packages folder. 
3.	Click the settings button (gear button) on PM UI or PMC window. 
4.	Click the NuGet Package Manager->General in the opened “Options” window. 
5.	Check the “Allow NuGet to download missing packages” and click OK button. 
6.	Right-click the solution node in Solution Explorer and click "Restore NuGet Packages".
7.	Go back to the path “C:\Users\[username]\.nuget\packages” and observe. 
  
#### Expected Result: 
The lost packages had been restored back to the global packages folder successfully.

#### Actual Result:
The lost package was not restored back to the global packages folder in step7 and the info “all packages are already installed an there is nothing to restore” in step6.

#### Notes:
1.Repro rate: 100%.
2.It doesn’t repro for iOS (Xamarin) App/C# UWP project.
3.It doesn’t repro when deleting the whole “packages” folder from the path “C:\Users\[username]\.nuget”.

nkolev92:
This is a known limitation of https://github.com/NuGet/Home/blob/dev/accepted/archive/VisualStudio-PartialRestoreOptimization.md. 

At this point, the perf benefits outweigh the likelihood of this edge case happening (we haven't heard any issues) around this. 

We're gonna leave this open to potentially collect upvotes, but we're not likely to change it.

SueSu01:
It still reproes on VS Main\35222.175 + NuGet Client Dev\6.12.0.82.

v-luzh:
It still reproes on VS Main\12206.68+NuGet Client Dev\7.12.0.4. So, I will reopen it as Nikolche mentioned above.

PR #13561, opened 2024-06-18
Spec: NuGet onboarding to Unified Settings in Visual Studio

donnie-msft:
[Rendered spec](https://github.com/NuGet/Home/blob/dev-donnie-msft-onboardUnifiedSettings/accepted/2024/unified-vs-settings.md)

Part of: https://github.com/NuGet/Client.Engineering/issues/2712


albarry4:
Looks great -- once the "clear NuGet cache" button is added back in, no concerns from me!

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,160 @@
...
+<!-- One-paragraph description of the proposal. -->
+
+Visual Studio Options are moving to a new user interface called "Unified Settings" (abbr. USX).
+The Visual Studio (VS) IDE team is responsible for the UI controls, and individual product teams will maintain a JSON file declaring the structure and data types for settings in the IDE in addition to any custom commands.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+The Visual Studio Editor team created Unified Settings to make the configuration experience have a Consistent UI, scoping capabilities, and the ability to share and sync settings across devices.
```
JoshuaBStevens: It's the IDE Experience team, not the editor team.

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,160 @@
...
+The `Clear local NuGet resources` button from legacy options will be shown as a hyperlink button in USX under the root section's title ("General").
+
+Pressing the hyperlink button will open a modal dialog window indicating the status of the operation of clearing the resources from disk.
+The mockup below shows the button and the effect of pressing it:
+
+![Mockup showing the Clear local NuGet resources hyperlink button and the modal dialog which opens once pressed](../../meta/resources/UnifiedSettings/clear-command-dialog.png)
+
+#### Error messages
+
+When the `NuGet.Config` has an invalid value, the UI will show a blank selection, or unchecked checkbox, and will be accompanied by a simple error message indicating that the value in the backing store was somehow invalid.
```
JoshuaBStevens: Minor correction: checkboxes will show an "indeterminate" state in this case.

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,160 @@
...
+
+Any action taken inside the USX page will overwrite that invalid value in the backing store (`NuGet.Config`) according to `NuGet.Configuration`.
+
+For any fatal errors, such as unable to load the `NuGet.Config` at all due to parsing errors, or similar, then an error message will be shown under the "General" section label.
```
JoshuaBStevens: Also worth mentioning for the "fatal error" scenario:
- All the edit controls in the section become disabled; the user cannot interact with them while in an error state.
- The error state can be cleared by your provider if you detect that the condition is fixed.
- Your provider can also choose to let us show the user a "Retry" link which will retry the failed operation.
- The error message you provide can contain embedded links which can point to files, so you can give the user a link to click in order to fix a bad NuGet.config in a text editor.

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,160 @@
...
+Therefore, our Visual Studio Command Table (VSCT) file will be updated with a new command, which executes `LocalsCommandRunner` in the following way:
+
+1. Open a modal dialog window.
+    - State that NuGet resources are being cleared.
+    - Prevent the user from accessing VS components like the Restore or the NuGet Package Manager since this could cause conflicts with the operation of clearing local resources.
+    - Show an indeterminate progress indicator.
+    - A Cancel button cannot be supported at this time.
+    (See Future Possibilities)
+1. Execute `LocalsCommandRunner` in the same manner as legacy VS Settings.
+1. Display text indicating the completion of the command (or any error messages) and a close button on the modal dialog.
```
JoshuaBStevens, line 109: Maybe the dialog should just close automatically when complete?  Just an idea, it's your call.

donnie-msft: I think I'll leave it for now as I'd like to show the completion information. Easy to change later, if necessary.

accepted/2024/unified-vs-settings.md
donnie-msft: Think I captured these except for the latter 2, which I feel are enhancements that may need a further spec beyond the scope of this one.

jeffkl (approved):

JoshuaBStevens (approved):

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,164 @@
...
+
+![Mockup showing a prototype of the NuGet General Options page in Unified Settings](../../meta/resources/UnifiedSettings/general-prototype.png)
+
+### Settings are stored in NuGet.Config
+
+Since the `NuGet.Config` is an external backing store, NuGet.Client can specify explanatory text and a hyperlink button that points to its location.
+The text shown in the mockup is "These settings are stored in NuGet.config".
+Pressing the hyperlink button on `NuGet.Config` will open the file in the VS editor.
+
+Given that there can be multiple `NuGet.Config` files, the plan for now is to have it open the first `NuGet.Config` which is closest to the solution, which would be the first file listed in the "Configuration Files" page of NuGet's legacy settings, today.
```
jgonz120, line 48: Is the "Configuration Files" page going to be migrated over at some point too?

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Registration
+
+The NuGet.Client repository will maintain a `registration.json` file to declare all VS Settings, including localizible names, data types, default values, command implementations, and any dependencies among the values of other settings.
+
+The `registration.json` will use the `"type": "external"` mode because our product stores configuration in a custom backing store: One or more `NuGet.Config` files.
+
+### Command: Clear Local Resources
+
+NuGet.Client will create a new modal dialog will be developed to indicate that the Clear Local Resources command is in-progress and has completed.
```
jgonz120:
```suggestion
NuGet.Client will create a new modal dialog to indicate that the Clear Local Resources command is in-progress and has completed.
```

jgonz120 (approved):

accepted/2024/unified-vs-settings.md
donnie-msft: Yes, all pages will be migrated. Just tackling General first.

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+All of NuGet's legacy VS Options pages will be migrated to USX.
+The General page will be tackled first, and is the main scope for this spec.
```
OliaG, line 33:
```suggestion
The General page will be tackled first, and is the main scope for this spec. The rest of the pages will be under More Settings tab for now that will open the old view and will be migrated later.
```

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Error messages
+
+When the `NuGet.Config` has an invalid value, the UI will show a blank selection, or an indeterminate state checkbox, and will be accompanied by a simple error message indicating that the value in the backing store was somehow invalid.
+
+Any action taken inside the USX page will overwrite that invalid value in the backing store (`NuGet.Config`) according to `NuGet.Configuration`.
+
+For any fatal errors, such as unable to load the `NuGet.Config` at all due to parsing errors, or similar, the following will occur:
+
+- An error message will be shown under the "General" section label.
```
OliaG, line 67: Should we add an instruction to modify/fix the NuGet.Config file externally (outside VS)?
Should we allow users to fix it somehow from the VS (like delete and create a blank file)? Just an idea, not sure if VS has a unified way of handling cases like that.

accepted/2024/unified-vs-settings.md
donnie-msft: I intentionally did not add too much detail here, as the way they are handling legacy pages is expected to change.

accepted/2024/unified-vs-settings.md
donnie-msft: Great idea! I've added an example message, with 2 hyperlinks. One is a docs link (to be determined which one, open to suggestions).
The other link is that pressing "NuGet.Config" will open the NuGet Configuration Files settings page showing all the files, and allowing the customer to open each directly in VS.

accepted/2024/unified-vs-settings.md
OliaG: As long as we are clear that "More Settings" tab is just a temporary state, I'm good. Initially it didn't look good to me until I realized it is just a transition state, not the final desired layout.

accepted/2024/unified-vs-settings.md
OliaG: For the docs reference maybe this one: https://learn.microsoft.com/nuget/consume-packages/configuring-nuget-behavior ?

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Registration
+
+The NuGet.Client repository will maintain a `registration.json` file to declare all VS Settings, including localizible names, data types, default values, command implementations, and any dependencies among the values of other settings.
+
+The `registration.json` will use the `"type": "external"` mode because our product stores configuration in a custom backing store: One or more `NuGet.Config` files.
+
+### Command: Clear Local Resources
+
+NuGet.Client will create a new modal dialog to indicate that the Clear Local Resources command is in-progress and has completed.
```
OliaG, line 93: My concern is that while resources are getting cleaned up, users might get their VS UI frozen for up to a few minutes. Would be great if we could allow users to navigate away from the dialog and Settings. For example, show the "Clearing NuGet local resources..." and after it is done "NuGet local resources have been cleared" in the VS lower left status panel:
 
image

OliaG: If it is not possible, I think we should have a message "Clearing local NuGet resources, this might take a minute...". Show progress bar if it is possible.

donnie-msft: I originally considered a separate notification like you describe; however, it will likely be problematic (a race condition) to allow restores and other NuGet gestures to happen when we are simultaneously deleting from package folders.

Today, the UI freezes while the delete is happening. In this spec, I do not propose freezing the UI in VS. Instead, I propose a modal dialog with a loading indicator. Ideally, we can design a cancellable Locals Command so that a cancel button can be shown on this dialog.

accepted/2024/unified-vs-settings.md
donnie-msft: That could work, also the nuget.config specification could work. Many options!

accepted/2024/unified-vs-settings.md
donnie-msft: NuGet doesn't own that UI or functionality. I'm not sure we need to put it in our own spec? My understanding is yes, they are moving away from "more settings". If you'd like me to add something specific, please let me know.

jeffkl (approved):

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,168 @@
...
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+No questions have been identified.
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+
+- `LocalsCommandRunner` could be enhanced to support Cancellation.
+This would allow the modal dialog shown from our USX command to provide a `Cancel` button in case the user decides not to wait on a large delete operation to complete and block usage of VS.
```
zivkan: Keep in mind that making delete support cancellation will significantly increase the risk that customers will experience:
* https://github.com/NuGet/Home/issues/13191

If we can make the button flush the 'fast no-op' cache, then that would be sufficient. Just something to remember if this goes ahead.

donnie-msft: Thanks! Added to Future Possibilities.

accepted/2024/unified-vs-settings.md
```diff
@@ -0,0 +1,169 @@
...
+
+The invocation of the `LocalsCommandRunner` which is supported by legacy VS settings will be adapted to a USX Command.
+A USX Command invokes a command in VS using the VS SDK and the extensible menu command system.
+Therefore, our Visual Studio Command Table (VSCT) file will be updated with a new command, which executes `LocalsCommandRunner` in the following way:
+
+1. Open a modal dialog window.
+    - State that NuGet resources are being cleared.
+    - Prevent the user from accessing VS components like the Restore or the NuGet Package Manager since this could cause conflicts with the operation of clearing local resources.
+    - Show an indeterminate progress indicator.
+    - A Cancel button cannot be supported at this time.
```
nkolev92, line 106: We discussed this during the PR review meeting, so worth adding a call out here that cancellation button that does partial deletion *probably* breaks fast no-op. 
This can be "solved" though, but forcefully invalidating the caches of fast no-op, but that's a future possibility anyways.

donnie-msft: Yes, I addressed that in Andy's comment here. Does this resolve it do you think? https://github.com/NuGet/Home/pull/13561/commits/1827b6e41464a7577bcdbcf9459d44009da5062c

nkolev92: Yeah, looks good.

accepted/2024/unified-vs-settings.md
OliaG: sounds good

accepted/2024/unified-vs-settings.md
OliaG: ok, let's proceed with this as MVP and see if we have any feedback from users

OliaG (approved):
LGTM

PR #13573, opened 2024-06-24
Create proposal to replace blue checkmark

albarry4:
[Rendered Spec](https://github.com/NuGet/Home/blob/f646a34da0b3b607383c9f061a64ea43ac606ee8/accepted/2024/bluecheckmark-replacement.md)


donnie-msft:
👀 does this also apply to Client?

albarry4:
> 👀 does this also apply to Client?

@donnie-msft yes! this would apply anywhere the current blue checkmark is shown

JonDouglas:
Wonderful proposal!

Another consideration is not needing any type of iconography and just create the status as a pill/badge HTML element that just says `Prefix Reserved`

One icon that comes to mind though is a tag + lock, which conveys the idea of a name/id & reserved/locked. But that may compete with the existing tagging functionality.

image

This could sit next to upcoming work revolving around build provenance which will have this type of icon: 
image

i.e. Forgive my 5 second mockup here:

image

or

image

albarry4:
@JonDouglas I really like the other icon that you proposed, I would definitely like to add that to the discussion of potential icons. I do, however, see the potential for people to associate it with the existing tags concept on NuGet mistakenly. Perhaps some customer research could be done here about the icon options? 

I did consider the idea of doing away with the icon all-together (probably should have mentioned this in the proposal, thank you for calling it out!) However, I thought that keeping an icon present would still retain the importance of prefix reservation for security reasons and keep things consistent with what users are used to. I do think that this is a good option to consider though! Would love to get some more opinions

nkolev92 (approved):
Great write-up! 
I like the icon in this proposal, but I like the icon in what @JonDouglas proposed as well. 
342529785-1072f446-c42b-40d5-8d8e-6c042fc41350

I don't have an opinion about the text beyond real estate considerations.

I have a slight preference towards the icon I linked.

jeffkl (approved):

JonDouglas (approved):
This lgtm. May want to do a quick team review to finalize the settled concept & merge.

aortiz-msft:
@albarry4 - How can we validate the icons with customers ahead of making the change?

accepted/2024/bluecheckmark-replacement.md
```diff
@@ -0,0 +1,81 @@
+# ***Replacing the Blue Checkmark***
+<!-- Replace `Title` with an appropriate title for your design -->
```
Frulfump, line 2: Should these comments really be left in?

Frulfump:
Why was this PR abandoned? Seems useful @albarry4

Frulfump:
@aortiz-msft this got automatically closed but you re-opened and self-assigned so hoping this would get some attention now

PR #13634, opened 2024-07-17
PrunePackageReference (fka Supplied by platform) design

nkolev92:
https://github.com/NuGet/Home/blob/dev/accepted/2024/prune-package-reference.md

### DIFF
--- a/accepted/2024/supplied-by-platform.md
+++ b/accepted/2024/supplied-by-platform.md
@@ -118,10 +118,10 @@ TODO NK
 
 See [Prior art](#net-sdk-package-pruning) for more details on how the .NET SDK prunes packages at runtime today.
 The particular pruning is based on the resulting list of shared frameworks.
-The .NET SDK will provide the list at the beginning of the restore operation.
+The .NET SDK will provide the list at the beginning of the restore operation, as such the .NET SDK *must* only consider direct shared frameworks, and not transitive ones.
 Given that packages are allowed to bring in a shared framework, and that is not known by the .NET SDK at the beginning of restore, package graphs brought in by packages are not going to be pruned with the current solution.
 
-A potential solution may be adding a way to prune additional frameworks as an opt-in. 
+A potential solution may be adding a way to prune additional frameworks as an opt-in.
 
 ## Drawbacks
 
--- a/accepted/2024/supplied-by-platform.md
+++ b/accepted/2024/supplied-by-platform.md
@@ -111,7 +111,7 @@ The list of PrunedPackageReference items must only include relevant packages.
 It will be included in the "project" section of the assets file, internally called the PackageSpec, similarly like the centralPackageVersions.
 
 ```json
-  "project": {
+  {
     "frameworks": {
       "net8.0": {
         "targetAlias": "net8.0",
@@ -129,6 +129,9 @@ It will be included in the "project" section of the assets file, internally call
         "prunedPackageReferences": {
           "System.Text.Json": "8.0.5"
         }
+      }
+    }
+  }
 ```
 
 The pruned packages won't be represented in the assets file in any other way.
--- a/accepted/2024/supplied-by-platform.md
+++ b/accepted/2024/supplied-by-platform.md
@@ -37,6 +37,7 @@ When NuGet encounters any of the specified packages or lower, it will simply rem
 This package id will not be downloaded and will not appear in the assets file libraries or targets section, but there will be a detailed verbosity message indicating that the package id has been pruned for the given framework.
 
 The feature is framework specific, and can be opted in/out using the `RestoreEnablePackagePruning` property.
+Pruning is only possible for transitive packages, if a direct package reference is attempted to be pruned, a warning will be raised.
 
 ### Technical explanation
 
--- a/accepted/2024/supplied-by-platform.md
+++ b/accepted/2024/supplied-by-platform.md
@@ -1,0 +1,70 @@
+# ***Pruning platform provided packages***
+
+- Nikolche Kolev <https://github.com/nkolev92>
+- [7344](https://github.com/NuGet/Home/issues/7344)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+
+## Motivation
+
+In the early versions of .NET (Core), individual assemblies from the runtime were shipped as packages.
+Starting with .NET (Core) 3.0, targeting packs were used for delivering references assemblies. The reference assemblies are no longer represented in graphs the way Microsoft.NETCore.App, Microsoft.AspNetCore.App, and NETStandard.Library were.
+When a .NET (Core) 3.0 or later project depends on these older platform packages, we want them to be ignored in the graph, as those APIs are supplied by the platform via other mechanisms, such as targeting packs at build time and runtime packs for self-contained deployments.
+
+Even .NET 9, there are certain assemblies that ship as both packages, but are also part of certain shared frameworks such as ASP.NET. See details <https://github.com/dotnet/aspnetcore/issues/3609>. An example of such assembly/package is `Microsoft.Extensions.Logging`, which is [published on nuget.org](https://www.nuget.org/packages/Microsoft.Extensions.Logging/), but also part of the Microsoft.AspNetCore.App shared framework. Currently there's conflict resolution in the .NET SDK to ensure that the latest version is chosen.
+
+There are a few benefits:
+
+It is trivial to make the assumption that the fewer packages need to be downloaded, the better the performance will be.
+Beyond that, the extra packages within the graph, do make the resolution step more challenging. Some of the targeting packs used to bring in such a large package graph, that affected the resolution performance significantly. See  and <https://github.com/NuGet/Home/issues/11993> for more details. Furthermore, certain versions of popular packages have taken dependencies on these targeting packs for historical reasons. Such an example is <https://www.nuget.org/packages/log4net/2.0.10> which performs significantly better than <https://www.nuget.org/packages/log4net/2.0.9> when installed in a .NET (Core) projects, see <https://github.com/NuGet/Home/issues/10030>.
+
+Certain versions of these packages are no longer  part of the project graph, thus reducing the chances of false positive scanning.
+This is especially important in the context of assemblies that are part of shared frameworks. The build time conflict resolution ensures that the most up to date version is used, despite the fact that the user referenced package may habe vulnerabilities.
+Examples: <https://github.com/dotnet/sdk/issues/30659#issuecomment-2072567192>.
+
+This changes significantly helps align what gets restored and what gets published. By avoiding packages that are simply not going to be used at runtime, what gets used at runtime becomes more transparent.
+
+## Explanation
+
+### Functional explanation
+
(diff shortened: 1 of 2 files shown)


accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,192 @@
...
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+A list of package id/versions to be pruned will be provided by the .NET SDK. The version will signify the highest version to be pruned.
+When NuGet encounters any of the specified packages or lower, it will simply remove the package from the graph. This package id will not appear in the assets file, but there will be a detailed verbosity message indicating that the package id has been pruned for the given framework.
+
+The feature is framework specific, and can be opted in/out using the `NuGetPrunePlatformPackages` property.
+In .NET 10 SDK, this feature will be enabled for all .NET (Core) target frameworks newer than netcoreapp3.0 by default.
```
ViktorHofer: Wouldn't .NET Framework benefit from this as well i.e. when referencing .NET Standard 1.x packages?

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,192 @@
...
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- MSBuild items/properties vs a file with the data
+- MSBuild item name ideas: `PrunedPackageReference` was used. `IgnoredPackageReference` and `SkippedPackageReference` are alternatives.
+- What if a project id is specified in PrunedPackageReference? Warn? Error? Skip? Prune & Warn?
+- What to do when a direct PackageReference is specified to be pruned?
+- Should the pruned package reference dissapear from the dependencies section completely? Strong preference towards no, since it aids visibility.
+- Should the assets file contain the list of pruned packages? Are only ids important, or do we need versions as well?
+- Should the version attribute of PrunedPackageReference be a version range instead? Should the attribute be named MaxVersion instead?
+- What's the list of packages that'll be removed by the .NET SDK? How do we verify it won't break things?
```
ViktorHofer: When referencing the Microsoft.NETCore.App 9.0 shared framework, the list of packages to prune is https://github.com/dotnet/runtime/blob/main/src/installer/pkg/sfx/Microsoft.NETCore.App/PackageOverrides.txt

When referencing the Microsoft.AspNetCore.App 9.0 shared framework, the list of packages is https://github.com/dotnet/aspnetcore/blob/main/eng/PackageOverrides.txt

Looks like there isn't a PackageOverrides.txt file when referencing the Microsoft.WindowsDesktop.App shared framework. We probably should add one.

Same for older frameworks. The PackageOverrides list is part of the shared framework targeting pack and the SDK will need that information before restore.

accepted/2024/supplied-by-platform.md
nkolev92: Potentially, yeah. 
I can add it.

accepted/2024/supplied-by-platform.md
nkolev92: Those are the ones that I used as dummy values in my tests. https://github.com/NuGet/NuGet.Client/blob/f4519cb5fa25de9080ce2a66ff580354c61c9c3e/src/NuGet.Core/NuGet.Build.Tasks/PopulatePrunedPackageReferencesTask.cs#L37. 

Given what you said about shared frameworks, I think we have  a challenge with the transitive framework references. 
In particular, a package can specify a shared framework even if the project itself doesn't require it. 
Pruning that would be challenging, since the .NET SDK cannot know about the framework refs brought through by packages. 

Pruning is a graph walk time feature. Transitive framework references are known after the graph walk has been completed. 

We don't have an easy solution (if any) to this particular problem. In theory, I think the .NET SDK can actually provide switches for pruning additional shared frameworks. However, there's the potential pitfall where pruning those packages may lead to the shared framework being pruned as well. It's detectable by the SDK at build time though. 

This also brings up a very important thing where the .NET SDK should only prune frameworks that are directly specified to avoid first and second restore being different.

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,192 @@
+# ***Pruning platform provided packages***
+
+- Nikolche Kolev <https://github.com/nkolev92>
+- [7344](https://github.com/NuGet/Home/issues/7344)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+Provide a means to prune certain packages from packages because those packages are not going to be used at runtime.
```
dsplaisted: > prune certain packages from packages

Is this the intended wording?  Would something like "prune certain packages from the package graph" be more clear?

dsplaisted (commented):
It doesn't look like the spec explicitly states whether pruned packages will be downloaded at all.  It would be good to explicitly never download pruned packages.  Some vulnerability scanners look at the global packages folder, so this would ensure that those scanners wouldn't have false positives for pruned packages.

accepted/2024/supplied-by-platform.md
nkolev92: Not the intended wording. I'll fix it.

nkolev92:
> It doesn't look like the spec explicitly states whether pruned packages will be downloaded at all. It would be good to explicitly never download pruned packages. Some vulnerability scanners look at the global packages folder, so this would ensure that those scanners wouldn't have false positives for pruned packages.

Apparently I only had that in the summary. 

> This helps avoid downloading unnecessary reference and implementation packages

I'll add that explicitly in the rest of the spec as well.

nkolev92:
Pushed some more changes to the spec based on the feedback. 
There is a limitation with transitive shared frameworks. Not sure if we have an automatic solution for that.

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,217 @@
...
+  - The solution explorer tree - Prototype shows the solution explorer is skipping the reference.
+  - PM UI tab - Prototype shows that since the reference does not really exist, it is not show at all.
+  - list package - Prototype shows that since the reference does not really exist, it is not show at all.
+  - dotnet nuget why - TODO NK
+
+### NET SDK - selecting packages to be pruned
+
+See [Prior art](#net-sdk-package-pruning) for more details on how the .NET SDK prunes packages at runtime today.
+The particular pruning is based on the resulting list of shared frameworks.
+The .NET SDK will provide the list at the beginning of the restore operation, as such the .NET SDK *must* only consider direct shared frameworks, and not transitive ones.
```
ericstj: Honestly I've only ever seen folks complain when they get transitive shared frameworks.  It's often accidental and it breaks deployment -- @dsplaisted what if the SDK encouraged people to make them direct?

IOW if the result of a restore or resolving project references changed the shared framework of the project the SDK would warn the user and ask them explicitly reference that framework.

dsplaisted: I'm wary of changing that behavior, because we don't know how many people we aren't hearing from who are relying on it (knowingly or not).

With respect to the supplied by platform design, I think it would probably be OK to only consider the direct framework references.

ericstj: What if we made it just a warning:
1. When transitive framework was brought in after restore.
2. We can see that the project has packages that overlap with that framework.
3. Warn that those packages could be dropped by directly referencing the framework.

nkolev92: I like that idea. 
I added it in future possibilities.

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,228 @@
...
+This is especially important in the context of assemblies that are part of shared frameworks. The build time conflict resolution ensures that the most up to date version is used, despite the fact that the user referenced package may habe vulnerabilities.
+Examples: <https://github.com/dotnet/sdk/issues/30659#issuecomment-2072567192>.
+
+This changes significantly helps align what gets restored and what gets published. By avoiding packages that are simply not going to be used at runtime, what gets used at runtime becomes more transparent.
+
+## Explanation
+
+### Functional explanation
+
+A list of package id/versions to be pruned will be provided by the .NET SDK. The version will signify the highest version to be pruned.
```
jaredpar: @ericstj would these be specified by the individual targeting packs or the SDK as a whole?

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,228 @@
...
+When NuGet encounters any of the specified packages or lower, it will simply remove the package from the graph.
+This package id will not be downloaded and will not appear in the assets file libraries or targets section, but there will be a detailed verbosity message indicating that the package id has been pruned for the given framework.
```
jaredpar: Will this diagnostic be displayed when the reference is a transitive dependency? If so what do we feel is the ideal state here for that? Should the .csproj 

1. Add a `<PackageReference Remove="..." />` to signify they understand the reference was provided but removed
2. Disable the warning with a note
3. etc ...

This seems like it comes up often enough in our code today that I'm curious what the _ideal_ state looks like when it's encountered cause that will reflect on the aesthetics of the .NET ecosystem.

accepted/2024/supplied-by-platform.md
```diff
@@ -0,0 +1,228 @@
...
+          },
+        },
+        "centralPackageVersions": {
+          "MyDependency": "2.3.0",
+          "Microsoft.Build": "17.10.4",
+        },
+        "prunedPackageReferences": {
+          "System.Text.Json": "8.0.5"
+        }
+```
```
jaredpar: Why does the background color go red after this point in the GitHub PR? Makes it much harder to read.

accepted/2024/supplied-by-platform.md
nkolev92: > Will this diagnostic be displayed when the reference is a transitive dependency

Only transitive dependencies can be pruned. Pruning direct ones is not allowed. 
This is not going to be a warning, but simply a message telling you, hey this got pruned. It should be something that happen often enough and not something we think people need to concern themselves with since that already happens for runtime anyways.

For direct dependencies, people would get an actual warning, and I'd expect that people remove that reference.

accepted/2024/supplied-by-platform.md
ericstj: Today this information already ships in the individual targeting packs.  @dsplaisted doesn't think that NuGet nor the SDK can directly consume it from here since they will not be available prior to restore, so the SDK build would need to capture them and store locally - similar to what it does with "KnownFramework" information.  

However nothing should prevent other components specifying these, so long as they can specify them _before_ restore.  IOW they can come from user files and SDKs but not packages.

accepted/2024/supplied-by-platform.md
nkolev92: GH's markdown renderer seems to break if the json file is not perfect. 

Fixed now.

nkolev92:
This spec has been reviewed and now all the unresolved questions are closed. 

@NuGet/nuget-client Can I get another look so I can merge this. 
I'm starting to PR the implementation and I want to merge the spec before that.

jeffkl (approved):

PR #13673, opened 2024-08-03
Add spec for Trusted Publishers (OIDC auth for NuGet)

joelverhagen:
See the [rendered functional description](https://github.com/NuGet/Home/blob/dev/accepted/2024/trusted-publishers-oidc-for-nuget-push.md) (main spec) and [rendered technical description](https://github.com/NuGet/Home/blob/dev/accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md) (supporting document).

This is a spec for https://github.com/NuGet/NuGetGallery/issues/9332.

It brings OpenSSF's https://repos.openssf.org/trusted-publishers-for-all-package-repositories document to NuGet.org.

Thank You 🎉


JonDouglas:
Here are some additional responses of ~n=100-200 on later questions on the recent survey we ran:

image

image

image

accepted/2024/meta/trusted-publishers-oidc-for-nuget-push/trust-policy-form.png
JonDouglas, line 1: Do we need a branch item here? i.e.

image

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,523 @@
...
+The screenshot above allows only new package versions to be pushed for packages owned by `joelverhagen`, having IDs
+starting in `Knapcode.`. New package IDs cannot be created in this case nor can packages be unlisted.
+
+Unlike API key definitions, there is no "Expires In" field, because a trust policy never expires.
+
+Multiple trust policies can be defined. If any of the trust policies match the incoming package and GitHub Action token,
+the operation will be allowed. If the package ID is new and there is ambiguity on which package owner should be assigned
+to the package (due to multiple matching trust policies), the most recently created trust policy will be used.
+
+### Step 2: add a GitHub Actions workflow with NuGet.org Trusted Publisher authentication
```
JonDouglas, line 300: I assume we will want to have a NuGet GH action template on the marketplace to simplify this beyond docs?

i.e. https://github.com/rubygems/rubygems.org/issues/4286

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,523 @@
...
+existing long-lived API key flow, but this can still be confusing. If we find this flow doesn't work well for users, an
+additional parameter can be added to the `nuget/login` step which is `owner`. If provided, this will assert that the
+desired package owner is `X` (such as an organization name). If no trust policy is found on the user for that package
+owner, a helpful error could be provided. We can wait on implementing this until we hear more from users.
+
+## Rationale and alternatives
+
+### Naming
+
+The name "trusted publishers" may seem overly vague. There has been conversations about this [on Hacker
```
JonDouglas, line 439: 💯 the name is so confusing but for consistency's sake may be best to use it or reference it regularly.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,523 @@
...
+- Trusted Publishers has been implemented by:
+  - Python Package Index, a.k.a. PyPI ([blog](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/),
+    [docs](https://docs.pypi.org/trusted-publishers/))
+  - RubyGems ([blog](https://blog.rubygems.org/2023/12/14/trusted-publishing.html),
+    [docs](https://guides.rubygems.org/trusted-publishing/))
+  - pub.dev ([docs](https://dart.dev/tools/pub/automated-publishing))
+
+## Unresolved Questions
+
+Ask away!
```
JonDouglas, line 703: There are no desired UI elements to indicate to end-users that a package has enabled this functionality today right?

Would we want to consider parts of the Build Provenance UI elements here?

image

i.e. Source Commit and Build File of the "trusted publisher"

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,523 @@
...
+
+## Future Possibilities
+
+- Enable additional CI/CD systems that support OIDC tokens
+  - Azure DevOps is next on the popularity list. Their workload identity federation is very focused on Azure today
```
JonDouglas: +1 to AzDo based on usage and qualitative data posted in this PR thread as a comment screenshot.

JonDouglas (approved):
Looks great. Awesome details and data driving this! Let me know when we can amplify and get more eyes on it. The functional bits look good to me and added comments as consideration points.

accepted/2024/meta/trusted-publishers-oidc-for-nuget-push/trust-policy-form.png
joelverhagen: The branch component of the `sub` identifier gets overridden when someone sets the `environment` in their YAML. But the environment isn't there by default. I'm not sure if we should only allow folks to specify which "channel" or "track" the packages come from via environment. It seems more `git` flavored to also allow a branch filter which will be possible on nearly all workflows out of the box vs. an environment specifier.

See https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment

We could have a series of checkboxes. "Filter by environment: ___________", "Filter by branch: __________", "Filter by workflow file: __________". At least one would be required but you could filter by all 3 if you want.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Are you referring to a workflow template like this?
https://docs.github.com/en/actions/using-workflows/creating-starter-workflows-for-your-organization
Or a GitHub Action to do pack + auth + push all in one?

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: As part of this design, no.

In your example screenshot there is a green check mark and the word "Signed" as well as a ledger entry. The green check mark only makes sense if there is some tamper protection or strong build provenance consideration (SLSA build level 2+ IMO). 

We could perhaps show "Published from GitHub Actions", with the source commit and build file links, but it's not strong in the same sense that npm build provenance is strong where the artifacts are signed inside the workload, rather than just published.

See the latter paragraphs in the [Auditing usage on NuGet.org](
https://github.com/NuGet/Home/blob/jver-oidc/accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md#auditing-usage-on-nugetorg) section to see what I think we should commit to in this spec. Essentially, it's just recording the token (or some useful part of it) for later mining.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
JonDouglas: More so on the public marketplace:

https://github.com/rubygems/release-gem?tab=readme-ov-file#requirements -> https://github.com/marketplace/actions/release-gem

joelverhagen: The `nuget/login` action in the design would be on the public marketplace. But it would only do the OIDC -> short-lived API key trade, and not the push step. ... but maybe it could? It would need a parameter to specify which .nupkgs to push. 

I think it's better to keep the public marketplace action to be minimal, and re-usable for multiple use cases. If the action does `nuget push` internally, then we may end up needing all of the CLI parameters that go into `nuget push`, and add them as push gets more complex.

I would prefer `nuget/login` public marketplace action and not `nuget/release` for now. This also makes using the short-lived API key for unlist/relist and even deprecate API easier since all `nuget/login` does is produce an API key. It's not hardcode to always push.

accepted/2024/meta/trusted-publishers-oidc-for-nuget-push/trust-policy-form.png
joelverhagen: I added clarification about the 3 filtering options (branch vs environment vs workflow file name)

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: I hope we can get Azure DevOps integration. That would help many Microsoft teams who publish from Azure DevOps to NuGet.org (e.g. NuGet client team, NuGet.* packages!).

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+| OIDC              | OpenID Connect, an authentication protocol build on OAuth 2.0, think of it as a way for NuGet.org to trust GitHub Actions via a JWT included in package upload                            |
+| Pack              | The name of the package creation operation, produces a .nupkg and optionally a .snupkg                                                                                                    |
+| Package source    | A NuGet.org package feed, or package registry (all synonymous), a destination for push operations                                                                                         |
+| Push              | The name of the package upload operation in NuGet ecosystem                                                                                                                               |
+| Trusted Publisher | An name for a CI/CD environment that can generate OIDC tokens, using workload identity, [described by OpenSSF](https://repos.openssf.org/trusted-publishers-for-all-package-repositories) |
```
dtivel:
```suggestion
| Trusted Publisher | A name for a CI/CD environment that can generate OIDC tokens, using workload identity, [described by OpenSSF](https://repos.openssf.org/trusted-publishers-for-all-package-repositories) |
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+the rich context and prior art.
+
+## Terminology
+
+There is a lot of lingo in this document, so here is a reference:
+
+| Term              | Definition                                                                                                                                                                                |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| CI/CD             | continuous integration/continuous deployment, technologies like GitHub Actions for building or deploying code automatically                                                               |
+| Forge             | A place to store code, often a SaaS offering like GitHub, GitLab, or Bitbucket, often has a CI/CD offering paired                                                                         |
```
dtivel: Consider lower casing terms like this which are neither acronyms nor proper nouns.  Capitalization suggests a proper noun, a specific instance of something, which this isn't.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+     temporary API key.
+
+Let's expand each of these steps for better understanding.
+
+### Prerequisite: set up a GitHub repository for the NuGet package
+
+This step is pretty self explanatory. Many developers, even hobbyist developers who are only producing packages for
+their own consumption, already have their source code stored in a Git repository, hosted on GitHub.
+
+Additionally, GitHub Actions must be enabled for your repository. For public repositories, there is a generous free
```
dtivel, line 121: You should be explicit on whether a repository can be private.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+So, it's a usability win and a security win!
+
+## Functional explanation
+
+At the core of this design is a requirement. The package author must publish their package to NuGet.org from an
+environment that is supported by the Trusted Publishers feature. The build environment must support OpenID Connect
+(OIDC) tokens because they take the place of long-lived NuGet API keys as the credential. The NuGet package source must
+also be configured to allow OIDC tokens from this specific build environment.
+
+For the sake of understanding, we'll use GitHub and GitHub Actions as the example below, but it's important to note that
```
dtivel, line 90: Are there any minimum expectations on a forge, in terms of security, transparency, or whatever?

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+To scope the ways in which a GitHub trust policy can be used, similar scoping rules to the current API key flow can be
+set on the GitHub Actions trust policy:
+
+![Scope form](meta/trusted-publishers-oidc-for-nuget-push/scope-form.png)
+
+Just like the existing API scoping feature, there are 3 dimensions which a trust policy can be scoped. For more
+information about scopes on API keys, see [Scoped API
+keys](https://learn.microsoft.com/en-us/nuget/nuget-org/scoped-api-keys). In summary:
+
+- **Package owner**: for users that are members of one or more organizations, it's important to know which package owner
```
dtivel: Suppose you transfer ownership of a package to me.  (You make me a coowner, and then I remove you.)  How would I remove your ability to continue publishing packages with your trusted publisher policy?

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,538 @@
...
+   - Publish the package!
+
+## Technical explanation
+
+See the [technical document](trusted-publishers-oidc-for-nuget-push.technical.md) in this case directory for more
+information.
+
+## Drawbacks
+
+This workflow only helps users who want to build and push their packages on a recognized, Trusted Publisher. Not all
```
dtivel:
```suggestion
This workflow only helps users who want to build and push their packages on a recognized Trusted Publisher. Not all
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: The moment you are removed as an owner, your trust policy will no longer work. The trust policy is moreso associated with a package owner rather than a package. In fact, the only this package-specific about a trust policy (or long lived API key) is a package ID scope.

Each time you upload a package version (via web, API key, or Trusted Publisher), whether or not the **Package owner** in question (the package owner property listed here) is a current, direct owner of the package. If it is not, the upload will be blocked.

I updated the spec to clarify this.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: I updated the future possibilities section with some vague details. I think the reality is that we will evaluate it on case-by-case basis. GitHub is a trusted, industry wide service so trusting their OIDC tokens is straight forward. If there is a request for a Trusted Publisher we, nor many of our users, know anything about, we will probably be a little skeptical and ask a lot of questions.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Added! As long as the token is valid and issued from GitHub Actions (the known issuer base URL `https://token.actions.githubusercontent.com`), we will accept it. This can include private repositories.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Done.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+- `repository_id` claim matches the numeric repository ID recorded at the time of the trust policy creation
+  - This is to avoid resurrection attacks.
+- If a branch filter is provided in the trust policy:
+  - `ref_type` claim must be `branch`
+  - `ref` claim must be `refs/head/{branch}` (case insensitive)
+- If an environment filter is provided in the trust policy:
+  - `environment` claim must be `{environment}` (case insensitive)
+- If a workflow path filter is provided in the trust policy:
+   - `job_workflow_ref` claim must be `{repo owner}/{repo name}/{workflow path}@.*` (case insensitive)
+   - The workflow path should be normalized to `/` path separators at the time of trust policy creation
```
dtivel: Is this NuGet.org's responsibility?  Is it as simple as replacing all `\` with `/`?

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+  - valid `aud` claim, being `nuget`
+
+The
+[Microsoft.IdentityModel.Protocols.OpenIdConnect](https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect)
+package can help us properly validate JWTs in our NuGet/NuGetGallery ASP.NET application.
+
+In addition to the general JWT checks, specific checks are made for each Trusted Publisher. For GitHub Actions, the
+following checks will be made:
+
+- `sub` claim matches `{repo owner}/{repo name}:.*` (case insensitive)
```
dtivel: It's a bit difficult for me to understand what's variable and what's constant.  For example, this might be reworded to:

`sub` claim matches the regular expression pattern `{repo owner}/{repo name}:.*` where `{repo owner}` is the repository owner's name and `{repo name}` is the repository's name.  Then, I would understand `dtivel/myrepo:somepath` as a matching pattern.

(I inferred that `*` is something of a wildcard or maybe `.*` is a regular expression pattern, but I can't be certain from the spec.)

What would an actual `sub` look like?

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+  - The suffix of the `sub` is implied by the other checks
+- `repository_owner` claim matches the `{repo owner}` name (case insensitive)
+- `repository_owner_id` claim matches the numeric owner ID recorded at the time of the trust policy creation
+  - This is to avoid resurrection attacks.
+- `repository` claim matches the `{repo name}` name (case insensitive)
+- `repository_id` claim matches the numeric repository ID recorded at the time of the trust policy creation
+  - This is to avoid resurrection attacks.
+- If a branch filter is provided in the trust policy:
+  - `ref_type` claim must be `branch`
+  - `ref` claim must be `refs/head/{branch}` (case insensitive)
```
dtivel: GitHub branch names are case sensitive.  You can have branches `test` and `TEST` simultaneously.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+   "expires": "{ISO 8601 timestamp of expiration}"
+}
+```
+
+Authorization failures on this endpoint must return HTTP 401 Unauthorized with an `WWW-Authenticate: Bearer` response
+header.
+
+The package source MUST NOT opt to return the an existing compatible API key (i.e. it must not cache the API key for
+subsequent calls). To do so would require the original API key to be stored in plain text. NuGet.org API keys are hashed
+prior to storage (much like standard recommendations around storing passwords). The the package source has concerns on
```
dtivel:
```suggestion
prior to storage (much like standard recommendations around storing passwords). The package source has concerns on
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+{
+   "api_key": "{short lived API key in clear text}",
+   "expires": "{ISO 8601 timestamp of expiration}"
+}
+```
+
+Authorization failures on this endpoint must return HTTP 401 Unauthorized with an `WWW-Authenticate: Bearer` response
+header.
+
+The package source MUST NOT opt to return the an existing compatible API key (i.e. it must not cache the API key for
```
dtivel:
```suggestion
The package source MUST NOT return an existing compatible API key (e.g. it must not cache the API key for
```

Returning an existing, long-lived API should also be forbidden.  (That's why I changed i.e. to e.g., because caching isn't the only reason.)

You could even explicitly state this:  The package source MUST NOT return an existing API key and MUST generate a new one on demand.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+subsequent calls). To do so would require the original API key to be stored in plain text. NuGet.org API keys are hashed
+prior to storage (much like standard recommendations around storing passwords). The the package source has concerns on
+scalability it must opt to rate limit the endpoint instead of caching. NuGet.org will rate limit the endpoint to 1 API
+key created per 30 seconds, per user.
```
dtivel: I don't grok how this limit aligns with someone publishing many packages in a short period of time.  Consider the .NET team publishing packages in bulk to NuGet.org.  Will that translate to hundreds or thousands of API calls within a short timeframe?  Or, is the API key cached and reused on the client side?

It would be good to discuss here how you see this limitation not being a problem for various publishing scenarios.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+```
+
+The response will look like this:
+
+```
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+   "api_key": "{short lived API key in clear text}",
```
dtivel: I didn't see any explicit mention of how long the short-lived API key will be valid.  I think that's intentional, since we should have some flexibility in adjusting its lifespan.  However, for the sake of argument, what is a reasonable starting point?  30 minutes?

What is the user experience if the short-lived API key expires before they're done publishing and how do they recover?  Or, is the client (NuGet client, here) supposed to automatically reacquire a new API key transparently so that the user never notices the expiration?

I'm trying to understand where this automagical flow might actually break down for users.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+[Microsoft.IdentityModel.Protocols.OpenIdConnect](https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect)
+package can help us properly validate JWTs in our NuGet/NuGetGallery ASP.NET application.
+
+In addition to the general JWT checks, specific checks are made for each Trusted Publisher. For GitHub Actions, the
+following checks will be made:
+
+- `sub` claim matches `{repo owner}/{repo name}:.*` (case insensitive)
+  - The suffix of the `sub` is implied by the other checks
+- `repository_owner` claim matches the `{repo owner}` name (case insensitive)
+- `repository_owner_id` claim matches the numeric owner ID recorded at the time of the trust policy creation
```
dtivel, line 41: This numeric owner ID is forge-assigned (here GitHub-assigned)?

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,189 @@
...
+  - This is to avoid resurrection attacks.
+- `repository` claim matches the `{repo name}` name (case insensitive)
+- `repository_id` claim matches the numeric repository ID recorded at the time of the trust policy creation
```
dtivel, line 44: This numeric repository ID is forge-assigned (here GitHub-assigned)?

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: I think it can be NuGet.org's responsibility, as a nicety. It would be very annoying to get auth errors, just because you copied the relative path from a Windows machine. Yes, it should be that simple.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: I expressed it as a regex. You're right, that's confusing. I'll clarify that and add a sample.

A branch `sub` would look like this: `repo:octo-org/octo-repo:ref:refs/heads/demo-branch`.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Great catch, I will make that case sensitive!

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: I think both the MUST NOT and MUST as you state is the clearest. Thanks!

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: > Will that translate to hundreds or thousands of API calls within a short timeframe? Or, is the API key cached and reused on the client side?

I expect the API key to be cached on the client side. In most workflows (no custom scripting), this will be done by a single `nuget/login` workflow step with a single output value, which is the short lived API key. This could be used one or more push operations (all referring to the single step output == single API key). I imagine most users will do `dotnet nuget push *.nupkg` which internally does multiple push operations with a single API key provided via CLI arg.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: > I didn't see any explicit mention of how long the short-lived API key will be valid. I think that's intentional, since we should have some flexibility in adjusting its lifespan. However, for the sake of argument, what is a reasonable starting point? 30 minutes?

I actually thought I documented. I did not 😂. I was planning on using 15 minutes. This is what PyPI uses and it should have good coverage for our push patterns. I will add supporting data.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Yes, GitHub assigned.

I should note that other forges may not have a numeric owner ID. We will need to assess resurrection attacks or any other impersonation risks on a case by case basis.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Yup, same as above https://github.com/NuGet/Home/pull/13673#discussion_r1711866688.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+      Owners = count_distinct(tostring(Owners)) by Forge
+   | order by ['Package IDs'] desc;
+   let TotalPackageIds = toscalar(Summary | summarize sum(['Package IDs']));
+   let TotalUniqueOwnersSets = toscalar(Summary | summarize sum(Owners));
+   Summary
+   | extend ["% of package IDs"] = round(100.0 * ['Package IDs'] / TotalPackageIds, 2)
+   | extend ["% of owners"] = round(100.0 * Owners / TotalUniqueOwnersSets, 2)
+-->
+
+| Forge        | Package IDs | Owners | % of package IDs | % of owners |
```
Hayden-IO: It might be helpful to also note which of these platforms support OIDC for their workflows.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+        with:
+          user: joelverhagen # this is the user profile which has the GitHub Actions trust policy
+          source: https://api.nuget.org/v3/index.json
+      
+      # this is exactly the same as before, but uses a short-lived API acquired on the fly
+      - name: NuGet push
+        run: dotnet nuget push artifacts/*.nupkg -k ${{steps.login.outputs.NUGET_API_KEY}} -s https://api.nuget.org/v3/index.json
+```
+
+This experience could be further improved with changes in the .NET SDK, such as allowing the push API key to come from
```
Hayden-IO, line 358: Have you considered accepting the OIDC credentials on the push API to avoid the need to exchange them for a short-lived API key?

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+See the [technical document](trusted-publishers-oidc-for-nuget-push.technical.md) in this case directory for more
+information.
+
+## Drawbacks
+
+This workflow only helps users who want to build and push their packages on a recognized Trusted Publisher. Not all
+CI/CD systems will be supported so some users will still need to use the older authentication flows such as long-lived,
+manually rotated API keys.
+
+The trust policy is configured once and works forever. An API key forces the package author to re-assess CI/CD
```
Hayden-IO, line 419: Thoughts on if you would let a package "downgrade" to move off trusted publishing? And a similar question, would you allow both to be configured, or only one?

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+- Is there transparency from the Trusted Publisher around the issued tokens?
+  - In other words, is it very clear to us and the even package consumers that a token is a sufficient replacement for
+    an API-key from a security perspective.
+
+### Enable NuGet authentication provider integration, to avoid token expiration for long workflows
+
+We will wait for evidence that our selected short-lived token lifetime duration is not sufficient for all workflow push
+operations. As mentioned in the technical document, we will start with 15 minutes and adjust later if needed.
+
+### Connect this effort with build provenance, tracked by [NuGet/Home#13581](https://github.com/NuGet/Home/issues/13581)
```
Hayden-IO, line 751: 🥳

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,244 @@
...
+  - valid signature via JWKS
+  - validate duration (`nbf` and `exp` claims)
+  - only used once, via `jti`
+  - valid `aud` claim, being `nuget`
+
+The
+[Microsoft.IdentityModel.Protocols.OpenIdConnect](https://www.nuget.org/packages/Microsoft.IdentityModel.Protocols.OpenIdConnect)
+package can help us properly validate JWTs in our NuGet/NuGetGallery ASP.NET application.
+
+In addition to the general JWT checks, specific checks are made for each Trusted Publisher. For GitHub Actions, the
```
Hayden-IO, line 33: If it's helpful, for build provenance for Sigstore, we mapped token claims to what they represent for each CI provider - https://github.com/sigstore/fulcio/blob/main/docs/oid-info.md#mapping-oidc-token-claims-to-fulcio-oids. This mapping might help if you want a table that maps claims for different providers.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,244 @@
...
+## Auditing usage on NuGet.org
+
+To help us understand the success of this feature and record priviledged actions for security auditing, we will at least
+record minimal information about the OIDC token trade, such as the repository owner, repository name, workflow path,
+etc. Existing auditing for API key usage will be used anywhere the short-lived API key is used.
+
+The metadata available in a GitHub Actions OIDC token is mentioned in GitHub's [Understanding the OIDC
+token](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token).
+
+These claims could offer useful indicators to package consumers about the package. In order to support a future effort
```
Hayden-IO: This comment is orthogonal for trusted publishing, but I might avoid discussing SLSA Build L1 - It doesn't add much given it's not a signed attestation and if the registry was constructing the provenance, then it requires trust in the registry that it properly recorded the claims, rather than the builder.

Hayden-IO (commented):
This is a fantastic write up! Excited to see this spec!

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+
+Once the package author has added this trust policy, NuGet.org will trust GitHub Actions OIDC tokens matching this
+pattern much like an existing long-lived API key, with the added benefit of the trust policy never expiring.
+
+NuGet package sources must ensure a secure, authenticated session for the user that is adding the trust policy.
+NuGet.org requires two-factor authentication for all web UI sign-ins.
+
+#### What NuGet API operations can be used inside the GitHub Actions workflow?
+
+To scope the ways in which a GitHub trust policy can be used, similar scoping rules to the current API key flow can be
```
skofman1: I recommend investigating the current usage of API Key scopes to understand how they are used. For example, are globs commonly used? Are there API keys that only allow pushing new versions? Are they common? I have a suspicion that this functionality isn't used, so we should check before carrying it forward. We might want to make changes / simplify the MVP.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,568 @@
...
+Repositories](https://repos.openssf.org/trusted-publishers-for-all-package-repositories), by Seth Larson
+- Trusted Publishers has been implemented by:
+  - Python Package Index, a.k.a. PyPI ([blog](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/),
+    [docs](https://docs.pypi.org/trusted-publishers/))
+  - RubyGems ([blog](https://blog.rubygems.org/2023/12/14/trusted-publishing.html),
+    [docs](https://guides.rubygems.org/trusted-publishing/))
+  - pub.dev ([docs](https://dart.dev/tools/pub/automated-publishing))
+
+## Unresolved Questions
+
```
skofman1, line 702: @joelverhagen , perhaps not directly related to this spec, but something to consider: we raise the security bar for package publishing only when we disable API Key publishing for this package. Following this logic, when a package owner opts-in to OIDC publishing, we should disable API Key publishing for affected packages. There can be many approaches here, and we should have a conversation when we talk about roll out strategy here.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Great point. I'll do that research and include a little note in the table.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: I did consider that, and even had that in my prototype. I brought this up with the OpenSSF folks and was convinced that the two-legged approach is better. I put some notes on this subject in the alternatives section at the bottom of the doc.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Hmm good question! I think starting out we would allow full flexibility in a "walk-crawl-run" approach. As we mature, we see how the feature goes in our ecosystem, we can lock things down more. We took a similar approach with requiring 2FA on our website. At first it was opt in, then opt out (new accounts had it enabled), then it was mandated. This allowed us to look at customer feedback as we went and not move so quickly that folks got stuck.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Good point. I think I will do this later when we consider a second Trusted Publisher platform. Then we can compare apples to apples with GitHub Actions.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: That's a fair point. I do want to mention SLSA Build levels to come extent, at least to say this does NOT provided L2+! I talked to Zach S. on the OpenSSF Slack. He said using this information would be L1 but it's not great since it is tamperable. I would say this metadata is better than package author-attested repo/project URL since the consumer needs to just trust the registry and GitHub Actions, but it's not as good as a proper signed provenance thing.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: Great point. I was assuming parity would be the easiest since we would be stamping out short-lived API keys from the OIDC token, so all of the existing API key scoping can apply. But there is still cost to adding these parameters to the trust policy definition and making the UI work and look pretty+accessible. I will check the usage and modify the spec if adoption is "low".

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: @haydentherapper (also participates in OpenSSF) mentioned this here:
https://github.com/NuGet/Home/pull/13673#discussion_r1714296911

I think we should add that as a future opportunity and not plan to do it immediately or in very short order after initial release. Perhaps 3 months+ after initial release or when we feel like we have a strong signal (positive or negative) from our users on how adoption is going in their projects.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: I looked at the data. I think we can definitely get by without package ID and allowed action scoping, especially since the API keys are short-lived.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen: I added a section about this.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -505,11 +485,195 @@ CI/CD workflows. In the end, this approach could support GitHub Actions via
...
+| ---- | ------------ | ------ | ---------- | -------------------------------------------------- |
+| 1    | 0            | 0      | 79.76      | Push new IDs and update existing IDs               |
+| 1    | 0            | 1      | 12.58      | Push new IDs and update existing IDs, allow unlist |
+| 0    | 1            | 0      | 6.77       | Update existing IDs                                |
+| 0    | 0            | 1      | 0.48       | Allow unlist                                       |
+| 0    | 1            | 1      | 0.41       | Update existing IDs, allow unlist                  |
+
+Note that "Push" implies "Push version" allowed action.
+
+The first two rows should be considered "high priviledge" API keys. They account for about 93% of the API keys. This
```
dtivel: priviledge -> privilege

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -505,11 +485,195 @@ CI/CD workflows. In the end, this approach could support GitHub Actions via
...
+| Exact match, multiple | 6.92       | `NuGet.Protocol`, `NuGet.Frameworks` |
+| Glob and exact match  | 0.65       | `Microsoft.*`, `NuGet.Protocol`      |
+
+This form of scoping based on package ID has a lot more variety than the previous section. Close to half of users go
+very broad with their scoping (`*`, allow all package IDs). The 2nd place scoping is very specific: only a single
+package ID can be modified with the API key.
+
+The current UI defaults to an empty glob pattern text field and a list of visible checkboxes for all of the package IDs
+the user has upload permissions for. This requires the user to provide *some* input. It's not clear whether this UI
+gently encourages users to click the checkboxes since that's probably quicker than typing. I were to guess, the adoption
```
dtivel:
```suggestion
gently encourages users to click the checkboxes since that's probably quicker than typing. If I were to guess, the adoption
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -125,33 +136,70 @@ These short-lived API keys will not be visible in the NuGet.org UI.
 
 NuGet.org MAY record the JWT and related details (e.g. JWKS) for auditing and feature adoption purposes.
 
+NuGet.org will produce API keys that last 15 minutes, but this value is subject to change as we learn more about how
+this feature is used in practice. PyPI uses this duration ([source](https://docs.pypi.org/trusted-publishers/)). Also,
+15 minutes will support about 99% of push sessions on NuGet.org. For the sake of this analysis, a push session is a
+sequence of push operations from a distinct package owner set, where the pushes are no more than 5 minutes apart. Below
+is a table of push sessions of various durations. The difference between 15 minute and 1 hour API key life givens less
```
dtivel:
```suggestion
is a table of push sessions of various durations. The difference between 15 minute and 1 hour API key life gives less
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
joelverhagen:
```suggestion
The first two rows should be considered "high privilege" API keys. They account for about 93% of the API keys. This
```

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
dtivel: Suppose `dotnet nuget push *.nupkg` takes longer than 15 minutes.  Imagine lots of packages, slow network, whatever.  Will NuGet client automatically re-request a new token via `nuget/login` midstream, as necessary?

dtivel (approved):
I'm partial to calling this "OIDC Publishers" over "Trusted Publishers", but otherwise this spec looks good.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Not with this version of the proposal. See the https://github.com/NuGet/Home/blob/jver-oidc/accepted/2024/trusted-publishers-oidc-for-nuget-push.md#enable-nuget-authentication-provider-integration-to-avoid-token-expiration-for-long-workflows.

The user will be able to work around the issue with multiple `nuget/login` GitHub Actions steps and breaking up the push steps. It's not a great solution, but the data suggests we can pick a single short lifetime (15 minutes) and cover the vast majority of cases.

accepted/2024/trusted-publishers-oidc-for-nuget-push.md
```diff
@@ -0,0 +1,750 @@
...
+2. The GitHub repository name.
+3. Some filter on which GitHub Action workflow runs should be allowed to push packages.
+
+The filter on which workflow runs should be allowed to publish packages can potentially be very specific. GitHub Actions
+supplies a variety of metadata in the OIDC token available to workflow runs. Refer to [Understanding the OIDC
+token](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token)
+to see all of the properties.
+
+For the sake of simplicity, we will start with the following workflow filters:
+- Filtering for a specific *branch*, as shown in the UI above (branch filter)
```
martincostello: Will this also support tags? For example, when a tag is created `v1.2.3` and a workflow is kicked off automatically on tag creation to build that version and then publish it.

joelverhagen: Hmmm generally a tag trust policy should be possible: https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect#filtering-for-a-specific-tag

Question for you: would you want this feature? It seems like you would need to update the trust policy for every release as your version tag increments.

joelverhagen: I'll note the existing filters are meant to be "invariants" you set once and change infrequently (branch name, env name, workflow path). Tag seems different in this regard.

martincostello: I guess more what I'd be after is a policy that matches a tag with pattern (e.g. `:ref:refs/tags/v*`) rather than exact match.

My typical release process is to draft a release for a tag in GitHub that has a specific version (e.g. `v1.2.3`) with the release notes prepared etc. Then when I'm ready to ship, I publish that release which creates the tag. The triggers on my CI workflow are then setup like this:

```yml
on:
  push:
    branches: [ main ]
    tags: [ v* ]
  pull_request:
    branches: [ main ]
  workflow_dispatch:
```

Publishing then happens based on a condition like this:

```yaml
if: |
  github.event.repository.fork == false &&
  startsWith(github.ref, 'refs/tags/v')
```

The net effect is that the "real" build happens after tagging, and that packages get shipped off to NuGet.org once produced.

So in a vacuum, I would match on the tag rather than a branch as I wouldn't want _any_ push to main to be allowed to publish.

If tags aren't supported, then I would use an environment rather than a branch, as I currently use environments to scope the NuGet API key secret to that environment so it's not available to all the workflow runs (it also lets me manually approve the push job).

baronfel: I heartily endorse @martincostello's comments above - this is how _a very large_ set of OSS projects are versioned, not just in the .NET ecosystem but broadly across other ecosystems as well. Not supporting pattern-based, tag-based workflows would be a huge miss IMO.

joelverhagen: Thank you @martincostello and @baronfel for your input.

The full example helps a lot! Given branch, workflow file path, and environment do not have wildcard support, I think we should explicitly call out that tag filtering would support wildcard/pattern matching on the tag name. This is a reasonable idea and I will update the doc accordingly.

martincostello: Thanks. If it's useful to you, here are two full workflows (not 100% identical, as one has code signing) used to do NuGet publishing for [Polly](https://github.com/App-vNext/Polly/blob/main/.github/workflows/build.yml) and [one of my personal libraries](https://github.com/martincostello/xunit-logging/blob/main/.github/workflows/build.yml).

baronfel: It may be useful looking into branch wildcard support as well - I know that e.g. in dotnet we use pattern-based rules in many other places to signify some special status - e.g. for the SDK `^release/\d\.\d\.\d\d\d$` is a stable release branch and may be able to trigger certain workflows.

joelverhagen: I updated the document with wildcard support on tags and branches. This aligns with the pattern matching on package ID scopes. I am not sure if a regex is the right UX but I'm happy to hear other opinions on that!

jimmylewis: > I am not sure if a regex is the right UX but I'm happy to hear other opinions on that!

Glob patterns vs. regex have both been used as examples here, so I think it's worth clarifying which since the syntax is so similar for trivial cases.  A lot of places use glob patterns for branch matching, so there may be some confusion if regex is used.

Glob pattern above: `:ref:refs/tags/v*` (typically converted to regex: `:ref:refs/tags/v.*` though this is not precisely the same; it would be more accurate as `:ref:refs/tags/v[!/]*` though that's hardly intuitive)
Regex pattern above: `^release/\d\.\d\.\d\d\d$` (converted to glob: `release/[0-9].[0-9].[0-9][0-9][0-9]`)

Worth noting in the glob pattern example, typically glob expansion only happens within the current directory level, so `v*` will match `v1.2.3` but not `version/1.2.3`.  The simple regex conversion `.*` does not respect path delimiters, so `v.*` would match both `v1.2.3` and `version/1.2.3`.

baronfel: to clarify - I'd be perfectly happy for just globs, the regex was just meant to be a related example of branch and ref names being signals based on structure. I'd expect users to use `release/*` for example.

joelverhagen: Good point @jimmylewis, in my experience the term globbing is most often used for file system/file paths. I am not sure if the concept cleanly translated in users' minds to tag/branch names with slashes. IMO we can do something simple where `*` expands to `.*` regex and not worry about including slashes in that that capture.

Today "globbing" on NuGet.org API key scopes is like this:
https://github.com/NuGet/NuGetGallery/blob/a4b3c64acb65ca68eb67ac6b51242356aa5e4da7/src/NuGetGallery.Services/Extensions/ScopeExtensions.cs#L54-L58

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,253 @@
...
+acceptable to trade for a short-lived NuGet API key.
+
+The following checks are made:
+
+- The token is issued by a known Trusted Publisher (only GitHub Actions at this time)
+- The token is validated per all JWT rules, such as but not limited to:
+  - valid signature via JWKS
+  - validate duration (`nbf` and `exp` claims)
+  - only used once, via `jti`
+  - valid `aud` claim, being `nuget`
```
di, line 27: Note that although it's not true for GHA, with other IdPs it's possible to generate/provide an OIDC token with multiple `aud` claims which is supported by the OIDC spec: https://openid.net/specs/openid-connect-core-1_0.html#:~:text=REQUIRED.%20Audience(s,case%2Dsensitive%20string.

You may want to additionally verify that _only_ the `nuget` audience is included in the token, to ensure that users aren't potentially sharing nuget-enabled OIDC tokens with additional audiences.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,253 @@
...
+- `repository` claim matches the `{repo name}` name (case insensitive)
+- `repository_id` claim matches the numeric repository ID recorded at the time of the trust policy creation
+  - This is to avoid resurrection attacks.
+- If a branch filter is provided in the trust policy:
+  - `ref_type` claim must be `branch`
+  - `ref` claim must be `refs/head/{branch}` (case sensitive, branch names of differing case can coexist, branch match supports wildcards)
+- If an environment filter is provided in the trust policy:
+  - `environment` claim must be `{environment}` (case insensitive)
+- If a workflow path filter is provided in the trust policy:
+   - `job_workflow_ref` claim must be `{repo owner}/{repo name}/{workflow path}@.*` (case insensitive)
```
di, line 52: I _think_ this should be case sensitive? The workflow path at least is definitely case sensitive.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
```diff
@@ -0,0 +1,253 @@
...
+
+```
+POST /api/v2/token HTTP/1.1
+Host: www.nuget.org
+Authorization: Bearer {OIDC token}
+Content-Type: application/json
+
+{
+   "username": "{username of user with trust policy}"
+}
```
di, line 98: Mostly curious: why is `username` necessary here? Shouldn't the OIDC token always map 1:1 with a configuration/policy?

di:
> I'm partial to calling this "OIDC Publishers" over "Trusted Publishers", but otherwise this spec looks good.

This was actually the original name planned for the PyPI implementation of this spec, but my concern was that "OIDC" as a term was too unfamiliar and low-level for the average user, which might discourage more widespread adoption.

Note that [RubyGems also calls this Trusted Publishing](https://blog.rubygems.org/2023/12/14/trusted-publishing.htm) so there is some consistency there as well. I'm definitely biased though, since I came up with the name. 🙂

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Ah, nice call out. I didn't know multiple audiences was possible. Yes, there should be exactly one and it should be a well-defined "nuget" value.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: I chose to relax case sensitivity here due to Windows being a case insensitive file system. I don't feel too strongly, but it seems like a silly reason to reject a token JUST because the case was wrong, assuming everything else matches.

di: Since GitHub is case sensitive, it does mean that more than one workflow could gain the ability to publish. In practice I think that would be fairly uncommon (probably mostly indicative of a typo) but might be a surprising side effect.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: In this design a trust policy is attached to a user. Inside the trust policy, the user specifies which user/org the trust policy is scoped to (to handle the `“Pending” Trusted Publishers` problem mentioned in Seth's doc) -- a user can publish packages as themselves or to an organization they are a memeber of. So this `username` parameter specifies the user profile than holds the trust policy. 

There is no enforcement of 1:1 mapping planned. If we did want to enforce a 1:1 mapping we'd have to decide who (which NuGet.org user profile) gets preference. A first come first service basis would be weird too -- who would win?

I think is it better to allow users to set up whatever trust policy they want and require this additional parameter during token trade. The token trade would first fetch all trust policies for that username and see if any of them apply to the incoming token.

As an added benefit this would allow a future model where the trust policy is attached to an organization directly (instead of one of the admins). But this is not planned yet and we're more approaching this as a way to mint short-lived API keys which are only owned by users today.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
joelverhagen: Yes, that could be weird. I am not sure if GH has mitigations in place for multiple files of the same case. I'll listen for other input on this, and we can make a call during implementation.

accepted/2024/trusted-publishers-oidc-for-nuget-push.technical.md
di: Got it, so the configuration of the `nuget/login` would need to include a single username as well, correct? I'm imagining that it might be hard to determine who this user would be in the case of a project with multiple maintainers. Does this also mean that the projects other maintainers have no way to administer the publisher?

(For context, in PyPI's implementation, publishers map to projects, not users, and all owners/maintainers of a given project can change/configure publishers for that project. But that does require 'pending publishers', which is a little awkward UX-wise, but seems to work)

joelverhagen: Thank you for testing my logic here! I really appreciate it!

Yes, `nuget/login` would have a username provided. I have a bit of sample YAML in the spec (exact shape TBD):
```yaml
      - name: NuGet login
        uses: nuget/login@v1
        id: login
        with:
          user: joelverhagen # this is the user profile which has the GitHub Actions trust policy
          source: https://api.nuget.org/v3/index.json
```

> I'm imagining that it might be hard to determine who this user would be in the case of a project with multiple maintainers.

Yes, this is not great. But it is the current problem we have with "who owns the long lived" API key, so it's not worse than what we have today. If a user is concerned about sharing their user profile name it could be a GitHub secret variable `NUGET_USERNAME`. 

> Does this also mean that the projects other maintainers have no way to administer the publisher?

The term `publisher` is not well defined in NuGet world, but when used that typically refers to the visible owner (user/org) of the package. But I think based on that PyPI definition I think that a PyPI project is equivalent to a NuGet.org package ID (sometimes called package registration). I don't think we would want to attach trust policies to a package ID since orgs have many many package IDs (hundreds) and they would need to set the policy on all those places (painful). I guess .NET people like a bunch of little packages for some reason 😆.

Suppose a core contributor on a GitHub repo disappears and another contributor needs to publish a new version of the package(s). They would add an equivalent trust policy to their user account scoped to the org (or user co-owner) that owns the GitHub repo's packages and then specify their username in the token trade as we discussed above.

I think this is the PyPI + NuGet term mapping (did this mainly to get my head straight)

PyPI | NuGet.org | Notes 
--- | --- | ---
Project | Package ID / Package Registration
Release | Package version | Sometimes just called "package" in NuGet, confusingly
Collaborator / Maintainer (?) | Owner | Can be a user or organization on NuGet.org, not sure about PyPI
Organization | Organization | Looks like creating these has an approval flow on PyPI, unlike NuGet.org
User | User | These seem to have equivalent usages

joelverhagen:
This spec is in good shape to merge and has several approvals. Thank you everyone for your input! 

There is no ETA yet for when this work will be complete but I have begun implementation of some common components needed for both first party (internal) package publishing and 3rd party (GitHub Actions) package publishing using OIDC.

Future design changes or clarifications can come as separate PRs.

Issue #13658, opened 2024-07-26
NuGetAudit should check PackageDownload


zivkan:
### NuGet Product(s) Involved

NuGet.exe, Visual Studio Package Management UI, MSBuild.exe, dotnet.exe

### The Elevator Pitch

NuGet has a feature intended for MSBuild SDKs to request NuGet to download a package on their behalf, but don't do any asset selection, package compatibility checks, or anything else. Just plain old download to a known location. The .NET SDK uses this to download platform specific runtime hosts, such as [Microsoft.NETCore.App.Runtime.linux-x64](https://www.nuget.org/packages/Microsoft.NETCore.App.Runtime.linux-x64/).

These packages can have known security vulnerabilities. Therefore, Audit should warn customers about these.

### Additional Context and Details

original feature spec, which might give some context for why it was added: https://github.com/NuGet/Home/wiki/%5BSpec%5D-PackageDownload-support

marcpopMSFT:
For this item, we may want to coordinate because some of the solutions are for customers to upgrade their SDK. We probably want to ensure customers get a specific error message for that case rather than one about a vulnerable package.

PR #13904, opened 2024-11-02
Enhancing NuGet Audit for Secure and Selective `<PackageDownload>` Vulnerability Reporting

### LINKED ISSUE #13905: Design Spec for Expanding NuGet Audit to Support <PackageDownload> Packages
Create a design spec detailing the expansion of NuGet Audit to include <PackageDownload> packages, with exclusions for SDK-managed dependencies.

Nigusu-Allehu:
Fixes: https://github.com/NuGet/Home/issues/13905

Issue https://github.com/NuGet/Home/issues/13658

Rendered: https://github.com/NuGet/Home/blob/ed9318029eda7ad534d58a0e6c44000da602bcef/accepted/2024/NuGet-Audit-for-PackageDownload.md


Nigusu-Allehu:
Related issue https://github.com/dotnet/sdk/issues/44421

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
+# ***Enhancing NuGet Audit for Secure and Selective `<PackageDownload>` Vulnerability Reporting***
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- Author Name [Nigusu](https://github.com/Nigusu-Allehu)
+- [GitHub Issue](https://github.com/NuGet/Home/issues/13658)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+In our effort to strengthen the security of NuGet, we would like to expand the support of NuGet Audit to `<PackageDownload>` packages. This will be done by introducing vulnerability errors for packages downloaded using `PackageDownload`. However, the .Net sdk also makes use of `PackageDownload` to implicitly download packages. And we would like to make sure these packages are not also receiving these errors, in order to prevent customers being confused. This will be done by introducing a new attribute to `PackageDownload` that makes it clear to NuGet Audit it should not write an error for these packages.
```
zivkan: Please put at new line at the end of each sentence, so it's one sentence per line.  This makes it much easier to comment on specific things

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Motivation 
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+We are expanding NuGet Audit to `<PackageDownload>` packages to improve security notifications for user-specified dependencies, addressing gaps in vulnerability alerts that currently miss these packages. This change prevents confusion by excluding SDK-managed PackageDownloads from error reports, ensuring users only see relevant security issues. The expected outcome is a more precise and useful vulnerability reporting experience, enabling users to address actual risks without unnecessary SDK-related alerts.
```
zivkan: > This change prevents confusion by excluding SDK-managed PackageDownloads from error reports, ensuring users only see relevant security issues.

The SDK adding a PackageDownload for a package with a known vulnerability is relevant. This is the purpose of Audit, to inform the customer that packages with known vulnerabilities are being used.

Unless customers get notified some other way that they're using an out of date SDK, that is including content with known vulnerabilities in their builds, I disagree fundamentally with this proposal.

zivkan (changes requested):

accepted/2024/NuGet-Audit-for-PackageDownload.md
Nigusu-Allehu: Regarding PackageDownload from the SDK here is the idea
> We should come up with an experience that notifies the user that the SDK is out of date and that they may be vulnerable to security issues. This experience should probably not even refer to PackageDownloads, as that’s an implementation detail of the SDK. We could have the same or similar experience for notifying that the SDK is out of date that isn’t driven by NuGet audit (and doesn’t depend on the project needing a PackageDownload).

 from the discussion in https://github.com/dotnet/sdk/issues/44421.

zivkan: The spec should explain this. To me, it currently reads like we're going to leave gaps in auditing. It should be clear that customers will get different notifications about these packages.

accepted/2024/NuGet-Audit-for-PackageDownload.md
nkolev92: nit: People generalize capitalize `.NET SDK`

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+<ItemGroup>
+    <PackageDownload Include="NuGet.Protocol" Version="[5.11.2]" />
+</ItemGroup>
+```
+
+If "NuGet.Protocol" is found to have a vulnerability, NuGet Audit will issue a warning to alert you so you can update to a safer version.
+
+##### Warning NU1906
+
+> warning NU1906: Package 'NuGet.Protocol' 5.11.2 has a known moderate severity vulnerability, https://github.com/advisories/GHSA-g3q9-xf95-8hp5
```
nkolev92: Did you consider using the same warning codes as PackageDownload?
We don't have strict rules around the warning codes, but we do it based off the UX we'd want people to have. 
Would we have situations in which people want to do something about PackageDownload but not about PackageReference or vice versa? Would the suppressions functionality cover that gap?

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+<ItemGroup>
+    <PackageDownload Include="System.Text.Json" Version="[1.0.0]" IsImplicitlyDefined="true"/>
+</ItemGroup>
+```
+
+Since this was implicitly added, no warnings are received by the user.
+
+#### Functional Design
+
+Here’s a pseudocode version of how this exclusion works:
```
nkolev92: nit: For future specs, you probably don't need go into this much detail regarding the code that "works as expected".
Would save you some time, and some visual real estate when the spec is read.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+The following technical components will support the proposal's functionality:
+
+1. **New Attribute for SDK-Specific PackageDownloads:** A new attribute, named `IsImplicitlyDefined="true"`, will be added to <PackageDownload> entries within the SDK. This will also be done in project-system to make sure VS nominations are working in a similar manner. Add the property [here](https://github.com/dotnet/project-system/blob/9f35656ad68aa1352d7b6b0fd01784f7aefe1005/src/Microsoft.VisualStudio.ProjectSystem.Managed/ProjectSystem/Rules/CollectedPackageDownload.xaml#L3) for project-system.
+
+1. **NuGet Audit Update to Handle Attribute:** NuGet Audit will be modified to recognize the new attribute and exclude SDK-flagged PackageDownloads from security alerts. This update ensures that only relevant <PackageDownload> packages, added directly by the user, are scanned for vulnerabilities, allowing users to remain informed without SDK-related alerts.
```
nkolev92: Historically IsImplicitlyDefined's impact had been *preventing* modification in the project file. 
PackageReference with IsImplicitlyDefined attributes will still report vulnerabilities correct? 

I'm wondering what other options we considered here.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
```
nkolev92, line 118: Most of the time, alongside a proposal, there a bunch of other ideas that we opt against. 
It'd be great to list some of those considerations.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+foreach (package in PackageDownloads):
+    if (package.HasAttribute("IsImplicitlyDefined") && package.IsImplicitlyDefined == true):
+        continue  // Skip auditing, as it's an SDK-managed package
+    else:
+        checkForVulnerabilities(package)
+```
+
+Now this ensures users are provided with NuGet Audit warnings that are not confusing and actionable.
+
+### Technical explanation
```
zivkan, line 97: I think we need to add the list of `PackageDownload` packages to the assets file.

The reason is that when we report packages as being vulnerable to the customer, they might try to use `dotnet nuget why` to get more information. We need to be able to have `dotnet nuget why` reprot back something sensible.  

Similarly, we might want to show PackageDownload packages in VS's solution explorer, although I think that's more of a dotnet/project-system implementation detail, rather than NuGet.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+
+<!-- Why should we not do this? -->
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
```
zivkan, line 142: on the topic of `IsImplicitlyDefined`, `PackageReference` is prior art. It's used to:

* Prevent VS (and maybe `dotnet add package`?) from allowing you to upgrade the PackageReference
* Prevent CPM from raising an error when the PackageReference has a Version attribute

I feel like there's another side effect, but I can't think of what it is at the moment.

These `PackageReference` functionality changes do not apply to `PackageDownload`.  Therefore, we should consider whether using the same metadata on `PackageDownload` is likely to confuse customers (particularly MSBuild SDK authors), or if it's helpful to reuse the same metadata name.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,106 @@
...
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
```
zivkan, line 149: Should we implement auditing of `PackageDownload`, ignoring `IsImplicitlyDefined` until the .NET SDK implements something to warn customers about out of date .NET SDKs?

According to [Microsoft's public Secure Futures Initiaive page](https://www.microsoft.com/en-us/trust-center/security/secure-future-initiative), we need to make our products secure by default.  If we have the capability of detecting that a PackageDownload has a vulnerable version, but nothing warns the customer, are we meeting expectations?

Similarly, thinking like a bad guy, what if I create my own MSBuild SDK, mark my PackageDownload as implicitly defined, but don't have any other mechanism to tell the customer they're using something that's out of date or vulnerable?  Will we do anything to try to mitigate this risk/attack vector?

If we go ahead with a design that allows malicious actors to disable NuGet's audit warnings, we'll definitely need to talk to the security experts to get sign off that this is acceptable.

accepted/2024/NuGet-Audit-for-PackageDownload.md
dsplaisted: I think we should use a different name for the metadata, to help make it clear that this should only be used by the .NET SDK.

accepted/2024/NuGet-Audit-for-PackageDownload.md
dsplaisted: If you're a bad guy and you get someone to use your malicious MSBuild SDK, you're already inside the trust boundary and can run arbitrary code during the build.  You don't need to do a PackageDownload on a vulnerable package.

If you're a good guy, I think we should use a different metadata name than `IsImplicitlyDefined` to make it clear that it isn't meant to be used outside of the .NET SDK.

We have some mechanisms to help keep the SDK up to date and will likely be adding more.  I don't think it's helpful to tell someone they have a vulnerable PackageDownload when they don't control the PackageDownload and the real problem is the SDK needs to be updated.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,123 @@
...
```
dsplaisted, line 149: I think we need to update the SDK to include the new attribute on its PackageDownloads in the same release that we add this NuGet feature.


Question: What happens when restoring from Visual Studio but using an older SDK that doesn't have those metadata values?  Do we need some sort of lightup so NuGet only does this if the SDK has the right metadata?

accepted/2024/NuGet-Audit-for-PackageDownload.md
nkolev92: IsImplicitlyDefined doesn't really do anything for audit in PackageReference either, so keeping the behavior consistent for metadata that exists in both consistent is important too

I know we're moving away from this but I'm just posting my way of thinking.

Nigusu-Allehu: I have updated the attribute. It's `AddedBy` now. The attribute will take in a string that specifies who the package was added by. More in the spec.

accepted/2024/NuGet-Audit-for-PackageDownload.md
Nigusu-Allehu: I have removed the IsImplicitlyDefined attribute. We will be warning for all PackageDownloads.

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,123 @@
...
+This means you’ll now receive warnings if a package added via `<PackageDownload>` has known security issues, so you can quickly address them.
+
+To prevent confusion about SDK-managed packages, we introduce a new attribute, `AddedBy`, on `<PackageDownload>` entries.
+This attribute will specify the source of the package, with values like `"dotnetsdk"` for SDK-managed packages and `"user"` for user-added dependencies.
+When a package marked with `AddedBy="dotnetsdk"` has a vulnerability, the warning will indicate that the package was added by the SDK.
+This will help users understand that the package originates from the SDK and is not something they explicitly added.
+
+| Attribute | Value     | Warning   |
+|-----------|-----------|-----------|
+| AddedBy   | dotnetsdk | warning NU1906: Package 'System.Text.Json' 1.0.0, added by '.NET SDK', has a known vulnerability. Refer to the vulnerability details at https://github.com/advisories/GHSA-g3q9-xf95-8hp5 to assess any necessary actions. |
```
zivkan, line 38: I think NuGet should special-case AddedBy=".NET SDK", and output a message along the lines "You should install a newer version of the .NET SDK"

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,123 @@
...
+| AddedBy   | user      | warning NU1906: Package 'System.Text.Json' 1.0.0, added by 'user', has a known vulnerability. Refer to the vulnerability details at https://github.com/advisories/GHSA-g3q9-xf95-8hp5 to assess any necessary actions. |
+| AddedBy   | undefined | warning NU1906: Package 'System.Text.Json' 1.0.0 has a known vulnerability. Refer to the vulnerability details at https://github.com/advisories/GHSA-g3q9-xf95-8hp5 to assess any necessary actions. |
+
+However, all of these warnings will be gated by the `SDKAnalysisLevel` property.
+The warnings will only be shown if the user is using a .NET SDK version that introduced these warnings or newer.
+For example, if these warnings are introduced in SDK version 9.0.300, the warnings will only appear if `SDKAnalysisLevel` is defined and set to 9.0.300 or higher.
```
zivkan, line 44: What we learned from the .NET 9.0.100 GA release is that `SDKAnalysisLevel` does not meet our customers or leaderships expectations. So, we can't use it as the gate.

I think we need to add a new config `<NuGetAuditPackageDownloads>`, which for now will default to `false`.

Once we have a plan for changing `NuGetAuditMode`'s defaults, we'll use the same plan to change this new config setting as well.

accepted/2024/NuGet-Audit-for-PackageDownload.md
marcpopMSFT: Good catch. Wouldn't want to flag user provided package downloads if they simply came from an older SDK. Should we try to get the metadata into 9.0.2xx just so there is some coverage there of the mismatch?

accepted/2024/NuGet-Audit-for-PackageDownload.md
```diff
@@ -0,0 +1,123 @@
...
+The expected outcome is a more precise and useful vulnerability reporting experience, enabling users to address actual risks without confusing SDK-related alerts.
+
+## Explanation
+
+### Functional explanation
+
+The NuGet Audit now supports `<PackageDownload>` packages, allowing it to detect and alert users about vulnerabilities in any user-specified dependencies downloaded via `<PackageDownload>`. 
+This means you’ll now receive warnings if a package added via `<PackageDownload>` has known security issues, so you can quickly address them.
+
+To prevent confusion about SDK-managed packages, we introduce a new attribute, `AddedBy`, on `<PackageDownload>` entries.
```
marcpopMSFT, line 31: CC @baronfel to weigh in on this. I can't think of anything better at the moment but don't want to say yes just because.

accepted/2024/NuGet-Audit-for-PackageDownload.md
marcpopMSFT: Per offline chat, tying packagedownload audit to level all is probably reasonable even though these are direct references. That way we're not opting people in earlier than they expect.

accepted/2024/NuGet-Audit-for-PackageDownload.md
Nigusu-Allehu: I think special-casing `AddedBy=".NET SDK"` should be handled by the SDK. NuGet will log the warnings and add them to the assets file, leaving it to the SDK to decide how to act. This ensures clear responsibilities and avoids overlap.

accepted/2024/NuGet-Audit-for-PackageDownload.md
Nigusu-Allehu: After considering the options, I decided to go with adding the `direct-with-packagedownload` value to `NuGetAuditMode` instead of introducing a separate `<NuGetAuditPackageDownloads>` configuration. This approach keeps audit-related settings centralized, avoids introducing new properties

Nigusu-Allehu:
Closing for now: we will collect some telemetry data on how many vulnerable `PackageDownlaod` packages are being used. That will give us an idea of how much the new warnings could affect customers.

ericstj:
> Closing for now: we will collect some telemetry data on how many vulnerable `PackageDownlaod` packages are being used. That will give us an idea of how much the new warnings could affect customers.

I think this would be a valuable way of communicating to folks that they are targeting a vulnerable .NET version in self-contained applications.  I suspect alone is a significant enough slice of customers to warrant this work.  

CC @richlander

nkolev92:
A quick clarification, we still want to do this, we just want to use data to drive the exact implementation we land on.

Issue #13372, opened 2024-04-03
dotnet update package --vulnerable (Audit fix)


nkolev92:
### NuGet Product(s) Involved

Visual Studio Package Management UI, dotnet.exe

### The Elevator Pitch

Provide an automated way for fixing project graphs with vulnerabilities in them. 

Frequently when transitive packages have vulnerabilities, updating the pasckages becomes a challenge.
Should I update the top level package? Update the vulnerable package only?
Is that enough?
Does that bring new vulnerabilities?

Doing this perfectly will be challenging, but something is better than nothing :D  

### Additional Context and Details

Mentioned in https://github.com/NuGet/Home/pull/11549 and part of the https://github.com/NuGet/Home/issues/8087 epic.

Frulfump:
When a package is promoted to a direct dependency to fix a vulnerability could the tool automatically add a comment that explains why it was added. Otherwise these dependencies will be left there forever when they might no longer be required. Ideally some automatic way of unpinning transitives that are no longer required would be really nice.

PR #13961, opened 2024-11-25
Add dotnet package update command, with --vulnerable option

OliaG:


accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
+# dotnet package update command
+
+- [Olia Gavrysh](https://github.com/OliaG/)
+- Start Date (2024-11-11)
+- [#8087](https://github.com/NuGet/Home/issues/8087)
```
nkolev92: We're overdue with some housekeeping here. 

- We shouldn't link specs to massive epics like this. This is a very specific feature that's only a part of the whole auditing experience. 
- This is a 2 in 1 spec. Adds `dotnet package update` and `dotnet audit fix`. We have some pre-existing specs on both that need tidied up/closed eventually when we think this is ready for the community to review. https://github.com/NuGet/Home/pull/11812 and https://github.com/NuGet/Home/pull/11549. 

There are 2 issues that in particular track audit fix & dotnet update package. https://github.com/NuGet/Home/issues/13372 and https://github.com/NuGet/Home/issues/4103. 
I think those are the issues that this spec should be referring to.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+This improvement will, in turn, bolster the overall security and reliability of .NET applications. The feature will build upon the existing NuGet Audit functionality.
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+Even though .NET developers now have the opportunity to learn about vulnerabilities and deprecated dependencies in their NuGet packages, addressing these issues remains a significant challenge. It requires a lot of manual work and knowledge about newer versions, which can be quite daunting. When the solution to these problems takes time, developers are less likely to act on it immediately, leading to these upgrades being postponed or never addressed. Additionally, developers might not always use the most efficient ways of resolving these issues, especially when dealing with multiple nested transitive dependencies. This is why developers will benefit greatly from a single command that attempts to automate all the work and update all the packages, efficiently eliminating vulnerabilities and deprecated versions.
```
nkolev92: There's a bunch of capabilities already that may not be given enough credit in this description here.

VS
- Has a means to show outdated packages.
- Has a means to show deprecate/vulnerable packages.
- Has a means to update all packages to latest. 
- Has a means to update a single package to latest (which can be challenging). 
- Has a vulnerability filter, which in turn can allow you to update all vulnerable packages to latest.

dotnet 
- Has a means to show outdated packages
- Has a means to show deprecated/vulnerable packages
- Has a means to update a single package
- Does not have means to update multiple packages at once. 

The gaps that exsist are: 

- No means to update to all latest packages on the CLI.
- No means for update heuristics, such as, update my vulnerable packages. This is essentially what the feature fka `audit fix` was intending to solve.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+
+### Updating all packages to the latest vesioin
+
+When users want to ensure their NuGet dependencies are up to date, they can run a command in .NET CLI `dotnet package update`.
+
+```CLI
+C:\> dotnet package update
+
+Fixing outdated packages in ContosoUniversity.sln
+ Updated UmbracoForms 8.4.1 to 8.7.1
```
nkolev92: Packages are installed on a per project basis, so this would probably be updating projects.

Solution updates & project updates can be different. 
There's some mechanics to work out potentially here, especially in non-CPM scenario.

For example, an update to package A from 1.0.0 to 2.0.0 in project 1 may succeed, but may fail in project 2.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+ Updated Newtonsoft.Json 11.0.1 to 13.0.3
+    
+Updated 2 packages in 36 scanned packages.
+```
+
+### Fixing vulnerabilities
+
+When users run `dotnet build` or `dotnet restore` commands in CLI, if they see any warnings related to vulnerabilities in their project’s NuGet dependencies, they can run `dotnet package update --vulnerable` CLI command to fix all the mentioned issues.
```
nkolev92: Does this address direct only or both direct & transitive? 
Is the audit mode considered at all?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+In the future this can be extended to fix deprecated versions.
+
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+#### dotnet audit
```
nkolev92: Reading this, not sure if this spec was quite ready for me to review. 

There's some mentions about `dotnet package update`, but now it's talking about `dotnet audit`. 

I'll continue my review, but just re-request review when the proposal is finalized.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+
+The `dotnet audit` command will scan your project and solution's dependencies and look for known vulnerabilities, deprecations, and outdated packages. If any vulnerabilities, deprecations, or outdated packages are found, a remediation will be calculated to then `fix` which will be applied to the resulting package graph.
+
+Example:
+
+![dotnet audit](/meta/resources/dotnetaudit/dotnetaudit.png)
+
+```
+$ dotnet audit
```
nkolev92: This effective combines `dotnet list package --outdated`, `dotnet list package --vulnerable`,  and `dotnet list package --deprecated`. 

I'm not saying the output should be the same, but it's worth considering them as inspiration. 
I think in general, consistency across the different commands helps.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+
+Run 'dotnet audit fix' to fix them.
+```
+
+##### Endpoints
+
+NuGet will use existing endpoints to optimize the speed of audit results.
+
+- [Deprecation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation)
+- [Vulnerability](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#vulnerabilities)
```
nkolev92: Restore uses a different endpoint for the vulnerability database. 
In general for package metadata, we use registration, which is what these links point to.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+- Outdated - No existing endpoint, will need to call a source.
```
nkolev92: The way outdated packages are also determined from the registration endpoint. 

It has a list of all available versions, listed/unlisted, so it can just determine from that.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+
+##### dotnet audit --json
+
+To get a detailed audit report in a JSON format.
+
+```
+{
+  "audit": {
```
nkolev92: I'd look at `dotnet list package --vulnerable`.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+
+##### Endpoints
+
+NuGet will use existing endpoints to optimize the speed of audit results.
+
+- [Deprecation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation)
+- [Vulnerability](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#vulnerabilities)
+- Outdated - No existing endpoint, will need to call a source.
+
+##### dotnet audit --json
```
nkolev92: Our prior art is `--format json`. 

dotnet list package and dotnet package search both. https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-package-search and https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-list-package

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+      ]
+    },
+    "scannedDependencies": "36"
+  }
+}
+```
+
+##### dotnet audit Exit Codes
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities, deprecations, or outdated packages were found.
```
nkolev92: Based on some of the latest conversations, i think we should drop the exist codes. 

The amount of information that can be communicated through exit codes is super low compared to the rich text we can communicate on the console or json output. 

We should just do `0` or `1`

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+There's a number of [dotnet tools](https://github.com/natemcmaster/dotnet-tools) that exist to solve many of these painpoints. Much of the information to put together this experience exists within NuGet.org's Server API & could created as a dotnet tool.
+
+There are a number of auditing tools in other ecosystems that are third-party / developed by the community. For other ecosystems, these features are built into the first-party CLI experiences.
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+There already exists a dotnet CLI commands called `dotnet list package` with options such as `--vulnerable, --deprecated, --outdated` which will list any known vulnerabilities, deprecations, and outdated packages in a project or solution. These options currently cannot be combined. Although this provides valuable information to understand the state of your dependencies, there does not exist a tool that allows you to quickly audit a project/solution & provide a way to act further on.
```
nkolev92: Consider: `dotnet list package --audit`

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+- [dotnet outdated](https://github.com/dotnet-outdated/dotnet-outdated)
+- [dotnet retire](https://github.com/retirenet/dotnet-retire)
+- [NuGet Defense](https://github.com/digitalcoyote/NuGetDefense)
+  
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+- Should the command be named `audit` or `check`?
```
nkolev92: `audit` matches the name of the NuGetAudit feature as well.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,333 @@
...
+  - `audit` is the more consistent name for package manager tooling with other ecosystems.
+  - `check` is the more consistent name with dotnet CLI.
+- Should this command only audit `vulnerabilities`? Or should it audit & be proactive as to `vulnerabilities`, `deprecations`, and `outdated` packages?
+- How much information should be present in the --json output?
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+- `dotnet audit` can be run on every `restore` which can throw warnings or errors to the user to take action against a vulnerable, deprecated, or outdated software supply chain.
```
nkolev92: That's exactly what NuGetAudit does today. 
It warns when people have vulnerability.

nkolev92 (commented):
Mid-way through I realized this was still a WIP, but I finished my review. 

hopefully these are helpful. 
Happy to chat more as needed.

accepted/2024/package-update-command.md
OliaG: That's great. I'll updates this part.

accepted/2024/package-update-command.md
OliaG: I apologize, I accidentally uploaded a completely different version of the file. fixing it now.

accepted/2024/package-update-command.md
OliaG: After talking to Dependabot engineers Andy and I agreed on the exit codes I'm going to upload with the new version, let me know how you feel about them.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,206 @@
...
+  "updated": [
+    {
+      "name": "UmbracoForms",
+      "version": "8.7.1",
+      "previousVersion": "8.4.1"
+    }
+  ],
+  "failures": [],
+  "warnings": []
+```
```
martinrrm: Since this command is going to be used in solutions too, I think we need to be more detailed in this json format. For example, let's say that dotnet audit fix resolves to different package version for different projects, or for different target frameworks, I think we need to reflect that in the JSON output.

Another edge case I can think of is when customer has project references. Let's say that ProjectA has a reference to ProjectB that contains a vulnerable package, since those packages are considered transitive will the solution be updating the package in ProjectB and install it in ProjectA?

With those considerations I was thinking something like this:

```json

{
	"version": "string",
	"feasible": true,
	"metadata": 
	{
		"added": "number",
		"removed": "number",
		"updated": "number",
		"promotedToTopLevel": "number",
	},
	"vulnerabilities":
	{
		"projectA":
		{
			"net8.0":
			{
				"total": "number",
				"newtonsoft.json": 
				{
					"originalVersion": "12.0.1", // Could be ranges too
					"newVersion": "13.0.1",
					"promotedToTopLevel": true,
					"projectReference": true, // Could be a path to the project reference
					...
				},
				"castle.core":
				{
					...
				},
			},
			"net7.0":
			{
				...
			},	
		},
		"projectB":
		{
			...
		},
	},
}
```
@OliaG @zivkan What do you think about this JSON proposal?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+- [Olia Gavrysh](https://github.com/OliaG/)
+- Start Date (2024-11-11)
+- [#13372](https://github.com/NuGet/Home/issues/13372)
+- [#4103](https://github.com/NuGet/Home/issues/4103)
+
+## Summary
+
+<!-- One-paragraph description of the proposal. -->
+The `dotnet package update` feature is designed to enhance the user experience by enabling .NET developers to address security vulnerabilities, deprecated packages, and outdated versions of their NuGet dependencies with a single command.
+This improvement will, in turn, bolster the overall security and reliability of .NET applications. The feature will build upon the existing NuGet Audit functionality.
```
zivkan:
```suggestion
This improvement will, in turn, bolster the overall security and reliability of .NET applications.
The feature will build upon the existing NuGet Audit functionality.
```

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+### Existing capabilities
+
+Visual Studio
+
+- Has a means to show outdated packages
+- Has a means to show deprecate/vulnerable packages
+- Has a means to update all packages to latest
+- Has a means to update a single package to latest (which can be challenging)
+- Has a vulnerability filter, which in turn can allow you to update all vulnerable packages to latest
+dotnet
```
zivkan:
```suggestion
```

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+
+CLI
```
zivkan, line 32: I think the comparison between VS and the dotnet CLI woud be easier to understand as a table. For example:

|Capability|VS|CLI|
|--|--|--|
|do thing|✔️|✔️|
|something else|✔️|❌|

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+This proposal is focused on bringing the CLI experience up to funcional parity with the experience in Visual Studio and make developers more productive by providing a single command to update all packages to the latest version and a way to update all vulnerable packages.
+
+## User experience
+
+### Updating all packages to the latest version
+
+When users want to ensure their NuGet dependencies are up to date, they can run a command in .NET CLI `dotnet package update` for a project or a solution.
+
+```CLI
+C:\> dotnet package update [<SOLUTION_PATH>|<PROJECT_PATH_>]
```
zivkan:
```suggestion
C:\> dotnet package update [<SOLUTION_PATH>|<PROJECT_PATH>]
```

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+
+In the "failures" and "warnings" should be the information about what projects were not updated due to failed pipeline or different reassons like no solution was found, the format of the project file is not supported, etc.
+
+##### Exit Codes
+
+Priority: P1
+
+- 0 - The command will exit with a 0 exit code if no vulnerabilities or outdated packages were found and no changes were made.
+- 1 - The command will exit with a 1 exit code if changes were successful and the PR is submitted.
+- 2 - The command will exit with a 2 exit code if any error has occurred, PR will not be submitted.
```
zivkan:
```suggestion
- 1 - The command will exit with a 1 exit code if changes were successful.
- 2 - The command will exit with a 2 exit code if any error has occurred.
```

`dotnet package update` won't automate creating pull requests, just modifying local files. The developer, or other tools that use `dotnet package update`, will be responsible for creating the pull request.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+- Outdated - no existing endpoint, will need to call a source.
+- [Deprecation](https://docs.microsoft.com/en-us/nuget/api/registration-base-url-resource#package-deprecation) for future possible improvements.
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+### Alternatives considrered
```
zivkan: `dotnet nuget fix` and `dotnet nuget audit fix` commands were considered to fix vulnerabilities, but we believe that it's more natural and more discoverable to have a generic update command that has different "modes" to choose all or just vulnerable packages.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+
+We could enable the option for the users to pick if they want only direct vulnerabilities to be resolved or direct and transitive. We decided to not do so and implement the resolution of all vulnerabilities as we believe that better corresponds to our goal of increasing security across all .NET applications.
```
zivkan:
```suggestion
We could enable the option for the users to pick if they want only direct vulnerabilities to be resolved or direct and transitive.
We decided to not do so and implement the resolution of all vulnerabilities as we believe that better corresponds to our goal of increasing security across all .NET applications. 
```

One sentence per line is more useful in our docs, where we're more likely to update the file some time in the future, but I like to practise it for all markdown documents

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+| `closest` | the algorithm will try to upgrade the direct package reference, and if that doesn't work, it will walk the graph one dependency level at a time trying to upgrade that, until it gets to the package with the known vulnerability. Once this mode is implemented, it should be set to be the default one. |
+
+#### --format json
+
+Priority: P2
+
+Provides a detailed report in a JSON format:
+
+```json
+  "added": [
```
zivkan: I'd like to see a complete, valid JSON, rather than a snippet of only part of a valid document.  In particular, this snippet doesn't show how a solution with multiple projects would be represented. The project path/filename will need to be declared somewhere.

accepted/2024/package-update-command.md
zivkan: I have a more comments than I thought I would:

1. The proposed schema doesn't provide a place to output warning or error messages. If the command fails, the error should still be in a valid JSON, to make it easier to script. I believe that having an `errors` or `messages` array removes the need for a `feasible` bool, since any result without any errors is by definition feasible.
2. Some languages make it hard to use json properties that are not valid language variable names, so property keys like "net8.0" is problematic, as is using project names. Even when your language allows you to parse it, it can be harder to use since property/key names are dynamic. In some languages it's much easier when the schema is static. For example, consider trying to use json query to search the result, or use `Convert-FromJson` in PowerShell.
3. On the topic of project names, it's possible to have multiple projects in a solution have the same name, as long as they're in different subdirectories, so we have to use the full path to the project, not just the name, otherwise in edge cases they won't be unique.
4. In the case of a package being transitive through a project reference, I think the tool should be smart enough to not update the package. In that case, since the package isn't being updated, there's no need for a `"projectReference"` property.
5. `dotnet package update` will be used for all kinds of updates, not just vulnerable packages. I think Olia's `"added":`, `"updated":`, and `"removed":` properties make more sense than a `vulnerabilities` property key. Although in that property you have a collection of projects, so `projects` is a more appropriate name.
6. I haven't been looking at JSON other than NuGet's own for years, so I'm quite out of touch with JSON best practises. But I don't understand the purpose of putting the counts within a `metadata` object. Is there a benefit over putting the counts in the parent object?
7. I feel like the counts should be per-project, not at the root object level. Having it at the root object level raises questions, like if multiple projects have to promote the same package to top level, does that count as 1, or is the promote to top level count going to equal the number of projects?  However, I don't understand the point of the count properties, since you can look at look at the length of the array of package updates.
8. JSON numbers aren't quoted. Some languages/parsers might be loosey-goosey, but strict parsers will fail. I feel like NuGet had this problem when migrating one of our parsers from Newtonsoft.Json (using JToken) to System.Text.Json.
9. While nuget.org doesn't have a package with id `total`, it is a valid package ID, so it should not be used in a place where package IDs are also expressed. However, I already mentioned we shouldn't have a dynamic schema, so this concern should automatically be solved when addressing where to put package ids.
10. Similar to the "I'm not familiar with JSON best practises" comment I made earlier, I'm not sure if long property names `originalVersion`, `newVersion`, `promotedToTopLevel` are commonly used. `from` and `to` I think are easy enough to understand, as are `original` and `new`. Assuming we have some docs somewhere that explains what it means, I think `promoted` might be a good enough name. `promotedToTopLevel` might be guessable without docs. However, I just had the thought that since it's possible to tell the difference between an "addition" vs an "upgrade", there could be a property `wasTransitive`, and I think that will be more clear to people than `promoted`

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+Fixed 2 packages in 36 scanned packages.
+```
+
+#### --mode
+
+Priority: P1 for `promote`; P2 for `closest` and switching the default to `closest`.
+
+| Mode | Explanation |
+|------|-------------|
+| `promote` | the algorithm will always promote transitive packages to top level. This is an easy to implement feature that we will release in MVP. |
```
zivkan: I'm not sure where in the document it should be written, but while talking to an internal partner, they suggested that packages that are promoted to top level in order to remove a vulnerable package should have some additional metadata to signal this. For example `<PackageReference Include="SomePackage" Version="1.2.3" PromotedBecauseOfVulnerability="true" />`.  A much, much better name is needed though, this is just an example.

If in the future the top level package that originally caused SomePackage to be included in the package graph has a new version that removes SomePackage from the package graph, this extra metadata will allow a future improvement to `dotnet update package` to detect that the PackageReference can now be removed.

I think this is a great idea for projects that aren't using CPM with transitive pinning. Although it's also a signal that transitive pinning without CPM would be a useful feature, but very much out of scope for this feature 😁

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+- Does not have means to update multiple packages at once
+
+### Gaps in CLI experience
+
+- No means to update to all latest packages on the CLI
+- No means for update heuristics, such as, update my vulnerable packages
+
+This proposal is focused on bringing the CLI experience up to funcional parity with the experience in Visual Studio and make developers more productive by providing a single command to update all packages to the latest version and a way to update all vulnerable packages.
+
+## User experience
```
zivkan: I think it would be nice to have a summary of all features, similar to what you'd expect to see when you run `dotnet package update -h`. The sub-feature examples below are good, but I had an idea for something, and a ctrl-f can't find it in the document. Without re-reading the whole thing top to bottom, I don't have an easy way to figure out if I'm searching for the wrong keyword or if it's something I need to suggest.

And that is that I think a `--dry-run` option would be good. But now I feel like this was originally proposed. Did we remove it for some reason that I can't remember?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+#### dotnet package update
+
+The `dotnet package update` command will update all the NuGet dependencies to the latest versions.
```
zivkan:
```suggestion
The `dotnet package update` command without any other arguments or options will update all the NuGet dependencies to the latest versions.
```

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+
+```CLI
+C:\> dotnet package update --help
+dotnet package update [<PROJECT>|<SOLUTION>] [--vulnerable] [--mode <MODE>] [-v|--verbosity <LEVEL>] [--json]
```
zivkan: above, the json option was `--format json`, but here it's just `--json`.  As someone else commented, we already have prior art with `--format json`

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,238 @@
...
+The `dotnet package update` command will update all the NuGet dependencies to the latest versions.
+
+```CLI
+C:\> dotnet package update --help
+dotnet package update [<PROJECT>|<SOLUTION>] [--vulnerable] [--mode <MODE>] [-v|--verbosity <LEVEL>] [--json] 
+```
+
+#### --vulnerable
+
+This flag will try to update only packages that have direct or transitive vulnerabilities. The remediation is calculated with an implicit dotnet audit to then apply directly to a resulting package graph. It can add packages, remove packages, and update packages depending on the problem it's attempting to resolve. It does not take into consideration downgrading to a compatible version if a higher one has already been specified.
```
zivkan:
```suggestion
This flag will try to update only packages that have direct or transitive vulnerabilities.
The remediation is calculated with an implicit dotnet audit to then apply directly to a resulting package graph.
It can add packages, remove packages, and update packages depending on the problem it's attempting to resolve.
It does not take into consideration downgrading to a compatible version if a higher one has already been specified.
```

I like one sentence per line, because it makes it easier to comment on specific things.

In particular:

1. In another comment, I wrote that if we tag packages promoted to top level, then later on we can remove it as a top level package when we detect it's no longer needed as top level. Without this feature, I'm not sure I can think of any scenarios where we will remove a package reference.
2. I shouldn't be working this late at night, but at the moment, I don't understand the last sentence. Is this in reference to packages where new versions no longer support target frameworks that the current project is using, so while the higher package version no longer has a known vulnerability, it's not compatible with the project? The "if a higher one has already been specified" part also throws me off. At this time of night, I can't guess what this means.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,266 @@
...
+This proposal is focused on bringing the CLI experience up to funcional parity with the experience in Visual Studio and make developers more productive by providing a single command to update all packages to the latest version and a way to update all vulnerable packages.
+
+## User experience
+
+### Updating all packages to the latest version
+
+When users want to ensure their NuGet dependencies are up to date, they can run a command in .NET CLI `dotnet package update` for a project or a solution.
+
+```CLI
+C:\> dotnet package update [<SOLUTION_PATH>|<PROJECT_PATH_>]
```
zivkan: It's helpful to also have a section that lists all the command arguments & options in one place, similar to what `dotnet package update -h` will show, because if both sln/proj and also package names are optional `dotnet package update [sln|proj] [package id]`, then it gets ambiguous when a single argument is provided. Does `dotnet package update Contoso.Utilities` mean update all package in a project with that name, or update that package to the latest version in whatever project/sln file happens to be in the current directory?

When a single argument is provided, we can do a Directory.Exists and File.Exists to guess, but I feel like that isn't great.  I think a `--project <path>` argument, which can accept a directory or sln file too, is my preferred design.

zivkan:
@NuGet/nuget-client and other interested parties (for example, @baronfel and @richlander), I pushed a significant update to this PR. Please review and provide comments.  I want to get this PR merged soon.

Keep in mind, it's a feature spec that shows the principals of the new command. It's not a requirement doc that will block implementation pull requests that stray from what's written here.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+Any package provided that does not have a known vulnerability will be skipped.
+
+The `--mode` option will allow customers to choose what strategy of resolving vulnerabilities will be used.
+The `direct` mode would only look for direct package references with known vulnerabilities and upgrade them.
+The `promote` mode would also check transitive packages (equivalent to restore's `<NuGetAuditMode>all`), and add a PackageReference to the package (therefore promoting it to a direct reference).
+Later, more advanced modes can be added.
+When a project is using CPM and transitive pinning is enabled, no PackageReference should be added, just the PackageVersion in the Directory.Packages.props file.
+
+In the future a similar option to fix deprecated versions can be added.
+
```
nkolev92, line 134: There's probably more to this as we implement vulnerable, but my question is what about partial success? 
What if we can't get a vulnerability free update, but we have fewer vulnerabilities than we did before?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+Some consideration may be needed to decide how much of the same information should be reported in the json (particularly when packages are not updated), and if the json schema should be the same for all commands, or if `--vulnerable` should have a different output schema than without `--vulnerable`.
+
+Please provide feedback at <https://github.com/NuGet/Home/issues/14360>
+
+#### Exit Codes
+
+Priority: P1
+
+- 0 - The command will exit with a 0 exit code if at least one package was updated without any error.
+- 1 - The command will exit with a 1 exit code if no packages were updated, but otherwise no error occurred.
```
nkolev92: No packages were updated, because we couldn't find appropriate solution or because there were not any updates available?

Basically, what if I run `dotnet package update` on a project with no packages to update?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+### Alternatives considered
+
+#### Extending dotnet package update
```
nkolev92:
```suggestion
#### Extending dotnet package add
```

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- CPM integration
+
+CPM adds more complexity than other NuGet features.
+Should `dotnet package update` upgrade `GlobalPackageReference`s?
```
nkolev92: This one is very interesting. I'd lean yes personally, because it is just syntactic sugar, but the way the item is used by customers might suggest otherwise.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+Should it allow you to modify Directory.Packages.props when only one project is provided, not the whole solution?
```
nkolev92: What does VS do?

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+When run on a solution, should it check for unused `PackageVersion`s and provide a way to remove them?
```
nkolev92: I'd say not yet, mostly cause of I dream of a command that can do that by itself later.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+  Updated UmbracoForms 8.4.0 to 8.7.1
+    
+Fixed 2 packages in 36 scanned packages.
+```
+
+Using the `--vulnerable` option should update the package to the lowest version that is higher than the currently reference version, which does not have a known vulnerability.
+For example, if `Contoso.Security` version 1.0.0 is currently referenced by the project, and the package source has versions 1.0.1 and 2.0.0 available, 1.0.1 should be selected.
+This is in contrast to `dotnet package update Contoso.Security` which will update to the highest version of the package.
+
+If package identifiers are provided in addition to the `--vulnerable` option, only those packages should be updated.
```
nkolev92, line 123: Feels like we're missing `--packages` option from the description.

nkolev92 (commented):
Th

accepted/2024/package-update-command.md
zivkan: packages are an argument, not an option. For example `dotnet package update pkg1 pkg2 pkg3 --vulnerable`.

nkolev92: For some reason I though we would add `--packages` option too, even though you clearly called out it's only an argument. 

Reading is hard.

accepted/2024/package-update-command.md
```diff
@@ -0,0 +1,286 @@
...
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- CPM integration
+
+CPM adds more complexity than other NuGet features.
+Should `dotnet package update` upgrade `GlobalPackageReference`s?
+Should it allow you to modify Directory.Packages.props when only one project is provided, not the whole solution?
+When run on a solution, should it check for unused `PackageVersion`s and provide a way to remove them?
+
+## Future Possibilities
```
Nigusu-Allehu, line 311: Another futurue possibility to consider:
- Update https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1901-nu1904 messages to suggest users to utilize the `package update --vulnerable` command to address the warnings.

Nigusu-Allehu (dismissed):
LGTM!

accepted/2024/package-update-command.md
zivkan: VS will let you update a package when in the single-project PM UI view, and the preview install will only check that this one project can complete a preview restore without failure. If updating the package version will cause a different project to start failing, VS will make the change and put restore into a failed state, which kind of defeats the intent of the preview restore, in my opinion.  But we can use it as precedent for having `dotnet package update` work in the same way.  I'd like to get people to think about the scenario, because when PM UI was created for packages.config years ago, transitive packages weren't a thing, so one project affecting another didn't really happen.

nkolev92: Yeah, I think preview restore doesn't run the whole solution thing was decided from a speed perspective. 
Not sure if we've validated how often this can lead to issues.

Nigusu-Allehu (approved):

Issue #13406, opened 2024-04-23
`dotnet list package` doesn't `restore`


richlander:
This behavior is incorrect, inconsistent, and inconvenient.

We have had implicit restore for `dotnet build` for many years. `dotnet list package` would benefit from this same behavior.

> You first need to build the project in order to have the assets needed for this command to process.

Per [docs](https://learn.microsoft.com/dotnet/core/tools/dotnet-list-package). That's unfortunate.

Observed behavior follows.

```bash
$ dotnet new console -o nugettest
$ cd nugettest
$ rm -r obj
$ dotnet list package
No assets file was found for `/Users/rich/nuget-test/nuget-test.csproj`. Please run restore before running this command.
$ dotnet restore
  Determining projects to restore...
  Restored /Users/rich/nuget-test/nuget-test.csproj (in 52 ms)
$ dotnet list package
Project 'nugettest' has the following package references
   [net8.0]: No packages were found for this framework.
$ dotnet add package System.Text.Json --version 8.0.3
$ dotnet list package                                
Project 'nuget-test' has the following package references
   [net8.0]: 
   Top-level Package       Requested   Resolved
   > System.Text.Json      8.0.3       8.0.3   
$ echo manually add package
manually add package
$ grep PackageRef nuget-test.csproj  

    <PackageReference Include="System.Text.Json" Version="8.0.3" />
    <PackageReference Include="System.Text.RegularExpressions" Version="4.3.0" />
dotnet list package              
Project 'nuget-test' has the following package references
   [net8.0]: 
   Top-level Package       Requested   Resolved
   > System.Text.Json      8.0.3       8.0.3
$ dotnet list package --vulnerable --include-transitive

The following sources were used:
   https://api.nuget.org/v3/index.json

The given project `nuget-test` has no vulnerable packages given the current sources.
```

nkolev92:
The original design was to make it offline, similar to other list commands, which is why it doesn't call restore implicitly.

richlander:
That doesn't make sense to me. The feature requires a restore to function and cannot detect when it will produce bad results (as I demonstrates).

Sounds like a broken design to me.

aortiz-msft:
Assigning to @Nigusu-Allehu since this is related to other work happening.

PR #14065, opened 2025-01-23
dotnet list package checks if restore is current

Nigusu-Allehu:
Rendered: https://github.com/NuGet/Home/blob/e767bd9abb67589d3892b1382f70e8fc72a8ca13/accepted/2025/dotnet-list-package-restore-no-op-check.md

### DIFF
--- a/accepted/2025/dotnet-list-package-restore-no-op-check.md
+++ b/accepted/2025/dotnet-list-package-restore-no-op-check.md
@@ -1,0 +1,105 @@
+# ***Dotnet list package checks when restore is not current***
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- Author Name [Nigusu](https://github.com/Nigusu-Allehu)
+- [GitHub Issue](https://github.com/NuGet/Home/issues/13703) <!-- GitHub Issue link -->
+
+## Summary
+
+The `dotnet list package` command is used to list packages referenced by a project, but it requires a restore operation to be performed beforehand to ensure accurate information. NuGet currently checks for the existence of an assets file to assert that restore has been performed.
+However, if a user modifies the project file after running restore, the assets file may become outdated, leading to misinformation from the dotnet list package command.
+This proposal discusses how to ensure that NuGet asserts the assets file is in sync with the project file to prevent such issues
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+`dotnet list package` command is used to list packages referenced by a project.
+Before this command is executed, users are expected to perform a restore operation.
+NuGet expects restore to be run both after project creation and every time a change has been made to the project.
+This is because the `dotnet list package` command is based on the project assets file providing accurate information.
+NuGet currently asserts restore has been performed by checking the existence of an assets file.
+However, it is possible a user might have run restore once and followed it up with a modification to the project file.
+That means the assets file has information that has expired.
+This is not detected/asserted by the `dotnet list package command`.
+As a result, in this specific scenario, `dotnet list package` command misinforms users about their projects.
+
+## Explanation
+
+### Functional explanation
+
+Assuming the functionality has been implemented, here is what the expected interaction with a user would look like.
+
+#### **Scenario 1: User runs `dotnet restore` and modifies the project file**
+
+*User Action:*
+
+- Runs `dotnet restore` to restore packages for their project.
+- Modifies the project file by adding or removing a package reference.
+- Runs `dotnet list package`.
+
+*Expected Behavior:*
+
+- The command checks if the assets file is in sync with the project file.
+- If the assets file is outdated, it prompts the user to run `dotnet restore` again.
+
+#### **Scenario 2: User runs `dotnet restore` and does not modify the project file**
+
+*User Action:*
+
+- Runs `dotnet restore`.
+- Runs `dotnet list package`.
+
+*Expected Behavior:*
+
+- The command lists packages accurately without any prompts.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+Comparing the contents of the project file and the assets file is the most important step here. We will achieve this by using no-op restore.
+
+- Create a DGSpec file from the evaluated project
+  - To create the DGSpec file, use an MSBuild target, [GenerateRestoreGraphFile](https://github.com/NuGet/NuGet.Client/blob/cb0b5561f4fd98347cedac11919f769da0b767f4/src/NuGet.Core/NuGet.Build.Tasks/NuGet.RestoreEx.targets#L46), that writes the DGSpec to a file and then load it.
+- Hash the DGSpec
+- Compare the hash value with the one saved in `project.nuget.cache` file
+
+#### If unable to load DGSpec file
+
+Log the following [similar to the dotnet sdk](https://github.com/dotnet/sdk/blob/b324c2e074f7691705b85b6ed69be23d350164b8/src/Cli/dotnet/commands/dotnet-add/dotnet-add-package/LocalizableStrings.resx#L147)
+
+> Error: Unable to create dependency graph file for project '{0}'.
+
+#### If the hash values are different
+
+Log the following
+
+> Error: The project's dependency graph has been altered. Please perform a restore for project '{0}'.
+
+## Drawbacks
+
+<!-- Why should we not do this? -->
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
+<!-- Do other features exist in other ecosystems and what experience have their community had? -->
+<!-- What lessons from other communities can we learn from? -->
+<!-- Are there any resources that are relevant to this proposal? -->
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
--- a/accepted/2025/dotnet-list-package-restore-no-op-check.md
+++ b/accepted/2025/dotnet-list-package-restore-no-op-check.md
@@ -80,12 +80,14 @@ Log the following
 ## Drawbacks
 
 <!-- Why should we not do this? -->
+- Performance Overhead: Generating and hashing the DGSpec file adds slight overhead to the dotnet list package.
 
 ## Rationale and alternatives
 
 <!-- Why is this the best design compared to other designs? -->
 <!-- What other designs have been considered and why weren't they chosen? -->
 <!-- What is the impact of not doing this? -->
+- **Force Restore Every Time**: Automatically run restore when dotnet list package is called.
 
 ## Prior Art
 


Nigusu-Allehu:
Thank you @zivkan for the ideas in our offline discussion.

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+
+<!-- Why should we not do this? -->
+- Performance Overhead: Generating and hashing the DGSpec file adds slight overhead to the dotnet list package.
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+- **Force Restore Every Time**: Automatically run restore when dotnet list package is called.
```
jeffkl, line 90: I think this should say "Implicitly run restore which might do a no-op with an option to disable via a command-line argument `--no-restore` which other commands have.  This would probably be me vote personally.

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
+# ***Dotnet list package checks when restore is not current***
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- Author Name [Nigusu](https://github.com/Nigusu-Allehu)
+- [GitHub Issue](https://github.com/NuGet/Home/issues/13703) <!-- GitHub Issue link -->
+
+## Summary
+
+The `dotnet list package` command is used to list packages referenced by a project, but it requires a restore operation to be performed beforehand to ensure accurate information. NuGet currently checks for the existence of an assets file to assert that restore has been performed.
```
zivkan, line 9:
```suggestion
The `dotnet list package` command is used to list packages referenced by a project, but it requires a restore operation to be performed beforehand to ensure accurate information.
NuGet currently checks for the existence of an assets file to assert that restore has been performed.
```

inconsistent new lines per sentence

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+However, if a user modifies the project file after running restore, the assets file may become outdated, leading to misinformation from the dotnet list package command.
+This proposal discusses how to ensure that NuGet asserts the assets file is in sync with the project file to prevent such issues
```
zivkan, line 11:
```suggestion
This proposal discusses how to ensure that NuGet asserts the assets file is in sync with the project file to prevent such issues.
```
nitpick 🤣

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
+# ***Dotnet list package checks when restore is not current***
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- Author Name [Nigusu](https://github.com/Nigusu-Allehu)
+- [GitHub Issue](https://github.com/NuGet/Home/issues/13703) <!-- GitHub Issue link -->
+
+## Summary
+
+The `dotnet list package` command is used to list packages referenced by a project, but it requires a restore operation to be performed beforehand to ensure accurate information. NuGet currently checks for the existence of an assets file to assert that restore has been performed.
+However, if a user modifies the project file after running restore, the assets file may become outdated, leading to misinformation from the dotnet list package command.
```
zivkan, line 10:
```suggestion
However, if a user modifies the project file, or a ProjectReference's project file, after running restore, the assets file may become outdated, leading to misinformation from the dotnet list package command.
```

packages are transitive

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+This proposal discusses how to ensure that NuGet asserts the assets file is in sync with the project file to prevent such issues
+
+## Motivation
+
+<!-- Why are we doing this? What pain points does this solve? What is the expected outcome? -->
+
+`dotnet list package` command is used to list packages referenced by a project.
+Before this command is executed, users are expected to perform a restore operation.
+NuGet expects restore to be run both after project creation and every time a change has been made to the project.
```
zivkan, line 19: nitpick: The first sentence is basically a less detailed duplicate of the second sentence. So one of these can be deleted.

accepted/2025/dotnet-list-package-restore-no-op-check.md
nkolev92: https://github.com/nuget/home/issues/13406 is the tracking issue.

It'd be worth to call out the challenges if we were to go that direction.

We can also see if @OliaG can help with some CD to see if the breaking is worth it.

nkolev92: fwiw, as per the spec guide, I think we should make a good argument about why we're not pursuing certain alternative. 
In this case, those drawbacks are not obvious.

nkolev92 (commented):
Spec looks great, complete and there's plenty of technical detail. 

We should explore @jeffkl's idea and get a sense of whether we'd want to make that breaking change (or if we've explored it in more detail, we should details about why we're not pursuing that direction.

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+
+*Expected Behavior:*
+
+- The command lists packages accurately without any prompts.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+Comparing the contents of the project file and the assets file is the most important step here. We will achieve this by using no-op restore.
```
zivkan, line 61:
```suggestion
Comparing the contents of the project file and the assets file is the most important step here.
We will achieve this by using no-op restore.
```

One sentence per line 😁

dsplaisted:
To me, it seems like the `list package` command should do an implicit restore (which will often be a no-op).  If it's already going to check if a restore is needed, why can't it do the restore if necessary rather than give you an error telling you to run restore?

Nigusu-Allehu:
> To me, it seems like the `list package` command should do an implicit restore (which will often be a no-op). If it's already going to check if a restore is needed, why can't it do the restore if necessary rather than give you an error telling you to run restore?

That is a great point. Currently the following is the counter argument to that: The `dotnet list package` command used to be offline, but now it would be potentially online due to the restore. For example, VS Code uses dotnet list package to list PackageReferences in a user’s project. This works offline, but with the restore, the command could fail if the user has no internet connection.

discussed here https://github.com/NuGet/Home/issues/13406#issuecomment-2073246250

jeffkl:
> > To me, it seems like the `list package` command should do an implicit restore (which will often be a no-op). If it's already going to check if a restore is needed, why can't it do the restore if necessary rather than give you an error telling you to run restore?
> 
> That is a great point. Currently the following is the counter argument to that: The `dotnet list package` command used to be offline, but now it would be potentially online due to the restore. For example, VS Code uses dotnet list package to list PackageReferences in a user’s project. This works offline, but with the restore, the command could fail if the user has no internet connection.
> 
> discussed here [#13406 (comment)](https://github.com/NuGet/Home/issues/13406#issuecomment-2073246250)

Thinking about this more, I'm imagine any scenario where my project XML just has:

```xml
<ItemGroup>
  <PackageReference Include="SomePackage" Version="1.0.0" />
</ItemGroup>
```

I run `dotnet list package` and it shows me that I reference `SomePackage`.  Then I edit the project XML and change the version and run `dotnet list package` again but it shows me the old version.  The proposal is to tell me that that what I just did to my project XML caused restore to be out of date and tells me that I need to run `dotnet restore` in order for `dotnet list package` to work properly.

What does that same scenario in Visual Studio feel like?  Implicit restore is on by default but if I turn it off, does PM UI show me what was previously restored?  Does it prompt me to manually restore before it will show me the packages?

I struggle because I defined my packages in XML and saved the file.  To me, `dotnet list package` should show me my current state but instead it just reads the `project.assets.json` (a file generated by restore that I may or may not understand).  So does `dotnet list package` list the packages I configured or does `dotnet list package` list the packages that are in a file that I do not control?  In my opinion, its the former, so implicit restore should have been how it worked from the beginning.  The fact that a restore has to take place in order for `dotnet list package` to give me what I want seems like an implementation detail that an end user should not have to care about.

The offline scenario is potentially something to care about of course.  But if I restore my project by simply building it or loading it in Visual Studio, the restore already happened and subsequent implicit restores with `dotnet list package` would no-op, making it work just fine.  If I edit my package graph while I'm disconnected, there's no guarantee `dotnet restore` would work unless I have all packages on disk.  In that case `dotnet list package` would tell me my restore is out of date, and restore wouldn't work either.  For scenarios we control, like VS Code, we could keep track of whether or not a restore completed and add `--no-restore` to the command-line arguments, just to make it faster.

The real question that I'm interested in knowing the answer to is: How many actual users would experience a negative experience if we changed `dotnet list package` to do an implicit restore?

Is that impact broad? If those users could just update their script with a `--no-restore` argument, would that be acceptable?  Is making this command a better experience for the majority a better approach?  Without more data telling us, its hard to make the decision.  Obviously, my opinion is to add the implicit restore but its definitely up for a vote.

donnie-msft:
> PM UI show me what was previously restored? Does it prompt me to manually restore before it will show me the packages?

PM UI shows a banner that "Some packages are missing" and offers a Restore gesture.
You'll also get Error List errors indicating the assets file isn't found. Or if it is found, errors about some packages being missing. 

An action (say, Update a package) still triggers restore regardless of whether the setting is disabled. Not sure if this is by design or a bug when an action triggers the restore versus a build.

I believe PM UI is already aligning with your suggestion for `dotnet list package` and PM UI will show users "the current state" and essentially "Work Offline" as much as it can.

accepted/2025/dotnet-list-package-restore-no-op-check.md
nkolev92, line 11: Feels like this should actually call out the proposal.

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+
+- Create a DGSpec file from the evaluated project
+  - To create the DGSpec file, use an MSBuild target, [GenerateRestoreGraphFile](https://github.com/NuGet/NuGet.Client/blob/cb0b5561f4fd98347cedac11919f769da0b767f4/src/NuGet.Core/NuGet.Build.Tasks/NuGet.RestoreEx.targets#L46), that writes the DGSpec to a file and then load it.
```
zivkan, line 64: I think I've given this feedback on a number of feature specs, but I think this is too much detail for a feature or design spec.

My concern is that some people find these design specs, wrongly assume they're still current, even years after they're written or implemented, and then assume that any behaviour differences are a bug. 

The first implementation might run msbuild as a child process to run this target, but this has a flaw when trying to get a PackageSpec or DGSpec, but I'm working on making it easier for all NuGet's `dotnet *` commands to create a DGSpec programatically, in-process. Depending on the timing of the two workstreams, maybe `dotnet list package` will not need to use this msbuild target.

Basically, I think it's sufficient to give the "high level" overview of how we'll detect if the restore is up to date, which is to use the same logic as no-op restores. How we achieve the no-op restore check is an implementation detail that is subject to change over time.

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+
+Log the following [similar to the dotnet sdk](https://github.com/dotnet/sdk/blob/b324c2e074f7691705b85b6ed69be23d350164b8/src/Cli/dotnet/commands/dotnet-add/dotnet-add-package/LocalizableStrings.resx#L147)
+
+> Error: Unable to create dependency graph file for project '{0}'.
+
+#### If the hash values are different
+
+Log the following
+
+> Error: The project's dependency graph has been altered. Please perform a restore for project '{0}'.
```
zivkan, line 78: I don't think "The project's dependency graph has been altered" makes any sense to customers. The DGSpec is a NuGet internal implementation detail, that even expert customers have no need to know about.

A more friendly message would be something like "The project has been modified since the last restore"

accepted/2025/dotnet-list-package-restore-no-op-check.md
```diff
@@ -0,0 +1,107 @@
...
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+- **Force Restore Every Time**: Automatically run restore when dotnet list package is called.
+
+## Prior Art
+
+<!-- What prior art, both good and bad are related to this proposal? -->
```
zivkan, line 94: * VS restores automatically on solution load for SDK style projects, or before every build for all other projects
* `dotnet build`, `dotnet publish`, and `dotnet test` all restore automatically by default, unless `--no-restore` is provided
* installing or upgrading packages either in VS or with `dotnet add package`, restores the project, in addition to modifying project files. `dotnet add package` has a `--no-restore` option.

Nigusu-Allehu:
We have reached an agreement to introduce implicit restore instead: https://github.com/NuGet/Home/pull/14119. Closing PR.

Issue #14002, opened 2024-12-17
Document PrunePackageReference and all the details around it 


nkolev92:
PR https://github.com/NuGet/NuGet.Client/pull/6142, design https://github.com/NuGet/Home/blob/dev/accepted/2024/prune-package-reference.md.

PR #14066, opened 2025-01-24
PrunePackageReference in .NET 10 Preview 1

nkolev92:
https://github.com/NuGet/Client.Engineering/issues/3110

[Rendered write-up](https://github.com/NuGet/Home/blob/dev/accepted/2025/prune-package-reference-rollout.md)

### DIFF
--- a/accepted/2024/prune-package-reference.md
+++ b/accepted/2024/prune-package-reference.md
@@ -155,7 +155,8 @@ Given that packages are allowed to bring in a shared framework, and that is not 
 
 ## Drawbacks
 
-- The risk of this feature is driven by the packages that are being pruned. Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there.
+- The risk of this feature is driven by the packages that are being pruned.
+Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there. However, there is a change we end up pruning more/fewer packages in the end. There's also a chance we end up getting new dependencies since pruning packages may have an effect on the outcome of the cousin dependency rule.
 
 ## Rationale and alternatives
 
--- a/accepted/2024/prune-package-reference.md
+++ b/accepted/2024/prune-package-reference.md
@@ -156,7 +156,7 @@ Given that packages are allowed to bring in a shared framework, and that is not 
 ## Drawbacks
 
 - The risk of this feature is driven by the packages that are being pruned.
-Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there. However, there is a change we end up pruning more/fewer packages in the end. There's also a chance we end up getting new dependencies since pruning packages may have an effect on the outcome of the cousin dependency rule.
+Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there. However, there is a chance we end up pruning more/fewer packages in the end. There's also a chance we end up getting new dependencies since pruning packages may have an effect on the outcome of the cousin dependency rule.
 
 ## Rationale and alternatives
 
--- a/accepted/2025/prune-package-reference-rollout.md
+++ b/accepted/2025/prune-package-reference-rollout.md
@@ -38,15 +38,15 @@ We validated the .NET SDK build that it *does not affect the builds negatively o
 - dotnet/runtime - <https://github.com/dotnet/runtime/pull/111767>
 - dotnet/aspnetcore - <https://github.com/dotnet/aspnetcore/pull/60025>
 
-### Observable build changes
+### Observable build changes and risks
 
 - Customers who expect the output of their project to produce identical results will notice smaller deps files. This is intentional.
-- Customers who use packages lock files (1% usage) with locked mode may notice fewer dependencies in their lock files. This is intentional.
 - Customers who observe the output of component governance / NuGet Audit will notice these components are no longer reported. This is intentional.
+- Customers who use packages lock files (1% usage) with locked mode may run into restore failures due to restore bringing in fewer dependencies. This is intentional.
 - Customers with builds that have customization may run into build failures if they are:
   - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
     - Manual assembly coping - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
-  - We don't foresee this is a common scenario given that the packages being pruned are usually within the framework and customers shouldn't need to anything additional to make things work.
+  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
 
 ## Rationale and alternatives
 
--- a/accepted/2025/prune-package-reference-rollout.md
+++ b/accepted/2025/prune-package-reference-rollout.md
@@ -45,7 +45,7 @@ We validated the .NET SDK build that it *does not affect the builds negatively o
 - Customers who use packages lock files (1% usage) with locked mode may run into restore failures due to restore bringing in fewer dependencies. This is intentional.
 - Customers with builds that have customization may run into build failures if they are:
   - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
-    - Manual assembly coping - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
+    - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
   - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
 
 ## Rationale and alternatives
--- a/accepted/2025/prune-package-reference-rollout.md
+++ b/accepted/2025/prune-package-reference-rollout.md
@@ -45,6 +45,7 @@ It may lead to some build issues for if a build has customization such as:
 - If a project manually references an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
   - Manual assembly coping - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
   - Test projects that are used for integration testing may have set-ups that manually copy assemblies to the build output folder for the project.
+- Issues with packages lock files - Customers using lock files may see fewer packages downloaded with the newer SDKs and if they're locked mode, they'll see failures. (Packages lock files are used in about 1% of all projects). For customers with lock files, we recommend that they lock down both their tooling and packages.
 
 ## Rationale and alternatives
 
--- a/accepted/2025/prune-package-reference-rollout.md
+++ b/accepted/2025/prune-package-reference-rollout.md
@@ -1,4 +1,4 @@
-# ***PrunePackageReference in .NET 10***
+# PrunePackageReference in .NET 10
 
 - Nikolche Kolev <https://github.com/nkolev92>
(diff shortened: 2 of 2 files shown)


accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,69 @@
...
+### Testing & Validation
+
+The feature has been development since November and the implementation has been extensively tested through automation, but we also ensured that our own repos can dogfood without any issues.
+
+We validated the .NET SDK build that it *does not affect the builds negatively of the following repos*:
+
+- NuGet/NuGet.Client
+- dotnet/sdk
+- dotnet/roslyn - <https://github.com/dotnet/roslyn/pull/76898>
+- dotnet/runtime - <https://github.com/dotnet/roslyn/pull/76896>
```
ericstj:
```suggestion
- dotnet/runtime - <https://github.com/dotnet/runtime/pull/111767>
```

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,69 @@
...
+- dotnet/aspnetcore - <https://github.com/dotnet/aspnetcore/pull/60025>
+
+## Risks and drawbacks
```
ericstj: I'm not really seeing these as risk and drawbacks.  I'd retitle this as "observable behavior changes".  I'd assert that all the cases we call out are desired behaviors.

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,69 @@
...
+
+It may lead to some build issues for if a build has customization such as:
+
+- If a project manually references an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
```
ericstj: Someone might be doing this for the purpose of static analysis against older packages.  If they wish to do this they can use PackageDownload.  That will be a more predictable result.  I don't expect anyone to actually do this - we haven't found any instances.

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,69 @@
...
+  - Manual assembly coping - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
+  - Test projects that are used for integration testing may have set-ups that manually copy assemblies to the build output folder for the project.
+- Issues with packages lock files - Customers using lock files may see fewer packages downloaded with the newer SDKs and if they're locked mode, they'll see failures. (Packages lock files are used in about 1% of all projects). For customers with lock files, we recommend that they lock down both their tooling and packages.
+
```
ericstj, line 50: Customers who expect the output of their project to produce identical results will notice smaller deps files.  This is intentional.

Customers who observe the output of component-detection / component-governance will notice that these components are no longer reported.  This is also intentional.

accepted/2025/prune-package-reference-rollout.md
nkolev92: They're desired so I'm adding that part in each line (great suggestion on that!), but i feel like anything that could lead ot a build failure is a risk in a way.

I like your suggestion better though.

accepted/2025/prune-package-reference-rollout.md
nkolev92: That's interesting. I would've expected people to do static analysis on framework assemblies unless I'm misunderstanding your point.

nkolev92:
Addressed your feedback, thanks @ericstj!

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
+# ***PrunePackageReference in .NET 10***
```
richlander: This isn't standard formatting. Go simple.

```suggestion
# PrunePackageReference in .NET 10
```

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Nikolche Kolev <https://github.com/nkolev92>
+- [#NuGet/Client.Engineering#3110](https://github.com/NuGet/Client.Engineering/issues/3110)
+
+## Summary
+
+`NuGetAudit` scans the dependencies in your package graph for vulnerabilities, reporting vulnerabilities for every package that ended up in your resulting graph.
+
+[`PrunePackageReference`](https://github.com/NuGet/Home/blob/dev/accepted/2024/prune-package-reference.md) allows NuGet to skip platform packages that are already available within the framework that the project is building.
+For example, if your project, targetting .NET 9 has a transitive dependency to System.Text.Json 8.0.4, which is vulnerable, enabling pruning will allow NuGet to skip it. This package will not be downloaded and it will not appear in the dependencies.
```
jeffkl: nit: The word "skip" seems too vague, I thought it would be better to really call out what the implementation would do.

```suggestion
For example, if your project, targetting .NET 9 has a transitive dependency to System.Text.Json 8.0.4, which is vulnerable, enabling pruning will allow NuGet to prune it from the defined list of dependencies during graph resolution. This package will not be downloaded and it will not appear in the resolved dependency graph.
```

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+
+We propose that we enable the `PrunePackageReference` feature by default for *all* .NET (Core) and .NET Standard 2.0 and above in .NET 10 preview 1 to reduce false positives by NuGetAudit and related scanning tools and improve performance.
```
jeffkl, line 13: nit: The motivation is spelled out nicely below so I think this should just be more brief.

```suggestion
We propose that we enable the `PrunePackageReference` feature by default for *all* .NET (Core) and .NET Standard 2.0 and above in .NET 10 preview 1.
```

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+
+This has 2 benefits:
+
+1. Reduces false positives by NuGet Audit, Component Governance and other dependency auditing tools.
+Our data investigations show that up 30% of transitive vulnerabilities are within packages that may be removed by pruning.
+1. Improves performance, by virtue of having smaller dependency graphs and fewer packages downloaded.
+
+For a feature like this, normally we'd need roll it out based on the target framework, but our telemetry shows that up to 30% of transitive vulnerability warnings may be addressed by enabling the pruning feature.
```
jeffkl, line 21: > normally we'd need roll it out based on the target framework

Is it worth explaining why here?  Or is everyone going to understand?

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+The list of packages being removed is the exact same that's part of the build time conflict resolution in the .NET SDK. In a way, this moves the conflict resolution to an earlier part of the build.
+The packages that are being pruned carried assemblies that were excluded from the final output of the project by the aforementioned conflict resolution.
+
+- As we are enabling this feature in .NET 10 Preview 1, it gives a very long runway should we eventually decide to go with a more convervative roll-out strategy.
```
jeffkl, line 25: It sounds like we're going to be careful to only prune packages that are being handled by conflict resolution at the moment, which reduces risk.  Is there a way to make sure that _we_ don't accidentally prune more/less than today?  Otherwise, I would call that out as a risk.  There might not be any enforcement in the initial rollout to guarantee that things are the same, only that we are trying to ensure that it happens that way.

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+- If there are any issues, users can easily disable the feature by setting `RestoreEnablePackagePruning` to `false` in their project/props.
```
jeffkl, line 26: Having an escape hatch is key, we should just make sure that we have a tracking issue to get this documented somewhere so people can find it.

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+- dotnet/aspnetcore - <https://github.com/dotnet/aspnetcore/pull/60025>
+
+### Observable build changes
+
+- Customers who expect the output of their project to produce identical results will notice smaller deps files. This is intentional.
+- Customers who observe the output of component governance / NuGet Audit will notice these components are no longer reported. This is intentional.
+- Customers who use packages lock files (1% usage) with locked mode may run into restore failures due to restore bringing in fewer dependencies. This is intentional.
+- Customers with builds that have customization may run into build failures if they are:
+  - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
+    - Manual assembly coping - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
```
jeffkl:
```suggestion
    - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
```

accepted/2025/prune-package-reference-rollout.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+- Enabling pruning for .NET Framework.
```
jeffkl, line 70: Just curious, we're making this call based on telemetry.  Is it worth calling out how many restores will be left out if we don't include .NET Framework?

jeffkl (commented):
Looks good, just a few comments.

accepted/2025/prune-package-reference-rollout.md
ericstj: For example - API compat.  Suppose I wanted to compare what an old package provided vs new. Definitely not customer facing thing.

accepted/2025/prune-package-reference-rollout.md
nkolev92: Great point. 

I have https://github.com/NuGet/Home/issues/14002, which I'll document with 9.0.200.

accepted/2025/prune-package-reference-rollout.md
nkolev92: We don't know that unfortunately.

accepted/2025/prune-package-reference-rollout.md
richlander: Another option is to enable it by default only when transitive is enabled.

nkolev92: @jeffkl 
I think it's ok in this case. We just want to roll out with an explicit opt-in.

@richlander 

Yep, that's added as an alternative: 
```md
- Enable pruning when NuGetAuditMode is set to `all` - reducing transitive false positives for audit.
```

accepted/2024/prune-package-reference.md
```diff
@@ -155,7 +155,8 @@ Given that packages are allowed to bring in a shared framework, and that is not
 
 ## Drawbacks
 
-- The risk of this feature is driven by the packages that are being pruned. Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there.
+- The risk of this feature is driven by the packages that are being pruned.
+Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there. However, there is a change we end up pruning more/fewer packages in the end. There's also a chance we end up getting new dependencies since pruning packages may have an effect on the outcome of the cousin dependency rule.
```
jeffkl:
```suggestion
Given that the list of packages that will be provided by the SDK is already being pruned in a different way, one could argue that the risks have been well mitigated there. However, there is a chance we end up pruning more/fewer packages in the end. There's also a chance we end up getting new dependencies since pruning packages may have an effect on the outcome of the cousin dependency rule.
```

jeffkl (approved):

PR #14119, opened 2025-02-20
Require implicit restore for commands that require restore

### LINKED ISSUE #13406: `dotnet list package` doesn't `restore`

Nigusu-Allehu:
Fixes: https://github.com/NuGet/Home/issues/13406

rendered https://github.com/NuGet/Home/blob/dev-nyenwork-implicit-restore/accepted/2025/when-to-require-implicit-restore.md

### DIFF
--- a/accepted/2025/when-to-require-implicit-restore.md
+++ b/accepted/2025/when-to-require-implicit-restore.md
@@ -57,6 +57,13 @@ If the `--no-restore` flag is specified, implicit restore should be bypassed.
 <!-- What is the impact of not doing this? -->
 1. **Requiring Manual Restore**: Users must manually run `dotnet restore` before commands. Rejected as it introduces usability friction.
 
+This was discussed in the following PR: https://github.com/NuGet/Home/pull/14065
+
+The proposal: 
+The proposal aims to improve the user experience of the `dotnet list package` command by ensuring it only displays accurate and up-to-date package information.
+If the project has been modified since the last restore, the command should detect the change and prompt the user to run `dotnet restore` again before listing packages.
+This prevents users from being misled by stale data and reinforces the expectation that a restore must reflect the current state of the project.
+
 ## Prior Art
 
 <!-- What prior art, both good and bad are related to this proposal? -->
--- a/accepted/2025/when-to-require-implicit-restore.md
+++ b/accepted/2025/when-to-require-implicit-restore.md
@@ -29,24 +29,22 @@ Ensuring an implicit restore will maintain synchronization between the assets fi
 
 When a user executes a NuGet command that requires restore, the system should:
 
-1. **Check if restore has been performed and is up to date**
-   - If it is missing, restore must be triggered implicitly.
-   - If it is outdated (i.e., the project file has changed since the last restore), implicit restore should also occur.
+1. **Perform restore**
 
 2. **Provide a `--no-restore` option**
    - Users who want to skip restore due to performance concerns or when they are confident the assets file is up-to-date should be able to pass this flag.
 
 3. **Handle scenarios where restore is unsuccessful**
    - If implicit restore fails due to network issues, package source unavailability, or conflicting dependencies, the command should fail with a clear error message.
-   - If `--no-restore` is specified and the assets file is missing or outdated, the command should either:
-     - Proceed with a warning that the results may be inaccurate.
+   - If `--no-restore` is specified, the command should:
+     - Assume the information from the assets file is correct.
 
 ### Technical explanation
 
 <!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
 
-The command should invoke restore **only when necessary**, checking the assets file has against the project dgspec has.
-The `--no-restore` flag should override automatic restore, and the command should verify the presence of a valid assets file before proceeding.
+The command must execute restore **prior to performing its own actions**. It should assume that restore will handle no-op restores when restore is unnecessary.
+If the `--no-restore` flag is specified, implicit restore should be bypassed.
 
 ## Drawbacks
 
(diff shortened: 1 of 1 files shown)


accepted/2025/when-to-require-implicit-restore.md
```diff
@@ -0,0 +1,81 @@
...
+   - If it is missing, restore must be triggered implicitly.
+   - If it is outdated (i.e., the project file has changed since the last restore), implicit restore should also occur.
+
+2. **Provide a `--no-restore` option**
+   - Users who want to skip restore due to performance concerns or when they are confident the assets file is up-to-date should be able to pass this flag.
+
+3. **Handle scenarios where restore is unsuccessful**
+   - If implicit restore fails due to network issues, package source unavailability, or conflicting dependencies, the command should fail with a clear error message.
+   - If `--no-restore` is specified and the assets file is missing or outdated, the command should either:
+     - Proceed with a warning that the results may be inaccurate.
```
jeffkl: Looks like this is missing other options.  I think we should detect if restore is out-of-date and log a warning if the user specified `--no-restore`.

accepted/2025/when-to-require-implicit-restore.md
```diff
@@ -0,0 +1,81 @@
...
+## Explanation
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
+<!-- Introduce new concepts, functional designs with real life examples, and low-fidelity mockups or  pseudocode to show how this proposal would look. -->
+
+When a user executes a NuGet command that requires restore, the system should:
+
+1. **Check if restore has been performed and is up to date**
```
nkolev92: Just call restore, it'll perform it's own up to date check. 
That's how build is implemented. It just calls restore, restore determines if anything needs to happen.

accepted/2025/when-to-require-implicit-restore.md
nkolev92: tbh, I wouldn't mess with the outdated check. 
The rest of the implicit commands also don't mess with it. 
If you skip the implicit command, we should just assume it's correct until we have a good reason to make it more costly.

In the worst case scenario, this works as well as today's list package does.

accepted/2025/when-to-require-implicit-restore.md
```diff
@@ -0,0 +1,81 @@
...
+3. **Handle scenarios where restore is unsuccessful**
+   - If implicit restore fails due to network issues, package source unavailability, or conflicting dependencies, the command should fail with a clear error message.
+   - If `--no-restore` is specified and the assets file is missing or outdated, the command should either:
+     - Proceed with a warning that the results may be inaccurate.
+
+### Technical explanation
+
+<!-- Explain the proposal in sufficient detail with implementation details, interaction models, and clarification of corner cases. -->
+
+The command should invoke restore **only when necessary**, checking the assets file has against the project dgspec has.
```
nkolev92: Determining whether restore is necessary is basically no-op. The first step of the restore command is no-op, so while there's some extra overhead of calling the msbuild process, I think in terms of correctness we're better off not duplicating the restore no-op logic.

nkolev92 (commented):
I think we should just follow what build does and just trigger restore and let it do it's own checks.

accepted/2025/when-to-require-implicit-restore.md
Nigusu-Allehu: Updated: If --no-restore is specified, assume the assets file is correct (up to date)

jeffkl (approved):

nkolev92 (approved):

accepted/2025/when-to-require-implicit-restore.md
```diff
@@ -0,0 +1,79 @@
...
+
+- `dotnet build` and `dotnet run` both trigger restore implicitly when needed.
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+- How should failure scenarios be communicated? Should errors halt execution or only show warnings?
```
nkolev92, line 82: I think errors halt execution, similarly as restore on build does.

Issue #12877, opened 2023-09-10
Support Bearer token auth in NuGet authentication plugins


joelverhagen:
### NuGet Product(s) Involved

MSBuild.exe

### The Elevator Pitch

Credential providers should be able to acquire bearer tokens (e.g. OAuth 2.0) tokens in response to a `WWW-Authenticate: Bearer ...` response header and inject an `Authorization: Bearer ...` request header for requests made to a package source.

Currently, this behavior works with `Basic` auth and probably others supported by .NET runtime's [`AuthenticationHelper`](https://github.com/dotnet/runtime/blob/de0ab156194eb64deae0e1018db9a58f7b02f4a3/src/libraries/System.Net.Http/src/System/Net/Http/SocketsHttpHandler/AuthenticationHelper.cs#L110-L129) but not for the `Bearer` scheme.

In my opinion, the .NET authentication model does not appear to cleanly fit bearer tokens since the `ICredentials` type (and the returned `NetworkCredential` instance) are username and password-based. A bearer token is a single credential without defined username and password values, although many bearer tokens contain username-like information as claims and the token itself (or the hashed component of a JWT) can be seen in some ways as a "password" or secret value.

If a `GetAuthenticationCredentialsResponse` is returned by a credential provider with a `Bearer` authentication method, it is never used by the .NET authentication stack, despite the NuGet `SecurePluginCredentialProvider` type receiving the response and adding it to the credential cache.

It is possible for the issue to be worked around by stuffing a token into a basic auth credential as the password and using an arbitrary value for the username. This is what the Azure DevOps credential provider does here: https://github.com/microsoft/artifacts-credprovider/blob/dc5f4a2da94af69f444b5fd709e889dce23ac026/CredentialProvider.Microsoft/CredentialProviders/Vsts/VstsCredentialProvider.cs#L162-L167 (this code stuffs a session token as a basic auth credential and uses `VssSessionToken` as the username)

It would allow better tooling support and 3rd party client support if the NuGet protocol supported `Bearer`  token auth instead of just `Basic`. For example, integrated OAuth 2.0 support in Postman or Insomnia could work seamlessly with a package source supporting bearer tokens. A first step in this would be to unlock credential provider support for bearer tokens.

### Additional Context and Details

[...]

PR #14174, opened 2025-03-10
[Tech Spec] Supporting Bearer Auth

jgonz120:
Tech spec for https://github.com/NuGet/Home/issues/12877


accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+To circumvent this Azure Artifacts sends their key using basic auth, passing the key as a password in a network credential.
+By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token.
+
+## Explanation
+
+### Functional explanation
+
+The push command will no longer display a warning when no API key is specified.
+
+When the NuGet client receives an unauthorized response, it will check to see if the response header includes bearer, if it does it’ll retry the request passing the credential as a bearer token in the authentication header.
```
donnie-msft: Would it be beneficial to have a way to check that a feed supports bearer tokens prior to making a request? Can a feed advertise that it supports bearer tokens in index.json which we would have already requested?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Technical explanation
+
+[HttpSourceAuthenticationHandler.SendAsync](https://github.com/NuGet/NuGet.Client/blob/313eecd3af442ee2eeed2e6decf310858934ab21/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs?plain=1#L68) will be updated to check if the previous response contains a WWW-Authenticate: Bearer header.
```
donnie-msft, line 36: Is the presumption that the previous response failed, and so now we'll try bearer tokens?
Should this knowledge be cached for feeds to avoid repeatedly sending unauthenticated requests followed by a second request with bearer?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+
+Additionally due to the [limitation in .NET](https://github.com/dotnet/runtime/issues/91867#issue-1889923702) we can’t send these types of credentials using the HttpClientHandler.
+To circumvent this Azure Artifacts sends their key using basic auth, passing the key as a password in a network credential.
+By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token.
+
+## Explanation
+
+### Functional explanation
+
+The push command will no longer display a warning when no API key is specified.
```
donnie-msft: Does this mean we're moving the CLI away from the ApiKey option?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+We would need to verify that these changes do not negatively impact feeds which use basic authentication as a workaround.
+
+## Prior Art
+
+## Unresolved Questions
+
+## Future Possibilities
+
+Encourage .NET to implement [Issue #91867](https://github.com/dotnet/runtime/issues/91867#issue-1889923702), allowing for native support of bearer tokens.
+This would allow us to remove the code required to pass the bearer token in the header and instead let the framework handle it.
```
donnie-msft, line 71: Nice! Sounds like in the future there'll be a PR with only removed lines of code and existing tests working as expected :)

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+This would require moving NuGet HTTP requests out of the devenv.exe into a service hub process or waiting for devenv.exe to move to .NET core.
```
donnie-msft, line 72: Interesting! Are there other reasons that we should start doing out of proc HTTP sooner than later?

accepted/2025/supporting-bearer-auth.md
joelverhagen: This is a good question. Personally, I am not sure. My understanding of auth in HTTP is that [`WWW-Authenticate`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) is the blessed way of advertising the supported authentication methods, but this is returned with a 401 challenge (essentially after a request has failed without proper credential). It seems what you're getting at is whether the client should be able to know about these available auth methods prior to hitting a 401. I think this only matters for public access (anonymous allowed) feeds like NuGet.org where an auth challenge will not be hit on the `index.json` or if the `index.json` auth challenge differs, say, per resource type. NuGet.org has auth challenge differing on resource type -- specifically the `PackagePublish` family of resource types requires auth but none of the other resources do.

I too wonder if it would be beneficial for the client impl to know this in advance, before seeing a 401.

accepted/2025/supporting-bearer-auth.md
joelverhagen: I think we should still show the warning if no valid credential was found, i.e. if our case (`WWW-Authenticate: bearer` is returned and a credential provider returns a `bearer` token) is NOT triggered. This will provide the same somewhat helpful warning message as before.

For Azure DevOps, I think they would opt out of the confusing warning by transitioning to a cred provider that provides `bearer` thus suppressing the warning in the same way as NuGet.org.

accepted/2025/supporting-bearer-auth.md
jgonz120: In this situation I think it's best to stick to the defined protocol. It's well documented and standard practice.

Also thinking of the push command, the user can provide the URL to push to, so we might not have the index.json to determine what auth the feed supports.

joelverhagen: Good point on the push URL difference. I don't know how widely people push to `https://api.nuget.org/v3/index.json` vs. `https://www.nuget.org/api/v2/package`. I think we can't know from server side data.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+The push command will no longer display a warning when no API key is specified.
+
+When the NuGet client receives an unauthorized response, it will check to see if the response header includes bearer, if it does it’ll retry the request passing the credential as a bearer token in the authentication header.
+
+### Technical explanation
+
+[HttpSourceAuthenticationHandler.SendAsync](https://github.com/NuGet/NuGet.Client/blob/313eecd3af442ee2eeed2e6decf310858934ab21/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs?plain=1#L68) will be updated to check if the previous response contains a WWW-Authenticate: Bearer header.
+If it does, then before sending the request the method will attempt to retrieve the token from the credentials provided and pass them in the authentication header if it’s available.
+
+[GetAuthenticationCredentialsRequest](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsRequest.cs?plain=1#L12C25-L12C60) will be updated to include a new field, AcceptedHeaders, that will tell the credential providers which auth headers can be used.
```
joelverhagen: The name could be `WwwAuthenticate` to align with the response header and match the naming convention in .NET
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.headers.httpresponseheaders.wwwauthenticate?view=net-9.0

Ideally client does minimal parsing and passes the value verbatim to the credential provider (perhaps as a string or array of strings, akin to https://learn.microsoft.com/en-us/dotnet/api/system.net.http.headers.headerstringvalues?view=net-9.0). It's the cred provider that knows how to interpret the auth scheme, realm, other auth parameters, etc.

joelverhagen: Maybe `WwwAuthenticateHeader` to be clear in context?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+This can be used by CredentialProviders to know if they’re communicating with a version of NuGet that supports Bearer tokens.
+If AcceptedHeaders is null, then the version of nuget they are communicating with does not support bearer tokens. Credential providers that want to use bearer auth will return a [GetAuthenticationCredentialsResponse](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs?plain=1#L14C1-L14C61) that has “bearer” in the AuthenticationTypes list.
+
+When deciding which protocol to use for authentication we will use a combination of the headers provided and the results from ICredentials.GetCredential.
+If the header indicates the server accepts basic, and GetCredential returns a result for basic, we will use that for the request.
+Similarly, if the header accepts bearer, and GetCredentials returns a result for bearer, we will use that.
```
joelverhagen, line 48: Where will the token be stuffed? In the password field? Seems like an alternative is to have a simple new `AuthorizationHeader` property. This essentially gives the cred provider full control on setting the `Authorization` header on the next request side stepping the `GetCredential` and [`ICredential`](https://learn.microsoft.com/en-us/dotnet/api/system.net.icredentials?view=net-9.0) interface entirely.

I am reading between the lines here, but it seems that `ICredential` flow will NOT be used in conjunction with `HttpClient` when `bearer` is returned. Instead https://learn.microsoft.com/en-us/dotnet/api/system.net.http.headers.httprequestheaders.authorization?view=net-9.0 will be used?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+Currently NuGet.org takes in the API key through the [X-NuGet-ApiKey](https://learn.microsoft.com/en-us/nuget/api/package-publish-resource#request-parameters) header.
+The CLI for pushing provides a warning if the ApiKey is not specified causing feeds, [like Azure Artifacts](https://learn.microsoft.com/en-us/azure/devops/artifacts/nuget/publish?view=azure-devops#publish-packages), to tell users to pass a placeholder value into the command.
+
+Additionally due to the [limitation in .NET](https://github.com/dotnet/runtime/issues/91867#issue-1889923702) we can’t send these types of credentials using the HttpClientHandler.
+To circumvent this Azure Artifacts sends their key using basic auth, passing the key as a password in a network credential.
+By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token.
+
+## Explanation
+
+### Functional explanation
```
joelverhagen, line 27: I want to confirm because this is a pleasant surprise to me -- the credential provider protocol works just today on NuGet push? I remember @kartheekp-ms mentioning something about a 401/403 difference in nuget push + credential providers. But I can't remember.

It might be worth expanding which status codes will NOT work in this flow (for example 403) and what happens -- unchanged by this design.

accepted/2025/supporting-bearer-auth.md
jgonz120: I know Artifactory allows for pushing to subdirectories in the package repo.

accepted/2025/supporting-bearer-auth.md
jgonz120: That's the plan, we would need to continue to support it since other feeds might still use it.

jgonz120: Would a new warning make sense? "We were unable to authenticate to the source: {uri} please verify your authentication for this source" or something.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,69 @@
...
+- Jonatan Gonzalez ([jgonz120](https://github.com/jgonz120))
+- Issue [#12877](https://github.com/NuGet/Home/issues/12877)
+
+## Summary
+
+Update the NuGet CLI to allow for bearer token authentication when publishing packages.
+
+## Motivation
+
+NuGet.org wants to shift to using OIDC authentication for publishing packages, enabling the use of short lived, 15-minute, API keys for authentication.
```
joelverhagen: I think we should clarify something here since I had a mild panic attack while I refreshed why we went down this general direction.

The `bearer` token that NuGet.org will acquire and refresh via the credential provider _is a short lived API key_ and not an OIDC token. An OIDC token exchange _will_ happen but it will be inside of the cred provider.

`GetAuthenticationCredentialsRequest` -> NuGet.org cred provider -> vendor specific OIDC token acquisition -> token trade via https://github.com/NuGet/NuGetGallery/pull/10308 -> return API key to NuGet client in `GetAuthenticationCredentialsResponse` based on the protocol defined in this doc -> NuGet client sends an `Authorization: bearer <short lived API key>` request header as part of push

Conceptually an API key is a bearer token, although it is not an OAuth bearer token. We are leaving the definition of what "bearer" means to the package source + cred provider combination.

accepted/2025/supporting-bearer-auth.md
jgonz120: Right, HttpClient doesn't support bearer tokens with ICredential. I love the idea of offloading all the responsibility of setting the headers to the cred provider! I'll check with @embetten to see what they think.

accepted/2025/supporting-bearer-auth.md
joelverhagen, line 27: We should explain that credential expiration should also trigger the cred provider flow again. Suppose a bearer token expires. What does server need to return so that the credential is refreshed? Perhaps this is just 401?

What if the credential is valid but has insufficient privileges? I think cred provider should NOT be invoked and the command should fail. Perhaps this is 403?

jgonz120: There's a configuration set that determines if we prompt on 403. So it depends on the configuration, do we want to always prompt on 403?

joelverhagen: I am not sure. I think to reduce risk we would keep current behavior for flows that don't use `bearer`. If the flow uses `bearer`, then I think we should prompt for 401 but not for 403 so we can differentiate "expired credential" = 401 from "valid credential, invalid permissions" = 403. No point in prompting when the user is trying to upload a package they do not have access to. But I am not confident on this approach.

accepted/2025/supporting-bearer-auth.md
jgonz120: For the first request this is expected, we keep track of the credential that was returned from cred provider as an ICredential. I can update the wording but Joel gave a suggestion I like below, where the cred provider will return the auth header. If we do that then we could store that and use it for subsequent requests.

accepted/2025/supporting-bearer-auth.md
donnie-msft: Nice! It's probably worth stating that this is the beginning of slowly encouraging a change in behavior. The motivation doesn't explicitly say "and it's a direction we want to begin moving towards with this spec."

donnie-msft: I thought similar to Joel initially and talked myself out of it since it felt like we're moving away from API key. Yet, now I'm back on the bandwagon that we need a warning. I like your idea of a new warning message. Should it even call out that your source "Doesn't support Bearer Token auth, and no API key was specified"? A way to hint that there's benefits to using sources that do support Bearer Tokens. 

And to be clearer so we don't get tickets like "why didn't bearer token auth work like your blog post said!?".

accepted/2025/supporting-bearer-auth.md
donnie-msft: Another place I was going with this as well is can we cache the knowledge in client that this feed supports bearer tokens or that it does not support it, and therefore not make requests with/without bearer depending on that knowledge?

This caching could be a consideration regardless of whether my comment about a feed advertising its support is implemented or not.

accepted/2025/supporting-bearer-auth.md
jgonz120: Looks like we set that value to !hasApiKey when publishing packages. Otherwise, it's set from the configuration. So, if the user is using bearer auth, we should expect them not to set the API key, and we would prompt on 403.

jgonz120: Looking at the code, if we get 401 it would automatically go through the cred provider again. I can add a line to make it clear that on 401 we get a new credential and don't use a cached one or something. 

For that second one I think returning 403 is what makes sense. Should probably decide what we want to do about prompting and 403 though: https://github.com/NuGet/Home/pull/14174/files/5b0ff0b203d3b3a91ae53713b26e704fcf1ad386#r1990090946

jgonz120: Also looking at the code it does appear that on retry we go to the cred provider. We could pass [CredentialRequestType](https://github.com/NuGet/NuGet.Client/blob/035e2ed98b082e68b27d926631d285b34795908c/src/NuGet.Core/NuGet.Configuration/Credential/CredentialRequestType.cs#L3) to the cred provider and let them decide what to do with a forbidden response.

accepted/2025/supporting-bearer-auth.md
jgonz120: If we do a new warning, I want it to be generic and just encourage users to verify how authentication is supposed to be done with the feed owner. 

For the old warning, we wouldn't know to log it until after we get the first 401 response.  
So we could say, we log the "NoApiKey" warning if we get a 401 response and are unable to get any authentication credentials?

joelverhagen: Sounds good. The existing warning is good for API key only feeds. And we should not suggest to the user that they can use a cred provider for the source (one may not exist) or use a bearer token (it may not be supported and shouldn't be provided manually). So your plan is good.

accepted/2025/supporting-bearer-auth.md
joelverhagen: Interesting. I wonder how we can improve this. If someone doesn't have the right permissions we shouldn't keep prompting cred provider over and over right?

joelverhagen: This is duplicate to thread https://github.com/NuGet/Home/pull/14174#discussion_r1989871212 now. I think.

accepted/2025/supporting-bearer-auth.md
jgonz120: Either way if we don't have credentials, we will be making the first request to get the initial 401 response. If we want to cache something, I think the response from the credential provider makes the most since, which we already do. I reworded it though so it doesn't sound so prescriptive on whether or not bearer is returned

accepted/2025/supporting-bearer-auth.md
jgonz120: Updated it to keep the old warning for now, seems easiest

accepted/2025/supporting-bearer-auth.md
jgonz120: Added some wording about the expiration.

jgonz120: Yea, I think we could focus the conversation on 403 here? Though it feels a little out of scope... but if we get 403 and request credentials again, my expectation would be that the user would try a different credential. If they were to use the same credential again then we should just fail saying auth failed. 

From the code it looks like when we don't get any credentials, we pop out of the loop that retries. 

So we could make sure to pass both the `wwwAuthenticate` and `[CredentialRequestType](https://github.com/NuGet/NuGet.Client/blob/035e2ed98b082e68b27d926631d285b34795908c/src/NuGet.Core/NuGet.Configuration/Credential/CredentialRequestType.cs#L3)` to the credential provider. At that point it's up to the credential provider to decide what to do. Could prompt again and if they get the same credential, they could just return null.

joelverhagen (dismissed):
Looks good. It is a relatively high level doc with some details left to the implementation. I am fine with this if your team is. Seems like some things are best figured out once you see how the code is factored.

accepted/2025/supporting-bearer-auth.md
jgonz120: Added a line in the motivation

accepted/2025/supporting-bearer-auth.md
joelverhagen: > Yea, I think we could focus the conversation on 403 here? Though it feels a little out of scope... but if we get 403 and request credentials again, my expectation would be that the user would try a different credential. If they were to use the same credential again then we should just fail saying auth failed.

Yes, perhaps in an interactive environment. There are non-interactive environments like CI too where the cred provider will be useful (as is Artifacts cred provider today). For interactive, it would be frustrating signing in over and over when the problem is not your password to, say, your Microsoft account (that turns into an OIDC token -> short lived API key) but is more about signing into the wrong user entirely. Or you as the user just don't have access to that package at all.

> Could prompt again and if they get the same credential, they could just return null.

This makes sense. If the cred provider has context to improve the UX and say "nope, sorry, that user was tried already and it didn't work due to the 2nd 401 and the cred provider being launched again, so we'll abort without asking for your username again) -- then this works. I think we should allow ourselves a little grace to iterate this tech design as we write the first cred provider that uses it, i.e. the nuget.org one for OIDC + short lived API keys.

jgonz120: We also have a retry cache, so before we make the request to the credential provider we check if an entry already exists using this [key](https://github.com/NuGet/NuGet.Client/blob/035e2ed98b082e68b27d926631d285b34795908c/src/NuGet.Core/NuGet.Credentials/CredentialService.cs?plain=1#L235) so for a non-interactive request if "IsRetry" is true and we got 403 you know not to try again.

jgonz120: For the interactive experience, who should be responsible for not prompting again? NuGet Client or the Credential Providers?

joelverhagen: If possible, NuGet Client so the same logic need not be replicated across multiple cred provider impls. Does that sound reasonable?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
+# Supporting Bearer Token Auth for Credential Providers
+
+- Jonatan Gonzalez ([jgonz120](https://github.com/jgonz120))
+- Issue [#12877](https://github.com/NuGet/Home/issues/12877)
+
+## Summary
+
+Update the NuGet CLI to allow for bearer token authentication when publishing packages.
```
zivkan: "the NuGet CLI" refers specifically to NuGet.exe.  Unless there's something later in the spec that makes it compelling, I'd expect auth changes to affect dotnet CLI, msbuild restore, and all NuGet functionality in VS as well.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+NuGet.org wants to shift to using bearer token authentication for publishing packages, enabling the use of short lived, 15-minute, API keys for authentication.
+This would reduce the security risk of a key leak, since currently the long-lived keys can be valid for up to a year.
+
+Currently NuGet.org takes in the API key through the [X-NuGet-ApiKey](https://learn.microsoft.com/en-us/nuget/api/package-publish-resource#request-parameters) header.
+The CLI for pushing provides a warning if the ApiKey is not specified causing feeds, [like Azure Artifacts](https://learn.microsoft.com/en-us/azure/devops/artifacts/nuget/publish?view=azure-devops#publish-packages), to tell users to pass a placeholder value into the command.
+By removing ApiKey as a requirement feeds will no longer have to document sending a placeholder.
+
+Additionally due to the [limitation in .NET](https://github.com/dotnet/runtime/issues/91867#issue-1889923702) we can’t send these types of credentials using the HttpClientHandler.
+To circumvent this Azure Artifacts sends their key using basic auth, passing the key as a password in a network credential.
```
zivkan: I suggest "basic auth", rather than "a network credential".

`NetworkCredential` is a .NET specific type that I don't think is well known. I think HTTP Basic auth is better known.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token.
```
zivkan:
```suggestion
By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token for push or get requests.
```

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Explanation
+
+### Functional explanation
+
+The push command will only display the no API key warning when the API key is not specified and we were unable to authenticate with the source.
+
+When the NuGet client receives an unauthorized response, it will pass the `WWW-Authenticate` header to the credential providers.
```
zivkan:
```suggestion
When the NuGet client receives an unauthorized response, it will pass the `WWW-Authenticate` headers to the credential providers.
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

> Multiple headers can also be sent in a single response

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+[GetAuthenticationCredentialsRequest](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsRequest.cs?plain=1#L12C25-L12C60) will be updated to include a new field, `WwwAuthenticateHeader`, that will tell the credential providers which auth headers can be used.
+This can be used by CredentialProviders to know if they’re communicating with a version of NuGet that supports Bearer tokens.
+If `WwwAuthenticateHeader` is null, then the version of nuget they are communicating with does not support bearer tokens. Credential providers that want to use bearer auth will return a [GetAuthenticationCredentialsResponse](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs?plain=1#L14C1-L14C61) that has “bearer” in the AuthenticationTypes list.
+
+When deciding which protocol to use for authentication we will use a combination of the headers provided and the results from ICredentials.GetCredential.
```
zivkan:
```suggestion
When deciding which auth scheme to use for authentication we will use a combination of the headers provided and the results from ICredentials.GetCredential.
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+If the header indicates the server accepts basic, and GetCredential returns a result for basic, we will use that for the request.
+Similarly, if the header accepts bearer, and GetCredentials returns a result for bearer, we will use that.
+
+To maintain the “PreAuthenticate” functionality, a cache can be used to keep track of URLs and the credential previously used to authenticate against it.
```
zivkan: NuGet disabled `PreAuthenticate` because:

* https://github.com/dotnet/runtime/issues/93340

I think it would be good to re-enable PreAuthenticate, but "maintain" doesn't seem right since we're not using it, so it can be a future improvement.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+When credential providers return a bearer token, the request will be retried by adding the bearer token to the authentication header.
+When we make a request with an expired token, the server should return a 401 Unauthorized respond, causing a new request for credentials.
+
+### Technical explanation
+
+[HttpSourceAuthenticationHandler.SendAsync](https://github.com/NuGet/NuGet.Client/blob/313eecd3af442ee2eeed2e6decf310858934ab21/src/NuGet.Core/NuGet.Protocol/HttpSource/HttpSourceAuthenticationHandler.cs?plain=1#L68) will be updated to check if the previous response contains a WWW-Authenticate: Bearer header.
+If it does, then before sending the request the method will attempt to retrieve the token from the credentials provided and pass them in the authentication header if it’s available.
+
+[GetAuthenticationCredentialsRequest](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsRequest.cs?plain=1#L12C25-L12C60) will be updated to include a new field, `WwwAuthenticateHeader`, that will tell the credential providers which auth headers can be used.
```
zivkan:
```suggestion
[GetAuthenticationCredentialsRequest](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsRequest.cs?plain=1#L12C25-L12C60) will be updated to include a new field, `WwwAuthenticateHeaders`, that will tell the credential providers which auth headers can be used.
```

As mentioned in another comment, there can be multiple headers when the server supports multiple auth schemes

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+This can be used by CredentialProviders to know if they’re communicating with a version of NuGet that supports Bearer tokens.
```
zivkan:
```suggestion
This can be used by Credential Providers to know if they’re communicating with a version of NuGet that supports Bearer tokens.
```

nitpick 😁 although perhaps credential providers don't need to be capitalised

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+If `WwwAuthenticateHeader` is null, then the version of nuget they are communicating with does not support bearer tokens. Credential providers that want to use bearer auth will return a [GetAuthenticationCredentialsResponse](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs?plain=1#L14C1-L14C61) that has “bearer” in the AuthenticationTypes list.
```
zivkan:
```suggestion
If `WwwAuthenticateHeader` is null, then the version of nuget they are communicating with does not support bearer tokens.
Credential providers that want to use bearer auth will return a [GetAuthenticationCredentialsResponse](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs?plain=1#L14C1-L14C61) that has “bearer” in the AuthenticationTypes list.
```

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+When deciding which protocol to use for authentication we will use a combination of the headers provided and the results from ICredentials.GetCredential.
+If the header indicates the server accepts basic, and GetCredential returns a result for basic, we will use that for the request.
+Similarly, if the header accepts bearer, and GetCredentials returns a result for bearer, we will use that.
```
zivkan, line 48: What if the server supports both bearer and basic, and the cred provider provides creds, also specifying both bearer and basic?  Which auth scheme wins?   Since you listed basic first, does that win?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+To maintain the “PreAuthenticate” functionality, a cache can be used to keep track of URLs and the credential previously used to authenticate against it.
+
+## Drawbacks
+
+Even though we are moving away from sending the bearer token as a password during the request, the ICredential.GetCredential explicitly returns a NetworkCredential, which requires a username and password.
```
zivkan: nitpick, but I don't like the first half of the sentence.

NuGet never supported bearer auth, so I just don't agree that "we're moving away" from anything. If cred providers "lie" to NuGet and give us HTTP basic auth credentials with what happens to be an OAuth token as the password, that's a decision of the cred provider, not NuGet.

Moreover, I feel like saying "moving away from" implies we're breaking backwards compatibility with something. But we're not. If a cred provider continues to give us http basic credentials that happen to use an OAuth token in the password field, that's completely opaque to NuGet, and will continue to work.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Rationale and alternatives
+
+### Alt - Use basic Authentication
+
+This option would have NuGet.org implement the credential provider like Azure Artifacts does for authentication.
+The client wouldn’t need to be updated, and the API Key would be sent over to the server using basic authentication.
+The username field would be filled with a placeholder.
```
zivkan: NuGet client wouldn't fill the username with a placeholder, the cred provider would need to.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+The client could still be updated to no longer require ApiKey.
+While this solution has minimal work for the client, it requires using an authentication mechanism that expects a username and password.
+
+We would need to verify that these changes do not negatively impact feeds which use basic authentication as a workaround.
```
zivkan: I don't understand this concern. If we don't make any changes to NuGet client, there wouldn't be any risk of regressions that need to be verified with other feeds.

If it's about the warning when an API key isn't provided, then it's up to the server to return an HTTP 401 response. Again, I don't understand what regression risk needs to be verified.

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Rationale and alternatives
+
+### Alt - Use basic Authentication
+
+This option would have NuGet.org implement the credential provider like Azure Artifacts does for authentication.
+The client wouldn’t need to be updated, and the API Key would be sent over to the server using basic authentication.
+The username field would be filled with a placeholder.
+The client could still be updated to no longer require ApiKey.
+While this solution has minimal work for the client, it requires using an authentication mechanism that expects a username and password.
```
zivkan:
```suggestion
While this solution has minimal work for the client, it requires using an authentication mechanism that `HttpClientHandler` supports via `ICredentials`, such as Basic, Digest, Negotiate (Kerberos), NTLM, etc.
```

nitpick

zivkan (commented):
just nitpick comments, but hardly surprising since I talked to you about this before you wrote the spec 😁

accepted/2025/supporting-bearer-auth.md
joelverhagen: The ICredential can return a network credential for digest and NTLM auth too right (not just basic)? When I was trying to get bearer too work it looks like those two schemes were supported too. But I may be wrong.

accepted/2025/supporting-bearer-auth.md
jgonz120: Oo good point, does this work? 
```suggestion
Update the NuGet authentication flow to allow for bearer token authentication when publishing packages.
```

accepted/2025/supporting-bearer-auth.md
jgonz120:
```suggestion
To circumvent this Azure Artifacts sends their key using basic auth, passing the key as a password.
```

jgonz120: Updated it to remove network credential, didn't want to say basic auth twice in one sentence.

accepted/2025/supporting-bearer-auth.md
jgonz120: Yea, basic would win

accepted/2025/supporting-bearer-auth.md
jgonz120: I moved it to future improvements

accepted/2025/supporting-bearer-auth.md
jgonz120: updated the wording

accepted/2025/supporting-bearer-auth.md
joelverhagen: (I guess what I'm getting at is that the network credential/`ICredential` abstraction is what restricting the flow today so it's probably good to call out that, rather than making it specific to `basic` auth). I don't feel strongly though.

jgonz120: updated~

accepted/2025/supporting-bearer-auth.md
jgonz120: That sounds reasonable to me, though right now we don't really account for this, and I don't these changes really make this that much worse, other than now 403 responses could fall under this loop.

accepted/2025/supporting-bearer-auth.md
jgonz120: fair, removed that line.

accepted/2025/supporting-bearer-auth.md
zivkan: > I guess what I'm getting at is that the network credential/ICredential abstraction is what restricting the flow today

Unless I misunderstand something, no, I wouldn't say `ICredential` or `NetworkCredential` is restricting Bearer auth. `HttpClientHandler` doesn't automatically handle it in the same way that it does for Basic, Digest, Negotiate or NTLM.

joelverhagen: Yup I agree with that wording! NetworkCredential can allow bearer if you pinch your nose about a bearer token in the username or password field 🤬

accepted/2025/supporting-bearer-auth.md
joelverhagen: Thinking about this more. Should there be symmetry between `WwwAuthenticateHeader` in the Request model and a `AuthorizationHeader` in the Response model?

So `WwwAuthenticateHeader` would provide most context the cred provider would need and `AuthorizeHeader` would explicitly side step the username+password+ICredential+NetworkCredential path.

But there would be two ways to do that same thing for, say, basic auth: ICredential+NetworkCredential+HttpClientHandler or an `AuthorizeHeader` = `basic base64(username : password)`

accepted/2025/supporting-bearer-auth.md
jgonz120: Good question, I'm not really sure. I believe @zivkan was the one that pointed this out to me?

accepted/2025/supporting-bearer-auth.md
jgonz120: Yea, if we went with this route I would say if we received an AuthorizeHeader from the cred provider we use that, if it's null, we check for the ICredential.

accepted/2025/supporting-bearer-auth.md
jgonz120: Updated wording again :P

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,72 @@
...
+This can be used by Credential Providers to know if they’re communicating with a version of NuGet that supports Bearer tokens.
+If `WwwAuthenticateHeader` is null, then the version of nuget they are communicating with does not support bearer tokens. Credential providers that want to use bearer auth will return a [GetAuthenticationCredentialsResponse](https://github.com/NuGet/NuGet.Client/blob/ab6d3b886b32e9b8f68a7f1b299c43b0c72487dc/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs?plain=1#L14C1-L14C61) that has “bearer” in the AuthenticationTypes list.
+
+When deciding which auth scheme to use for authentication we will use a combination of the headers provided and the results from ICredentials.GetCredential.
+If the header indicates the server accepts basic, and GetCredential returns a result for basic, we will use that for the request.
```
nkolev92, line 47: Do we expect the cred provider to provide bearer?
https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Protocol/Plugins/Messages/GetAuthenticationCredentialsResponse.cs#L35C30-L35C50

Not changes in the cred provider protocol I'm guessing?

accepted/2025/supporting-bearer-auth.md
```diff
@@ -0,0 +1,71 @@
...
+
+By working around the .NET limitation for bearer token usage, we can allow a more technically correct solution and allow other feeds to implement a credential provider that returns a bearer token for push or get requests.
+
+## Explanation
+
+### Functional explanation
+
+The push command will only display the no API key warning when the API key is not specified and we were unable to authenticate with the source.
+
+When the NuGet client receives an unauthorized response, it will pass the `WWW-Authenticate` headers to the credential providers.
```
embetten, line 31: If the nuget credprovider plugin API is being updated to pass in the WWW-Authenticate, are there any other API changes being considered?  

In the past, we have requested changes such as passing in preferred account hints, returning and displaying better exception messaging, etc. In these discussions, it was said that it's easier to update the API with all changes at once instead of incrementally.

accepted/2025/supporting-bearer-auth.md
zivkan: out of proc has a bunch of pros and cons, and it's entirely possible that the cons outweigh the pros. I don't want to be the only person in the team with this knowledge (especially if I'm wrong about some things, I need other people to review my ideas), so anyone interested please let me know and we can chat about it. But it's too much to write down in a random github comment.

However, even if we take NuGet in VS out of devenv and into a .NET (Core) process, VS's out of proc infrastructure doesn't exist in msbuild. So, we still wouldn't be able to take out any bearer token handling out of Nuget unless we also want to write our own out of proc infrastructure for msbuild.

nkolev92: Another thing is going out of proc for this feature as the main motivator is kind of like using a sledgehammer to crack a nut.

accepted/2025/supporting-bearer-auth.md
nkolev92: Great catch. 

This passing back of headers probably needs some sort of a plugin protocol change. 

> such as passing in preferred account hints

Isn't this for VS? 
How would NuGet even know the preferred account?

accepted/2025/supporting-bearer-auth.md
jgonz120: There's a separate thread about whether or not we want the cred provider to return the headers, which would require a protocol change. 

But yea, cred providers would provide bearer in the auth types. If they return multiple, we would bias towards the ones supported by the HttpClientHandler winning over "bearer"

accepted/2025/supporting-bearer-auth.md
embetten: > Isn't this for VS?

Preferred account hints were discussed mainly in relation to the visual studio NuGet credprovider, but I thought it was the same protocol? Due to the keychain integration, the credprovider layer does not have enough information to know which account credentials to return if there are multiple accounts in the vs keychain with access to a feed. This is similarly true for the artifacts-credprovider, however, unlike VS, in that context we can provide users an environment variable `NUGET_CREDENTIALPROVIDER_MSAL_LOGIN_HINT`. We also played with the idea of passing back multiple account credentials.

> How would NuGet even know the preferred account?

I know this is not a feature available to users through vs and/or NuGet today, but the cred provider layer isn't given enough context in these multi feed situations either.

But I don't want to completely distract from other protocol suggestions, these were just the changes I know of from previous discussions.

accepted/2025/supporting-bearer-auth.md
embetten: The artifacts-credprovider needs to call the ADO server to retrieve custom headers, including the WwwAuthenticateHeader. Even if the `WwwAuthenticateHeader` was provided, we would need to call our server, and we typically know what type of auth our feeds support. But I'm not sure how unique our situation is here.

Returning a token and auth type would be preferred to a new `AuthorizationHeader` property. In our experience, its advantageous to have the credprovider layer be as simple as possible. Considering that clients may change and place tokens in different locations over time (such as HTTP headers, API keys, directly in NuGet configs, or in URLs like pip), I think keeping the format of the returned token consistent enables more flexibility across versions. For all the other packaging clients we integrate with we return just the token, and this would increase the NuGet specific code we would need to maintain.

accepted/2025/supporting-bearer-auth.md
nkolev92: The protocols are different. 
One is a MEF integration, the other one is inter process communication. 

VS: https://learn.microsoft.com/en-us/nuget/reference/extensibility/nuget-credential-providers-for-visual-studio
CLI Cred Provider: https://learn.microsoft.com/en-us/nuget/reference/extensibility/nuget-cross-platform-plugins.

accepted/2025/supporting-bearer-auth.md
nkolev92: We should explicitly call out whatever the decision ends up being. 
It's the one piece of information that's relevant for non-NuGet.org implementers.

nkolev92: actually it's written down. disregard.

accepted/2025/supporting-bearer-auth.md
zivkan: > Even if the WwwAuthenticateHeader was provided, we would need to call our server, and we typically know what type of auth our feeds support

I don't know if you can get into it publicly, but the point of the `WWW-Authenticate` header is for servers to tell clients what auth schemes the server supports. I'm guessing by this comment that ADO is therefore using some non-standard way of sharing the same information.

Also, similar lines, sometimes I think about writing a cred provider for github, but it's impossible to know, solely from the URL, whether the server is a self-hosted GitHub Enterprise server.  So, if NuGet can pass all HTTP headers to the cred provider, then it will help all cred providers for any server that can be hosted on-prem (or on a cloud hosted, but on a private domain).

zivkan: FWIW, I just tried a `curl -v` to an Azure Artifacts private feed, and I see 3 www-authenticate headers:

> www-authenticate: Bearer authorization_uri=https://login.windows.net/*****
> www-authenticate: Basic realm="https://*******/"
> www-authenticate: TFS-Federated

But, I also see a bunch of `x-tfs-*` and `x-vss-*` headers about auth realms, and redirect URLs.

So, I think this reinforces the point that for some cred providers to know how to auth correctly, they need the full list of headers, because the server might be using custom headers that the cred provider needs to make decisions.

joelverhagen: It seems like there is a spectrum of options, ranging from:
- Simple cred provider interface, where access to 401 response headers and subsequent request heads is minimal. **Example:** stuff bearer token in the current password, no new WwwAuthenticateHeader/AuthorizationHeader properties.

*to*

- Powerful cred provider interface, where the cred provider can get and set a lot of headers. **Example:** put all 401 response headers in the `GetAuthenticationCredentialsRequest` and allow any request header to be set on the subsequent responses with a property bag on `GetAuthenticationCredentialsResponse` (this would allow custom auth models beyond WWW-Authenticate + Authorization).

Personally, I would like to move more to that 2nd extreme than stick to the 1st (which is where we are right now prior to this design). Putting all headers in the protocol is a more flexible protocol allowing custom auth schemes, but there are a lot of options in between that would meet the goals of OIDC.

What's not clear to me is what flavor of the protocol change aligns with NuGet client team's long-term goals and what would accommodate enhancements or "purifications" in the Artifacts cred provider.

accepted/2025/supporting-bearer-auth.md
jgonz120: Added to future

accepted/2025/supporting-bearer-auth.md
jgonz120: @embetten I'll reach out offline to discuss any additional changes to the protocol and we can see if they will be quick enough to implement while this is worked on.

accepted/2025/supporting-bearer-auth.md
jgonz120: I think it could make sense to provide all the response headers we received to the cred provider, regardless of which option we pick.

We could optimize for flexibility and allow the credential providers to return a network credential or the full auth header. 

Update `GetAuthenticationCredentialsRequest ` to include `HttpResponseHeaders ResponseHeaders`

Update `GetAuthenticationCredentialsResponse` to include `HttpRequestHeaders/NameValueCollection RequestHeaders` 

Update `CredentialResponse` to include `RequestHeaders`

This will allow the credential provider to specify any headers needed to auth, even custom ones.

NuGet.Client will still need to check for values in Username and Password fields since that will be used by legacy credential providers. 

So if we see that `RequestHeaders` is empty from the credential provider we use Username/Password + AuthenticationTypes combination to decide what we do. 

Something like this

```C#
if(CredResponse.RequestHeaders is not null && CredResponse.RequestHeaders.Count > 0) {
    foreach(var header in CredResponse.RequestHeaders) {
        request.AddHeader(header.key, header.value);
    }
}
else if(CredResponse.Credential.Get("basic") == null) { 
    var bearer = CredResponse.Credential.Get("bearer");
    if(bearer is not null)  {
        request.AddHeader("bearer", bearer");
    }
}
//do other stuff and make a request 🤷
```
We can handle both, and we can message clearly that we are adding support for bearer tokens but moving forward if credential providers want to handle another type of authentication, they need to be updated to request the request headers necessary.

joelverhagen: I like this solution a lot. It means a new provider could even add `X-NuGet-ApiKey` for an old service. I recommend picking a simpler model for `RequestHeaders`/`ResponseHeaders` the more clearly maps to the JSON stdout/stdin bidirectional pipe. Perhaps a List of { name: "", value: "" } POCOs or an Dictionary<string, string[]>. Technically header names are not unique in the list of request/response headers. The `HttpHeaders` type does validation on HTTP header which can cause pain. I think treating them as strings and doing `TryAddWithoutValidation` in client code on the returned `RequestHeaders` would be most flexible. But I'm touching on impl. details.

LGTM.

Frulfump:
Seems useful could it get some new attention?

aortiz-msft:
@zivkan - Could you please help understand next steps on this one?

zivkan (approved):
@aortiz-msft I think it's acceptable to merge as is, and then schedule the implementation.  If I were writing the design spec, I would have worded some things a bit differently, but that would just be for clarity. The fundamental substance of the spec is good.

aortiz-msft (approved):

PR #14211, opened 2025-03-26
Add rule for issues labeled as breaking-change

aortiz-msft:
Add rule per new .NET SDK Breaking change policy

### DIFF
--- a/.github/policies/resourceManagement.yml
+++ b/.github/policies/resourceManagement.yml
@@ -523,7 +523,7 @@ configuration:
       then:
       - addReply:
           reply: >-
-            Refer to the [.NET SDK breaking change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#required-process-for-all-net-sdk-breaking-changes)
+            For changes in NuGet functionality (such as NuGet Restore changes), refer to the [.NET SDK breaking change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#required-process-for-all-net-sdk-breaking-changes)
       description: Add breaking change doc instructions to issue
 onFailure: 
 onSuccess: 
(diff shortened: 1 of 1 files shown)


nkolev92:
Copying comments from the other side: 

1. Not every breaking change we have is a .NET SDK breaking change, so might be something to consider.

2. We also should update our release notes tool.
https://github.com/NuGet/Entropy/blob/f5c27465a5dfc85ee39b690a79d0c76df7138049/NuGetReleaseTool/NuGetReleaseTool/GenerateReleaseNotesCommand/IssueLabels.cs#L17

Putting the comments from the client PR since they're actually more relevant here.

.github/policies/resourceManagement.yml
```diff
@@ -516,5 +516,14 @@ configuration:
...
           label: Status:No recent activity
       description: '[stale community PR] [5-4] Remove "Status:No recent activity" if there are any reviews.'
+    - if:
+      - payloadType: Issues
+      - labelAdded:
+          label: breaking-change
+      then:
+      - addReply:
+          reply: >-
+            Refer to the [.NET SDK breaking change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#required-process-for-all-net-sdk-breaking-changes)
```
nkolev92: Maybe put an "if" or something to qualify that this is only needed when an issue is affecting .NET SDK?

nkolev92 (dismissed):
Minor suggestion.

.github/policies/resourceManagement.yml
donnie-msft: +1

aortiz-msft:
> Copying comments from the other side:
> 
> 1. Not every breaking change we have is a .NET SDK breaking change, so might be something to consider.
> 2. We also should update our release notes tool.
>    https://github.com/NuGet/Entropy/blob/f5c27465a5dfc85ee39b690a79d0c76df7138049/NuGetReleaseTool/NuGetReleaseTool/GenerateReleaseNotesCommand/IssueLabels.cs#L17
> 
> Putting the comments from the client PR since they're actually more relevant here.

https://github.com/NuGet/Entropy/pull/120

nkolev92 (approved):

zivkan (approved):

PR #14218, opened 2025-03-29
Update icebox-watcher.yml

jeffkl:
Remove invalid arguments

### DIFF
--- a/.github/workflows/icebox-watcher.yml
+++ b/.github/workflows/icebox-watcher.yml
@@ -15,6 +15,10 @@ jobs:
   Run-IceBox-Watcher:
     runs-on: ubuntu-latest
     steps:
+      - name: Install .NET 8.x
+        run: |
+          curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin -version 8.x --install-dir /Users/runner/hostedtoolcache/dotnet
+          dotnet --info
       - name: Set up NuGet authentication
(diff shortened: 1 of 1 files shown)


zivkan (dismissed):
it must be available since the .NET 9 SDK, but the agent still only has .NET 8 SDK pre-installed

jeffkl:
> it must be available since the .NET 9 SDK, but the agent still only has .NET 8 SDK pre-installed

Should I add a UseDotNet task instead?

I didn’t look into the problem much either. I saw this issue: https://github.com/dotnet/sdk/issues/10375

Still not sure why it stopped working though.

nkolev92:
Is there no way to roll forward a tool?

jeffkl:
Longer term, we can make the tool roll forward via the MSBuild property 

```xml
<PropertyGroup>
  <RollForward>Major</RollFoward>
</PropertyGroup>
```

For now this works, the tool targets .NET 8 and the action YAML now installs .NET 8.  The agent has .NET 8.0.1xx on it which didn't support the `--allow-rollforward` command-line argument which is why it was broken.  But these changes fix it now, I tested it by running it manually.

jeffkl:
image
https://github.com/NuGet/Home/actions/runs/14178473497/job/39718921565

nkolev92 (approved):

zivkan (approved):

Issue #14196, opened 2025-03-20
Pruning should prune and not warn for a direct reference in a multi-targeting project that does not prune on all TargetFrameworks


ericstj:
### NuGet Product(s) Affected

dotnet.exe

### Current Behavior

Multi-targeting projects may reference a package that is required in some frameworks, but part of the framework in others.

For example: a project targeting netstandard2.0;net8.0 and referencing System.Text.Json 8.0.x

Today we warn in this case, and forcing the user to write a condition.  So in the example, user would condition out the reference to STJ on net8.0.

Now what happens when they decide to upgrade that package without retargeting?  Say the version is managed centrally and they upgrade to System.Text.Json 9.0.0.

Now they *must* bring back the package reference so that their net8.0 library is compatible with the netstandard2.0 library, but nothing will catch this and remind them to do it.  That will result in runtime errors for their consumers who may have built libraries on referencing them.

### Desired Behavior

Don't warn when there is a direct reference, and it is not pruned on some TargetFrameworks.  Prune it for the user.

If the direct reference would be pruned on all TargetFrameworks, warn.

Consider *always* pruning, even when direct, to make things consistent.  Add a user gesture on the reference to opt out of pruning (rather than using a direct reference to indicate this).

richlander:
There are likely folks who have a bunch of PackageRefs that are not needed. It would be good to retain a way to get a diagnostic for references that can be safely removed.

I can imagine people merging dependabot PRs for years not realizing that it isn't needed.

ericstj:
> There are likely folks who have a bunch of PackageRefs that are not needed. It would be good to retain a way to get a diagnostic for references that can be safely removed

That's **exactly** why I mentioned that the warning should still be raised if the package is pruned for all TargetFrameworks.  I updated the title to make it clear.

ericstj:
We've had multiple instances of folks hitting this with new packages too - like System.Linq.AsyncEnumerable.  People don't understand that they need to add conditions, nor do they understand how to write those.  Folks outside framework/NuGet team aren't really experts at MSBuild conditions.  They prefer the package manager UI to help them.

richlander:
Are we considering a fixer to remove unnecessary packagerefs? I suspect people would greatly appreciated that? I can imagine running a fixer on a solution.

nkolev92:
Thanks for bringing this up Eric!

There's a couple of ideas being proposed here, my initial take on them:

[...]

Issue #14251, opened 2025-04-15
NU1510: PackageReference System.IO.Pipelines will not be pruned


glen-84:
### NuGet Product Used

dotnet.exe

### Product Version

10.0.100-preview.3.25201.16

### Impact

It's more difficult to complete my work

### Repro Steps & Context

> warning NU1510: PackageReference System.IO.Pipelines will not be pruned. Consider removing this package from your dependencies, as it is likely unnecessary.

I can't remove it, it's being used.

We're using `TreatWarningsAsErrors` as well. I can add it to `WarningsNotAsErrors` (or disable entirely with `RestoreEnablePackagePruning false`), but why would I be seeing this warning if the package doesn't seem to be part of the SDK?

### Verbose Logs

```shell

```

nkolev92:
> but why would I be seeing this warning if the package doesn't seem to be part of the SDK?

This warning is raised for packages that already available within the runtime.

How many frameworks do you have in your project?
What frameworks are those and which version of System.IO.Pipelines are you using?

glen-84:
> What frameworks are those

```xml
<TargetFrameworks>net10.0; net9.0; net8.0</TargetFrameworks>
```

> ... which version of System.IO.Pipelines are you using?

```xml
<ItemGroup Condition="'$(TargetFramework)' == 'net10.0'">
  <PackageVersion Include="System.IO.Pipelines" Version="10.0.0-preview.3.25171.5" />
</ItemGroup>

<ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
  <PackageVersion Include="System.IO.Pipelines" Version="9.0.2" />
</ItemGroup>

<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
  <PackageVersion Include="System.IO.Pipelines" Version="8.0.0" />
</ItemGroup>
```

Using CPM.

nkolev92:
What the warning is telling you is that you do not need this reference because the APIs are already available.

I think you can safely remove those references and you won't have any issues building.

cc @ericstj @viktorhofer

glen-84:
Okay, it seems that the package is available in net9.0 and net10.0, but not net8.0.

I've referenced the package conditionally for net8.0:

```xml
<ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
  <PackageReference Include="System.IO.Pipelines" />
</ItemGroup>
```

Thanks.

ericstj:
[...]

Issue #14126, opened 2025-02-21
Warning rollout for PrunePackageReference


nkolev92:
See some concerns 
- https://github.com/NuGet/NuGet.Client/pull/6142#discussion_r1889037638
- https://github.com/NuGet/NuGet.Client/pull/6142#discussion_r1889247176

Jeff said:

> I think this whole feature should be off by default but when enabled, we should consider a mode like "strict" where these errors are logged. I'm thinking of the scenario someday when I update to the latest .NET SDK and now I can't restore. If we make this on by default but just silently ignore misconfigurations, that would be a fine experience. If I want to opt into getting these errors, that would at least be a change I made. I think good documentation would help users know that they can opt into the more strict mode.

Andy said:

> Given my experience with NuGetAuditMode, I'm now a huge opponent of SdkAnalysisLevel. It doesn't work the way a lot of customers expect it to, and it doesn't work in the way that our division's leadership want us to provide backwards compatibility to customers. I think SdkAnalysisLevel doesn't work and should be considered a failed experiment we no longer use.
> 
> Unfortunately, I also don't have a suggestion on how we'll ever turn on new features by default. It's something we're going to have to have discussions about (and when we have a proposal, take it to leadership) early next year.

Furthermore in the P1 roll-out we're discovering that this is going to be common enough where it's worth gating it on the .NET 10 target to allow seamless adoption of the feature without any warnings.

nkolev92:
I have a gist with some ideas for the concept of a `strict` mode.

https://gist.github.com/nkolev92/251bf51416c85bc70f9b91e737e28397

nkolev92:
Reopening since we're still having conversations.
The conversation may still lead to this issue being closed.

nkolev92:
NU1510 and NU1511 are only raised when .NET 10 is targeted. 
So this is really P2. 

There are 2 ways forward for this issue.

1. Implement the design suggested in https://gist.github.com/nkolev92/251bf51416c85bc70f9b91e737e28397. 
2. Create a new issue for https://gist.github.com/nkolev92/251bf51416c85bc70f9b91e737e28397 and close this once is implemented https://github.com/NuGet/NuGet.Client/pull/6672. 

I'm leaning the 2nd one for now, but waiting on a final decision for https://github.com/NuGet/Home/issues/14345 first.

nkolev92:
Created https://github.com/NuGet/Home/issues/14494.

Issue #6279, opened 2017-12-08
project.assets.json should indicate "analyzer" assets like "compile" and "runtime"


nguerrera:
project.assets.json doesn't actually indicate which analyzer references a project should use. 

Unlike "compile" and "runtime", the SDK is just pattern matching on all files in the package to find analyzers: https://github.com/dotnet/sdk/blob/634680ab0ae045ba45977b6293b46f95a0385aac/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs#L251

"analyzer" assets should work like "compile" and "runtime" and the SDK should resolve them the same way. The current mechanism blocks respecting PrivateAssets and ExcludeAssets=analyzers

cc @emgarten

emgarten:
Analyzers were added without work being done on the NuGet side originally. We should add it to project.assets.json so that the SDK can switch over to using that instead of reading packages directly to find the analyzers folder.

nguerrera:
Related to this, is there a spec for how analyzer assets should be selected based on project language?

Current implementations are here:
https://github.com/NuGet/NuGet.BuildTasks/blob/1d8af3499f94a32c1d128a42faceade39c1f4337/src/Microsoft.NuGet.Build.Tasks/ResolveNuGetPackageAssets.cs#L444-L470
https://github.com/dotnet/sdk/blob/634680ab0ae045ba45977b6293b46f95a0385aac/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageDependencies.cs#L251-L272

I believe they are semantically equivalent, but both have some strange behavior if I'm reading correctly.

* Leading "analyzers" is matched without separator 
    * So "analyzersBanana/myanalyzer.dll" would match
* Leading "analyzers" is case-sensitive. 
    * Is that how the others like "lib" and "ref" are matched too?
* C# projects matches `("*/cs/*" or not "*/vb/*")` and VB matches `("*/vb/*" or not "*/cs/*")`?
    * This seems to ignore the possibility of other languages gaining analyzers. e.g. If there were F# analyzers, they would match VB and C# projects.
* Looking for language in any path component seems like a hammer vs how other nuget assets are matched.


cc @jinujoseph @natidea @jasonmalinowski

jasonmalinowski:
Adding @jmarolf too.

> This seems to ignores the possibility of other languages gaining analyzers. e.g. If there were F# analyzers, they would match VB and C# projects.

[...]

PR #14325, opened 2025-05-20
PrunePackageReference improvements, NU1510 and directs pruning

nkolev92:
Design for https://github.com/NuGet/Home/issues/14196

### DIFF
--- a/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
+++ b/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
@@ -11,6 +11,7 @@ Change the NU1510 heuristic in multi-targeting scenarios to only be raised when 
 This would ensure customers don't end up in a potentially more challenging to manage conditional PackageReference scenario. As of <https://github.com/NuGet/NuGet.Client/pull/6469>, this change is implemented.
 
 This proposal explores various ideas for pruning direct package references, the technical details around and the benefits of doing the work.
+The proposed handling of direct PackageReference that are within pruning range is to privatize them, by marking them with `PrivateAssets='all'` and `IncludeAssets='none'`.
 
 ## Motivation
 
@@ -154,8 +155,7 @@ Only 5% of .NET SDK based projects are suspected to be multi-targeted, and it wo
 
 #### Direct PackageReference in the pruning range are treated with `PrivateAssets=all` and `IncludeAssets=none`
 
-If a direct PackageReference is within the pruning range, we can treat with `PrivateAssets=all` and `IncludeAssets=none`.
-
+If a direct PackageReference is within the pruning range, we will treat them `PrivateAssets=all` and `IncludeAssets=none`.
 The direct packages are still resolved. No need to introduce special concepts for direct packages that are not resolved.
 
 `PrivateAssets=all` means that they will not be:
@@ -163,11 +163,14 @@ The direct packages are still resolved. No need to introduce special concepts fo
 - Transitive project through project - this is pretty inconsequential to the larger story, since as a transitive package, it would be pruned in the referencing project as well (assuming pruning wasn't explicitly disabled).
 - Be part of the resulting package - This is beneficial in scenarios like 2a, 2b and 2c where we can't raise an NU1510 warning due to the risks of conditional PackageReference. We would effectively be removing this package for the user, without them having to apply it themselves.
 Additionally, in single framework scenarios, prunable packages would be removed from the package definition completely, which is effectively NuGet applying the `NU1510` heuristic for the customer. Given that the `NU1510` warning is enabled for `.NET 10` only, this would clean-up packages targeting net9.0 and earlier of unnecessary dependencies.
+- Overall, `PrivateAssets=all` means for direct PackageReference prunable packages private, for the project only.
+It completely *stops* the propagation of unnecessary dependencies within packages (what the original NU1510 was aiming at, but without the gotchas that come with the customer manually trying to get the references correct).
 
 `IncludeAssets=none` will be used to clearly indicate that the package would not be used at compile or runtime.
 
-The one very clear con of this proposal is that this is a breaking change for the pack operation.
-You upgrade the .NET SDK and now your package has different dependencies without a clear indication.
+This has the following side-effects:
+
+- Breaking change for the pack operation. You upgrade the .NET SDK and now your package has different dependencies without a clear indication.
 The whole PrunePackageReference feature is a breaking change itself, but in the case of transitive package pruning, the benefits are no audit warnings, since the package is not used at all, so very low chances of things going wrong.
 In the directs case, we'd be changing the resulting package.
 
@@ -185,10 +188,14 @@ A practical example would be, 2a's nuspec is:
 
 Note that we could consider gating the `PrivateAssets=all` addition to net10.0 being the used framework, but the impact would significantly smaller. If we gate this behavior on net10.0 being used, we could consider adding a separate property as well.
 
+- Aliases for PackageReference that will be pruned will stop working. This is *unlikely* to be a problem, since we normally wouldn't expect customers to alias platform packages.
+
+This changes will automatically happen in *all* pruning scenarios, if pruning is enabled, you would get the behavior for both transitives & directs.
+
 ## Drawbacks
 
-- Additional UX concepts for both 1st and 3rd party tooling that manage packages.
-- Cost of direct PackageReference pruning vs the potential benefit since customers in affected scenarios with a vulnerable package would either get a NU1510 or get a legit audit warning.
+- Breaking change for the pack operation (albeit, a warranted one)
+- Breaking change for aliases on direct PackageReference from the platform.
 
 ## Rationale and alternatives
 
@@ -233,30 +240,31 @@ For the components mentioned above, we need to define their behavior for pruned 
 
 Given that pruning is an offline operation, ie NuGet does not talk to the sources for pruned packages, we never validate that the requested package exists and the specified version may be incorrect. This is likely to not be of too much consequence, since it's only really possible with hand-editing, and the impact of it is minimal since the package is not used anyways.
 
+There are some unresolved questions with this approach:
+
+- What should `dotnet list package` show for direct pruned packages? Should it show the version specified as requested?
+  - It can be marked in a similar way that auto-referenced packages are marked. `> Newtonsoft.Json           (A)   [13.0.2, )   13.0.2`
+- What should the GetInstalledPackages return for the pruned direct packages?
+- Should we add a setting to prune direct package references?
+- Validate how 3rd party components would behave with unresolved direct package references.
+
 ## Prior Art
 
 N/A
 
 ## Unresolved Questions
 
-- How should we handle direct packages? `PrivateAssets=all`, prune completely or no action.
+- How should we handle direct packages? `PrivateAssets=all` or no action.
   - If the answer is `PrivateAssets=all`, should we gate it behind the net10.0 framework being used?
     - The impact is more minimal, but we're fully following the [Breaking Change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#tie-potentially-impactful-changes-to-the-target-tfm).
-  - If the answer is prune completely, the following unresolved questions apply:
-    - What should `dotnet list package` show for direct pruned packages? Should it show the version specified as requested?
-      - It can be marked in a similar way that auto-referenced packages are marked. `> Newtonsoft.Json           (A)   [13.0.2, )   13.0.2`
-    - What should the GetInstalledPackages return for the pruned direct packages?
-    - Should we add a setting to prune direct package references?
-    - Validate how 3rd party components would behave with unresolved direct package references.
+- Do we have to apply `IncludeAssets=none` or is making prunable packages private good enough?
+- Should we add a switch for opting out of the direct package pruning behavior?
 
 ## Future Possibilities
 
-- If we prune direct or we apply PrivateAssets=all and IncludeAssets=none to those packages:
-  - Visualize pruned packages in the PM UI Installed/Updates tabs. Note that the PM UI does not have visualization support for multi-targeting, so the heuristic here should be similar to the one for the NU1510 warning.
-- If we prune direct packages:
-  - Change the warning icon for a missing direct package reference to something that accounts for the pruned packages.
-- If we apply PrivateAssets=all and IncludeAssets=none to those packages:
-  - Add an icon indicating a direct package would be pruned.
+- Opt-out for the pruning behavior of direct PackageReference (pack time and restore time.)
+- Visualize prunable direct packages in the PM UI Installed/Updates tabs. Note that the PM UI does not have visualization support for multi-targeting, so the heuristic here should be similar to the one for the NU1510 warning.
+- Add an icon indicating a direct package would be pruned.
 - Block installation of packages that would be pruned. Effectively, do not allow System.Text.Json 8.0.0 to be installed in a project targeting `net10.0` only. The heuristic could be similar to the one of NU1510, block the installation if a NU1510 is raised for the package id being installed.
 - [#14126: Warning rollout for PrunePackageReference](https://github.com/NuGet/Home/issues/14126) - We may want a way to enable how this warning is surfaced for customers that do not target .NET 10.0.
 - NuGet vulnerability warnings for direct PackageReference can account for the fact that a package would be pruned and provide that information. We could also choose to raise a NU1510 warning when a vulnerability is reported for that package during NuGet Audit.
--- a/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
+++ b/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
@@ -207,8 +207,10 @@ N/A
 
 <!-- What future possibilities can you think of that this proposal would help with? -->
 
-- Visualize pruned packages in the PM UI Installed/Updates tabs. Note that the PM UI does not have visualization support for multi-targeting, so the heuristic here should be similar to the one for the NU1510 warning.
-- Change the warning icon for a missing direct package reference to something that accounts for the pruned packages.
+- If we prune direct packages:
+  - Visualize pruned packages in the PM UI Installed/Updates tabs. Note that the PM UI does not have visualization support for multi-targeting, so the heuristic here should be similar to the one for the NU1510 warning.
+  - Change the warning icon for a missing direct package reference to something that accounts for the pruned packages.
 - Block installation of packages that would be pruned. Effectively, do not allow System.Text.Json 8.0.0 to be installed in a project targeting `net10.0` only. The heuristic could be similar to the one of NU1510, block the installation if a NU1510 is raised for the package id being installed.
 - [#14126: Warning rollout for PrunePackageReference](https://github.com/NuGet/Home/issues/14126) - We may want a way to enable how this warning is surfaced for customers that do not target .NET 10.0.
 - NuGet vulnerability warnings for direct PackageReference can account for the fact that a package would be pruned and provide that information. We could also choose to raise a NU1510 warning when a vulnerability is reported for that package during NuGet Audit.
+- [[DCR]: ExcludeAssets="all" should exclude the package from the restore graph](https://github.com/NuGet/Home/issues/11567) covers a similar idea of removing packages from restore, like direct pruning would. Treating pruned packages with `IncludeAssets="none"` followed by excluding this package from all auditing may yield similar results.
--- a/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
+++ b/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
@@ -197,10 +197,11 @@ N/A
 
 ## Unresolved Questions
 
-- What should `dotnet list package` show for direct pruned packages? Should it show the version specified as requested?
-- What should the GetInstalledPackages return for the pruned direct packages?
-- Should we add a setting to prune direct package references?
-- Validate how 3rd party components would behave with unresolved direct package references.
+- Should we prune direct packages? If the answer is yes, the following unresolved questions apply:
+  - What should `dotnet list package` show for direct pruned packages? Should it show the version specified as requested?
+  - What should the GetInstalledPackages return for the pruned direct packages?
+  - Should we add a setting to prune direct package references?
+  - Validate how 3rd party components would behave with unresolved direct package references.
 
 ## Future Possibilities
 
--- a/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
+++ b/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
@@ -77,7 +77,7 @@ b. .NET & .NET Standard scenario
   </ItemGroup>
 ```
 
-For .NET, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues as talked about in the [Motivation](#motivation).
+For .NET 9, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues as talked about in the [Motivation](#motivation).
 
 c. Multiple .NET frameworks, package specified for all frameworks.
 
--- a/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
+++ b/accepted/2025/PrunePackageReference-with-direct-PackageReference.md
@@ -10,7 +10,7 @@
 Change the NU1510 heuristic in multi-targeting scenarios to only be raised when the package would be completely removed from the project.
 This would ensure customers don't end up a potentially more challenging to manage conditional PackageReference scenario.
 
-This proposal also explores the idea of pruning direct package references, the technical details around and the benefits of doing the work.
+This proposal explores various ideas for pruning direct package references, the technical details around and the benefits of doing the work.
 
 ## Motivation
 
@@ -44,11 +44,12 @@ As a long as a package is specified within a project but it's not needed, it add
 - It may be accidentally updated by tooling such as dependabot
 - It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
 - It is still part of auditing functionality such as NuGetAudit, deprecation etc.
+- It will be listed as a dependency of packages (thus needing them to be pruned.)
 
 Given that pruning is a per framework feature, there a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
 
 1. Single framework
-If the direct package reference is within the pruning range, we would like for the customer to remove the PackageReference and reduce the overhead.
+If the direct package reference is within the pruning range, we the customer to remove the PackageReference and reduce the management overhead.
 
 2. Multi framework scenario
 a. .NET & .NET Framework scenario
@@ -63,7 +64,7 @@ a. .NET & .NET Framework scenario
   </ItemGroup>
 ```
 
-For .NET, the package is unnecessary, and ideally we'd want the customer to consider conditionally referencing the System.Text.Json reference.
+For .NET, the package is unnecessary, and ideally we'd want the customer to consider conditionally reference the System.Text.Json package.
 Note that even in these situations, there's a theoretical way to end up with runtime issues, but it would require that a .NET Framework package references a package that references both, and then a .NET application reference this package.
 
 b. .NET & .NET Standard scenario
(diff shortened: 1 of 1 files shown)


nkolev92:
@jeffkl @zivkan @aortiz-msft @jebriede 

Can I review from you 4 whenever you have a chance?
There are some UX considerations here that I would appreciate your input on.

jeffkl (dismissed):

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+- Update the NU1510 heuristic only. Retain the current behavior for pruning.
+- Instead of removing the direct PackageReference from the graph completely, treat it as `IncludeAssets="none"`. This would allow get us benefits from Component Governance, in a more cost-effective way. `NuGetAudit` would also need to be updated to stop auditing unused packages, similarly to how Component Governance approaches the problem. [Make ExcludeAssets visible in Audit and nuget why](https://github.com/NuGet/Home/issues/13860)
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions
+
+- Should we prune direct packages? If the answer is yes, the following unresolved questions apply:
```
nkolev92: Thanks for taking a look @jeffkl!

What's your though on this?
Is it worth going through the effort of direct pruning?

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+This design builds on the previous pruning specs, [PrunePackageReference roll-out](../2025/prune-package-reference-rollout.md) and [PrunePackageReference](../2024/prune-package-reference.md).
+
+#### Scenarios for direct PackageReference items
+
+The motivation for the pruning overall is that we'd want to reduce the overhead of customers managing components that can be managed by the runtime.
+We want customers to remove unnecessary PackageReference items.
+As a long as a package is specified within a project but it's not needed, it adds maintenance overhead:
+
+- It may be accidentally updated by tooling such as dependabot
+- It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
```
zivkan, line 44:
```suggestion
- It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
- The package might become vulnerable in the future, causing NuGetAudit warnings that could be avoided by letting the SDK use the target's reference assemblies (update the runtime to fix the vulnerability)
```

Perhaps you were thinking about vulnerable packages in the "accidentally updated by tooling" example, but just in case it's not obvious to people reading this, I thought it would be useful to point out that removing package references reduces risk of vulnerable packages.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+
+Given that pruning is a per framework feature, there a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
```
zivkan: If you're giving background on pruning as a feature, is it worthwhile also explaining that targeting .NET 8, but referencing a 9.0.x version of a package like System.Text.Json will not prune that package?

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+
+#### How do we tell the customers we want them to remove a PackageReference
+
+We raise a NU1510 warning.
+
+For multi-targeted projects, the warning will be raised if at least one of the frameworks is .NET 10.
+We will raise an NU1510 when the PackageReference can be completely removed from the project file.
+In other words, when a package would be pruned for every target framework it is specified in, raise NU1510.
+
+In order to provide continuity, and minimize disruption that would be caused by raising NU1510 for existing projects, as per the [Breaking Change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#tie-potentially-impactful-changes-to-the-target-tfm) NU1510 will only be raised for projects targeting net10, requiring an explicit customer action to get this diagnostic.
```
zivkan, line 125: off topic for this spec, but what target frameworks does the SDK have pruning data for?

If I have a project targeting only .NET Framework 4.8, can I opt-in to pruning, and have it remove packages like NETStandard.Library 1.6, and all its dependencies?

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+  <PropertyGroup>
+    <TargetFramework>net9.0;netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="9.0.4" />
+  </ItemGroup>
+```
+
+For .NET, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues.
```
zivkan: Isn't the same true for .NET Framework, because of Asset Target Fallback?  That would make scenario 2a the same as 2b

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+
+- It may be accidentally updated by tooling such as dependabot
+- It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
+
+Given that pruning is a per framework feature, there a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
+
+1. Single framework
+If the direct package reference is within the pruning range, we would like for the customer to remove the PackageReference and reduce the overhead.
+
+1. Multi framework scenario
```
zivkan: Reading the raw markdown, rather than rendered, is confusing because this gets rendered as 2. I'd normally ignore it, but later in the file you have table where you refer to 2a, 2b, 2c, and I was having trouble understanding what it was referring to, since it just says `1.` here

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+In order to provide continuity, and minimize disruption that would be caused by raising NU1510 for existing projects, as per the [Breaking Change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#tie-potentially-impactful-changes-to-the-target-tfm) NU1510 will only be raised for projects targeting net10, requiring an explicit customer action to get this diagnostic.
+
+Referring back to the [scenarios](#scenarios-for-direct-packagereference-items) listed above:
+
+| Scenario | NU1510 | Notes |
+|----------|--------|-------|
+| 1 | Yes | |
+| 2a | No | The warning may be beneficial, but complex heuristic would be harder to reason about. |
+| 2b | No | Eventual runtime failure risks |
+| 2c | No | Eventual runtime failure risks |
```
zivkan, line 134: Neither 2b nor 2c are runtime failure risks as the projects are presented, because the package major version matches the target framework, right?

My understanding is that the risk is when a "lower" TFM references a package version that's higher than what the runtime provides.  In other words, ConsoleAppA references ClassLibB, which in turn references ClassLibC.  ConsoleAppA targets .net8, classLibB targets netstandard2.0, and ClassLibC targets both, and references System.Text.Json version 9.0.0, but only for netstandard2.0 (getting the assembly via the target framework's ref assemblies for net8.0).  ClassLibB will get System.Text.Json 9 transitively, and can therefore call APIs on it.  But ConsoneApp will not get the package, therefure loading the assembly from the runtime. Both ClassLibC and ConsoleApp will be fine, but there will be a runtime failure if ClassLibB tries to call a json API.

Maybe it's just me, but I don't feel like this failure scenario is clearly explained in this document.  Or maybe there are more failure modes than I'm aware of.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+
+```json
+"prunedPackages" : {
+        "net9.0": {
+          "System.Text.Json": "[9.0.0, )"
+        }
+}
+```
+
+Direct packages tend to be minimal and what users first update when they transitive vulnerabilties.
```
zivkan:
```suggestion
Direct packages tend to be minimal and what users first update when they have transitive vulnerabilties.
```

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
zivkan: As a customer, yes, I would want it to prune direct references.  That way, when I need to use a package, I can have a single unconditioned PackageReference (scenario 2b) and this is the most simple way for me to manage my package (to install in the first place, and to upgrade later).   Conditioned PackageReferences are more work, and isn't even possible through PM UI.

As a NuGet Client Team member, no, it sounds difficult 😆   You brought up a lot of good questions below. So, while I feel like the feature is "obviously" good, the low impact of projects that will benefit, plus these subtle issues that may cause other things to not work as expected has me doubtful.

In the end, I don't know which way to vote.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+- Instead of removing the direct PackageReference from the graph completely, treat it as `IncludeAssets="none"`. This would allow get us benefits from Component Governance, in a more cost-effective way. `NuGetAudit` would also need to be updated to stop auditing unused packages, similarly to how Component Governance approaches the problem. [Make ExcludeAssets visible in Audit and nuget why](https://github.com/NuGet/Home/issues/13860)
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions
+
+- Should we prune direct packages? If the answer is yes, the following unresolved questions apply:
+  - What should `dotnet list package` show for direct pruned packages? Should it show the version specified as requested?
```
zivkan: It could list a symbol, or something like `#1`, and under the table say that it means pruned (and an aka.ms link to learn what pruning is?)

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,216 @@
...
+  - What should the GetInstalledPackages return for the pruned direct packages?
```
zivkan: I'm aware of two usages of VS's GetInstalledPackages API from other built-in VS components, and possibly one more usage.

1. Test Explorer gets all packages, to scan the package install directory for test adapters. It doesn't need to know about pruned packages
2. Roslyn has a code fix so that when calling APIs that can't be found in the current project, but is available in another project, and that API comes from a package, it can propose installing the package in the current project.  Honestly, I think we're best off talking to the roslyn team to ask them for their opinion. But since pruned packages means the API will come from the reference assemblies, not the package reference, I guess that knowing about the pruned package reference is not helpful.
3. Special project types that ship build tools via nuget packages. They want to know what version of the build tools package the project is already using, so it can upgrade to a newer version if available.  These packages will never be pruned.

So, I'm leaning towards not returning pruned direct packages from GetInstalledPackages

zivkan (commented):
I stopped and restarted reviewing this document many times, so I've forgotten all the comments I've written.  Sorry if this ends up being a duplication of something I wrote earlier.

One piece of feedback I have is that I think I understood the runtime failure risk before reading this document, but trying to read the document, I don't feel like it's very clear. In particular, the 2c and 2d examples seem a bit artificial, because it seems so unlikely that customers would target a different version of a package per framework. I know it's something NuGet.Client does for MSBuild's packages, but NuGet is such a special case.  However, I really don't understand the dotnet/runtime repo requirements.

Also, the most obvious failure scenario is similar to scenario 2b, where there's a multi-targeting project, but with a single, unconditioned PackageReference.  But while 2b references a package that would be pruned in one of the two TFMs, there's no example where the package reference is for a higher version that will not be pruned.

Additionally, the current implementation, shipped in .NET 10 P4, encourages customers to go from a 2b scenario of an unconditioned PackageReference to a conditioned one just for netstandard2.0.  The problem is when that conditioned package reference is upgraded to a version higher than what net8.0 ships.  I don't remember this error scenario being called out explicitly, or if it is, perhaps there was not enough emphasis on it. 

To me, this document feels like an experts technical analysis, and so I guess my suggestion is to make it feel more "casual conversation" by starting out with "here's the most likely scenario", discussing it as necessary, and later have a "here are some technically possible scenarios" and don't go too in depth, because it blurs the focus on the common scenario.

As someone who writes far too verbosely, I hope you, or anyone else, gives me similar advice on how to make my next design spec easier to read 🤣

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: netcoreapp3.0 and newer
ns2.0 and newer.

We considered net framework as a potential future possibility, but assembly loading is different enough that there's some potential gotchas, so we didn't want to risk it early on.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: Fair, I usually prefer this tooling because it ensures we don't get out of order, but in this case it can be annoying, so I'll just put 2.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: This spec explicitly calls out that the reader should be familiar with the previous 2 specs, so I was thinking let's not clutter it, but let me know if you think it's important to add.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: This list is already referring to packages that would be pruned.

That being said, we *can* change the auditing behavior for direct packages that are part of the framework, as called out in future possibilities.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: You are correct, they are very very similar.

But with AssetTargetFallback you get a warning, so the switch in compat between net472 and .NET based frameworks is not quiet, so makes it less likely to happen and not as critical because customers won't switch frameworks through simple updates.

It's described here: https://github.com/NuGet/Home/issues/14196.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: The explanation right before the examples talks about it:

> For .NET, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues.

In general, the more narrow framework should not have newer APIs than the more broader framework (ie, NS2.0 surface area should always be *exactly* or smaller than that of net8.0). 

I can try to clarify it more.  
Your example is pretty good and pretty similar to the main failure scenario. 
You can always imagine ClassLibB is a secondary package. 
Meaning ConsoleApp never knows anything about ClassLibC

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: It'd have to be something similar to (A) that we get for auto referenced.

I'll explicitly add a view of how that looks like today.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: I feel like you're taking a bit of a glass box approach and considering all the known scenarios to us. 

Say a component like NuGet.Solver is using GetInstalledPackagesAsync. 
In that case, it'd need to know. 

I also feel like there's always going to be risk if we show it one place, but not show it in another, makes for a complicated decision matrix.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: > That way, when I need to use a package, I can have a single unconditioned PackageReference (scenario 2b) and this is the most simple way for me to manage my package (to install in the first place, and to upgrade later). Conditioned PackageReferences are more work, and isn't even possible through PM UI.

I might be misunderstanding your explanation, I feel like this reasoning doesn't really touch on the benefits of direct pruning because of a few reasons:

* By not pruning, we're not asking customers to reference it conditionally. You reference the package, and you'd get it downloaded, and at runtime you'd end up using whatever the runtime has. 
* Even if we removed the package from 1 framework, audit will still mark it vulnerable for the other framework, so any reduction benefit is not there. 

By not pruning, we'd effectively say, NU1510 only when package is completely removable. 
Build time conflict resolution if it is.

We can always choose to make audit special case these packages anyways and stop raising warnings, be it by making IncludeAssets="none" or just special casing it altogether.

> In the end, I don't know which way to vote.

Why do you think my spec is so long? :D

nkolev92:
Thanks for the detailed review @zivkan! I really appreciate it. 

> One piece of feedback I have is that I think I understood the runtime failure risk before reading this document, but trying to read the document, I don't feel like it's very clear. 

Fair. I'll add a more explicit example of the runtime failure. 

> In particular, the 2c and 2d examples seem a bit artificial, because it seems so unlikely that customers would target a different version of a package per framework. I know it's something NuGet.Client does for MSBuild's packages, but NuGet is such a special case. However, I really don't understand the dotnet/runtime repo requirements.

2c and 2d were a literal customer example haha, https://github.com/NuGet/Home/issues/14251. Or rather 2c was, a 2d was a variation that I used to explain the approach that the package doesn't have to be pruned in every framework, but that it rather needs to not be a reference in every framework.

> Also, the most obvious failure scenario is similar to scenario 2b, where there's a multi-targeting project, but with a single, unconditioned PackageReference. But while 2b references a package that would be pruned in one of the two TFMs, there's no example where the package reference is for a higher version that will not be pruned.

I used words, but I'll use an example instead in a future commit, great points.

> Additionally, the current implementation, shipped in .NET 10 P4, encourages customers to go from a 2b scenario of an unconditioned PackageReference to a conditioned one just for netstandard2.0. The problem is when that conditioned package reference is upgraded to a version higher than what net8.0 ships. I don't remember this error scenario being called out explicitly, or if it is, perhaps there was not enough emphasis on it.

I feel like, this is what all the 2.* scenarios talk about. I think the example will help.

> To me, this document feels like an experts technical analysis, and so I guess my suggestion is to make it feel more "casual conversation" by starting out with "here's the most likely scenario", discussing it as necessary, and later have a "here are some technically possible scenarios" and don't go too in depth, because it blurs the focus on the common scenario.

I think it needs to be both.  I ended up with just explanation rather than technical vs functional because they're just so damn close. I'll push something later today that should hopefully read easier.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: Pushed a bunch of updates.
Ready for another look.

jeffkl (dismissed):

nkolev92:
cc @ericstj 
I can't add you as a reviewer for some reason. I guess it's tied to org membership.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
ericstj: Is there a compromise that doesn't prune the package when direct, but makes it PrivateAssets=all and ExcludeAssets=all?  Just trying to think outside the box after being away for a bit.

nkolev92: PrivateAssets=all would act like NU1510 does and could potentially lead to the same runtime issues.

I did add `ExcludeAssets=all` as a possible alternative: https://github.com/NuGet/Home/pull/14325/files#diff-6888bfdb5da9045e294e7b479ab046591fbd8472b45c072fd81f823b713e859aR201.

ericstj: What runtime issues?  These packages are trimmed by conflict resolution so should be absent from runtime assets completely.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Drawbacks
+
+- Additional UX concepts for both 1st and 3rd party tooling that manage packages.
+- Cost of direct PackageReference pruning vs the potential benefit since customers in affected scenarios with a vulnerable package would either get a NU1510 or get a legit audit warning.
+
+## Rationale and alternatives
+
+- Update the NU1510 heuristic only. Retain the current behavior for pruning.
+- Instead of removing the direct PackageReference from the graph completely, treat it as `IncludeAssets="none"`. This would allow get us benefits from Component Governance, in a more cost-effective way. `NuGetAudit` would also need to be updated to stop auditing unused packages, similarly to how Component Governance approaches the problem. [Make ExcludeAssets visible in Audit and nuget why](https://github.com/NuGet/Home/issues/13860)
```
ericstj: This doesn't work for packages with analyzers today, since the SDK raises those from the libraries section (due to no native support).  Would be nice to fix that.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: I think I combined 2 concepts into one. 

PrivateAssets is interesting just on its own as well. 
The *risk* with partial pruning is because of what the user defines, but it's not really a concern if we remove it. 
In a world where pruning is enabled everywhere, these packages would be transitively pruned anyways, so no changes there.
This way, even if the end user doesn't have pruning enabled, they'd get some of the benefits.

I'll work that into the spec.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,248 @@
...
+# PrunePackageReference handling of direct PackageReference items
+
+- Nikolche Kolev <https://github.com/nkolev92>
+
+- [#14196: Pruning should prune and not warn for a direct reference in a multi-targeting project](https://github.com/NuGet/Home/issues/14196)
+
+## Summary
+
+Change the NU1510 heuristic in multi-targeting scenarios to only be raised when the package would be completely removed from the project.
+This would ensure customers don't end up a potentially more challenging to manage conditional PackageReference scenario.
```
donnie-msft:
```suggestion
This would ensure customers don't end up in a potentially more challenging to manage conditional PackageReference scenario.
```

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,248 @@
...
+
+- It may be accidentally updated by tooling such as dependabot
+- It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
+- It is still part of auditing functionality such as NuGetAudit, deprecation etc.
+- It will be listed as a dependency of packages (thus needing them to be pruned.)
+
+Given that pruning is a per framework feature, there a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
+
+1. Single framework
+If the direct package reference is within the pruning range, we the customer to remove the PackageReference and reduce the management overhead.
```
donnie-msft:
```suggestion
If the direct package reference is within the pruning range, we ask the customer to remove the PackageReference and reduce the management overhead.
```

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,248 @@
...
+The motivation for the pruning overall is that we'd want to reduce the overhead of customers managing components that can be managed by the runtime.
+We want customers to remove unnecessary PackageReference items.
+As a long as a package is specified within a project but it's not needed, it adds maintenance overhead:
+
+- It may be accidentally updated by tooling such as dependabot
+- It may appear as a direct reference in the Package Manager UI or list package or others tools that list the project dependencies.
+- It is still part of auditing functionality such as NuGetAudit, deprecation etc.
+- It will be listed as a dependency of packages (thus needing them to be pruned.)
+
+Given that pruning is a per framework feature, there a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
```
donnie-msft:
```suggestion
Given that pruning is a per framework feature, there are a few scenarios to consider for the handling of direct package references, primarily revolving around multi-targeting:
```

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,248 @@
...
+  - What should the GetInstalledPackages return for the pruned direct packages?
+  - Should we add a setting to prune direct package references?
+  - Validate how 3rd party components would behave with unresolved direct package references.
+
+## Future Possibilities
+
+<!-- What future possibilities can you think of that this proposal would help with? -->
+
+- If we prune direct packages:
+  - Visualize pruned packages in the PM UI Installed/Updates tabs. Note that the PM UI does not have visualization support for multi-targeting, so the heuristic here should be similar to the one for the NU1510 warning.
```
donnie-msft: What would the visualization support look like? I can imagine the Installed Packages List could show the pruned packages in some way that indicates they're not applicable due to pruning (perhaps a different group for those, or just a different background and some indicator).
For Package Update, perhaps a prompt or acknowledgement pop-up when a pruned package is being promoted? 
Other ideas? We can discuss later once your improvements are shipped as well.

donnie-msft (dismissed):

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: > I can imagine the Installed Packages List could show the pruned packages in some way that indicates they're not applicable due to pruning

Yeah, probably something like that. 
I think the urgency of the future work is also different depending on whether we prune completely, orf if we apply IncludeAssets and PrivateAssets (more in the former one, less critical in the latter one).

nkolev92: > For Package Update, perhaps a prompt or acknowledgement pop-up when a pruned package is being promoted?

The update one is tricky, you could be updating a package out of the range, so we'd need to mindful there. 
I don't really want prompts tbh. 
I feel like I'd rather fail the operation completely since it wouldn't have an effect. 

>  We can discuss later once your improvements are shipped as well.

Yes for sure. This is all assuming any of the direct pruning changes are approved.

nkolev92:
The NU1510 heuristic is final. 

@ericstj, @ViktorHofer, @zivkan, @dsplaisted @jeffkl @baronfel

Looking for your thoughts on new proposal that includes `PrivateAssets=all` being automatic vs no action. 
Also curious what you think about the potential roll-out.

donnie-msft (dismissed):

dsplaisted (dismissed):

jeffkl (dismissed):

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: it seems like `ExcludeAssets=all` to a direct PackageReference with an Alias does break it (the .NET SDK doesn't have to forward). 

We have 2 options:
* Have the SDK somehow figure out to apply the aliases. 
* Don't do anything.
* Document it as another breaking change (I don't really expect this to be a common scenario, having an alias on a PR that's part of the .NET runtime).

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,264 @@
...
+  <PropertyGroup>
+    <TargetFramework>net9.0;netstandard2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.Text.Json" Version="9.0.4" />
+  </ItemGroup>
+```
+
+For .NET, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues as talked about in the [Motivation](#motivation).
```
ericstj: For `net9.0` the package is unnecessary.
```suggestion
For .NET 9, the package is unnecessary, but if the customer removes the reference and then at a later point point updates the PackageReference without bumping the target framework, it may lead to a situation where the netstandard2.0 version of their package has a broader API surface area than net9.0 and potentially leading to runtime issues as talked about in the [Motivation](#motivation).
```

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,264 @@
...
+
+### How do we tell the customers we want them to remove a PackageReference
+
+We raise a NU1510 warning.
+
+For multi-targeted projects, the warning will be raised if at least one of the frameworks is .NET 10.
+We will raise an NU1510 when the PackageReference can be completely removed from the project file.
+In other words, when a package would be pruned for every target framework it is specified in, raise NU1510.
+
+In order to provide continuity, and minimize disruption that would be caused by raising NU1510 for existing projects, as per the [Breaking Change guidelines](https://github.com/dotnet/sdk/blob/main/documentation/project-docs/breaking-change-guidelines.md#tie-potentially-impactful-changes-to-the-target-tfm) NU1510 will only be raised for projects targeting net10, requiring an explicit customer action to get this diagnostic.
```
ericstj, line 125: > will only be raised for projects targeting net10

For policy like this, you may want to avoid hardcoding the versions in NuGet but instead let it be defined/overridden in MSBuild.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,264 @@
...
+The primary motivation for pruning is that the audit functionality and external scanners no longer sees those package references and raise false positives.
+Transitively pruned packages do not appear in the assets file as dependencies of the packages that originally referenced them.
+They have effectively been selected as unnecessary during the framework selection step of dependency resolution.
+The only place the id appears in the assets file is the `project` section under the `packagesToPrune`.
+
+When the NU1510 warning is raised, we are giving the customer a clear signal that they need to remove the PackageReference.
+In these scenarios, what happens to the direct package references is not very critical.
+
+In the partial prune scenarios such as 2b and 2c, pruning directs like transitive could be beneficial in the .NET leg, helping avoid additional audit warnings.
+However, due to the fact that the PackageReference is in other frameworks, the benefit is more minimal, since if the package itself had a vulnerability, the audit warning will be warranted since the customer is using it in a case where it is not being pruned.
```
ericstj, line 148: This is a very good point that I don't remember us discussing before.  This convinces me that direct pruning has minimal benefit.  The package will already be downloaded for use in other frameworks.  All of its dependencies will end up being pruned if itself would have been pruned.

Conflict resolution should still apply in this case.  @dsplaisted we may want to test that and ensure conflict resolution will follow the same rules as pruning, or at least document the behavior.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,264 @@
...
+
+The direct packages are still resolved. No need to introduce special concepts for direct packages that are not resolved.
+
+`PrivateAssets=all` means that they will not be:
+
+- Transitive project through project - this is pretty inconsequential to the larger story, since as a transitive package, it would be pruned in the referencing project as well (assuming pruning wasn't explicitly disabled).
+- Be part of the resulting package - This is beneficial in scenarios like 2a, 2b and 2c where we can't raise an NU1510 warning due to the risks of conditional PackageReference. We would effectively be removing this package for the user, without them having to apply it themselves.
+Additionally, in single framework scenarios, prunable packages would be removed from the package definition completely, which is effectively NuGet applying the `NU1510` heuristic for the customer. Given that the `NU1510` warning is enabled for `.NET 10` only, this would clean-up packages targeting net9.0 and earlier of unnecessary dependencies.
+
+`IncludeAssets=none` will be used to clearly indicate that the package would not be used at compile or runtime.
```
ericstj, line 169: One annoying thing about this is that if the package has Analyzers the SDK will still raise those :-(.  I wish we could clean that up and make nuget responsible for raising analyzer items and delete the hack in the SDK.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
```diff
@@ -0,0 +1,264 @@
...
+
+Given that pruning is an offline operation, ie NuGet does not talk to the sources for pruned packages, we never validate that the requested package exists and the specified version may be incorrect. This is likely to not be of too much consequence, since it's only really possible with hand-editing, and the impact of it is minimal since the package is not used anyways.
+
+## Prior Art
+
+N/A
+
+## Unresolved Questions
+
+- How should we handle direct packages? `PrivateAssets=all`, prune completely or no action.
```
ericstj: I've come around on not pruning directs.  I think treating as private is "better" with some sort of escape hatch.  I don't think it should be a hardcoded TFM version, but instead some property that folks could toggle if they wish.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
ericstj: I'm not too worried about aliases, at runtime folks will be forced to use the latest copy even if they tried to alias an older trimmed version.  They could always workaround with PackageDownload + Reference.

What's your thinking in general about someone opting out of behavior on a direct reference?  Should we use some existing syntax as opt-out (IncludeAssets/Aliases/etc) or new syntax (Preserve=true) or none?  

Previously we said that a direct reference would be the way to opt out of pruning.  Now if we're having pruning apply in some way to direct references we may want an opt-out gesture.

ericstj (approved):
I'm good with this general direction - some comments aimed at clarification and ongoing discussion - but those shouldn't block progress.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: I was leaning towards none. 
A bit of "this is how pruning works", and then listen for feedback. 
For now I'll add it to future possibilities.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: The challenge is that this setting is a combination of the values of inner builds. 
I am assuming the question is more about people enabling it for more scenarios, correct?
I have a follow-up in https://github.com/NuGet/Home/issues/14126 that covers that.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: https://github.com/NuGet/Home/issues/6279 is scheduled for July/August. @martinrrm and @kfikadu are gonna work on it.

We won't roll it out by default in .NET 10, but we'll soon cover that gap.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: I was originally not leaning towards allowing a switch for how directs are handled, kind of a this is how pruning works thing. 

I agree with regards to the TFM. I'll need to send details about this to the breaking change aliases.

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
nkolev92: > This is a very good point that I don't remember us discussing before

We did not discuss it yeah. It was an idea that I was able to clearly put together after I rewatch the recording of the meeting we had for direct pruning.

> Conflict resolution should still apply in this case. @dsplaisted we may want to test that and ensure conflict resolution will follow the same rules as pruning, or at least document the behavior.

In the current proposal I'm applying `IncludeAssets=None` in https://github.com/NuGet/NuGet.Client/pull/6545. 
The benefit of that is that there's no conflict resolution to be done. The con is the aliasing changes. 

If we want to allow conflict resolution to handle it, we *could* just do PrivateAssets=all, but that'll make things like analyzers apply long term.

zivkan (approved):

accepted/2025/PrunePackageReference-with-direct-PackageReference.md
ericstj: Thank you 🙇  That's great news.

PR #14344, opened 2025-06-02
Add support for packaging MCP servers with NuGet

joelverhagen:
Describes https://github.com/NuGet/NuGetGallery/issues/10461.

A lot of the work here is already done (MCP SDK for C#/.NET, designs for single-shot execution, existing MCP registry efforts). This document is meant to describe the big picture of what's missing and zoom in on the NuGet-specific pieces.

[Rendered Proposal](https://github.com/NuGet/Home/blob/dev-joelverhagen-mcp/accepted/2025/nuget-mcp.md)


accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### New package type
+
+To address the filtering and discoverability problem, we will introduce an additional NuGet package type of `McpServer`.
+
+Currently, .NET tools get the `DotnetTool` package type, defined in the resulting package manifest (.nuspec file) generated by `dotnet pack`. [As of .NET 10](https://github.com/dotnet/sdk/pull/48039), it is possible to add additional package types to .NET tool packages using the `<PackageType>McpServer</PackageType>` MSBuild property in the tool .csproj.
+
+The resulting MCP server .nupkg will have two package types: `DotnetTool` and `McpServer`. For more information on package types, [see NuGet documentation](https://learn.microsoft.com/en-us/nuget/reference/nuspec#packagetypes).
+
+#### Startup instructions
```
JonDouglas, line 62: We should emphasize the `dotnet new mcpserver` or similar template that will have this already there.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+      "args": ["Contoso.MySuperMcp", "--", "mcp", "start"],
+      "env": {
+        "CONTOSO_API_KEY": "${input:contoso-api-key}"
+      }
+    }
+  }
+}
+```
+
+NuGet.org will scrape the README.md for a JSON code block with a `servers` JSON property containing a property matching the current package ID. If found, it will place the JSON in the command palette for easy copying. If not found, a default MCP JSON will be generated:
```
JonDouglas: love this. i am worried about only `servers` though. there's many variations of this that NuGet.org will need to account for. Perhaps a string containing `servers` will address the `mcpServers` and other weird syntax people are using.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+      "command": "dnx",
+      "args": ["Contoso.MySuperMcp", "--version", "1.0.1"]
+    }
+  }
+}
+```
+
+In a future phase, we can introduce a more machine-readable format for the MCP server's inputs, args, and environment variables. This would require alignment across other programming ecosystems (Python or npm MCP servers have the same need) and IDEs (VS Code and other MCP-enabled tools).
+
+#### Project template
```
JonDouglas, line 126: Should also setup the `README.md` with the section we will scrape by default.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+
+The [official MCP registry](https://github.com/modelcontextprotocol/registry/discussions/11) is not yet live, but the discussion and protocol description around it clearly list [several supported underlying registries](https://github.com/modelcontextprotocol/registry/blob/a4cefcf05f81466ad65e7c3971e76d0f6d60783e/docs/openapi.yaml#L173-L176) for hosting the local server's code.
+
+Current list: `npm`, `docker`, `pypi`, `homebrew`
+
+#### Add NuGet to the list
+
+We will work with the MCP server registry team to add `nuget` to the list.
+
+We will publish/link to guidance on how to publish a NuGet-based MCP server to the MCP registry. This step is performed after the package is published to NuGet.org.
```
JonDouglas, line 224: love this!

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+   - As of June 2025, there are 28 .NET tools on NuGet.org that use the MCP SDK. Newer versions should have the `McpServer` package type.
+   - Some Microsoft MCP servers are implemented in .NET but distributed via npm, for example [@azure/mcp](https://www.npmjs.com/package/@azure/mcp).
+3. Wait for `dnx` to land in a .NET 10 preview.
+4. Work with the MCP registry and VS Code to get NuGet support, perhaps providing PRs/OSS contributions ourselves.
+5. Enable MCP server UI on NuGet.org.
+6. Publish a document on learn.microsoft.com on how to create your own NuGet MCP server. 
+
+## Future Possibilities
+
+We will wait to publish an MCP server template until the .NET MCP SDK has announced a stable API surface area. It is currently in prerelease.
```
JonDouglas:
```suggestion
We will wait to publish an MCP server template until the .NET MCP SDK has announced a stable API surface area. It is currently in prerelease. In addition we will continue to keep eyes on the development of the [MCP specification](https://modelcontextprotocol.io/development/updates) and [MCP Registry](https://github.com/modelcontextprotocol/registry).
```

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,234 @@
...
+We should replicate what is already working for npm, PyPI, and Docker.
+
+## Unresolved Questions
+
+- What guidance should be provided for private MCP server implementations?
+  - For example, if you publish an `McpServer` to your Azure DevOps feed, what MCP registry should be used?
+- What schema should be used for a machine-readable MCP server startup instruction?
+  - The package author could include an `mcp.json` in the NuGet package [matching the MCP registry OpenAPI spec](https://github.com/modelcontextprotocol/registry/blob/a4cefcf05f81466ad65e7c3971e76d0f6d60783e/docs/openapi.yaml#L183-L201).
+- How are client runtime requirements expressed?
+  - For example, if an MCP server needs a certain .NET version, how is this communicated to the end user before failure occurs?
```
JonDouglas, line 276: Super interesting question here. Didn't think about this as I was under the impression this would be a .NET SDK version/compatibility issue.

JonDouglas (dismissed):
I couldn't have wrote this better to be honest. This is gold!

accepted/2025/nuget-mcp.md
joelverhagen: I think we'll need to keep an eye on how this is solved in other ecosystems. I will do some research and ask in the MCP/registry repo if needed.

accepted/2025/nuget-mcp.md
joelverhagen: Fixed.

accepted/2025/nuget-mcp.md
joelverhagen: Fixed.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,261 @@
...
+It is not possible to download and run a .NET tool in a single command today. This is in contrast to npm, Python, and Docker. For example, an MCP server on npm can be started by VS Code using `npx`. 
+
+#### New experience: `dotnet tool exec` / `dnx`
+
+The .NET team is working on a single-shot experience similar to `npx`. This is not work done by the NuGet team, so we'll just link to the existing efforts.
+
+- GitHub issues: [dotnet/sdk#31103](https://github.com/dotnet/sdk/issues/31103), [dotnet/sdk#47517](https://github.com/dotnet/sdk/issues/47517)
+- Design document: [dotnet/designs#334 - Add a design proposal for dotnet tool exec and dnx](https://github.com/dotnet/designs/pull/334) 
+- Design document: [dotnet/designs#333 - Add proposal for RID-specific .NET Tool packages](https://github.com/dotnet/designs/pull/333)
+
```
jeffkl, line 207: Proposed pull request: https://github.com/dotnet/sdk/pull/48443

jeffkl (dismissed):
Great work, looking forward to seeing these changes implemented!

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,261 @@
...
+      "command": "dnx",
+      "args": ["Contoso.MySuperMcp", "--version", "1.0.1"]
+    }
+  }
+}
+```
+
+#### Project template
+
+To improve the project setup experience, we will introduce a new project template. The template will be available via `dotnet new mcp-server`. This will create a .csproj for a CLI tool, with a stable MCP SDK dependency version, and an `McpServer` server package type.
```
baronfel, line 128: Where is the distribution for this template? Will it be shipped in the SDK? Or will it be a 'normal' NuGet-delivered template? I'd suggest that this could live entirely in the https://github.com/modelcontextprotocol/csharp-sdk repo and be published as part of releases there.

baronfel: I _suspect_ that it would also be nice if we had [one-shot template install + invocation](https://github.com/dotnet/templating/issues/3418) like we'll soon have for tool execution for getting-started and always-latest purposes as well...

joelverhagen: > Where is the distribution for this template?

I am not sure. I think we should ship it out of band early on, the moment the blockers are out of the way (single shot, stable MCP SDK API shape) and then be part of another release vehicle later (either SDK template or template from MCP SDK repo). Another TBD is whether we want this to be a default template or one installed manually.

accepted/2025/nuget-mcp.md
joelverhagen: Added a link. Thanks!

accepted/2025/nuget-mcp.md
baronfel: We have this problem with tools already - there are a few ways to tackle it

Available today:
* the _tool author_ can set rollforward on their tool at build time to allow it to run on newer runtimes than the one being targeted
* _tool consumers_ can use the roll forward flags on the tool run commands to explicitly tell a tool to run on a runtime that it doesn't explicitly support

Available soon:
* when we support RID-specific tools, the RID-specific tools can compile self-contained. This means they won't need a separate runtime. This solves the problem at the cost of needing to bump package versions to get runtime updates (which one-shot execution _also_ solves).
* when a RID-agnostic MCP server is used, the runtime already has knobs for controlling 'roll forward'. We expose those knobs on `tool run` and related commands (see above), and would likely need to do the same on one-shot invocation as well. This would allow _tool consumers_ to opt into running a tool on newer runtimes than the one it was built for

Available later:
* the CLI bootstrapper could auto- or interactively-acquire runtimes that tools require to run

baronfel (dismissed):
Left a few notes to consider, but overall I love the end to end.

accepted/2025/nuget-mcp.md
joelverhagen: Added to an open question in the doc. I think we can decide this after the first cut since we're waiting on a stable MCP SDK API (say that ten times fast) surface area.

baronfel: FWIW I think a .NET SDK-shipped package is going to be farrrr to slow for this space. My vote would be the MCP SDK once it settles as you noted.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+
+Currently, .NET tools get the `DotnetTool` package type, defined in the resulting package manifest (.nuspec file) generated by `dotnet pack`. [As of .NET 10](https://github.com/dotnet/sdk/pull/48039), it is possible to add additional package types to .NET tool packages using the `<PackageType>McpServer</PackageType>` MSBuild property in the tool .csproj.
+
+The resulting MCP server .nupkg will have two package types: `DotnetTool` and `McpServer`. For more information on package types, [see NuGet documentation](https://learn.microsoft.com/en-us/nuget/reference/nuspec#packagetypes).
+
+#### Startup instructions
+
+We have two options to encode the startup instructions into the package:
+
+1. Instruct the package author to include the desired consumer MCP JSON in the readme, and allow NuGet.org to scrape the JSON from the README markdown (a code block matching a certain pattern).
```
AArnott: Why would nuget.org need to scrape the JSON? It already presents the README, which the user sees and can copy from and paste into their mcp.json file.

joelverhagen: This would make the command palette on nuget.org MCP aware so it's clearer how to "install" the thing. The command palette for .NET tools would just be to install it, not run it (let alone run it with needed args/env).

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+2. Embed standardized, machine-readable information in the package to allow tooling to know of the startup instructions required for the MCP server
```
jeffkl: I prefer an explicit mcp.json in the package in case my README contains several examples of configuration where I show how to use certain environment variables etc.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+
+#### New package type
+
+To address the filtering and discoverability problem, we will introduce an additional NuGet package type of `McpServer`.
+
+Currently, .NET tools get the `DotnetTool` package type, defined in the resulting package manifest (.nuspec file) generated by `dotnet pack`. [As of .NET 10](https://github.com/dotnet/sdk/pull/48039), it is possible to add additional package types to .NET tool packages using the `<PackageType>McpServer</PackageType>` MSBuild property in the tool .csproj.
+
+The resulting MCP server .nupkg will have two package types: `DotnetTool` and `McpServer`. For more information on package types, [see NuGet documentation](https://learn.microsoft.com/en-us/nuget/reference/nuspec#packagetypes).
+
+#### Startup instructions
```
jeffkl, line 66: We could also consider metadata in the `.nuspec` like how you specify git repository information.  We could then render an `mcp.json` on nuget.org based on that metadata on behalf of the user.  We'd technically already know the package ID and if its an MCP server.  Maybe that would be best since perhaps the "one shot" command could change over time and we could update how its rendered on nuget.org rather than it being statically declared in each individually pushed package?

accepted/2025/nuget-mcp.md
joelverhagen: Yes, I agree. The risk with mcp.json (now renamed to [server.json apparently](https://github.com/modelcontextprotocol/registry/discussions/118#discussioncomment-13357690)) is that we ware buying into a specific schema as a contract. To me it feels like more of a commitment. If we scrape the README, it's a UI affordance only and not a contract with real package content consequences.

I would like more input on this area. I appreciate your comment!

accepted/2025/nuget-mcp.md
joelverhagen: We can compute a basic consumption JSON just from the ID and version. And then single-shot command changing over time can be done there. But this misses the args/env settings you can set in the mcp.json (now server.json as mentioned above). For example, @ErikEJ's MCP .NET tool needs an `-mcp` argument to launch:
https://github.com/ErikEJ/SqlServer.Rules/blob/95f72d575734f59ce2be3e1117e8a8d970d472fd/tools/SqlAnalyzerCli/readme.md#github-copilot-mcp-server-preview

_Something_ in the package needs to define that. 

I prefer the mcp/server.json approach over .nuspec because it doesn't invent a new scheme and instead inherits the schema used by the MCP registry project which the author will likely need to construct anyway to publish to the MCP registry. 

Publish tool for MCP registry: https://github.com/modelcontextprotocol/registry/tree/main/tools/publisher (seems prototypical, the space is evolving)

accepted/2025/nuget-mcp.md
joelverhagen: We spoke offline a bit about this but wanted to share in this forum:
Long term VS Code/VS IDE/other MCP clients may have UI to connect to the MCP registry, making the MCP consumption JSON less useful. But I think the consumption JSON is still a good thing to have on NuGet.org because there is a ecosystem of tools that can use MCP and maybe advanced scenarios where people still edit their client mcp.json manually. I think the copy-paste of MCP consumption JSON is useful long term, just decreasing over time is MCP registry adoption goes up.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+
+The .csproj will have the following shape:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0</TargetFramework>
+    <RootNamespace>Contoso.MySuperMcp</RootNamespace>
```
AArnott: `RootNamespace` isn't customary for a minimal SDK style project these days, is it?

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+
+Once Visual Studio has a corresponding experience ([the current experience is manually editing the `mcp.json`](https://learn.microsoft.com/en-us/visualstudio/ide/mcp-servers?view=vs-2022)), we will ensure NuGet is supported similarly.
+
+## Rollout plan
+
+1. Enable `McpServer` support on NuGet.org search and the package details page, behind a feature flag.
+2. Work with Microsoft and community MCP server implementers to package as NuGet.
+   - As of June 2025, there are 28 .NET tools on NuGet.org that use the MCP SDK. Newer versions should have the `McpServer` package type.
+   - Some Microsoft MCP servers are implemented in .NET but distributed via npm, for example [@azure/mcp](https://www.npmjs.com/package/@azure/mcp).
+3. Wait for `dnx` to land in a .NET 10 preview.
```
AArnott, line 248: Given `dnx` will be a critical part to this, VS Code and VS should consider what SDK is installed and what `global.json` may be present in the open folder/solution before using it. VS Code itself doesn't guarantee any SDK version is present. VS can guarantee a given .NET SDK is present _if_ the .NET workload is installed, but that SDK may not be available within the open solution if global.json specifies a different (possibly older) SDK version.
If `dnx` isn't available, what should these clients do?

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+4. Work with the MCP registry and VS Code to get NuGet support, perhaps providing PRs/OSS contributions ourselves.
+5. Enable MCP server UI on NuGet.org.
+6. Publish a document on learn.microsoft.com on how to create your own NuGet MCP server. 
+
+## Future Possibilities
+
+We will wait to publish an MCP server template until the .NET MCP SDK has announced a stable API surface area. It is currently in prerelease. In addition we will continue to keep eyes on the development of the [MCP specification](https://modelcontextprotocol.io/development/updates) and [MCP Registry](https://github.com/modelcontextprotocol/registry).
```
AArnott: MCP may be turbulent for quite some time. But MCP support in clients isn't slowing down. Why wait to expose this for a stable .NET MCP SDK? Changes to the MCP SDK won't break servers, since they ship with all their dependencies of a version that they require.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+
+As MCP servers are run in more and more places, we can consider enhacing the MSBuild project file to enable an MCP server dependency. This could allow the MCP server to be available to the editor (instead of defined in client `mcp.json` configuration) or on a CI for build-time tasks. For example, an MCP server could be used inside an analyzer to produce or fix build warnings. This could work much like the existing [build integration that NuGet has to ship MSBuild props and targets](https://learn.microsoft.com/en-us/nuget/concepts/msbuild-props-and-targets). Thanks Jeff Kluge ([@jeffkl](https://github.com/jeffkl)) for the idea!
```
AArnott: typo: "enhacing"

AArnott: This is an interesting idea, but I wonder ... what process would be the 'client' of such an MCP server? Would the compiler/analyzer act as the MCP client itself? Or would the editor/IDE be expected to read these project files and add the server to the user's list? 
Perhaps the answer would depend on whether you expect _user_ prompts to use this server or only the compiler/analyzer. But if only the compiler/analyzer, how would it use the MCP server without a connection to an LLM? And if we want to expose this through the IDE as another MCP server for the user to use in their chat window, ... that adds another dimension to how we discover MCP servers that is worth discussing with the impacted teams.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,265 @@
...
+## Prior Art
+
+We should replicate what is already working for npm, PyPI, and Docker.
+
+## Unresolved Questions
+
+- What guidance should be provided for private MCP server implementations?
+  - For example, if you publish an `McpServer` to your Azure DevOps feed, what MCP registry should be used?
+- What schema should be used for a machine-readable MCP server startup instruction?
+  - The package author could include an `mcp.json` in the NuGet package [matching the MCP registry OpenAPI spec](https://github.com/modelcontextprotocol/registry/blob/a4cefcf05f81466ad65e7c3971e76d0f6d60783e/docs/openapi.yaml#L183-L201).
```
AArnott: An mcp.json file in the package makes a lot of sense to me. And it would allow an mcp.schema.json file to be used in editing it to make sure it's valid.

accepted/2025/nuget-mcp.md
joelverhagen: This would need agreement with the .NET MCP SDK team. If we fix to a specific version (not floating in the template) and release new template versions frequently, it seems like it could work.

accepted/2025/nuget-mcp.md
joelverhagen: The same problem exists for npm, Python, or docker based MCP servers. I think this is dependent on the client to check and smooth the experience (instead of just hard failing). Options are prompting to install a proper SDK version, or shipping one in box, I think.

accepted/2025/nuget-mcp.md
Redth: My preference would definitely be Option 2.  Given the schema is pretty ubiquitous, I don't see the commitment to it as being overly problematic.  If there's a new thing in the future, that could always be _additive_.

The advantage to having something more prescriptive is obviously easier detection and could enable more scenarios like auto-including these configs from referenced nuget packages in tooling to be contributed to the tool/ide itself.

accepted/2025/nuget-mcp.md
Redth: While in a way it feels helpful to try and construct the JSON on behalf of the package author, I think as you've illustrated there's just too many things to consider accounting for.  The format is really not that hard, and if the (eventual) template contains the mcp.json wired into the package build properly, it's pretty simple to _change_ as opposed to having to create from scratch.

accepted/2025/nuget-mcp.md
Redth: I do really think the template should ship either alongside the MCP SDK (ideally), or on its own, but not in the SDK.  The more we can decouple templates from the SDK, the better and more agile we become in the future.  While obviously out of scope here, and a bit of a tangent, this is a concept MAUI has struggled with too.  Would be great if there was a more dynamic way to manage template versions and notify the user of available updates to them within tools like VS instead of having to insert them into VS updates.

accepted/2025/nuget-mcp.md
Redth: Without considering the 'how'....

One scenario for this I can imagine is taking the MAUI MCP server I built with tools for querying developer environment state (installed xcode version, android sdk components, jdk, etc) and tools for installing/configuring those things, and shipping it as a dependency of the MAUI SDK itself (just the fact that the nuget package reference makes it into the project is the important part of course).

VS/VSWin could then advertise the MCP server availability for Copilot chat to consume, at which point the MCP Server's prompts, tools, resources, etc. would be available to a .NET MAUI developer to help diagnose and fix environment configuration issues within the chat (and any other agents).

This could help with having a sort of 'per project/solution' set of MCP servers referenced just as a happy side effect of which nuget packages you happen to reference.

joelverhagen: > Would the compiler/analyzer act as the MCP client itself?

In a CI context -- yes, it could work that way, or some other automated part of the build pipeline. This is a very forward looking "future opportunity". 

> Or would the editor/IDE be expected to read these project files and add the server to the user's list?

This is another option, but perhaps a bit more controversial. This would be an alternative to defining MCP servers in a client mcp.json and instead define them in an "MSBuild place". Maybe a .NET only shop could find use in this but given MCP servers can be non-.NET and a user might want to use a mix of MCP servers (not just .NET tools, they may not even know or care how they are implemented), shifting this into a project definition makes it so there are not two places to depend on an MCP server for editing, and makes the MCP dependency "project specific" whereas it's more of a user-wide setting, let alone solution-wide.

This idea is here more to help us think creatively about where an MCP server (packaged in NuGet or otherwise) might be used in the workflow. A lot of the conversation right now is around IDE + MCP integration, but in the future the LLM + MCP will likely run in a lot more places, including non-interactive ones.

accepted/2025/nuget-mcp.md
joelverhagen: There are two categories of MCP server schema in question here (subpoints of option 2, to be clear):

a. "client" or "consumer" mcp.json which is specific to the client tool. The format between VS Code and Claude are a bit different (`mcpServers` vs. `servers` JSON node for example)
b. "registry" mcp.json, now called server.json, which used for publishing to the MCP registry and exposed at browse time on MCP registry endpoints.

The problem with option 2 is that the industry has not settled on a single format for client mcp.json or registry mcp.json, at least from what I can tell.

If we were to pick a schema now, we could end up being wrong. I am leaning towards option 2b since I think we should "get behind" the MCP registry effort and therefore adhere to their schema. By embedding server.json in the package we get two birds with one stone: we use an existing schema and don't invent something new (e.g. nuspec elements) and we get the package author closer to publishing their NuGet package to the MCP registry which they probably want to do next in the workflow anyway,

joelverhagen: By choosing option 1, we are letting the author define the JSON shape they want to communicate to consumers more flexibly, and we're just kicking option 2 down the road a bit, until things are settled (a clear, ubiquitous standard).

accepted/2025/nuget-mcp.md
joelverhagen: I think you're right. We should be skeptical of generating JSON for the consumer, on behalf of the package author unless there is a clear hint inside the package. It's hard to say if most MCP servers will be at all useful with the default JSON we generate.

accepted/2025/nuget-mcp.md
joelverhagen: Good point. Will remove. This is just a sample of what the csproj COULD look like. The actual template definition can be implemented later.

accepted/2025/nuget-mcp.md
joelverhagen: Let's merge the conversation with https://github.com/NuGet/Home/pull/14344#discussion_r2125004566.

JonDouglas (approved):

accepted/2025/nuget-mcp.md
joelverhagen: I added a pro's/con's table for the option. I think we should do option 2b which is embedding the MCP Registry server.json file in the .nupkg and reading it on NuGet.org.

accepted/2025/nuget-mcp.md
skofman1: I imagine we won't make the consumption json mandatory for the publishers, so we won't be able to show it for all MCP packages. In cases where it's missing, is there a default we can show? If so, what would it look like?

AArnott: I think a default would cause more confusion than do good if it wasn't guaranteed to work. To guarantee it would work, you'd need to know the basic command line to use as well as any switches and env vars that are required (which will be common for MCP servers that access network resources).

accepted/2025/nuget-mcp.md
AArnott: I think there are more ways for dnx to fail than those others, because global.json can get in the way. But yes, let's certainly explore how we can predict failure and guide users to resolving it.

accepted/2025/nuget-mcp.md
AArnott: Oh, I missed that this was just about the project template. Sure, you can wait on that. I thought the spec was saying that you were waiting on this whole document until that milestone was reached.

accepted/2025/nuget-mcp.md
AArnott: >In a CI context -- yes, it could work that way

I don't think CI builds should ever call an LLM. LLMs are slow, expensive, and non-deterministic. None of those things belong in a CI. All of them together? I really think we should focus on _other_ scenarios.

> VS/VSWin could then advertise the MCP server availability for Copilot chat to consume, at which point the MCP Server's prompts, tools, resources, etc. would be available to a .NET MAUI developer

That's interesting. I'll start a conversation with my VS copilot team to get folks thinking about this. 

>  perhaps a bit more controversial

Definitely this. But maybe it'll fly. We'll see.

accepted/2025/nuget-mcp.md
joelverhagen: I proposed https://github.com/modelcontextprotocol/registry/issues/125. This would upfront some of the context needed to improve the acquisition experience. But your point about global.json is great. The MCP client will need to handle other error modes also. I don't think we need a perfect story here to say "this is how you should shape your MCP NuGet package" (the immediate goal of this proposal) and leave some of there error handled for client-specific implementations. Ideally VS Code and VS IDE can share error handling strategies.

accepted/2025/nuget-mcp.md
joelverhagen: > In cases where it's missing, is there a default we can show?

I think we would show a user message not a JSON configuration or something meant to be machine readable. Something like "This NuGet package contains an MCP server. Check the package README and project documentation for details on how to launch the MCP server." 

The command palette tabs for global and local tools would still be there.

We could show a warning to the author (not the end user) indicating that the `server.json` is missing, much like we do when a package is missing a README.md.

accepted/2025/nuget-mcp.md
```diff
@@ -0,0 +1,285 @@
...
+| 2a. Embed server.json | *"align with MCP registry schema"*<br>- Would adhere to a schema defined in the ecosystem<br>- Allows other .nupkg readers to see it easily<br>- Shares effort with MCP registry publishing<br>- Likely to be stable after MCP registry launch | - There are other machine-readable formats like VS Code mcp.json<br>- Diverges from npm/pypi, which do not have rich MCP support<br>- Needs mapping code to generate consumption JSON (mcp.json)
+| 2b. Embed mcp.json | *"align with VS Code config schema"*<br>- Would adhere to a schema defined in the ecosystem<br>- Allows other .nupkg readers to see it easily<br>- Makes VS Code consumption very clear | - There are other machine-readable formats like Claude format<br>- Picks a format that may change over time (VS Code config)
+
+**I propose we start with option 2a since it is client-agnostic and positions the package author for publishing to the MCP registry as a next step.**
```
nkolev92, line 84: Whatever option we choose, would aligning it to the version of package type help? 
I feel like with as fast moving as everything is, we're bound to need to update/change at some point.

joelverhagen: Yes, is was planning on having `McpServer` v1.0 once things are clearer, if we need it. We can start with `McpServer` (default version 0.0) for the initial cut. If this is the only version we need, great, If not, we can do a 0.1 or 1.0 based on what we see in the future.

For the initial version, the embedded file (server.json) will not be required. We can phase in requirements later when we know better, on the existing package type version or a new one.

joelverhagen: In short "wait and see".

nkolev92: Makes sense. 
I don't know if you want to add the package type call out in the spec at all or just leave it out. 
Feel free to make a call and resolve this thread.

Great job on the write-up!

Nigusu-Allehu (approved):

joelverhagen:
This is close enough to merge. The areas we will be focusing on for implementation are:

- Required: Track the .NET preview release that brings single-shot execution
- Required: Enhance NuGetGallery with MCP awareness
- Required: Write a tutorial, blog
- Add support for NuGet on MCP registry, like npm/pypi
- Make a project template
- Work with VS Code for integration like npm/pypi
- Work with MSFT and community authors to use the new NuGet package type

baronfel:
> Required: Track the .NET preview release that brings single-shot execution

We just merged the PR for this, it'll be in 10.0.100-preview.6! As will RID-specific/AOT tools.

PR #14383, opened 2025-06-24
Allow dotnet pack to pack .nuspec files without requiring a .csproj file

### LINKED ISSUE #4254: allow dotnet pack to pack nuspec file without needing a csproj file
Follow up to https://github.com/NuGet/Home/issues/3904#issuecomment-267671637

Scenario:
I want to pack a nuspec file in an x-plat build system.

Inputs: 
- nuspec
- files

NB: I am not packing a csproj, just files.

In <= 4.0.0-rc2, this was available as `dotnet nuget pack (nuspec)` file. This was removed in 4.0.0-rc3 https://github.com/NuGet/NuGet.Client/pull/1065 without a suitable replacement.

More data to consider:

- This is is a common scenario for the ASP.NET Core runtime team.
- Although possible to use nuget.exe, this requires running 'mono nuget.exe' on xplat, which is less than ideal, not to mention buggy.

cc @emgarten @rohit21agrawal 

Thank You 🎉

kfikadu:
## Summary

This PR adds a proposal document for supporting `.nuspec`-only packaging with `dotnet pack`. The proposal outlines the motivation, technical options, and prior art for enabling developers to pack NuGet packages using only a `.nuspec` file, without requiring a `.csproj`.

[Rendered MD File](https://github.com/NuGet/Home/pull/14383/files?short_path=eddfbcd#diff-eddfbcda2c7d5bb1ef6cbfdfd6a0201ddaab05f21006cd8467d532ac33d8d7e3)  
## Motivation

Currently, `dotnet pack` does not support packing from a `.nuspec` file alone, forcing developers to use `nuget.exe` (Windows-only) or create dummy project files. This change aims to improve cross-platform workflows and reduce friction for legacy and modular packaging scenarios.

## Technical Details

- Describes two main options for implementation:
  - Allowing `.nuspec` as a direct argument to `dotnet pack`
  - Introducing a new subcommand (`dotnet pack nuspec`)
- Includes screenshots comparing current `nuget.exe pack -h` and `dotnet pack -h` outputs.

## Related Issue

Fixes #4254

## Additional Context

See the included images and prior art section for more background.


kfikadu:
@dotnet-policy-service agree company="Microsoft"

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+when using the command `dotnet pack` in files or non-sdk-style projects.  
+
+## Explanation
+
+### Functional explanation
+
+Currently, arguments for `dotnet pack` are [\<PROJECT\>|\<SOLUTION\>]. This feature will enable customers to also use a .nuspec to pack files. 
+`dotnet pack` is available for projects that are using the SDK-style format. Adding the ability to use a .nuspec path instead  
+will enable customers to pack projects or files that aren't SDK-style. In these cases, `dotnet pack` with a nuspec will not run  
+`NuGet.Build.Tasks.Pack.targets` and instead use the current implementation of `nuget.exe` [pack commands](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs)
```
nkolev92: which is in turn running: PackCommandRunner. 

https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs is just a couple hundred lines gathering inputs and that's it.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+Currently, arguments for `dotnet pack` are [\<PROJECT\>|\<SOLUTION\>]. This feature will enable customers to also use a .nuspec to pack files. 
+`dotnet pack` is available for projects that are using the SDK-style format. Adding the ability to use a .nuspec path instead  
+will enable customers to pack projects or files that aren't SDK-style. In these cases, `dotnet pack` with a nuspec will not run  
+`NuGet.Build.Tasks.Pack.targets` and instead use the current implementation of `nuget.exe` [pack commands](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs)  
+
+### Technical explanation
+
+**Option 1: .nuspec file path as an arugment dotnet pack [\<PROJECT\>|\<SOLUTION\>|\<NUSPEC\>]**  
+
+Currently, dotnet pack options and nuget.exe pack options are very different, meaning that we will need to add new options
```
nkolev92: We should explicitly enumerate the new options we would need.

For what it's worth, for the `dotnet pack` command, the `--property` option is already available, so the parsing for that is already done, so I don't imagine we'd need to be adding new property, maybe we'd just be interpreting them now while before they were just converted to msbuild properties by the SDK when the targets were invoked.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+## Explanation
+
+### Functional explanation
+
+Currently, arguments for `dotnet pack` are [\<PROJECT\>|\<SOLUTION\>]. This feature will enable customers to also use a .nuspec to pack files. 
+`dotnet pack` is available for projects that are using the SDK-style format. Adding the ability to use a .nuspec path instead  
+will enable customers to pack projects or files that aren't SDK-style. In these cases, `dotnet pack` with a nuspec will not run  
+`NuGet.Build.Tasks.Pack.targets` and instead use the current implementation of `nuget.exe` [pack commands](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs)  
+
+### Technical explanation
```
nkolev92, line 36: It's great that we have 2 options. We should call out which one we're leaning towards.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+
+**Option 1: .nuspec file path as an arugment dotnet pack [\<PROJECT\>|\<SOLUTION\>|\<NUSPEC\>]**  
+
+Currently, dotnet pack options and nuget.exe pack options are very different, meaning that we will need to add new options  
+that would not apply when packing an SDK project. There are a couple that are shared but the main difference is that when   
+using nuget.exe you can use the option -Properties to do token replacement in the .nuspec (ex. $variable$).  
+All new options and existing ones will need to have a check to see if they are applicable to the user scenario.  
+
+This implementation doesn't require us to create a new command or subcommand, making the command look cleaner and
```
nkolev92: nit: We can add some examples of how we'd call the command.

Let's add one with the property option too.

nkolev92 (commented):
Great start for this write-up. 

I'm hopeful we actually don't have to add new UX but rather reuse existing one.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Summary
+
+
+`dotnet pack` currently does not support packing a .nuspec file without a corresponding .csproj file.  
+As a result, developers are forced to either use `nuget.exe` (which is Windows-only) or create  
+a dummy .csproj file solely for packaging purposes. This creates friction for cross-platform  
+workflows, disrupts legacy systems that rely on .nuspec-only packaging, and undermines modular packaging strategies.  
+This proposal aims to add support for .nuspec-only packaging in dotnet pack, eliminating the need for a .csproj file and  
+improving developer experience across platforms.
```
zivkan: One sentence per line please, and don't split sentences across multiple lines.  Please update the entire document, not just this paragraph.

reasons: If edits are made, adding or removing more than a small number of characters, then the lines are no longer going to be "nicely justified". In order to maintain lines being similar length, then all following lines in the paragraph have to also be edited. But then a small change addings four or five words to a sentence becomes a large diff, making it harder to see the real content change.

Plus markdown is for editing. Reading is intended to be done on the HTML rendered view. So, there's no benefit to optimizing the markdown for equal line lengths.

Finally, if a line was split because a sentence was too long making it hard to read on a very long line, then splitting it into multiple short sentences probably improves readability, which is an additional benefit to having shorter lines in the raw markdown.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+**Option 2: New subcommand dotnet pack nuspec [\<NUSPEC\>]**  
+
+This new subcommand will help us add all current nuget.exe pack options more easily and don't worry about existing pack options  
+
+
+**Current result of running `nuget.exe pack -h` and `dotnet pack -h`**  
+`nuget.exe pack -h`: ![output of running nuget.exe pack -h](../../meta/resources/dotnet-pack-nuspec/nuget%20pack%20-h%20ss.png)
+
+
+`dotnet pack -h`: ![output of `dotnet pack -h`](../../meta/resources/dotnet-pack-nuspec/dotnet%20pack%20-h%20ss.png)
```
zivkan: Please don't take screenshots of text. The second screenshot in particular is wide, but since GitHub limits the width of content, the image gets squashed and the text is too small to easily read. But more importantly, text in screenshots can't be searched with ctrl-f, the text doesn't get indexed by search engines, and it's impossible to copy-paste snippets.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+`dotnet pack -h`: ![output of `dotnet pack -h`](../../meta/resources/dotnet-pack-nuspec/dotnet%20pack%20-h%20ss.png) 
+
+
+
+## Prior Art
+
+Previously, in 4.0.0-rc2, you were able to pack a .nuspec file without a csproj file by using 
+`dotnet nuget pack (nuspec)` file. This was removed in 4.0.0-rc3.
```
zivkan: Can you provide a link to the PR?  I wasn't aware that it was ever in the product.  I'm curious how it was implemented, and what options/arguments it had, but also if any justification was provided about why it was removed.

accepted/2025/dotnet-pack-nuspec.md
martinrrm: @baronfel Do you have any feedback on the command design? I prefer the second option which is `dotnet pack nuspec`

baronfel: I _strongly_ prefer `dotnet pack`. I think a lot of the existing options that you would need from `nuget pack` are already present (see @nkolev92 's note on `-p` below). Even if you do go with `dotnet nuget pack nuspec` you _definitely_ should take the opportunity to audit the flags that are present on the command today and:
* ensure that the flags you expose match the style of the rest of the `dotnet` cli,
* consider if you need to support all of the existing flags 
  * e.g. `-MsBuildVersion`: for the `dotnet` cli it is probably more reasonable to say "you only get to use the MSBuild that's present in this dotnet installation"
  * should any of the flags that are related to _building the project_ not be exposed through the `dotnet` CLI command surface at all? If this command is _only_ be about packaging a nuspec file then building/publishing should be done through the 'normal' CLI commands, no?
* consider if any flags you currently expose are already implemented in the CLI in other ways/with other defaults:
  * `-interactive` - the dotnet CLI has a very smart `--interactive` option now that is user-activity aware
  * maybe some of the output-path-related options can use `-o/--output` instead to align to CLI conventions

and so on!

nkolev92: We should to the exercise of enumerating all the new options we'd need in either case. 

IMO, 1. is the better UX and we should do that, unless the cost of 1 is really big or completely breaking with the rest of the command.

accepted/2025/dotnet-pack-nuspec.md
baronfel: Definitely agree with auditing/enumerating the options and calling out if there is an existing dotnet CLI facility for that same/similar concept.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+This runs `PackCommandRunner` which manages the packaging process in the `dotnet pack` command. It collects inputs, processes them, and generates NuGet packages. 
+This component supports packing from .nuspec files, enhancing cross-platform workflows and eliminating the need for a .csproj file. 
+
+### Technical explanation
+
+#### Option 1: .nuspec file path as an argument dotnet pack [\<PROJECT\>|\<SOLUTION\>|\<NUSPEC\>]
+
+Currently, dotnet pack options and nuget.exe pack options are very different, meaning that we will need to add new options that would not apply when packing an SDK project.
+There are a couple that are shared but the main difference is that when using nuget.exe you can use the option -Properties to do token replacement in the .nuspec (ex. \$variable\$).
+All new options and existing ones will need to have a check to see if they are applicable to the user scenario.
```
zivkan: Reading this document from top to bottom makes it sound like effort comparing the options haven't been done yet, but when I got further down the document, there's a table with the info.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+In these cases, `dotnet pack` with a nuspec will not run `NuGet.Build.Tasks.Pack.targets` and instead use the current implementation of `nuget.exe` [pack commands](https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Clients/NuGet.CommandLine/Commands/PackCommand.cs).
+This runs `PackCommandRunner` which manages the packaging process in the `dotnet pack` command. It collects inputs, processes them, and generates NuGet packages. 
+This component supports packing from .nuspec files, enhancing cross-platform workflows and eliminating the need for a .csproj file. 
+
+### Technical explanation
+
+#### Option 1: .nuspec file path as an argument dotnet pack [\<PROJECT\>|\<SOLUTION\>|\<NUSPEC\>]
+
+Currently, dotnet pack options and nuget.exe pack options are very different, meaning that we will need to add new options that would not apply when packing an SDK project.
+There are a couple that are shared but the main difference is that when using nuget.exe you can use the option -Properties to do token replacement in the .nuspec (ex. \$variable\$).
```
zivkan: The table below comparing arguments contradicts this sentence, as `-p:SomeProperty=SomeValue` exists (for any MSBuild command)

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+| `--property` / `-p:`           | Set or override a property for token replacement in `.nuspec`.   | Reused; interpret for token replacement, not just MSBuild.       | `-Properties` in `nuget pack`, `--property` in `dotnet pack`   |
+| `--output`                     | Specifies the output directory for the generated package.         | Reused; already available.                                       | `-OutputDirectory` in `nuget pack`, `--output` in `dotnet pack`|
+| `--basepath`                   | Specifies the base path for resolving relative file paths.        | New; needed for file path resolution in `.nuspec`.               | `-BasePath` in `nuget pack`, not present in `dotnet pack`      |
+| `--version-suffix` / `--version` | Sets the version or version suffix for the package.             | Reused/Clarify; clarify how it applies to `.nuspec` versioning.  | `-Version` in `nuget pack`, `--version-suffix` in `dotnet pack`|
+| `--exclude`                    | Exclude specific files or patterns from the package.              | New; needed for file exclusion support.                          | `-Exclude` in `nuget pack`, not present in `dotnet pack`       |
```
zivkan: devil's advocate: How confident are we that `--exclude` is really needed?  

Unfortunately we don't have telemetry, so we have no way of knowing how often it's actually used currently.

But, if the nuspec has a `<files>` section, then it's the author's responsibility to make sure the list is correct. I don't know about pack to know if `-Exclude` even works when the nuspec explicitly lists its files.

If the nuspec doesn't have a file list, then pack will look at the base path (or current directory) and pack all the files it can find, unless excluded.  But can we say "this is an advanced scenario, convert it to a csproj, use `<IncludeOutputInPackage>false`, and use msbuild pack instead"?

Perhaps not having an exclude option is not customer friendly enough, but I'd like to make sure we check assumptions.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+| `--nologo`                     | Suppresses display of the startup banner and copyright.           | Reused; already available.                                       | `-NoLogo` in `nuget pack`, `--nologo` in `dotnet pack`         |
+| `--min-client-version`         | Sets the minimum NuGet client version required to install the package. | Consider; may need to add.                                 | `-MinClientVersion` in `nuget pack`, not in `dotnet pack`      |
```
zivkan: can't min client version be set in the nuspec itself?  I don't understand why this would need to be a CLI option

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+| `--no-package-analysis`        | Skips package analysis after building the package.                | Consider; may need to add.                                       | `-NoPackageAnalysis` in `nuget pack`, not in `dotnet pack`     |
+| `--no-default-excludes`        | Disables default file exclusion patterns.                         | Consider; may need to add.                                       | `-NoDefaultExcludes` in `nuget pack`, not in `dotnet pack`     |
+| `--serviceable`                | Marks the package as serviceable.                                 | Consider; may need to add or clarify.                            | `-Serviceable` in `nuget pack`, `--serviceable` in `dotnet pack`|
```
zivkan: I'm very confident servicable is settable in the nuspec. I see no need for it to be a CLI option

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,75 @@
...
+| `--interactive`                | Enables interactive prompts during packing.                      | Consider; may need to add.                                       | `-Interactive` in `nuget pack`, not in `dotnet pack`           |
```
zivkan: interactive is useful for nuget.exe pack, because when you pack a project, it will restore and build the project, and restoring might need authentication.  But packing a nuspec file has no need to make any HTTP requests, so I can't think of any reason why there would need to be an interactive option.

accepted/2025/dotnet-pack-nuspec.md
zivkan: Now that this spec has been updated, it's still not clear which option the spec is actually proposing.  That's fine while this spec is still under discussion, but I won't be approving until it's clear what the proposal to be implemented actually is.

note, the spec template has an alternatives section to share what else was considered, but not accepted: https://github.com/NuGet/Home/blob/a601e1360bcf3de42c75c1110d4387642053aafb/meta/template.md?plain=1#L30-L34

accepted/2025/dotnet-pack-nuspec.md
nkolev92: I feel like we should support token replacement and that's about it. 

We don't have to enumerate every single piece of metadata. 
We can call that out as a limitation.

accepted/2025/dotnet-pack-nuspec.md
kfikadu: This is a great point, seeing as it is a relatively niche case. It is useful when .nuspec omits <files> and relies on default file discovery, amongst other things. This mainly enhances ease-of-use for developers. I am going to be adding it to the future possibilities section and based on demand, it can be added on later.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Options Relevant to .nuspec-only Support in `dotnet pack`
+
+| Option / Flag        | Description                                                        | Notes / Status for .nuspec-only support                        | Comparison (`nuget pack -h` / `dotnet pack -h`)                |
+|----------------------|--------------------------------------------------------------------|---------------------------------------------------------------|---------------------------------------------------------------|
+| `--property` / `-p:` | Set or override a property for token replacement in `.nuspec`.     | Reused; interpret for token replacement, not just MSBuild.     | `-Properties` in `nuget pack`, `--property` in `dotnet pack`   |
+| `--output`           | Specifies the output directory for the generated package.           | Reused; already available.                                     | `-OutputDirectory` in `nuget pack`, `--output` in `dotnet pack`|
+| `--basepath`         | Specifies the base path for resolving relative file paths.          | New; needed for file path resolution in `.nuspec`.             | `-BasePath` in `nuget pack`, not present in `dotnet pack`      |
```
nkolev92: If this is not specified, do we use the nuspec path?
I feel like if that's the case, then we shouldn't add this option either. 
It's a bit of a crutch and something that I'd say can be easily migrated away from.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,123 @@
...
+
+> **Note:**  
+> This table focuses on options that are new, reused with modified behavior, or require special handling for .nuspec-only support.
+> Project-specific and unrelated options (like `--configuration`, `--no-build`, etc.) are omitted for clarity and relevance.
```
nkolev92: So if someone has a thing like $configuration$ in their nuspec, --configuration wouldn't work and you';d need to do `dotnet package --property configuration=release` instead of  `dotnet pack nuspec --configuration release`

@baronfel @zivkan do you think we should support the latter on too? 
I don't imagine it's *too* complicated.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Limitations
+
+- Only a subset of options are exposed as CLI flags for .nuspec-only packaging.
+- Most .nuspec metadata, including advanced fields like `<serviceable>`, must be set directly in the .nuspec file.
+- For dynamic or automated scenarios, use token replacement in the .nuspec and pass values via `--property`.
+- The `--version` flag does not override a hardcoded `<version>` in the `.nuspec` file. To dynamically set the version, define a token in the `.nuspec` file and use `--property Version=<value>` to replace it.
```
nkolev92: I think we should consider supporting this and define a behavior in case someone passes both.

nkolev92 (commented):
The spec is looking much much much better. 

I think you should be able to start the implementation. 
In my opinion there's only a few things to figure related to the options which is something that should be easy to maneuver around.

accepted/2025/dotnet-pack-nuspec.md
zivkan: yeah, good suggestion. With the SDK changing pack's default configuration from Debug to Release, I think it would be good to maintain the same default.  Although [the way the default is set](https://github.com/dotnet/sdk/blob/2ed26645f68c5280848488ef47bd719afb77a537/src/Cli/dotnet/ReleasePropertyProjectLocator.cs#L57) feels very weird to me. I was expecting a System.CommandLine.Option default value factory. Hopefully we can easily reuse this default calculation method, but I'm not motivated to read the implementation right now to understand it.

edit: we might not be able to reuse it because I see it's doing some `ProjectInstance` stuff, and this feature is all about a nuspec without a project.

nkolev92: yeah, that doesn't apply to this scenario. 

When using a replacement token, you need to provide the value in this implementation. 

Based on your comment and @baronfel's thumbs up on https://github.com/NuGet/Home/pull/14383#discussion_r2180757031. sounds like we're in agreement of support --configuration and --version as priority.

So if someone passes `dotnet pack stuff.nuspec --version 1.0.0 --property version=2.0.0` we'd respect --version.  (I'm assuming that's what NuGet.exe pack would do too, so worth double checking)

I don't think this is a thing that'll be done by a user on purpose, but it clarifies the way the implementation would be done.

baronfel: the behavior of `--version 1.0.0` and `--property version=2.0.0` is something that you should be able to control and define. In the SDK, we generally have a rule for MSBuild-forwarded properties that

* our nice-UX wrappers like `--version`, `--configuration`, `--runtime`, etc come 'first' in the command line. 
* then, any 'defaults' for the command specified. For example, `dotnet pack` forwards `/t:Pack`.
* Finally, any user-supplied `--property` settings or unknown arguments are applied - this allows the user to override defaults because MSBuild has a 'last-wins' approach to merging properties.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,122 @@
...
+
+Currently, dotnet pack options and nuget.exe pack options are very different, meaning that we will need to add new options that would not apply when packing an SDK project.
+Some options, such as `--property` (or `-p:`), are already supported in MSBuild, `dotnet pack`, and `nuget.exe pack` for setting build properties.
+For .nuspec-only support, `--property` will be reused and interpreted for token replacement in the `.nuspec` file, similar to how `nuget.exe pack -Properties` works.
+Other options, such as `--output`, are already standard in the dotnet CLI and will be reused for .nuspec-only support, with clarified behavior where necessary.
+All new options and existing ones will need to be checked to see if they are applicable to the user scenario.
+This implementation does not require a new command or subcommand, making the command look cleaner and ensuring customers won't need to learn a new command.
+This is the ideal implementation.
+
+#### Example: Packing a .nuspec File with `dotnet pack`
```
nkolev92, line 46: doc feedback: It'd argue this is functional documentation. 
The commands and how they work and all that is functional and tells the end user how to use the command.

accepted/2025/dotnet-pack-nuspec.md
```diff
@@ -0,0 +1,122 @@
...
+> This table focuses on options that are new, reused with modified behavior, or require special handling for .nuspec-only support.
+> Project-specific and unrelated options (like `--no-build`) are omitted for clarity and relevance.
+> Options that are standard in the dotnet CLI, such as `--configuration` and `--output`, will be reused without modification.
+
+## Limitations
+
+- Only a subset of options are exposed as CLI flags for .nuspec-only packaging.
+- Most .nuspec metadata, including advanced fields like `<serviceable>`, must be set directly in the .nuspec file.
+- For dynamic or automated scenarios, use token replacement in the .nuspec and pass values via `--property`.
+- `--no-package-analysis`: Skips package analysis after building the package. This is not currently supported in .nuspec-only scenarios. As a result, this flag will not be available for .nuspec-only support.
```
nkolev92, line 71: This is probably simple enough to add, but I'm not a fan of the mega switch to begin with, so I'm ok with it being listed as a future possibility.

nkolev92 (approved):

aortiz-msft (approved):

Issue #14046, opened 2025-01-16
Enable packing legacy PackageReference projects without the need for a package (ie merge NuGet.Build.Tasks.Pack into NuGet.Build.Tasks)


nkolev92:
The PackageReference functionality has 2 targets, restore and pack, which are deliver by 2 separate packages/targets files created by NuGet.Build.Tasks and NuGet.Build.Tasks.Pack and 

We have 2 types of PackageReference functionality, SDK projects and classic csproj projects.

For SDK projects, they're both available by default, although they're imported from 2 different places.
For legacy projects, the restore target is available , but the pack target is not. It requires you to add a PR to NuGet.Build.Tasks.Pack.

Another feature of NuGet.Build.Tasks.Pack is that it allows you to overwrite the default pack  targets, allowing for faster onboarding for pack fixes. 

Beyond the UX, the issue with pack and restore being separate is that pack is ilmerged, so we're effectively shipping duplicates of most assemblies. 
We should decide whether we want to preserve the bootstrapping ability of NuGet.Build.Tasks(.Pack) or if it's sufficient to just make it available in all relevant scenarios. 
Final consideration is that making the pack targets available for all legacy projects could cause a different behavior when people run the "pack" target, so we'd need to consider whether we condition the import for legacy projects on something.

nkolev92:
cc @zivkan @jeffkl 

If we do this, we don't need: https://github.com/NuGet/Home/issues/13079.

nkolev92:
Either way, we should probably stop publishing https://www.nuget.org/packages/NuGet.Build.Tasks#versions-body-tab, like https://github.com/NuGet/client.engineering/issues/2582 suggests.

nkolev92:
Design completed.

We might still need https://github.com/NuGet/Home/issues/13079 for the .NET Framework but that should be temporary.

nkolev92:
Probably best to do this work when https://github.com/NuGet/NuGet.Client/pull/6539 is completed, since it touches signing & loc.

aortiz-msft:
@zivkan - Can we close this issue and create new issues for work pending in other repos?

zivkan:
This issue tracks the end to end experience. I created subtasks for what's already been done and what's remaining.

zivkan:
Will be available from VS 2026 preview 2

PR #14401, opened 2025-07-03
Make the pack command available in classic csproj, reduce .NET SDK size by merging pack & restore functionality

nkolev92:
Design for https://github.com/NuGet/Home/issues/14046

[Rendered](https://github.com/NuGet/Home/blob/dev-nkolev92-consolidatePackaRestore/accepted/2025/consolidate-pack-and-restore.md)

cc @zivkan

### DIFF
--- a/accepted/2025/consolidate-pack-and-restore.md
+++ b/accepted/2025/consolidate-pack-and-restore.md
@@ -87,7 +87,6 @@ To summarize, we will need to do the following technical work:
 - When the insertion into the .NET SDK happens, make the imports and layout changes in the .NET SDK.
 - Make the Pack target project style aware, don't invoke it on non-PackageReference based projects.
 - Add an import for the NuGet.Build.Tasks.Pack.targets into MSBuild, but make it .NET SDK aware to avoid double loading the targets.
-  - When we have new msbuild and old SDK, we should ensure the pack targets are loaded from new msbuild by setting NuGetBuildTasksPackTargets to the MSBuild version and setting ImportNuGetBuildTasksPackTargetsFromSdk to false.
 
 ## Drawbacks
 
--- a/accepted/2025/consolidate-pack-and-restore.md
+++ b/accepted/2025/consolidate-pack-and-restore.md
@@ -33,11 +33,8 @@ The following table contains details about the behavior of how pack is run for e
 | msbuild | classic csproj | Fails with `The target "pack" does not exist in the project` |
 
 When using the dotnet commands, there are no proposed changes in behavior.
-When using the msbuild commands, there are 2 changes:
+When using the msbuild commands, there is 1 change:
 
-- When running msbuild /t:pack on an .NET SDK based project, the pack targets from MSBuild will be imported.
-This allows us to avoid shipping the .NET Framework version of pack in the .NET SDK and unifies the restore & pack functionality compared to what we currently have which is build & pack are aligned.
-I don't think there's an easy answer here.
 - When running classic csproj, we will need to make the pack target project style aware.
 What that means is that we will change the pack target to no-op when the project style is not PackageReference.
 This is a *change* compared to the previous behavior.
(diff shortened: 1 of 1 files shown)


accepted/2025/consolidate-pack-and-restore.md
```diff
@@ -0,0 +1,127 @@
...
+
+In the .NET SDK, we will keep the import order the same to minimize chances of breaking anyone.
+In the classic csproj projects, we will add an import in <https://github.com/dotnet/msbuild/blob/194c2e94475106f9727ede721bb057caf46dc0d3/src/Tasks/Microsoft.Common.CurrentVersion.targets#L7005-L7011>.
+The NuGet.targets are imported in the solution context as well, so this avoids any weirdness due to that.
+
+To summarize, we will need to do the following technical work:
+
+- Merge NuGet.Build.Tasks.Pack into NuGet.Build.Tasks.
+- When the insertion into the .NET SDK happens, make the imports and layout changes in the .NET SDK.
+- Make the Pack target project style aware, don't invoke it on non-PackageReference based projects.
```
zivkan, line 88: There's already a [IsPackable property](https://github.com/NuGet/NuGet.Client/blob/a0631a17ada04e95f6af93b00ea20a52318b1da3/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L35-L36), which [makes pack no-op](https://github.com/NuGet/NuGet.Client/blob/a0631a17ada04e95f6af93b00ea20a52318b1da3/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L211-L212).  So, changing this property's default value based on project style == package reference gives customers an easy opt-in for when they know what they're doing.

zivkan (commented):
LGTM.  I'm interested what our partner teams will think, especially about the risk of breaking changes.

accepted/2025/consolidate-pack-and-restore.md
nkolev92: ProjectStyle can only correctly be determined by running a target because of the add PackageReference opt-in, so this becomes more challenging.

nkolev92:
@dsplaisted @jeffkl @rainersigwald @baronfel 
Thoughts?

jeffkl (dismissed):

accepted/2025/consolidate-pack-and-restore.md
zivkan: On one hand, I feel like we could move IsPackable's default value to a target that depends on the get project style target. Having a default value (not null or empty) allows dotnet/project-system's project properties to populate a value. But it gets its information from a design time build, so if the property gets set as part of one of the DTB targets, it'll still be populated. I can't think of any other reason why moving the default value from evaluation to a target would be breaking.  Although with enough customers, there's likely to be at least one doing something unexpected and somehow have a dependency on this property value being set already.

On the other hand, pack via msbuild is designed explicitly for PackageReference, so I guess it's acceptable to block developers who are trying to hack some super custom MSBuild project to create a NuGet package from something that's not a .NET project.

zivkan (dismissed):

accepted/2025/consolidate-pack-and-restore.md
```diff
@@ -0,0 +1,127 @@
...
+| dotnet | classic PackageReference | N/A |
+| dotnet | classic csproj | N/A |
+| msbuild | .NET SDK | Uses .NET SDK pack, which contains .NET Framework assemblies |
+| msbuild | classic PackageReference | Fails with `The target "pack" does not exist in the project` |
+| msbuild | classic csproj | Fails with `The target "pack" does not exist in the project` |
+
+When using the dotnet commands, there are no proposed changes in behavior.
+When using the msbuild commands, there are 2 changes:
+
+- When running msbuild /t:pack on an .NET SDK based project, the pack targets from MSBuild will be imported.
```
dsplaisted: This seems to contradict the table below.

I think that the .NET SDK version of the pack (and restore) targets should always be used for SDK-style projects.  MSBuild is working on a .NET task host which will allow using tasks which target .NET (Core) from Full Framework MSBuild.  This should eliminate the need to multi-target tasks.  I'm not sure when this is supposed to be ready, but until it is I think you should continue to ship the .NET Framework versions of the tasks in the SDK.

/cc @baronfel

dsplaisted (dismissed):

accepted/2025/consolidate-pack-and-restore.md
baronfel: The basic direction we are moving across the .NET tooling stack is "SDK-style projects as much as possible use components from the resolved .NET SDK". (dotnet/msbuild#11142). This ensures consistency of behavior for all .NET SDK projects regardless of build tool (msbuild.exe/dotnet) or environment (IDE/CLI/CI). This suggestion does run counter to that, and I agree with Daniel that the SDK layout should remain the canonical location for logic for all SDK-style projects.

nkolev92: > This ensures consistency of behavior for all .NET SDK projects regardless of build tool (msbuild.exe/dotnet) or environment (IDE/CLI/CI).

This is gonna be challenging in VS scenarios, and would've made some more recent changes more difficult. Anyway, we can take that offline.

>  I'm not sure when this is supposed to be ready, but until it is I think you should continue to ship the .NET Framework versions of the tasks in the SDK.

We can certainly do that, I'll push an update to the spec for that tomorrow.

I'd love to have this done for .NET 10, since it'll allow us to stop publishing the package and reduce the number of assemblies we target 3.x. 
It's honestly less controversial to have the pack targets be used from the SDK, since there's no breaking changes whatsoever for SDK customers and it's just a .NET SDK layout change.

nkolev92: Pushed a change.

accepted/2025/consolidate-pack-and-restore.md
nkolev92: Move IsPackable to a target will break anyone conditioning on IsPackable though.

I feel like it has to be something specific to the target execution only.

nkolev92:
fyi @aortiz-msft 

This change is: 
- A functionality win for anyone that still has to use old style csproj
- A tech debt win for our build/testing, since we can finally drop the .NET Standard 2.0 leg of our builds.
- A tech debt win reducing the size of the .NET SDK.

jeffkl (approved):

nkolev92:
I'll merge this PR as the design has been signed-off. 
We can still decided how/when to roll it out.

PR #14409, opened 2025-07-09
Add issue template for MCP Server feedback tickets

kartheekp-ms:
Fixes: https://github.com/NuGet/Client.Engineering/issues/3328

### DIFF
--- a/.github/ISSUE_TEMPLATE/MCPSERVER.yml
+++ b/.github/ISSUE_TEMPLATE/MCPSERVER.yml
@@ -6,7 +6,6 @@ body:
   - type: markdown
     attributes:
       value: |
-        ## 🧠 NuGet MCP Server Feedback
         Thank you for taking the time to file an issue!  
         Please fill out the sections below to help us triage and resolve your issue effectively.
 
--- a/.github/ISSUE_TEMPLATE/MCPSERVER.yml
+++ b/.github/ISSUE_TEMPLATE/MCPSERVER.yml
@@ -6,9 +6,8 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thank you for taking the time to file an issue!  
-        Please fill out the sections below to help us triage and resolve your issue effectively.
-
+        Thanks for taking the time to file an issue!
+        
   - type: dropdown
     id: issue-type
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/MCPSERVER.yml
```diff
@@ -0,0 +1,91 @@
+name: NuGet MCP Server Issue
+description: Report a bug, suggestion, or behavior discrepancy for the NuGet MCP Server (Model Context Protocol)
+title: "[MCP Server]: "
+labels: ["Product:NuGetMCPServer", "Triage:Untriaged"]
```
kartheekp-ms: We have to create `Product:NuGetMCPServer` if we agree on this label.

jeffkl: I would vote for just `Product:MCPServer`

jeffkl (dismissed):
Thanks for taking care of this!

.github/ISSUE_TEMPLATE/MCPSERVER.yml
kartheekp-ms: I agree. The word `NuGet` is redundant. Removed it in the latest commit.

jeffkl (approved):

martinrrm (approved):

Nigusu-Allehu (approved):

.github/ISSUE_TEMPLATE/MCPSERVER.yml
jeffkl: I have created the label

nkolev92 (dismissed):
Why not just add the MCP Server in the dropdown for Bug/Feature/DCR and all other templates?

jeffkl:
> Why not just add the MCP Server in the dropdown for Bug/Feature/DCR and all other templates?

We want the template to ask details very specific to Copilot, like what the prompt was and what other Copilot extensions are installed.  Do issue templates allow us to show/hide sections based on the selection in this Product dropdown?

nkolev92:
Should be possible as far as I know. 

I also don't have time to dig deeper (it's a bit of trial & error thing).

kartheekp-ms:
> Should be possible as far as I know.
> 
> I also don't have time to dig deeper (it's a bit of trial & error thing).

YAML Lint says the following is valid whereas GitHub raises a [critical error for the same code in the issue template](https://github.com/NuGet/Home/blob/dev-kmps-test-yml/.github/ISSUE_TEMPLATE/test.yml). There are no examples for using conditionals in [GitHub docs](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema) also. I am assuming it is unsupported feature for now.

image

Issue #14345, opened 2025-06-02
Package pruning is enabled for all projects targeting .NET 10 including multi-targeted ones


nkolev92:
ie, maybe pruning should be enable when nuget audit mode is all.

Frulfump:
I don't think package pruning is related to nuget audit mode or nuget audit. I view pruning as a completely separate concept and nuget audit being related to security auditing and the mode just controlling the depth of the dependency graph that you look at.

(You obviously know this better than me)
NuGet Audit Mode is `all` for .NET 9.0.100 SDK and then that was reverted to `direct` in 9.0.101 and onwards to .NET 10 when it was again changed to `all`.

https://learn.microsoft.com/en-gb/dotnet/core/compatibility/sdk/10.0/nugetaudit-transitive-packages#previous-behavior "In .NET 9 preview 6, NuGetAuditMode's default was changed to all for all projects, and this change was reverted back to direct in the .NET 9.0.101 SDK"
https://learn.microsoft.com/en-gb/dotnet/core/compatibility/sdk/10.0/nugetaudit-transitive-packages#new-behavior
"When projects target .NET 10 or higher, then NuGetAuditMode defaults to all if it hasn't been explicitly set."

With that said maybe it should be a new separate setting to control this? It's easy to warn not as error for NU1510 for those that have warnings as errors enabled. And NU1510 is only raised for direct dependencies https://learn.microsoft.com/en-gb/dotnet/core/compatibility/sdk/10.0/nu1510-pruned-references which makes it a bit strange to tie it to the NuGet AuditMode `all` option.

nkolev92:
Thanks for the feedback @Frulfump. 

> I don't think package pruning is related to nuget audit mode or nuget audit. I view pruning as a completely separate concept and nuget audit being related to security auditing and the mode just controlling the depth of the dependency graph that you look at.

They are definitely separate features, but audit is the biggest use-case for pruning, so that's why they're so related. 

Note that enabling pruning for existing projects is a breaking change (as opposed to tying it nuget audit mode which is coming in as a new default in .NET 10 or tying it to .NET 10 adoption), which what this issue is tracking, whether we're ok with how broad we're making the breaking change.

We do have some documentation updates to before P7 of .NET 10 though. 

- https://github.com/NuGet/Home/issues/14254 
- https://github.com/dotnet/docs/issues/47292
- https://github.com/NuGet/Home/issues/14410

[...]

Issue #14371, opened 2025-06-18
Audit sources info incorrect in relation to `dotnet list package --vulnerable` usage


Frulfump:
### Docs Page URL

https://learn.microsoft.com/en-gb/nuget/concepts/auditing-packages

### Description of the Issue

> Audit sources are not used by dotnet list package --vulnerable at this time.
https://learn.microsoft.com/en-gb/nuget/concepts/auditing-packages#audit-sources

But this was added in NuGet 6.14 according to the release notes https://learn.microsoft.com/en-us/nuget/release-notes/nuget-6.14
> dotnet list package --vulnerable should support auditSources - [#13767](https://github.com/NuGet/Home/issues/13767)

And is included in the 9.0.300 SDK as per the same page

(the wording of `should` is a bit funny, but it stems from the issue title) the PR that implemented it is here https://github.com/NuGet/NuGet.Client/pull/6237

### Suggested Fix

Include some info about it being available in 9.0.300+ SDKs

### Additional Context

N/A

nkolev92:
https://github.com/NuGet/Home/issues/14021 was incomplete.

Issue #13954, opened 2024-11-21
NuGet AuditSources support in the Package Manager UI


nkolev92:
Repos that don't use nuget.org directly, can define an auditSource to be used for NuGetAudit at restore time. 
See audit sources: https://learn.microsoft.com/en-us/nuget/concepts/auditing-packages#audit-sources. 

An example is the NuGet.Client repo, https://github.com/NuGet/NuGet.Client/blob/dev/NuGet.Config

![image](https://github.com/user-attachments/assets/29e4d990-b95a-4c65-b3d8-153ade125f84)

The solution explorer and error list are restore based and use audit sources, but the PM UI has it's own logic. 

This issue tracks making Visual Studio UIs support audit sources the exact same way restore does.

nkolev92:
May be done in conjunction with #13050

Issue #13050, opened 2023-11-29
"Manage NuGet Packages" in vulnerability gold bar could take you to a better PM UI view


joelverhagen:
### NuGet Product(s) Affected

Visual Studio Package Management UI

### Current Behavior

When you click "Manage NuGet Packages" in the vulnerability gold bar
![image](https://github.com/NuGet/Client.Engineering/assets/94054/781d1895-bca8-4759-904f-ca929ef6b3d5)

It takes you to the solution view of PM UI. This might be hard or slow to track down the package you need to fix.
![image](https://github.com/NuGet/Home/assets/94054/68e53fcb-d8c4-4f33-aaf1-05bb0551ce12)

### Desired Behavior

It could take you to the Installed tab on the PM UI for the specific project that has the vulnerability. If there are multiple projects with vulnerable packages then the solution view of PM UI makes sense.

It could also pre-check the "Show only vulnerable" checkbox. 

It could select the lowest or highest non-vulnerable version in the version selector.

![image](https://github.com/NuGet/Home/assets/94054/890cec72-e9ee-4822-98e0-cbaa9a8cd810)

nkolev92:
I'm tweaking this to a feature, while it's technically a change in behavior, I think the proposal is superior and a next iteration on what we have today.

PR #14442, opened 2025-07-24
Pruning roll-out take 2

nkolev92:
Design for https://github.com/NuGet/Home/issues/14345.

### DIFF
--- a/accepted/2025/prune-package-reference-rollout-take-2.md
+++ b/accepted/2025/prune-package-reference-rollout-take-2.md
@@ -1,0 +1,161 @@
+# PrunePackageReference in .NET 10
+
+- Nikolche Kolev <https://github.com/nkolev92>
+- <https://github.com/NuGet/Home/issues/14345>
+
+## Summary
+
+In this document, the various roll-out options for [PrunePackageReference](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-References-in-Project-Files.md#prunepackagereference).
+
+The ideas explored are:
+
+- Enabling pruning for all .NET versions + .NET Standard 2.0+ (.NET 10 Preview 1 behavior)
+- Enabling pruning for .NET 8+ and .NET Standard 2.0+ (proposed .NET 10 Preview 7 behavior)
+- Enabling pruning for .NET 10 only (new)
+- Enabling pruning when NuGetAuditMode is all. (new)
+
+## Motivation
+
+The main benefits of pruning
+
+1. Reduces false positives by NuGet Audit, Component Governance and other dependency auditing tools.
+Our data investigations show that up 30% of transitive vulnerabilities are within packages that may be removed by pruning.
+This improves the confidence in NuGetAudit and reduces unnecessary alerts by scanners that do not ship as part of the .NET SDK tooling.
+1. Improves performance, by virtue of having smaller dependency graphs and fewer packages downloaded.
+
+The following list summarizes the list of per framework changes that would happen with pruning enabled.
+
+### Observable build changes for pruning
+
+- Transitive packages:
+  - Packages are removed and not downloaded. Packages are excluded from the project.assets.json and deps.json files, compared to the previous experience.
+- Direct packages, `PrivateAssets='all'` and `IncludeAssets='none'` are implicitly applied.
+  - `IncludeAssets='none'` ensures that the assemblies from this package are not used during the build. Before pruning existed, the conflict resolution during the build ensured that platform assemblies were preferred over those coming from the packages.
+  - `PrivateAssets='all'` ensures that the packages aren't included in packages or through project references.
+- Customers with builds that have customization may run into build failures if they are:
+  - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
+    - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
+  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
+- If the project targets .NET 10, [NU1510](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/errors-and-warnings/NU1510.md) and [NU1511](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/errors-and-warnings/NU1510.md) may be raised.
+
+## Explanation
+
+### Pruning in .NET 10 historical log
+
+- .NET 10 Preview 1
+  - default for every framework when using the .NET 10. SDK Details in the following design: <https://github.com/NuGet/Home/blob/dev/accepted/2025/prune-package-reference-rollout.md>.
+  Given that this was an early preview, we considered 
+  - NU1510, warning for prunable direct PackageReference was raised on a per framework level in every situation pruning was enabled.
+- .NET 10 Preview 3
+  - NU1510, warning for prunable direct PackageReference, was gated behind .NET 10 frameworks only. This meant that it was no longer a breaking change.
+- .NET 10 Preview 5
+  - NU1510 heuristic was updated
+- .NET 10 Preview 7
+  - Direct PackageReference within pruning range are automatically applied with `PrivateAssets=all` and `IncludeAssets=none`. <https://github.com/NuGet/Home/blob/dev/accepted/2025/PrunePackageReference-with-direct-PackageReference>.
+
+### What went wrong
+
+During the insertion into .NET 10 SDK Preview 7, <https://github.com/dotnet/sdk/pull/49740>, we discovered a bug in the .NET Core App 2.2 and below pruning behavior.
+In particular, pruning applied `IncludeAssets=none` to `Microsoft.NETCore.App`, since it was specified for pruning.
+This is a conflict in the input for restore, since `Microsoft.NETCore.App` is actually needed for building netcoreapp2.2 and below. <https://github.com/dotnet/sdk/issues/49917>
+
+### What did we do
+
+Given that .NET 10 Preview 7 was really close to shipping, we decided to enable pruning for .NET8+ as an immediate remediation. The motivation behind our policy was to enable by default for the frameworks in support, while minimizing the risk for unexpected bugs.
+We are currently reevaluating what the best strategy in terms of feature benefit + risk is.
+
+### Enable per framework comparison
+
+For the following list comparing the approaches a few things are important to consider:
+
+- NuGetAuditMode=all for projects targeting .NET 10 or above.
+What this means is that if at least 1 framework is .NET 10, the whole projects auditing warnings.
+NuGetAudit cannot be partially enabled.
+| Frameworks | NuGetAuditMode |
+|------------|----------------|
+| net10.0 | all |
+| net9.0 | direct |
+| net9.0;net10.0 | all |
+- Pruning runs per framework, or in other words, it can be partially enabled.
+- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting `RestoreEnablePackagePruning=true`
+- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of the it. We can always say pruning is not supported for <= .NET (Core) 2.2.
+
+#### All .NET & .NET Standard 2.0 >=
+
+Pros:
+
+- NET 10 P1 behavior. Nearly completely eliminates NuGetAudit, component governance and 3rd party scanner false positives.
+  - We have had pruning for transitive enabled since .NET 10 preview 1, and we have not received any breaking bugs aside from the NU1510 diagnostics, which has been fixed, and is disabled for projects targeting .NET 9 and earlier)
+- Unnecessary platform packages will not be a part of any package built with the .NET 10 SDK (due to direct PrivateAssets=all)
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+
+#### >= .NET 3 & .NET Standard 2.0 >=
+
+Pros:
+
+- Nearly completely eliminates NuGetAudit and component governance false positives. .NET CoreApp 2.2 comprise under 10% of projects.
+- Bug free zone. This is the broadest set of frameworks we can enable it for that has not encountered pruning bugs.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+- If we are going to go with this approach, we are better off fixing the bug and enabling it for all .NET frameworks or just declaring pruning unavaiable for <= .NET Core 2.2 and not fixing the bug.
+
+#### >= .NET 8 & .NET Standard 2.0 >=
+
+Pros:
+
+- Enables pruning for all in support frameworks. Providing benefits to projects staying current.
+- Does not address false positives for every customer currently getting false positive warnings (see example: <https://github.com/dotnet/sdk/issues/49226>). 
+- Still impacts a considerable amount of projects, about 2/3 of projects.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects. These are projects target actively supported runtimes, so these users may be more receptive of the risk.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+- We are making a TFM based enablement on the runtime supported, but that logic really only applies for applications, but pruning and audit are enabled for both apps and libraries. Library authors will often target lower version of the framework because that's all they need.
+
+### >= .NET 10 only
+
+Pros:
+
+- Not a breaking change.
+- Addresses false positives for .NET 10 and above
+
+Cons:
+
+- Does not address false positives for every customer actively getting false positive warnings (0% of projects in 17.14 get the benefit.)
+- Projects multi targeting .NET 10 and above, get only partial benefits, since pruning will only work for .NET 10. Say they System.Text.Json 8.0.0 in project targeting .NET 10, and .NET 8, they'll get a false positive warning because pruning is not enabled for .NET 8.
+- NU1510, unnecessary package warning, is likely to never be triggered for multi-targeted projects.
+
+### NuGetAuditMode=all enables pruning
+
+Pros:
+
+- Nearly completely eliminates NuGetAudit false positives
+- Potential breaking change only for 1% of total .NET SDK projects.
+- NU1510, unnecessary package warning, will be triggered correctly in all scenarios (.NET 10 only warning, so no breaking change.)
+- Can choose to combine this with `>= .NET 10 only` and `>= .NET 8 & .NET Standard 2.0 >=`.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects. However, the benefits for these customers are more obvious, since they are likely to see fewer vulnerability warnings. They made an explicit decision to adopt NuGetAudit, and they will likely be receptive of a feature improve NuGetAudit itself. (see example: <https://github.com/dotnet/sdk/issues/49226>)
+
+## Prior Art
+
+- Pruning is the restore equivalent of [the build time conflict resolution](<https://github.com/dotnet/sdk/blob/262b9c3d6cf67287f649e38d83e6c5d9d08feb8a/src/Tasks/Common/ConflictResolution/ResolvePackageFileConflicts.cs#L178-L182>), that is current enable for every framework.
+
+## Unresolved Questions
+
+- Precise Roll-out strategy for PrunePackageReference for .NET 10 SDK.
--- a/accepted/2025/prune-package-reference-rollout-take-2.md
+++ b/accepted/2025/prune-package-reference-rollout-take-2.md
@@ -86,7 +86,7 @@ NuGetAudit cannot be partially enabled.
 
 Pros:
 
-- NET 10 P1 behavior. Nearly completely eliminates false positives om NuGetAudit, component governance and 3rd party scanners.
+- NET 10 P1 behavior. Nearly completely eliminates false positives of NuGetAudit, component governance and 3rd party scanners.
   - We have had pruning for transitive enabled since .NET 10 preview 1, and we have not received any breaking bugs aside from the NU1510 diagnostics, which has been fixed, and is disabled for projects targeting .NET 9 and earlier)
 - Unnecessary platform packages will not be a part of any package built with the .NET 10 SDK (due to direct PrivateAssets=all)
 
--- a/accepted/2025/prune-package-reference-rollout-take-2.md
+++ b/accepted/2025/prune-package-reference-rollout-take-2.md
@@ -37,7 +37,7 @@ The following list summarizes the list of per framework changes that would happe
   - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning does not remove direct dependencies, so this would need to be a transitive package.
     - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
   - Aliases for PackageReference that will be pruned will stop working. This is *unlikely* to be a problem, since we normally wouldn't expect customers to alias platform packages.
-  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
+  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need to do anything custom to make things work.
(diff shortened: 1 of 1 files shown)


accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,161 @@
...
+- Direct packages, `PrivateAssets='all'` and `IncludeAssets='none'` are implicitly applied.
+  - `IncludeAssets='none'` ensures that the assemblies from this package are not used during the build. Before pruning existed, the conflict resolution during the build ensured that platform assemblies were preferred over those coming from the packages.
+  - `PrivateAssets='all'` ensures that the packages aren't included in packages or through project references.
+- Customers with builds that have customization may run into build failures if they are:
+  - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning is not allowed for direct dependencies, so this would need to be a transitive package.
+    - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
+  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
```
ViktorHofer: I agree with that sentiment. We might need to react to this in the few repositories that build the shared frameworks: aspnetcore, runtime and windowsdesktop but I don't expect any other components to be significantly impacted.

ViktorHofer (commented):
Great summary of the current state. I read through it and still believe that the option that we initially chose "> All .NET & .NET Standard 2.0 >=" is the right one.

I agree that we should fix the entries for frameworks that don't have a targeting pack and use the meta package concept: .NETCoreApp2.x and .NETStandard2.0. The latter might not be affected.

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,161 @@
...
+- Enabling pruning when NuGetAuditMode is all. (new)
+
+## Motivation
+
+The main benefits of pruning
+
+1. Reduces false positives by NuGet Audit, Component Governance and other dependency auditing tools.
+Our data investigations show that up 30% of transitive vulnerabilities are within packages that may be removed by pruning.
+This improves the confidence in NuGetAudit and reduces unnecessary alerts by scanners that do not ship as part of the .NET SDK tooling.
+1. Improves performance, by virtue of having smaller dependency graphs and fewer packages downloaded.
```
aortiz-msft, line 27: Isn't savings in size of deployments also a motiviation?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,161 @@
...
+
+Given that .NET 10 Preview 7 was really close to shipping, we decided to enable pruning for .NET8+ as an immediate remediation. The motivation behind our policy was to enable by default for the frameworks in support, while minimizing the risk for unexpected bugs.
+We are currently reevaluating what the best strategy in terms of feature benefit + risk is.
+
+### Enable per framework comparison
+
+For the following list comparing the approaches a few things are important to consider:
+
+- NuGetAuditMode=all for projects targeting .NET 10 or above.
+What this means is that if at least 1 framework is .NET 10, the whole projects auditing warnings.
```
aortiz-msft, line 76: Is it technically possible to replicate this behavior for Pruning?

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: No, build time conflict resolution was already taking care of that.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Yes, https://github.com/NuGet/Home/blob/dev-nkolev92-pruningRolloutTake2/accepted/2025/prune-package-reference-rollout-take-2.md#nugetauditmodeall-enables-pruning one of options is that.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Yeah, our repos are definitely the most complex ones from this perspective.

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+## Summary
+
+In this document, the various roll-out options for [PrunePackageReference](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-References-in-Project-Files.md#prunepackagereference).
+
+The ideas explored are:
+
+- Enabling pruning for all .NET versions + .NET Standard 2.0+ (.NET 10 Preview 1 behavior)
+- Enabling pruning for .NET 8+ and .NET Standard 2.0+ (proposed .NET 10 Preview 7 behavior)
+- Enabling pruning for .NET 10 only (new)
+- Enabling pruning when NuGetAuditMode is all. (new)
```
Frulfump, line 16: I'll add part of my comment from the linked issue here

"My only concern is that features get tangled and opted into when enabling another setting that could cause confusion. For me ideally both pruning and auditmode=direct (direct is default in .NET 10 so that's good) are the defaults going forward but I don't want them to be toggled based on each other (but if there's potential for breaking changes I understand pruning needs to be handled a bit more carefully but not that it would need to be tied to another option and that would be even more reason to keep them separate)."

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Summary
+
+In this document, the various roll-out options for [PrunePackageReference](https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/consume-packages/Package-References-in-Project-Files.md#prunepackagereference).
+
+The ideas explored are:
+
+- Enabling pruning for all .NET versions + .NET Standard 2.0+ (.NET 10 Preview 1 behavior)
+- Enabling pruning for .NET 8+ and .NET Standard 2.0+ (proposed .NET 10 Preview 7 behavior)
+- Enabling pruning for .NET 10 only (new)
```
Frulfump, line 14: This seems like the safest option if its assumed there could be potential bugs in the implementation causing breaking changes but I like progress so I still vote for option 1 or 2 there's one more preview and 2 RC's for it to be battle tested, dealing with out of support versions of .NET I think is of lesser interest but could be included if it comes at a very low cost.

(I assume this means when the TFM is `net10.0`? Should maybe be clarified if so)

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Observable build changes for pruning
+
+- Transitive packages:
+  - Packages are removed and not downloaded. Packages are excluded from the project.assets.json and deps.json files, compared to the previous experience.
+- Direct packages, `PrivateAssets='all'` and `IncludeAssets='none'` are implicitly applied.
+  - `IncludeAssets='none'` ensures that the assemblies from this package are not used during the build. Before pruning existed, the conflict resolution during the build ensured that platform assemblies were preferred over those coming from the packages.
+  - `PrivateAssets='all'` ensures that the packages aren't included in packages or through project references.
+- Customers with builds that have customization may run into build failures if they are:
+  - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning does not remove direct dependencies, so this would need to be a transitive package.
```
Frulfump, line 39: Worth calling out NU1510 here?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+What this means is that if at least 1 framework is .NET 10, the whole projects auditing warnings.
+NuGetAudit cannot be partially enabled.
+| Frameworks | NuGetAuditMode |
+|------------|----------------|
+| net10.0 | all |
+| net9.0 | direct |
+| net9.0;net10.0 | all |
+- Pruning runs per framework, or in other words, it can be partially enabled.
+- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting `RestoreEnablePackagePruning=true`
+- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of the it. We can always say pruning is not supported for <= .NET (Core) 2.2.
```
Frulfump: "Typo"
>criticality of the it

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### All .NET & .NET Standard 2.0 >=
+
+Pros:
+
+- NET 10 P1 behavior. Nearly completely eliminates NuGetAudit, component governance and 3rd party scanner false positives.
```
Frulfump: Little bit unclear but this means it eliminated the mentioned things?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+  - We have had pruning for transitive enabled since .NET 10 preview 1, and we have not received any breaking bugs aside from the NU1510 diagnostics, which has been fixed, and is disabled for projects targeting .NET 9 and earlier)
+- Unnecessary platform packages will not be a part of any package built with the .NET 10 SDK (due to direct PrivateAssets=all)
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
```
Frulfump: Is this not `all` in .NET 10 / `net10.0` TFM (unsure which one it is right now)

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+
+#### >= .NET 3 & .NET Standard 2.0 >=
```
Frulfump: Core 3 & >= .NET 5? I went 👀 when I saw .NET 3 :P Haven't really seen .NET used without using .NET 5+ when it refers to post Core drop and not seen it without pre Core drop

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+Pros:
+
+- Nearly completely eliminates NuGetAudit and component governance false positives. .NET CoreApp 2.2 comprise under 10% of projects.
+- Bug free zone. This is the broadest set of frameworks we can enable it for that has not encountered pruning bugs.
```
Frulfump: yet

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
```
Frulfump: I assume we would see a drop in warnings over time?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+- If we are going to go with this approach, we are better off fixing the bug and enabling it for all .NET frameworks or just declaring pruning unavaiable for <= .NET Core 2.2 and not fixing the bug.
+
+#### >= .NET 8 & .NET Standard 2.0 >=
+
+Pros:
+
+- Enables pruning for all in support frameworks. Providing benefits to projects staying current.
+- Does not address false positives for every customer currently getting false positive warnings (see example: <https://github.com/dotnet/sdk/issues/49226>).
```
Frulfump, line 122: Hoping this will get addressed, false positives are highly annoying.

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+- Does not address false positives for every customer currently getting false positive warnings (see example: <https://github.com/dotnet/sdk/issues/49226>). 
+- Still impacts a considerable amount of projects, about 2/3 of projects.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects. These are projects target actively supported runtimes, so these users may be more receptive of the risk.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects (Only 1% of restores have NuGetAuditMode=all)
+We will see reduction in downloads of these packages and reduction of these packages being part of newly published packages, but we won't necessarily reduce the warnings count (overall we do want to eliminate vulnerable package usage).
+The primary benefit is prunable packages not appearing in newly published packages due to automatic PrivateAssets=all.
+- We are making a TFM based enablement on the runtime supported, but that logic really only applies for applications, but pruning and audit are enabled for both apps and libraries. Library authors will often target lower version of the framework because that's all they need.
```
Frulfump, line 131: How come the logic only applies for apps in this case?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+Pros:
+
+- Nearly completely eliminates NuGetAudit false positives
+- Potential breaking change only for 1% of total .NET SDK projects.
+- NU1510, unnecessary package warning, will be triggered correctly in all scenarios (.NET 10 only warning, so no breaking change.)
+- Can choose to combine this with `>= .NET 10 only` and `>= .NET 8 & .NET Standard 2.0 >=`.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects. However, the benefits for these customers are more obvious, since they are likely to see fewer vulnerability warnings. They made an explicit decision to adopt NuGetAudit, and they will likely be receptive of a feature improve NuGetAudit itself. (see example: <https://github.com/dotnet/sdk/issues/49226>)
```
Frulfump, line 158: "Typo"
>of a feature improve NuGetAudit

But this is not an opt-in anymore for .NET 10 SDK?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Prior Art
+
+- Pruning is the restore equivalent of [the build time conflict resolution](<https://github.com/dotnet/sdk/blob/262b9c3d6cf67287f649e38d83e6c5d9d08feb8a/src/Tasks/Common/ConflictResolution/ResolvePackageFileConflicts.cs#L178-L182>), that is current enable for every framework.
```
Frulfump: "Typo"
>that is current enable for
(is it also multi space after the comma?)

Frulfump (approved):
I liked the overall pros/cons and learned a bit more about pruning while reading it as well.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Thanks, added feature tangling causing confusion in the cons.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: It's called out on line 40.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: it eliminates false positives. 

I'll actually switch the order and it'll read better. Thanks!

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: This is referring to non .NET 10 projects.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: cleaned it up. 

NetCoreApp3.0 is the first version where platform as packages was removed.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Fair.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Clarified it a bit. 

What I am trying to say here is that there are actually no warnings to eliminate becuase transitive auditing is not happening.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Apps care about the runtime they're targeting and when a runtime goes out of support, they're likely to move their project to a newer runtime since they'll get a warning. 

For libraries, the tfm targeted just defines the APIs is has available, but the questions around vulnerabilities are super relevant. 

See some context in https://github.com/dotnet/sdk/issues/28518.

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: Added the 2 separately.

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,178 @@
...
+  - `PrivateAssets='all'` ensures that the packages aren't included in packages or through project references.
+- Customers with builds that have customization may run into build failures if they are:
+  - Manually referencing an assembly or props/targets from a package from the global packages folder. Note that pruning does not remove direct dependencies, so this would need to be a transitive package.
+    - Manually copying assembly during a build - at runtime, the platform version will be preferred anyways, but it could lead to the build failing if the package cannot be found on disk anymore.
+  - Aliases for PackageReference that will be pruned will stop working. This is *unlikely* to be a problem, since we normally wouldn't expect customers to alias platform packages.
+  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need do to anything custom to make things work.
```
Frulfump: Typo
```suggestion
  - We don't expect this to be a common scenario given that the packages being pruned are within the framework and customers do not need to do anything custom to make things work.
```

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,182 @@
...
+Pros:
+
+- NET 10 P1 behavior. Nearly completely eliminates false positives om NuGetAudit, component governance and 3rd party scanners.
+  - We have had pruning for transitive enabled since .NET 10 preview 1, and we have not received any breaking bugs aside from the NU1510 diagnostics, which has been fixed, and is disabled for projects targeting .NET 9 and earlier)
+- Unnecessary platform packages will not be a part of any package built with the .NET 10 SDK (due to direct PrivateAssets=all)
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects.
+- NuGetAuditMode is gonna be direct for a large majority of existing projects, which target .NET 9 and older (Only 1% of restores have manually enabled NuGetAuditMode=all)
```
Frulfump, line 98: One of the reasons we didn't change AuditMode at work was that first audit source couldn't be set separately and Azure Artifacts don't have a vuln db (so even `direct` showed no vulns) and after audit source became a thing there was buggy behavior in VS where you just got a yellow label saying you had vulnerable packages and when you clicked it didn't tell you which packages were vulnerable anyway, I haven't checked back on the current state of things but will do that during fall.

The third reason is that we use GitHub Advanced Security for Azure DevOps so we still get the information but I would like to shift left if possible. I introduced listing vulnerable packages via dotnet CLI to the other developers so they at least manually could check before pushing and it would show in CI instead.

But maybe adoption is hurt by the fact that it's not that known that the option exists?

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,182 @@
...
+| net9.0;net10.0 | all |
+- Pruning runs per framework, or in other words, it can be partially enabled.
+- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting `RestoreEnablePackagePruning=true`
+- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of it. We can always say pruning is not supported for <= .NET (Core) 2.2.
+
+#### All .NET & .NET Standard 2.0 >=
+
+Pros:
+
+- NET 10 P1 behavior. Nearly completely eliminates false positives om NuGetAudit, component governance and 3rd party scanners.
```
Frulfump: Typo

```suggestion
- NET 10 P1 behavior. Nearly completely eliminates false positives of NuGetAudit, component governance and 3rd party scanners.
```

Frulfump (commented):
I didn't mention this prior but the different stats being provided is very interesting (and also very relevant)

accepted/2025/prune-package-reference-rollout-take-2.md
nkolev92: fyi @aortiz-msft @zivkan 

Your context around why you haven't enabled NuGetAuditMode all is super helpful! 

I do have answers to some of your concerns!

> I haven't checked back on the current state of things but will do that during fall.

We want to address that in the next version of Visual Studio. https://github.com/nuget/home/issues/13954.

> I introduced listing vulnerable packages via dotnet CLI using the nuget.org source (not sure if audit source can now be set as an option on the CLI)to the other developers so they at least manually could check before pushing and it would show in CI instead.

dotnet list package --vulnerable does support audit sources. 
One of my colleagues recently updated the documentation calling out that 9.0.300 supports AuditSources. https://github.com/NuGet/docs.microsoft.com-nuget/commit/ec65226fd6375f5d2a56a44f7779fda1582d14bd.

Frulfump: Thanks! (I remembered I created the issue #14371 to get that fixed in the docs now 😅).

In the issue you linked you linked to #13050 as well which covers the warning label in VS.

accepted/2025/prune-package-reference-rollout-take-2.md
ericstj: For the most part - as a side-effect of the pruning feature we changed/made-consistent the data used for conflict resolution which will help drop all major-version-matching packages.  https://github.com/dotnet/sdk/issues/44566

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,182 @@
...
+Given that .NET 10 Preview 7 was really close to shipping, we decided to enable pruning for .NET8+ as an immediate remediation. The motivation behind our policy was to enable by default for the frameworks in support, while minimizing the risk for unexpected bugs.
+We are currently reevaluating what the best strategy in terms of feature benefit + risk is.
+
+### Enable per framework comparison
+
+For the following list comparing the approaches a few things are important to consider:
+
+- NuGetAuditMode=all for projects targeting .NET 10 or above.
+What this means is that if at least 1 framework is .NET 10, the whole projects auditing warnings.
+NuGetAudit cannot be partially enabled.
```
donnie-msft, line 77: The table doesn't render for some reason. Maybe a new line will work?
```suggestion
NuGetAudit cannot be partially enabled.

```

donnie-msft (dismissed):

accepted/2025/prune-package-reference-rollout-take-2.md
Frulfump: Nice catch! To make it a part of the list it needs to be indented with four spaces (then no new line is required, but if it's to be viewed correctly outside of GitHub a leading and trailing new line might be required in addition to the indentation.)

## Markdown Example
```md
- NuGetAuditMode=all for projects targeting .NET 10 or above.
  What this means is that if at least one framework is .NET 10, the whole project surfaces auditing warnings.
  NuGetAudit cannot be partially enabled.
    | Frameworks       | NuGetAuditMode |
    |------------------|----------------|
    | net10.0          | all            |
    | net9.0           | direct         |
    | net9.0; net10.0  | all            |
- Pruning runs per framework, or in other words, it can be partially enabled.
- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting `RestoreEnablePackagePruning=true`.
- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of it. We can always say pruning is not supported for <= .NET (Core) 2.2.
```
## Live example
- NuGetAuditMode=all for projects targeting .NET 10 or above.
  What this means is that if at least one framework is .NET 10, the whole project surfaces auditing warnings.
  NuGetAudit cannot be partially enabled.
    | Frameworks       | NuGetAuditMode |
    |------------------|----------------|
    | net10.0          | all            |
    | net9.0           | direct         |
    | net9.0; net10.0  | all            |

- Pruning runs per framework, or in other words, it can be partially enabled.
- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting `RestoreEnablePackagePruning=true`.
- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of it. We can always say pruning is not supported for <= .NET (Core) 2.2.

## Just one new line issue example
Otherwise it looks like this when just the leading new line is added.

- NuGetAuditMode=all for projects targeting .NET 10 or above.
What this means is that if at least 1 framework is .NET 10, the whole projects auditing warnings.
NuGetAudit cannot be partially enabled.

| Frameworks | NuGetAuditMode |
|------------|----------------|
| net10.0 | all |
| net9.0 | direct |
| net9.0;net10.0 | all |
- Pruning runs per framework, or in other words, it can be partially enabled.
- Every customer can enable pruning in their projects already (shipped originally in 9.0.200) by setting RestoreEnablePackagePruning=true
- We should fix <https://github.com/dotnet/sdk/issues/49917>, regardless of the defaults, but some defaults increase the criticality of it. We can always say pruning is not supported for <= .NET (Core) 2.2.

aortiz-msft (dismissed):

accepted/2025/prune-package-reference-rollout-take-2.md
```diff
@@ -0,0 +1,183 @@
...
+- Potential breaking change only for 1% of total .NET SDK projects.
+- NU1510, unnecessary package warning, will be triggered correctly in all scenarios (.NET 10 only warning, so no breaking change.)
+- Can choose to combine this with `>= .NET 10 only` and `>= .NET 8 & .NET Standard 2.0 >=`.
+
+Cons:
+
+- Potential of a breaking change for pre-existing projects. However, the benefits for these customers are more obvious, since they are likely to see fewer vulnerability warnings. They made an explicit decision to adopt NuGetAudit, and they will likely be receptive of a feature improve NuGetAudit itself. (see example: <https://github.com/dotnet/sdk/issues/49226>)
+- Features get a bit tangled and enabling one feature opting you into another one could cause confusion.
+
+### >= .NET 10 enables pruning for all frameworks
```
ericstj, line 161: We should have explicitly called out .NETFramework.  I thought we had previously decided that pruning on .NETFramework wasn't a good idea (which is why it was excluded in both past iterations), however it looks like the current implementation enables this.

PR #14455, opened 2025-07-29
[spec] Indicate analyzer assets and respect analyzer filtering in project.assets.json

### LINKED ISSUE #6279: project.assets.json should indicate "analyzer" assets like "compile" and "runtime"

kfikadu:
Fixes: https://github.com/NuGet/Home/issues/6279

[Rendered MD](https://github.com/NuGet/Home/blob/dev-kfikadu-analyzerAssetsSpec/accepted/2025/indicate-analyzer-assets.md)

### DIFF
--- a/accepted/2025/indicate-analyzer-assets.md
+++ b/accepted/2025/indicate-analyzer-assets.md
@@ -26,7 +26,7 @@ When enabled via `<RestoreEnableAnalyzerAssets>true</RestoreEnableAnalyzerAssets
 - PrivateAssets/ExcludeAssets will correctly filter analyzers, preventing them from being included in projects that should not have them.
 
 This feature will initially be behind a feature flag defined as `<RestoreEnableAnalyzerAssets>`. 
-This property should be set as true in the project file. 
+This property should be set to true in the project file. 
 This is done to avoid a breaking change.
 
 ### Technical explanation
@@ -190,12 +190,30 @@ It also maintains backward compatibility by being opt-in via a feature flag, all
 
 NuGet's existing asset filtering system (compile, runtime assets) provides the model.
 
+**Content Files**
 Language-specific content selection (`contentFiles/{language}/`) demonstrates similar filtering patterns already work in the ecosystem.
-For example, a package can include assets such as `empty.txt` in both `contentFiles/cs/net9.0/` and `contentFiles/vb/net9.0/`, mapped via the `packagepath` property.
 
-To validate this behavior, a test was performed using a .NET 9 project that packed `empty.txt` into both C# and VB.NET content file folders. 
-The resulting NuGet package contained the file in both locations. When referenced by a C# project, only the C# content file was selected by the SDK; the VB.NET file was ignored.
-This was confirmed by inspecting the `project.assets.json`, which included both files in the package metadata but only used the language-appropriate asset.
+The SDK is able to determine the project language by the type of project file: 
+- `.csproj` files are treated as C# projects
+- `.vbproj` files are treated as VB.NET projects
+
+The NuGet package can contain analyzers and content files for multiple languages, and they are sorted into folders such as `contentFiles/cs/` and `contentFiles/vb/`.
+During the restore process, NuGet lists all of the content files into the `project.assets.json` file, regardless of the language. 
+
+In the `project.assets.json` file, each content file entry includes metadata, such as `"buildAction"`, `"codeLanguage"`, `"copyToOutput"`, etc. 
+- The `codeLanguage` property is what enables the SDK to filter and include the files that are relevant to the project's language
+
+For each file, MSBuild checks the `"codeLanguage"` property to determine if it should be included in the build
+- If `"codeLanguage"` matches the project language, the file is included in the build. 
+As a result, only the correct assets for the current project's are compiled, copied, or referenced. 
+
+This is similar to language-specific analyzer selection (`analyzers/dotnet/{language}/`).
+- The SDK determines the project language from the project file type (e.g., `.csproj`, `.vbproj`, `.fsproj`)
+
+Example: 
+- If a package contains `analyzers/dotnet/cs/Analyzer.dll` and `analyzers/dotnet/vb/VBAnalyzer.dll`:
+  - The C# project will only include `Analyzer.dll`
+  - The VB.NET project will only include `VBAnalyzer.dll`
 
 ## Unresolved Questions
 
@@ -205,7 +223,7 @@ This was confirmed by inspecting the `project.assets.json`, which included both 
 
 ## Future Possibilities
 
-This is a feature that is going to be enabled by default for all builds in the future
+This is a feature that is going to be enabled by default for all builds.
 
 Something that would be valuable would be an analysis of NuGet packages that have analyzers, and focusing on analyzers as dependencies. 
 It would also look at the impact of the `IncludeAssets` and `ExcludeAssets` settings. 
@@ -215,6 +233,12 @@ Ideally, this analysis would:
 - Provide data on how many packages (and consumers) would be impacted by a change in default behavior, helping to mitigate surprises or unintended consequences.
 - Affirm the correctness and robustness of the proposed design prior to broad deployment.
 
+Also, provide analysis on source generators as analyzers. Since it is possible for people to depend on source generators in a way that their build will fail without them, there should be data on which ones may fail. 
+
+Additional follow up should also be done in component governance. Currently, packages that exclusively contribute to `Compile` assets are treated as development dependencies.
+There are special case analyzers, such as in the SDK, that should be removed once this feature is implemented. 
+
+We should also consider creating a dedicated design time package graph. This will explicitly resolve and track packages that are meant to run as analyzers or tasks, making the process more accurate and maintainable
 <!-- What future possibilities can you think of that this proposal would help with? -->
 
 ### References
--- a/accepted/2025/indicate-analyzer-assets.md
+++ b/accepted/2025/indicate-analyzer-assets.md
@@ -7,61 +7,69 @@
 ## Summary
 
 `project.assets.json` does not currently indicate which analyzers are active, and analyzers are added even when `PrivateAssets` or `ExcludeAssets` should prevent this.
-This leads to analyzers being included in projects, which are not the expected behavior, especially when defaults for private/excluded assets are not respected
+This leads to analyzers being included in projects, which is not the expected behavior, especially when defaults for private/excluded assets are not respected
 
 ## Motivation
 
-Currently, analyzer assets in NuGet packages do not respect asset filtering options like ExcludeAssets and PrivateAssets, unlike other asset types (compile, runtime).
-Analyzers are included based on pattern matching in package files, not by explicit references in project.assets.json.
-Developers who try to exclude analyzers using asset filters still see analyzers loaded and executed during builds, leading to unexpected warnings or errors.
-This causes wasted time, frustration, and confusion, as configuration changes appear ineffective and troubleshooting becomes more difficult.
-Package authors cannot reliably control analyzer distribution or exclusion in downstream projects, making support and package design more complex.
-Real-world feedback shows users resorting to custom MSBuild scripts, manual artifact modification, or abandoning packages due to inability to control analyzer inclusion.
+- Analyzer assets currently don't respect standard NuGet asset filtering options like `ExcludeAssets` and `PrivateAssets`.
+- Developers can't control analyzer inclusion in their projects, leading to unexpected warnings or errors.
+- Users are required to use custom MSBuild scripts or abandon packages due to inability to control analyzer inclusion.
+- Package authors can't reliably control analyzer distribution
 
 ## Explanation
 
 ### Functional explanation
 
-Analyzer assets in NuGet packages will now respect asset filtering options like ExcludeAssets and PrivateAssets, just as other asset types do.
-- When excluding analyzers from a package, they will no longer be included or executed during the project's build.
-- Configurations will accurately control which analyzers are present, eliminating unexpected analyzer warnings or errors.
-- Both package authors and consumers will have clear, reliable control over analyzer inclusion without needing custom workarounds.
+Enable analyzer assets to respect asset filtering options like other asset types.
+When enabled via `<EnableIndicateAnalyzerAssets>true</EnableIndicateAnalyzerAssets>`:
+- Analyzers will be tracked in the `project.assets.json` file under a new "analyzers" group.
+- PrivateAssets/ExcludeAssets will correctly filter analyzers, preventing them from being included in projects that should not have them.
+- Developers will gain proper control over analyzer inclusion
 
-This feature will initially be behind a feature flag defined as `<EnableIndicateAnalyzerAsset>`. 
+This feature will initially be behind a feature flag defined as `<EnableIndicateAnalyzerAssets>`. 
 This property should be set as true in the project file. 
 This is done to avoid a breaking change.
 
 ### Technical explanation
 
-#### NuGet changes
+#### File Structure
 
-Update the NuGet restore process to:
-- Identify analyzer DLLs in each package.
-- Filter out analyzers that are excluded by PrivateAssets/ExcludeAssets
-- Add an "analyzers" group under the appropriate "targets" node for each package in project.assets.json.
+|Pattern|Description|Example|
+|---|---|---|
+|`analyzers/dotnet/cs/{name}.dll`|C# analyzers|`analyzers/dotnet/cs/MyAnalyzer.dll`|
+|`analyzers/dotnet/vb/{name}.dll`|VB.NET analyzers|`analyzers/dotnet/vb/MyAnalyzer.dll`|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzers|`analyzers/dotnet/{language}/MyAnalyzer.dll`|
 
-Example Output
+**NuGet Changes**
+
+- Add "analyzers" group to project.assets.json during restore.
+- Respect `PrivateAssets` and `ExcludeAssets` when populating this group.
+
+**SDK Changes**
+
+Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
+- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
+- Only load analyzers that are listed in this group.
+- When `<EnableIndicateAnalyzerAssets>` is set to true, the SDK will only use the analyzer group that is in the assets file to determine which analyzers to include. 
+- If the analyzers group is missing from the assets file and the feature flag is enabled, the SDK won't fall back to legacy scanning. 
+  - In this case, no analyzers will be selected, even if analyzer assets exist in the packages.
+- If the feature flag is not set or is false, the SDK will use the legacy scanning behavior to discover analyzers and preserve compatibility. 
+
+**Example Output**
 ```json
 "targets": {
   ".NETCoreApp,Version=v8.0": {
     "My.Analyzer.Package/1.0.0": {
-      "analyzers": [
-        "analyzers/dotnet/cs/MyAnalyzer.dll"
-      ]
+      "type": "package",
+      "compile": { ... },
+      "runtime": { ... },
+      "analyzers": {
+        "analyzers/dotnet/cs/MyAnalyzer.dll": {}
+      }
     }
   }
 }
-```
-
-#### SDK Changes
-
-Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
-- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
-- Only load analyzers are listed in this group.
-- When <EnableIndicateAnalyzerAssets> is set to true, the SDK will only use the analyzer group that is in the assets file to determine which analyzers to include. 
-- If the analyzers group is missing from the assets file and the feature flag is enabled, the won't fall back to legacy scanning. 
-  - In this case, no analyzers will be selected, even if analyzer assets exist in the packages.
-- If the feature flag is not set or is false, the SDK will use the legacy scanning behavior to discover analyzers and preserve compatibility. 
+``` 
 
 #### Testing
 
@@ -70,48 +78,118 @@ Validate correct behavior with:
 - Multi-targeted packages (ensure correct analyzer/TFM pairing).
 - Projects with the feature flag enabled, confirming that included/excluded analyzers match expectations.
 
-## Analyzer and Compile Asset Transitivity
+## Private Assets and Transitive Behavior
 
-NuGet handles different asset types (compile, analyzers, etc.) with specific rules for transitivity and filtering.
+### Current Behavior (Without Feature Flag)
 
-**Analyzers**
-- Treated as private assets by default. They do not flow transitively, meaning that analazers from a package will not be included in projects that reference the package unless explicitly included.
+| Attribute | What It Should Do | What Actually Happens | project.assets.json |
+|-----------|-------------------|----------------------|---------------------|
+| `PrivateAssets="analyzers"` | Prevent analyzers from flowing to dependent projects | analyzers still flow transitively | No analyzer information tracked |
+| `ExcludeAssets="analyzers"` | Exclude analyzers from this project entirely | analyzers still included | No analyzer information tracked |
+| `IncludeAssets="compile;runtime"` | Only include compile and runtime assets (exclude analyzers) | **Analyzers still included** - not controlled by asset filtering | No analyzer information tracked |
 
-**Compile Assets**
-- These are transistive. Unless you explicity exclude them, they will flow through the entire dependency chain.
+### New Behavior (With EnableIndicateAnalyzerAssets=true)
 
-### Test Results
-- Setting `ExcludeAssets="analyzers"` or `PrivateAssets="all"` on a `PackageReference` only affects that specific reference; analyzers do not flow through dependencies.
-- Compile assets are transitive. If you set `ExcludeAssets="compile"` on a reference, compile-time assemblies from that reference are excluded, but will still flow from other dependencies unless also excluded.
-- The risk of old packages unintentionally redistributing analyzers is low because analyzers must be explicitly included by the package author.
+**Scenario 1: Library with PrivateAssets**
+```xml
+<!-- Library.csproj -->
+<PackageReference Include="MyAnalyzer" Version="1.0.0" 
+                  PrivateAssets="analyzers" />
+```
+
+**Library's project.assets.json:**
+```json
+"MyAnalyzer/1.0.0": {
+  "type": "package",
+  "analyzers": {
+    "analyzers/dotnet/cs/MyAnalyzer.dll": {}
+  }
+}
+```
+**App's project.assets.json (references Library):**
+```json
+"MyAnalyzer/1.0.0": {
+  "type": "package"
+  // No "analyzers" section - blocked by PrivateAssets
+}
+```
+
+**Scenario 2: App with ExcludeAssets**
+```xml
+<!-- App.csproj -->
+<PackageReference Include="LibraryWithAnalyzers" Version="1.0.0" 
+                  ExcludeAssets="analyzers" />
+```
+Result: App's project.assets.json will not contain analyzer entries for LibraryWithAnalyzers
+
+### How Transitivity Works
+
+```
+Project A -> Project B -> Package C (with analyzers)
+```
+
+- **Without PrivateAssets:** Analyzers flow from C to B to A
+- **With PrivateAssets on B's reference to C:** Analyzers stop at B
+
+## Breaking Changes
+
+### No Breaking Changes Initially 
+
+This feature is opt-in via `<EnableIndicateAnalyzerAssets>true</EnableIndicateAnalyzerAssets>`, so existing projects will not be affected until they explicitly enable it.
+
+### When Enabled by Default
+
+Projects using `PrivateAssets="analyzers"` or `ExcludeAssets="analyzers"` will start working as intended:
+
+**Build Behavior Changes:**
+- Missing analyzer diagnostics in dependent projects
+- Different warning/error counts in build output
+- TreatWarningsAsErrors outcomes may change
+
+**CI/CD Pipeline Impact:**
+- Builds expecting certain analyzers may fail
+- Builds may pass that previously failed on analyzer warnings
+
+**Mitigation:**
+```xml
+<PropertyGroup>
+  <EnableIndicateAnalyzerAssets>false</EnableIndicateAnalyzerAssets>
+</PropertyGroup>
+```
+
+### Rollout Strategy
+
+1. **Current:** Opt-in via feature flag
+2. **Future:** Enable by default with clear communication
+3. **Later:** Remove feature flag
 
 
 ## Drawbacks
 
-- Requires changes to both NuGet restore logic and .NET SDK asset resolution, which may increase maintenance complexity and introduce risk of regression.
-- May require updates to existing tooling, documentation, or custom scripts that expect the current analyzer handling behavior.
-- Potential for minor performance impact during restore and build, as asset filtering and lock file generation become more granular for analyzers.
-- Older packages or tools that do not adopt the new pattern may continue to exhibit legacy behavior, requiring a fallback or compatibility layer that adds to system complexity.
+- Requires coordinated changes to NuGet and SDK
+- Breaking change when enabled by default
+- Potential minor performance impact during restore
+- Older tools may not understand the new project.assets.json format
 
 ## Rationale and alternatives
 
+This design is chosen to align with existing NuGet asset filtering patterns while providing a clear, structured way to manage analyzers.
+By introducing a new "analyzers" group in `project.assets.json`, it allows for consistent handling of analyzers similar to other asset types, while respecting existing filtering options like `PrivateAssets` and `ExcludeAssets`.
+It also maintains backward compatibility by being opt-in via a feature flag, allowing developers to adopt it gradually without breaking existing projects.
 <!-- Why is this the best design compared to other designs? -->
 <!-- What other designs have been considered and why weren't they chosen? -->
 <!-- What is the impact of not doing this? -->
 
 ## Prior Art
 
-### Language Selection 
+NuGet's existing asset filtering system (compile, runtime assets) provides the model.
 
-The .NET SDK and NuGet packaging system provide established support for language-specific content file selection using the `contentFiles/{language}/` convention. 
+Language-specific content selection (`contentFiles/{language}/`) demonstrates similar filtering patterns already work in the ecosystem.
 For example, a package can include assets such as `empty.txt` in both `contentFiles/cs/net9.0/` and `contentFiles/vb/net9.0/`, mapped via the `packagepath` property.
 
 To validate this behavior, a test was performed using a .NET 9 project that packed `empty.txt` into both C# and VB.NET content file folders. 
 The resulting NuGet package contained the file in both locations. When referenced by a C# project, only the C# content file was selected by the SDK; the VB.NET file was ignored.
 This was confirmed by inspecting the `project.assets.json`, which included both files in the package metadata but only used the language-appropriate asset.
-
-### Asset Filtering
-Existing NuGet asset filtering (ExcludeAssets, PrivateAssets) works for compile, runtime, and other assets groups, providing a good model for how analyzer filtering should work
 
 ## Unresolved Questions
 
--- a/accepted/2025/indicate-analyzer-assets.md
+++ b/accepted/2025/indicate-analyzer-assets.md
@@ -6,8 +6,10 @@
 
 ## Summary
 
-`project.assets.json` does not currently indicate which analyzers are active, and analyzers are added even when `PrivateAssets` or `ExcludeAssets` should prevent this.
-This leads to analyzers being included in projects, which is not the expected behavior, especially when defaults for private/excluded assets are not respected
+`project.assets.json` does not currently record which analyzers a package contributes.
+Because that information is missing, the .NET SDK has to work around the gap by scanning each package's files to discover analyzers itself.
+That workaround has no knowledge of asset filtering, so analyzers are applied to the project's compilation even when `PrivateAssets` or `ExcludeAssets` should exclude them.
+Surfacing analyzer assets in `project.assets.json` closes this gap and gives the SDK the information it needs to decide which analyzers to apply.
 
 ## Motivation
 
@@ -33,15 +35,22 @@ This is done to avoid a breaking change.
 
 #### File Structure
 
-All `.dll` files under `analyzers/` are tracked, at any depth, excluding satellite `.resources.dll` assemblies. This mirrors the SDK's analyzer discovery, so the assets file does not assume a fixed folder layout.
+All `.dll` files under `analyzers/` are tracked, at any depth, excluding satellite `.resources.dll` assemblies.
+This mirrors the SDK's analyzer discovery, so the assets file does not assume a fixed folder layout.
 
-|Pattern|Description|
+NuGet decides whether a file is an analyzer purely from the `analyzers/` prefix and the `.dll` extension.
+The remaining path segments are not used to decide whether a file is tracked, so every matching analyzer is listed regardless of its language or compiler version.
+NuGet inspects the optional language and compiler-version segments only to derive selection metadata:
+
+|Path segment|Metadata derived|
 |---|---|
-|`analyzers/**/{name}.dll`|Any analyzer assembly, at any depth under `analyzers/`|
-|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzer (`cs`, `vb`, `fs`)|
-|`analyzers/dotnet/{compilerApiVersion}/{language}/{name}.dll`|Analyzer for a specific compiler API version (`roslynX.Y`)|
+|A `cs`, `vb`, or `fs` segment (e.g. `analyzers/dotnet/cs/{name}.dll`)|`codeLanguage`; defaults to `any` when no language segment is present|
+|A `roslynX.Y` segment (e.g. `analyzers/dotnet/roslyn4.0/cs/{name}.dll`)|`compilerApiVersion`; omitted when no such segment is present|
 
-The `{language}` and `{compilerApiVersion}` segments are optional and may appear at any depth. NuGet derives `codeLanguage` and `compilerApiVersion` metadata from them (see the example below) so the SDK can select the applicable analyzers, the same way `codeLanguage` is used for content files. Analyzers with no language segment get a `codeLanguage` of `any`, and `compilerApiVersion` is omitted when the path has no `roslynX.Y` segment. Excluded analyzers (for example via `PrivateAssets`) are written as a bare `analyzers/.../_._` placeholder with no metadata.
+The `{language}` and `{compilerApiVersion}` segments are optional and may appear at any depth.
+The SDK uses this metadata to select which analyzers to apply, the same way `codeLanguage` is used for content files.
+Because all analyzers are listed, an `analyzers/dotnet/fs/MyAnalyzer.dll` entry still appears for a C# project, but the SDK does not apply it since its `codeLanguage` is `fs`.
+Excluded analyzers (for example via `PrivateAssets` or `ExcludeAssets`) are written as a `analyzers/.../_._` placeholder with no metadata.
 
 Examples: 
 - `analyzers/dotnet/cs/MyAnalyzer.dll` (C# analyzer)
@@ -85,13 +94,6 @@ Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
   }
 }
 ``` 
-
-#### Testing
-
-Validate correct behavior with:
-- Analyzer packages with and without exclusion flags.
-- Multi-targeted packages (ensure correct analyzer/TFM pairing).
-- Projects with the feature flag enabled, confirming that included/excluded analyzers match expectations.
 
 ## Private Assets and Transitive Behavior
 
@@ -139,7 +141,16 @@ Validate correct behavior with:
 <PackageReference Include="LibraryWithAnalyzers" Version="1.0.0" 
                   ExcludeAssets="analyzers" />
 ```
-Result: App's project.assets.json will not contain analyzer entries for LibraryWithAnalyzers
+**App's project.assets.json:**
+```json
+"LibraryWithAnalyzers/1.0.0": {
+  "type": "package",
+  "analyzers": {
+    "analyzers/dotnet/cs/_._": {}
+  }
+}
+```
+The excluded analyzers are written as a `_._` placeholder, consistent with how other excluded asset groups (such as `compile` and `runtime`) are represented.
 
 ### How Transitivity Works
 
--- a/accepted/2025/indicate-analyzer-assets.md
+++ b/accepted/2025/indicate-analyzer-assets.md
@@ -27,7 +27,9 @@ Analyzer assets in NuGet packages will now respect asset filtering options like 
 - Configurations will accurately control which analyzers are present, eliminating unexpected analyzer warnings or errors.
 - Both package authors and consumers will have clear, reliable control over analyzer inclusion without needing custom workarounds.
 
-This feature will initially be behind a feature flag to avoid doing a breaking change.
+This feature will initially be behind a feature flag defined as `<EnableIndicateAnalyzerAsset>`. 
+This property should be set as true in the project file. 
+This is done to avoid a breaking change.
 
 ### Technical explanation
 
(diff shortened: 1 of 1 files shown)


accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+## Explanation
+
+### Functional explanation
+
+Analyzer assets in NuGet packages will now respect asset filtering options like ExcludeAssets and PrivateAssets, just as other asset types do.
+- When excluding analyzers from a package, they will no longer be included or executed during the project's build.
+- Configurations will accurately control which analyzers are present, eliminating unexpected analyzer warnings or errors.
+- Both package authors and consumers will have clear, reliable control over analyzer inclusion without needing custom workarounds.
+
+This feature will initially be behind a feature flag to avoid doing a breaking change.
```
nkolev92: nit: this description should ideally call out what the feature flag is.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+```
+
+#### SDK Changes
+
+Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
+- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
+- Only load analyzers are listed in this group.
+- If the "analyzers" group is absent (e.g., from older lock files), fall back to legacy scanning to preserve compatibility
```
nkolev92: We should probably "teach" the SDK how to skip the scanning when the feature is enabled but it just let to no analyzers being selected since analyzers don't exist in any of the packages.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+- Filter out analyzers that are excluded by PrivateAssets/ExcludeAssets
+- Add an "analyzers" group under the appropriate "targets" node for each package in project.assets.json.
+
+Example Output
+```json
+"targets": {
+  ".NETCoreApp,Version=v8.0": {
+    "My.Analyzer.Package/1.0.0": {
+      "analyzers": [
+        "analyzers/dotnet/cs/MyAnalyzer.dll"
```
nkolev92: I'd love to understand the format of this string. 
Are subfolders supported, should they be? 

Basically, each part of the folder, enumerate whether it's static or variable. If it's variable what's supported and how the matching is done. Finally, whether we have the data during restore already. 

Some examples is this: 

https://github.com/NuGet/NuGet.Client/blob/495cb33e3915e16f09f5d42b0b1b312c5f5ce44a/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs#L497

In this case, i'm assuming
- analyzers - static
- dotnet - static (albeit maybe a bit redudant, but if that's what the guidelines for creating analyzers say good enough)
- cs - language

Would subfolders be supported?

`analyzers/dotnet/cs/stuff/MyAnalyzer.dll` does it match? 

Looking at some analyzers, it seems like a bunch have resources, so those can probably be selected the same way we do resources for the lib folder. 
In  the lib folder, we don't select subfolders (resources are an add on and must match the assembly name in the main folder).

There are no framework parameters for analyzers correct?

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
+
+- Existing NuGet asset filtering (ExcludeAssets, PrivateAssets) works for compile, runtime, and other assets groups, providing a good model for how analyzer filtering should work
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
```
nkolev92, line 247: I think we have things like the format, and then we can talk roll-out strategy.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
```
nkolev92, line 251: Some things I'd love to see is an analysis on analyzers as dependencies and whether Include/Exclude assets are selected and basically a way to figure out whether there are packages that would be affected by enabling this feature and how many. 

This shouldn't block this spec, the spec should focus on the analyzer mechanics, not the roll-out, but it makes it more complete by adding it in future possibilites and possibly explicitly calling out that it'd be an analysis that drives the default roll-out decisions. 

It is also something we can use to affirm that our design is the right one.

nkolev92 (commented):
Apologies for the delay, great first write-up. 
I added some comments about some specifics I'd love to see + some ideas for future possibilities.

accepted/2025/indicate-analyzer-assets.md
zivkan: `contentFiles` supports code language as well: https://github.com/NuGet/NuGet.Client/blob/495cb33e3915e16f09f5d42b0b1b312c5f5ce44a/src/NuGet.Core/NuGet.Packaging/ContentModel/ManagedCodeConventions.cs#L592

This lets a package contain cs, vb, fs source code in the package. I'd like to know how the assets file represents these, and choose only the language suitable for the current project.

Similarly, say an package contains analuzers for at least two languages. How will the assets file represent them?

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Rationale and alternatives
+
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
+
+- Existing NuGet asset filtering (ExcludeAssets, PrivateAssets) works for compile, runtime, and other assets groups, providing a good model for how analyzer filtering should work
```
zivkan: As mentioned earlier, contentFiles already supports programming language specific files. It should be considered prior art given that analyzers appear to have language specific subdirectories.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,93 @@
...
+- Analyzer packages with and without exclusion flags.
+- Multi-targeted packages (ensure correct analyzer/TFM pairing).
+- Projects with the feature flag enabled, confirming that included/excluded analyzers match expectations.
+
+## Drawbacks
+
+- Requires changes to both NuGet restore logic and .NET SDK asset resolution, which may increase maintenance complexity and introduce risk of regression.
+- May require updates to existing tooling, documentation, or custom scripts that expect the current analyzer handling behavior.
+- Potential for minor performance impact during restore and build, as asset filtering and lock file generation become more granular for analyzers.
+- Older packages or tools that do not adopt the new pattern may continue to exhibit legacy behavior, requiring a fallback or compatibility layer that adds to system complexity.
```
zivkan: I guess this falls under the general umbrella of "older packages", but one of my biggest concerns is that Nuget has been putting analyzers as private assets by default since the beginning: https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

This dramatically increases the likelihood hood that there will be packages unknowingly having dependencies to analyzer packages where the nuspec says `exclude="analyzers;build"`. Therefore, this feature will break these packages.

My memory is that include/excludes flow transitively as well, which adds more risk.  For example, say my package has a dependency on Microsoft.VisualStudio.Threading and I use the default `PrivateAssets`, which therefore means `exclude="analyzers;build"`.  Even though Microsoft.VisualStudio.Threading's dependency on Microsoft.VisualStudio.Threading.Analyzers has `Include="all"`, does the previous exclude cause the analyzers to be excluded?   If so, breaking changes are pretty much inevitable.

Something I just tried to assess risk is created a new project with `dotnet new xunit`, edit the csproj to add `ExcludeAssets="compile"` to the xunit reference, and tried to build. It failed even though the xunit package doesn't contain any assets, it just contains dependencies to other packages. But I don't know if ExcludeAssets on the PackageReference work differently to exclude assets in package nuspecs.

I think it would be good to validate if package exclude assets flows transitively via a test using compile assets and for this spec to acknowledge the risk and what the outcome of the test is.

accepted/2025/indicate-analyzer-assets.md
zivkan: > This shouldn't block this spec, the spec should focus on the analyzer mechanics, not the roll-out,

Unfortunately, I'm a bit more pessimistic. I just wrote another comment which I'm too lazy to edit/copy, but I'm concerned that the current design will have large number of breaking changes due to PrivateAssets instructing analyzers to be transitively excluded by default.

For example, if it will cause the VSSDK to remove all analyzers from projects until relevant pacakges are updated.

If so, then we might need to change the design (slightly), and therefore it will impact this spec.

nkolev92: > For example, if it will cause the VSSDK to remove all analyzers from projects until relevant pacakges are updated.

Many MS packages are authored with include all, SDK in particular.

If you go to https://www.nuget.org/packages/Microsoft.VisualStudio.SDK, you'll see 
>         <dependency id="Microsoft.VisualStudio.SDK.Analyzers" version="17.7.79" include="All" />

You are right though, it's possible that analysis tells us we need to change the design. 
In theory, we might need to add a new assets group for example or something like that. 

I don't think it changes the folder though.

accepted/2025/indicate-analyzer-assets.md
nkolev92: > This dramatically increases the likelihood hood that there will be packages unknowingly having dependencies to analyzer packages where the nuspec says exclude="analyzers;build". Therefore, this feature will break these packages.

Analyzers tend to be treated as deve dependencies, so them not flowing is I'd argue the preferred experience.

https://www.nuget.org/packages/Microsoft.VisualStudio.Threading.Analyzers#readme-body-tab
https://www.nuget.org/packages/Microsoft.CodeAnalysis.Analyzers/5.0.0-1.25277.114

Packages that want to redistribute analyzers do it explicitly.

Dev dependencies by design are not supposed to redistribute. 

Note that I'm not saying we should be disrupting builds regardless, but the fact that include/exclude assets is respected may be preferred by customers. 

Fortunately, these are all things we can detect.

accepted/2025/indicate-analyzer-assets.md
kfikadu: > Would subfolders be supported?

Subfolders aren't currently supported. Only the analyzer assemblies that are placed inside `analyzers/dotnet/cs` would be recognized and selected. 


> `analyzers/dotnet/cs/stuff/MyAnalyzer.dll` does it match?

No, `analyzers/dotnet/cs/stuff/MyAnalyzer.dll` would not be matched. Only assemblies that are located at `analyzers/dotnet/cs/MyAnalyzer.dll` would be matched. 


> There are no framework parameters for analyzers correct?

Thats right, analyzers don't have specific framework parameters.

kfikadu: > This lets a package contain cs, vb, fs source code in the package. I'd like to know how the assets file represents these, and choose only the language suitable for the current project.
>
> Similarly, say an package contains analuzers for at least two languages. How will the assets file represent them?

From my understanding, the analyzers would be placed into language-specific subfolders, such as `analyzers/dotnet/cs/MyAnalyzer.dll` and `analyzers/dotne/cs/MyAnalyzer.dll`. In the `project.assets.json` file, it would look like this: 
```json
"analyzers": [
  {
    "type": "analyzer",
    "path": "analyzers/dotnet/cs/MyAnalyzer.dll"
  },
  {
    "type": "analyzer",
    "path": "analyzers/dotnet/vb/MyAnalyzer.dll"
  }
]
```

accepted/2025/indicate-analyzer-assets.md
kfikadu: I agree, before enabling this feature by default we should have analysis/telemetry on how the analyzers as dependencies interact with Private/Include/Exclude Assets, and how many packages would be affected

accepted/2025/indicate-analyzer-assets.md
zivkan: How is the .net sdk supposed to know which of the the current project is supposed to use?  use the `/cs/` analyzer for C# projects, and `/vb/` for VB.NET projects?   ContentFiles must have solved this already, so the same pattern should be used for analyzers.

kfikadu: Ahh, I see what you're asking. From what I was able to find, during the restore process ContentFIles uses the project file itself to determine what language to use, for example a csproj is going to use `/cs/` and vbproj is going to use `/vb/`. It uses the layout `contetFiles/{language}/{tfm}/...`, and you are right, the same pattern is going to be used for analyzers, just a bit simpler since there aren't any target frameworks and no subfolders, so the pattern would be `analyzers/dotnet/{language}/<analyzer.dll>`. 

I do agree that it should be added into the prior art, since it is a good foundation to build upon.

accepted/2025/indicate-analyzer-assets.md
kfikadu: I did some tests to get some more clarity, and what I've found is:
- Analyzers are treated as private assets by default, even if someone were to use `IncludeAssets` or `ExcludeAssets`, those settings wouldn't flow transitively. They only apply to the specific reference where they are set. 
- Compile assets on the other hand, are transitive by default, so unless they are explicitly excluded, they are going to flow through the dependency chain.

I'll add my test results and make sure the risks are reflected in the spec.

nkolev92: > Analyzers are treated as private assets by default, even if someone were to use IncludeAssets or ExcludeAssets, those settings wouldn't flow transitively. They only apply to the specific reference where they are set.

Just to make sure I understand your point, the current .NET SDK powered implementation of analyzers treats as them PrivateAssets=all?

kfikadu: Yes, at least that is my current understanding of them

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+## Rationale and alternatives
+
+This design is chosen to align with existing NuGet asset filtering patterns while providing a clear, structured way to manage analyzers.
+By introducing a new "analyzers" group in `project.assets.json`, it allows for consistent handling of analyzers similar to other asset types, while respecting existing filtering options like `PrivateAssets` and `ExcludeAssets`.
+It also maintains backward compatibility by being opt-in via a feature flag, allowing developers to adopt it gradually without breaking existing projects.
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
```
kartheekp-ms, line 216: I’m not sure if this is the right section for it, but there’s some prior art on [`DevelopmentDependency`](https://github.com/NuGet/Home/wiki/DevelopmentDependency-support-for-PackageReference) support in NuGet.

In simple terms, **DevelopmentDependency** means “this dependency is only used when building the package that declares it, and should not flow to downstream consumers.”

For example:

**Case 1 – Direct dependency**
```text
Package A
└─ Package B
   └─ Analyzers (DevelopmentDependency)
```
**Effect:** The analyzers run for **B** during its build, but **A** is **not** affected.

**Case 2 – Transitive dependency**
```text
Package A
└─ Package C
   └─ Package B
      └─ Analyzers (DevelopmentDependency)
```
**Effect:** Even when going through **C**, analyzers still don’t affect **A**.

**Case 3 – Without DevelopmentDependency (for contrast)**
```text
Package A
└─ Package B
   └─ Analyzers (regular dependency)
```
**Effect:** The analyzers currently flows transitively and impact **A** and there is no way to exclude analyzers from transitive dependencies.

**Takeaway:** If **B** marks its analyzers as a development dependency, then **A** (directly or indirectly) won’t get them.

kfikadu: This is a great find, this also explains transitivity a bit better than I was able to. Thanks!

donnie-msft (dismissed):
Looks good! 
Wondering what you think about renaming `EnableIndicateAnalyzerAssets` to be more concise, maybe `IndicateAnalyzerAssets`?

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+- Developers can't control analyzer inclusion in their projects, leading to unexpected warnings or errors.
+- Users are required to use custom MSBuild scripts or abandon packages due to inability to control analyzer inclusion.
+- Package authors can't reliably control analyzer distribution
+
+## Explanation
+
+### Functional explanation
+
+Enable analyzer assets to respect asset filtering options like other asset types.
+When enabled via `<EnableIndicateAnalyzerAssets>true</EnableIndicateAnalyzerAssets>`:
```
nkolev92: Normally, we would prefix the property names with nuget or restore whichever is more appropriate. 

In this case, I think restore fits better. 
See examples in the property names in https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Build.Tasks/NuGet.targets. 

A proposal would be; 
`RestoreEnableAnalyzerAssets`

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+- Analyzers will be tracked in the `project.assets.json` file under a new "analyzers" group.
+- PrivateAssets/ExcludeAssets will correctly filter analyzers, preventing them from being included in projects that should not have them.
+- Developers will gain proper control over analyzer inclusion
+
+This feature will initially be behind a feature flag defined as `<EnableIndicateAnalyzerAssets>`.
```
nkolev92: This is something we want to have enabled globally at some point. I'd add that in the future opportunities section.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+
+### Technical explanation
+
+#### File Structure
+
+|Pattern|Description|Example|
+|---|---|---|
+|`analyzers/dotnet/cs/{name}.dll`|C# analyzers|`analyzers/dotnet/cs/MyAnalyzer.dll`|
+|`analyzers/dotnet/vb/{name}.dll`|VB.NET analyzers|`analyzers/dotnet/vb/MyAnalyzer.dll`|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzers|`analyzers/dotnet/{language}/MyAnalyzer.dll`|
```
nkolev92: nit; this is all you need, then you can call the other 2 above as examples.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+- Respect `PrivateAssets` and `ExcludeAssets` when populating this group.
+
+**SDK Changes**
+
+Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
+- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
+- Only load analyzers that are listed in this group.
+- When `<EnableIndicateAnalyzerAssets>` is set to true, the SDK will only use the analyzer group that is in the assets file to determine which analyzers to include. 
+- If the analyzers group is missing from the assets file and the feature flag is enabled, the SDK won't fall back to legacy scanning. 
+  - In this case, no analyzers will be selected, even if analyzer assets exist in the packages.
```
nkolev92: > even if analyzer assets exist in the packages

This can be misleading. The only case in which these 2 conditions will be true if the analyzers are only available for another language.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+- If the feature flag is not set or is false, the SDK will use the legacy scanning behavior to discover analyzers and preserve compatibility.
```
nkolev92, line 73: Makes sense. 
We should explicitly set the property in the NuGet.targets

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+  "analyzers": {
+    "analyzers/dotnet/cs/MyAnalyzer.dll": {}
+  }
+}
+```
+**App's project.assets.json (references Library):**
+```json
+"MyAnalyzer/1.0.0": {
+  "type": "package"
+  // No "analyzers" section - blocked by PrivateAssets
```
nkolev92: I think we often indicate this with `_._`, but I think the code will lead you into the right solution, so not too worried about how exact the spec is.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+Validate correct behavior with:
+- Analyzer packages with and without exclusion flags.
+- Multi-targeted packages (ensure correct analyzer/TFM pairing).
+- Projects with the feature flag enabled, confirming that included/excluded analyzers match expectations.
+
+## Private Assets and Transitive Behavior
+
+### Current Behavior (Without Feature Flag)
+
+| Attribute | What It Should Do | What Actually Happens | project.assets.json |
```
nkolev92: nit: you don't need the last column, since it';s all the same.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+```
+Result: App's project.assets.json will not contain analyzer entries for LibraryWithAnalyzers
+
+### How Transitivity Works
+
+```
+Project A -> Project B -> Package C (with analyzers)
+```
+
+- **Without PrivateAssets:** Analyzers flow from C to B to A
```
nkolev92, line 161: Worth calling out that the default behavior of Analyzers is `PrivateAssets=build;contentfiles;analyzers`

To be honest, I'm not certain that should be the case.
Something for us to consider for later as we're implementing this feature.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Breaking Changes
+
+### No Breaking Changes Initially 
+
+This feature is opt-in via `<EnableIndicateAnalyzerAssets>true</EnableIndicateAnalyzerAssets>`, so existing projects will not be affected until they explicitly enable it.
+
+### When Enabled by Default
+
+Projects using `PrivateAssets="analyzers"` or `ExcludeAssets="analyzers"` will start working as intended:
```
nkolev92: We should probably enable it for the latest TFM proejcts, like we've enabled a bunch of other features. 

We want the default to be true for all eventually, so it makes sense to have it default to true for new projects.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Build Behavior Changes:**
+- Missing analyzer diagnostics in dependent projects
+- Different warning/error counts in build output
+- TreatWarningsAsErrors outcomes may change
+
+**CI/CD Pipeline Impact:**
+- Builds expecting certain analyzers may fail
```
nkolev92, line 182: Appreciate how thorough this is. 

I think this scenario is probably not very likely.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+- Builds may pass that previously failed on analyzer warnings
```
nkolev92, line 183: This could be a problem I agree.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Unresolved Questions
+
+<!-- What parts of the proposal do you expect to resolve before this gets accepted? -->
+<!-- What parts of the proposal need to be resolved before the proposal is stabilized? -->
+<!-- What related issues would you consider out of scope for this proposal but can be addressed in the future? -->
+
+## Future Possibilities
+
+Something that would be valuable would be an analysis of NuGet packages that have analyzers, and focusing on analyzers as dependencies.
```
nkolev92, line 255: We should move ahead with the implementation defining the folder structure, the filtering etc, but in order to be confident in how it's enabled and the default, we'll need to do this analysis.

nkolev92 (dismissed):
This is looking great. 

I may have left some feedback on the spec, but general direction seems fine. 
Please go through the feedback and lmk if you have any questions. 

Even if this spec is not fully merged yet, I think you can start working on the implementation

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Explanation
+
+### Functional explanation
+
+Enable analyzer assets to respect asset filtering options like other asset types.
+When enabled via `<EnableIndicateAnalyzerAssets>true</EnableIndicateAnalyzerAssets>`:
+- Analyzers will be tracked in the `project.assets.json` file under a new "analyzers" group.
+- PrivateAssets/ExcludeAssets will correctly filter analyzers, preventing them from being included in projects that should not have them.
+- Developers will gain proper control over analyzer inclusion
```
jeffkl: This line is redundant with the one above it, I would just remove it.

jeffkl (commented):
My vote for the property name is `RestoreEnableAnalyzerIncludeExcludeAssets`

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
+- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
+- Only load analyzers that are listed in this group.
+- When `<EnableIndicateAnalyzerAssets>` is set to true, the SDK will only use the analyzer group that is in the assets file to determine which analyzers to include. 
+- If the analyzers group is missing from the assets file and the feature flag is enabled, the SDK won't fall back to legacy scanning. 
+  - In this case, no analyzers will be selected, even if analyzer assets exist in the packages.
+- If the feature flag is not set or is false, the SDK will use the legacy scanning behavior to discover analyzers and preserve compatibility. 
+
+**Example Output**
+```json
```
ericstj, line 76: Should we also version the assets file format?
```diff
- "version": 3,
+ "version": 4,
```

It's possible that this is an easier thing for the SDK to use when deciding to raise analyzers itself.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+      "compile": { ... },
+      "runtime": { ... },
+      "analyzers": {
+        "analyzers/dotnet/cs/MyAnalyzer.dll": {}
+      }
+    }
+  }
+}
+``` 
+
```
ericstj, line 97: I don't think this happens today, but we might want to talk about if there are any changes to the deps file, or to runtime compilation scenarios that consume nuget packages - like interactive.  Could also be out of scope or handled in separate doc.

accepted/2025/indicate-analyzer-assets.md
ericstj: Keep in mind that analyzers are also source generators.  Folks can definitely depend on source generators in a way that the build will fail without them.  I think the most likely failure would be someone who depended on a transitive analyzer and that's no longer applying due to respecting *Assets on reference.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,215 @@
...
+|`analyzers/dotnet/cs/{name}.dll`|C# analyzers|`analyzers/dotnet/cs/MyAnalyzer.dll`|
+|`analyzers/dotnet/vb/{name}.dll`|VB.NET analyzers|`analyzers/dotnet/vb/MyAnalyzer.dll`|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzers|`analyzers/dotnet/{language}/MyAnalyzer.dll`|
+
+**NuGet Changes**
+
+- Add "analyzers" group to project.assets.json during restore.
+- Respect `PrivateAssets` and `ExcludeAssets` when populating this group.
+
+**SDK Changes**
```
ericstj, line 65: I'd also like to include follow up work in component governance.  Package which exclusively contribute Compile assets are treated as development dependencies, however we had to special case analyzers, like the SDK.  We should remove that special case when this mode is enabled.  
https://github.com/microsoft/component-detection/blob/e53a7eb2b31cb03202e15d8ed1755bdf97cb952c/src/Microsoft.ComponentDetection.Detectors/nuget/NuGetProjectModelProjectCentricComponentDetector.cs#L77-L82

accepted/2025/indicate-analyzer-assets.md
ericstj: Note that `DevelopmentDependency` is a specific thing in addition to *Assets and transitivity.  *Assets allow the referencer to decide what to do with a reference.  `DeveloementDependency` allows the package itself to decide what happens when it is referenced.

I've never liked NuGet's definition of `DevelopmentDependency` because its actually different than other ecosystems, and it is even counter-intuitive in that it will include runtime (but not compile) assets in it's directly referenced project.

accepted/2025/indicate-analyzer-assets.md
ericstj, line 255: Speaking of analyzers and dependencies ... today analyzers with dependencies (and tasks with dependencies) redistribute (!!!) them.  This creates a blind spot for package audit/inventory systems and makes the dependencies non-serviceable by the referencer.  An alternative design would be to have a "deisgn time" package graph where a  set of packages can be resolved for execution as an analyzer or task.  This is a major undertaking but it is more correct than the hack that is used for tasks and analyzers today.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,223 @@
...
+```json
+"version": 4,
+"targets": {
+  ".NETCoreApp,Version=v8.0": {
+    "My.Analyzer.Package/1.0.0": {
+      "type": "package",
+      "compile": { ... },
+      "runtime": { ... },
+      "analyzers": {
+        "analyzers/dotnet/cs/MyAnalyzer.dll": {}
```
zivkan: I think I commented on this elsewhere, but this example is missing the information on how the SDK is supposed to know which analyzers is relevant to the current project. Say a package contains analyzers for multiple programming languages, how does the SDK know which one to use?

My memory is I did a screen sharing session with you, and we found that `contentFiles/` puts some item metadata, something like:

```json
"contentFiles": {
   "contentFiles/cs/any/whatever.txt": {
     "lang": "c#"
  }
}
```

I think this is an important part of the feature that needs to be explicitly called out in the spec. Unless I missed it, it seems to me that the spec doesn't currently explain how multi-language packages have their correct assets selected for the current project.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,223 @@
...
+<!-- Why is this the best design compared to other designs? -->
+<!-- What other designs have been considered and why weren't they chosen? -->
+<!-- What is the impact of not doing this? -->
+
+## Prior Art
+
+NuGet's existing asset filtering system (compile, runtime assets) provides the model.
+
+Language-specific content selection (`contentFiles/{language}/`) demonstrates similar filtering patterns already work in the ecosystem.
+For example, a package can include assets such as `empty.txt` in both `contentFiles/cs/net9.0/` and `contentFiles/vb/net9.0/`, mapped via the `packagepath` property.
```
zivkan: How `contentFiles/` filters out assets (languages) that are not relevant for the current project is a key part of the prior art, in my opinion.  How is it represented in the assets file?

accepted/2025/indicate-analyzer-assets.md
kfikadu: Yes! I tried to summarize it a bit as the spec was getting long, but I can definitely add in more detail and explain it better

accepted/2025/indicate-analyzer-assets.md
kfikadu: After some conversation, we came to the conclusion that if a package is only providing analyzers, then it should not affect the deps.json file

For cases like interactive, do you mean C# interactive?

accepted/2025/indicate-analyzer-assets.md
kfikadu: This is a great idea, I can add this into the future possibilities as something we can do.

accepted/2025/indicate-analyzer-assets.md
nkolev92: Great, point, we should try to account for source gens in our nuget.org data anylasis.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -215,6 +233,12 @@ Ideally, this analysis would:
 - Provide data on how many packages (and consumers) would be impacted by a change in default behavior, helping to mitigate surprises or unintended consequences.
 - Affirm the correctness and robustness of the proposed design prior to broad deployment.
 
+Also, provide analysis on source generators as analyzers. Since it is possible for people to depend on source generators in a way that their build will fail without them, there should be data on which ones may fail. 
+
+Additional follow up should also be done in component governance. Currently, packages that exclusively contribute to `Compile` assets are treated as development dependencies.
```
aortiz-msft: Component Governance should be capitalized.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+| `ExcludeAssets="analyzers"` | Exclude analyzers from this project entirely | Analyzers still included |
+| `IncludeAssets="compile;runtime"` | Only include compile and runtime assets (exclude analyzers) | **Analyzers still included** - not controlled by asset filtering |
+
+### New Behavior (With RestoreEnableAnalyzerAssets=true)
+
+**Scenario 1: Library with PrivateAssets**
+```xml
+<!-- Library.csproj -->
+<PackageReference Include="MyAnalyzer" Version="1.0.0" 
+                  PrivateAssets="analyzers" />
```
aortiz-msft: This is the default value, so adding a comment <!-- default value> would be helpful.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+**Mitigation:**
+```xml
+<PropertyGroup>
+  <RestoreEnableAnalyzerAssets>false</RestoreEnableAnalyzerAssets>
+</PropertyGroup>
+```
+
+### Rollout Strategy
+
+1. **Current:** Opt-in via feature flag
```
aortiz-msft: Earlier in the doc it is mentioned that this feature will be on by default for projects targeting the new TFM. We should incorporate that into this section.

aortiz-msft: Proposal: Opt-in only in .NET 10. For .NET 11, it is opt-in for all frameworks and it is on by default for projects targeting .NET 11.

aortiz-msft (changes requested):

aortiz-msft (approved):

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+- Developers can't control analyzer inclusion in their projects, leading to unexpected warnings or errors.
+- Users are required to use custom MSBuild scripts or abandon packages due to inability to control analyzer inclusion.
+- Package authors can't reliably control analyzer distribution
+
+## Explanation
+
+### Functional explanation
+
+Enable analyzer assets to respect asset filtering options like other asset types.
+When enabled via `<RestoreEnableAnalyzerAssets>true</RestoreEnableAnalyzerAssets>`:
```
nkolev92, line 26: We might need to *or* the valuies of this property across ~projects~ tfms. 

Basically, if one of them is true, then it's enabled, but if they're all empty or false, then it's disabled.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+This property should be set to true in the project file. 
+This is done to avoid a breaking change.
+
+### Technical explanation
+
+#### File Structure
+
+|Pattern|Description|
+|---|---|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzers|
```
nkolev92: Is the name actually used for matching?

Basically the lib folder says use a *.dll, but the build folder requires the package id.

I think for analyzers it makes sense to just include all, but double checking.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+
+- Add "analyzers" group to project.assets.json during restore.
+- Respect `PrivateAssets` and `ExcludeAssets` when populating this group.
+
+**SDK Changes**
+
+Update SDK asset resolution logic (e.g., ResolvePackageAssets.cs) to:
+- Read analyzer assets directly from the "analyzers" group in "targets", instead of scanning all files in "libraries".
+- Only load analyzers that are listed in this group.
+- When `<RestoreEnableAnalyzerAssets>` is set to true, the SDK will only use the analyzer group that is in the assets file to determine which analyzers to include.
```
nkolev92, line 71: Given that we'll want to or this value, we should probably express that in the assets file, but we can talk about it in the implementation, no need to address it at this point.

nkolev92 (dismissed):

accepted/2025/indicate-analyzer-assets.md
ericstj: Yeah - C# interactive was just an example.  Jupyter notebooks is another.  I was trying to think if there was anything else that tried to resolve packages or consume the project.assets.json that might need to change.  I don't expect the design to change as a result of those, but it might be good to capture a list of things that might change when project.assets.json changes and inform those folks as part of this, so that they might make changes to react if necessary.

accepted/2025/indicate-analyzer-assets.md
ericstj: I see you incremented the version.  It would be good to call that out explicitly.  I'm not sure what changing the version entails WRT to clients reading the assets file.  Presumably NuGet team knows more about this since it's already at 3 😄.  I suspect it's more than just update the number.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### File Structure
+
+|Pattern|Description|
+|---|---|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzers|
+
+Examples: 
+- `analyzers/dotnet/cs/MyAnalyzer.dll` (C# analyzer)
+- `analyzers/dotnet/vb/MyAnalyzer.dll` (VB.NET analyzer)
```
baronfel, line 57: (some of this is feedback about the existing analyzers selection)

This path doesn't mention TargetFramework at all - my intuition is that analyzers are specific to 
* TFM
* language
* compiler name (and optionally version)

as well. Does this need to be modeled here? And how does _selection_ of analyzers assets get impacted by TFM? cc @ericstj / @jaredpar for thoughts.

accepted/2025/indicate-analyzer-assets.md
baronfel: I'm curious about the 'or'-ing - project.assets.json are per-project, so why would a per-project flag not be enough to satisfy the detection criteria? It seems like Restore itself can know across all of the projects if it needs to track analyzer data or not, but can know per-project if it needs to emit the analyzer data to the assets file.

accepted/2025/indicate-analyzer-assets.md
baronfel: agree - it the shape of the path is well-known, it would help all downstream consumers if the variation points were populated by the assets file (see my above comment about the current pivots).

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+<PropertyGroup>
+  <RestoreEnableAnalyzerAssets>false</RestoreEnableAnalyzerAssets>
+</PropertyGroup>
+```
+
+### Rollout Strategy
+
+The current rollout will be to opt-in for .NET 10.
+For .NET 11, it is going to be opt-in for all frameworks and enabled by default for projects that target .NET 11.
+The last step will be to remove the feature flag.
```
baronfel, line 197: At what point would the "last step" happen? A future SDK version?

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,247 @@
...
+For .NET 11, it is going to be opt-in for all frameworks and enabled by default for projects that target .NET 11.
+The last step will be to remove the feature flag. 
+
+
+## Drawbacks
+
+- Requires coordinated changes to NuGet and SDK
+- Breaking change when enabled by default
+- Potential minor performance impact during restore
+- Older tools may not understand the new project.assets.json format
```
baronfel, line 205: it would be nice to investigate this a bit further - are older versions of the nuget libraries around parsing the assets file resilient to additive changes to the json content? if not, it might be good to make them forward-compatible in this way so that future changes aren't breaking in this way.

accepted/2025/indicate-analyzer-assets.md
kfikadu: The name itself isn't used for the matching; I put it there as a placeholder for the actual name of the analyzers.

accepted/2025/indicate-analyzer-assets.md
kfikadu: As of right now, we want to keep it as a feature flag for .NET 11, and we are considering enabling it as a default for .NET 12.

accepted/2025/indicate-analyzer-assets.md
ericstj: Here's a reference https://learn.microsoft.com/en-us/nuget/guides/analyzers-conventions#analyzers-path-format I

Framework selection is not all that useful, and I suspect might lead to loading issues since Roslyn will prefer a cached assembly of the same name (so folks would need to be sure to name / version each analyzer assembly differently). 
We recommend that analyzers and source-generators just test for supporting API in order to have behavior that differs on different frameworks.
Still the docs imply that at one point we imagined having a framework selection even though we don't use it.  It's worth revisiting that now that we're changing how this works.

We do have compiler API version and make use of it today -- https://github.com/dotnet/sdk/blob/cdf3a40287e5e21a09cb44a6214e3efa8e032e66/src/Tasks/Microsoft.NET.Build.Tasks/ResolvePackageAssets.cs#L1078

accepted/2025/indicate-analyzer-assets.md
donnie-msft: Do we have a naming pattern to indicate a `Global` setting? I'm still not sure reading this setting's name how customers will know it applies to all projects once enabled in any project.

nkolev92: There's no concept of global, I meant to write across TFMs.

Basically similar to what we do NuGetAudit and package pruning.

accepted/2025/indicate-analyzer-assets.md
nkolev92: > per-project flag not be enough to satisfy the detection criteria?

The spec says we want to enable it for the new projects based on the tfms targeted, so we need to `or` it for the multi targeted projects.

accepted/2025/indicate-analyzer-assets.md
nkolev92: I'd just add *.dll in this case.

accepted/2025/indicate-analyzer-assets.md
jaredpar: > We recommend that analyzers and source-generators just test for supporting API in order to have behavior that differs on different frameworks.

Agree. Essentially an analyzer that has TFM specific diagnostics should always load but only offer the diagnostics when the TFM is being targeted by the compilation. 

> We do have compiler API version and make use of it today --

Believe the compiler API version should be modeled above.

@jjonescz

aortiz-msft:
@martinrrm - Could you please help understand next steps on this one?

accepted/2025/indicate-analyzer-assets.md
martinrrm: Updated the spec, language and compiler metadata is now showing in the assets file.

martinrrm:
Updated the spec and created a PR for Client change https://github.com/NuGet/NuGet.Client/pull/7464 and for SDK change https://github.com/dotnet/sdk/pull/54646.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,263 @@
+# project.assets.json should indicate analyzer assets
+<!-- Replace `Title` with an appropriate title for your design -->
+
+- Author Name: [kfikadu](https://github.com/kfikadu)
+- GitHub Issue: [Issue #6279](https://github.com/NuGet/Home/issues/6279)
+
+## Summary
+
+`project.assets.json` does not currently indicate which analyzers are active, and analyzers are added even when `PrivateAssets` or `ExcludeAssets` should prevent this.
```
zivkan: nitpick: this reads like analyzers are added to the assets file, which is not the case. I think it'd be good to use more precise language.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,263 @@
...
+
+#### File Structure
+
+All `.dll` files under `analyzers/` are tracked, at any depth, excluding satellite `.resources.dll` assemblies. This mirrors the SDK's analyzer discovery, so the assets file does not assume a fixed folder layout.
+
+|Pattern|Description|
+|---|---|
+|`analyzers/**/{name}.dll`|Any analyzer assembly, at any depth under `analyzers/`|
+|`analyzers/dotnet/{language}/{name}.dll`|Language-specific analyzer (`cs`, `vb`, `fs`)|
+|`analyzers/dotnet/{compilerApiVersion}/{language}/{name}.dll`|Analyzer for a specific compiler API version (`roslynX.Y`)|
```
zivkan: Does the first `analyzers/**/{name}.dll` pattern exclude the `analyzers/dotnet/` directory? If not, what prevents `analyzers/dotnet/fs/MyAnalyzer.dll` from being used in a C# project?

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,263 @@
...
+
+The `{language}` and `{compilerApiVersion}` segments are optional and may appear at any depth. NuGet derives `codeLanguage` and `compilerApiVersion` metadata from them (see the example below) so the SDK can select the applicable analyzers, the same way `codeLanguage` is used for content files. Analyzers with no language segment get a `codeLanguage` of `any`, and `compilerApiVersion` is omitted when the path has no `roslynX.Y` segment. Excluded analyzers (for example via `PrivateAssets`) are written as a bare `analyzers/.../_._` placeholder with no metadata.
```
zivkan: nitpick: one line per sentence please

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,263 @@
...
+          "codeLanguage": "cs",
+          "compilerApiVersion": "roslyn4.0"
+        }
+      }
+    }
+  }
+}
+``` 
+
+#### Testing
```
zivkan: nitpick: personally I don't feel like listing a few testing scenarios adds to the design spec, but that's very opinionated. FWIW, it's not part of our template. I'd remove it, and if you want to keep the list of scenarios that need testing, add it to the feature issue

accepted/2025/indicate-analyzer-assets.md
martinrrm: This change will add all of the files that start with `analyzers/**/{name}.dll`, reading this table again it suggests otherwise, but wanted to show that we are gathering `language`, `compilerApiVersion` and `name` from the path and using it in the metadata. 

NuGet code won't prevent language selection; this list will be passed to the sdk and the sdk (which currently does) will decide which language to use.

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,274 @@
...
+|A `cs`, `vb`, or `fs` segment (e.g. `analyzers/dotnet/cs/{name}.dll`)|`codeLanguage`; defaults to `any` when no language segment is present|
+|A `roslynX.Y` segment (e.g. `analyzers/dotnet/roslyn4.0/cs/{name}.dll`)|`compilerApiVersion`; omitted when no such segment is present|
+
+The `{language}` and `{compilerApiVersion}` segments are optional and may appear at any depth.
+The SDK uses this metadata to select which analyzers to apply, the same way `codeLanguage` is used for content files.
+Because all analyzers are listed, an `analyzers/dotnet/fs/MyAnalyzer.dll` entry still appears for a C# project, but the SDK does not apply it since its `codeLanguage` is `fs`.
+Excluded analyzers (for example via `PrivateAssets` or `ExcludeAssets`) are written as a `analyzers/.../_._` placeholder with no metadata.
+
+Examples: 
+- `analyzers/dotnet/cs/MyAnalyzer.dll` (C# analyzer)
```
nkolev92, line 56: Everything is `analyzers/dotnet/` right? 

Do we automatically match dotnet or is somehow configurable?

accepted/2025/indicate-analyzer-assets.md
nkolev92: This wasn't clsoed, but I imagine the answer is yes, the reader should be resilient.

nkolev92 (approved):

accepted/2025/indicate-analyzer-assets.md
```diff
@@ -0,0 +1,274 @@
...
+**Mitigation:**
+```xml
+<PropertyGroup>
+  <RestoreEnableAnalyzerAssets>false</RestoreEnableAnalyzerAssets>
+</PropertyGroup>
+```
+
+### Rollout Strategy
+
+The initial rollout is opt-in via `<RestoreEnableAnalyzerAssets>true</RestoreEnableAnalyzerAssets>`, and the opt-in is only available for projects targeting .NET 11 or greater; on earlier target frameworks it is ignored.
```
nkolev92, line 195: > on earlier target frameworks it is ignored.

Does this mean we cannot be enable analyzers for .NET 10 even as opt in?

Issue #14047, opened 2025-01-17
It's not possible to push to HTTP sources specified via command line


kekekeks:
### NuGet Product Used

dotnet.exe

### Product Version

9.0.102

### Worked before?

Worked in .NET 8

### Impact

I'm unable to use this version

### Repro Steps & Context

We are specifying the local package source in our CI scripts. The URL is only accessible from CI machines and is dynamically selected based on the branch, so it can't be set in NuGet.config file.

We are running `dotnet nuget push ./artifacts/packages/**/*.nupkg  --api-key *** --source http://1.2.3.4/v3/index.json`.
It worked previously, but now complains about HTTP sources on our local network being insecure.

The command suggest using allowInsecureConnections, but it's not possible to set that option via CLI.

Am I missing something or is that use case now broken with .NET 9 SDK?

BTW, why are you requiring an SSL cert for an IP address? How is one even supposed to get one for local IP.

jebriede:
It looks like the option for --allow-insecure-connections does not exist for the `dotnet nuget push` command.
Similar to https://github.com/NuGet/Home/issues/13396, the consideration would be to add this option to `dotnet nuget push`.

Nigusu-Allehu:
`dotnet nuget push` allows setting up your sources by using your configuration files. In addition, you can also do the same to disable TLS certificate validation.
example
```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="src" value="http://test" allowInsecureConnections="true" disableTLSCertificateValidation="true"/>
  </packageSources>
</configuration>

```

Nigusu-Allehu:
to learn more about configuration files: https://learn.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior

koryphaee:
I can confirm this is broken in .NET 9. Pushing to a HTTP NuGet server worked fine in .NET 8 but not anymore.

Nigusu-Allehu:
`dotnet nuget push` allows configuration files to define package sources and related settings. You can use these files to specify your source and disable TLS certificate validation. Here’s an example configuration:

```xml
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="src" value="http://test" allowInsecureConnections="true" disableTLSCertificateValidation="true"/>
  </packageSources>
</configuration>
```

[...]

Issue #12513, opened 2023-03-29
Add "dotnet nuget install" command


akoeplinger:
### NuGet Product(s) Involved

dotnet.exe

### The Elevator Pitch

Related to https://github.com/NuGet/Home/issues/4855

We had multiple internal teams at MSFT who are doing a migration to Mariner reach out to us when moving from Mono's `nuget`(which is just the nuget.exe CLI) to the `dotnet` CLI and one common pain point was the missing support for the `install` command.

Most of these teams are not even using .NET but have some tool which is hosted on an internal AzDO feed and they were using `nuget install <tool>` e.g. from CMake to grab the assets. Right now you need to create a dummy .csproj like this so you can run `dotnet restore` on it which makes this a lot more complicated:

```xml
<Project Sdk="Microsoft.NET.Sdk">

<PropertyGroup>
   <OutputType>Exe</OutputType>
   <TargetFramework>net6.0</TargetFramework>
   <RestorePackagesPath>packages</RestorePackagesPath>
</PropertyGroup>

<ItemGroup>
  <PackageDownload Include="Some.Internal.Tool" Version="[7.0.7]" />
</ItemGroup>

</Project>
```

It'd be great if we could get `dotnet nuget install` supported in the future 😄 

Please 👍 or 👎 this comment to help us with the direction of this feature & leave as much feedback/questions/concerns as you'd like on this issue itself and we will get back to you shortly.

JonDouglas:
Adding a quick comment here.

We've had a number of requests for this specific feature. Basically the concept is to acquire a NuGet package without a .csproj and without being so tied down to a project system to get an asset to disk.

We want to continue to collect upvotes and scenarios for this experience, so I've edited the description to help us do so.

Some of the additional scenarios we have heard regarding this are:

1. Wanting to download a NuGet package to run security health checks with (See https://github.com/ossf/scorecard)
2. Wanting to download a NuGet package without a dummy .csproj / project system attachment.
3. Wanting to download a NuGet package as the equivalent of NuGet.org's raw nupkg download (i.e. [Download package](https://www.nuget.org/api/v2/package/Newtonsoft.Json/13.0.3))
4. Wanting to download a NuGet package without any client or server-side validation performed.
5. Wanting to download a NuGet package similar to `nuget.exe install <packageid> -OutputDirectory <directory>`, `npm install <package id> --prefix <prefix>`, `pip install <package id> --target <target>`, etc.

Please feel free to share your scenarios if you're reading this comment!

[...]

PR #14495, opened 2025-08-20
Introduce package downloading feature to dotnet 

Nigusu-Allehu:
This document proposes adding a `dotnet package download` command to the .NET CLI.

Rendered: https://github.com/NuGet/Home/blob/dev-nyenework-install/accepted/2025/dotnet-package-download.md


accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
+# Dotnet Package Install Command
+
+* Author: [Nigusu](https://github.com/Nigusu-Allehu)
+* GitHub Issues: [#12513](https://github.com/NuGet/Home/issues/12513), [#14491](https://github.com/NuGet/Home/issues/14491)
+
+## Summary
+
+This document proposes adding a `dotnet package install` command to the .NET CLI.
```
baronfel: This document needs to answer one very important question:

> What does it _actually mean_ to install packages in a non-project context?

Is this literally just "download package X, version Y to the local package cache?"

Once installed, how is someone expected to interact with the package - is there a way for tools to request the path to the 'root' of the package in some structured way?

If the goal is to prepopulate caches, how will you handle multiple packages/multiple versions of packages? Multiple args? Would you accept some kind of manifest file to use as input? If so, how would users _create_ such a file?

zivkan: Would `dotnet package download` feel more natural than install? 

If I install in a project directory, I might expect it to add it to the project. Or maybe install is used to install system tools, so should run an install script?

I think download makes it more clear that the command shouldn't do much more than download.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+The command would bring parity with `nuget.exe install`, enabling package download and extraction scenarios without requiring a project file.
```
zivkan: unless we're planning on having the new command have every feature that nuget.exe install does, I'd avoid the word parity.  Something along the lines "this command will prevent customers needing to use obtain nuget.exe  just for extracting packages without a project file"

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+  install <PACKAGE_ID>   Install a NuGet package (and dependencies) to a folder, without modifying a project.
+  list                   List all package references of the project or solution.
+  remove <PACKAGE_NAME>  Remove a NuGet package reference from the project.
+```
+
+`dotnet package install <packageId> [options]`
+
+**Supported options:**
+
+* `--version <semver>`: A version (eg. 13.0.1).
```
zivkan: `dotnet package add` started supporting the `packageid@version` syntax last year, I think. For `dotnet package update` it only supports that syntax, it has no `--version` option, in order to reduce confusion when multiple packages are provided in one CLI.

I think this new command should support multiple packages, and so should use the `id@version` syntax, and not have the `--version` option.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+When not specified the command searches all sources defined in hierarchical selected `NuGet.config` or passed via `--configfile`.
+It can also be used to specify a package source key that is found in a config file.
+* `--output-directory <path>`: Sets the directory where packages will be downloaded.
+The default is the global packages folder.
+* `--exclude-version`: Omits version number from the package folder name.
+* `--prerelase`: Allows pre-release packages to be considered when resolving versions.
+By default only stable versions are selected.
+* `--framework <tfm>`: Specifies the target framework.
+* `--download-only`: Downloads the `.nupkg` file to the output directory without extracting it.
+No dependencies are downloaded either.
```
zivkan: Does this mean that when `--download-only` is not provided, dependencies are downloaded?   When a package has multiple dependencies groups (one per TFM), which TFM will the dependency be downloaded for?

If so, and I want to use this command to download a nupkg and its dependencies so I can immediately push it to a private feed, I will want the dependencies. Why does `--download-only` prevent downloading dependencies?

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Explanation
+
+### Functional explanation
+
+The new command lives under `dotnet package`.
+
+`dotnet package` will then be the natural home for **search -> install -> add/remove** since it unifies all package-related workflows under one surface.
+
+Here is what `dotnet package` would look like
```
zivkan: When we create specs for the new command, it's typical we provide a sample help output for the new command, not its parent command. IMO this is just making the spec longer 15-20 lines longer, and only 1 line has new information relevant to the spec, so it's information sparse.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+  search <SearchTerm>    Searches one or more package sources for packages that match a search term. If no sources are specified, all sources defined in the NuGet.Config are used.
+  add <PACKAGE_ID>       Add a NuGet package reference to the project.
+  install <PACKAGE_ID>   Install a NuGet package (and dependencies) to a folder, without modifying a project.
+  list                   List all package references of the project or solution.
+  remove <PACKAGE_NAME>  Remove a NuGet package reference from the project.
+```
+
+`dotnet package install <packageId> [options]`
+
+**Supported options:**
```
zivkan: will any of these has single letter aliases? if so, what are they?  If a single letter is proposed for one of these options, but it's a letter that is commonly used by other commands, and it's used in a different context in this command, we might want to avoid that letter to reduce confusion.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+
+* `--version <semver>`: A version (eg. 13.0.1).
+If omitted, the highest version is chosen from the available sources.
+* `--source <package source>`: Specifies one or more package sources to install from.
+When not specified the command searches all sources defined in hierarchical selected `NuGet.config` or passed via `--configfile`.
+It can also be used to specify a package source key that is found in a config file.
+* `--output-directory <path>`: Sets the directory where packages will be downloaded.
+The default is the global packages folder.
+* `--exclude-version`: Omits version number from the package folder name.
```
zivkan: extracting the package without a version should absolutely not be allowed in the global packages folder.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+* `--prerelase`: Allows pre-release packages to be considered when resolving versions.
```
zivkan: are customers going to understand what "resolving versions" means?  I think "when finding the latest version" is much more clear.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+By default only stable versions are selected.
+* `--framework <tfm>`: Specifies the target framework.
+* `--download-only`: Downloads the `.nupkg` file to the output directory without extracting it.
+No dependencies are downloaded either.
+* `--no-http-cache`: Prevents the usage of http cached packages.
```
zivkan:
```suggestion
* `--no-http-cache`: Refresh downloads of any packages already in the HTTP cache.
```

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+Values: Lowest(default), HighestPatch, HighestMinor, and Highest.
+* `--no-audit`: opts out of vulnerability auditing. Be default it is on.
+* `--allow-insecure-connections`: allows downloading from http sources(insecure).
+
+### Technical explanation
+
+The `NuGet.CommandLine.XPlat` project currently exposes `dotnet package search` and `dotnet package update` in `Program.cs`.
+The upcoming `install` sub‑command fits naturally alongside these operations.
+
+### Algorithm: `dotnet package install <packageId> [options]`
```
zivkan: IMO, when a feature spec have very low technical complexity, it's ok to leave this section empty, or very minimal. This sounds like implementation details that I'll wait for the implementation pull request, so maybe I'm being lazy, but I'm not going to read in a feature spec.

accepted/2025/dotnet-package-install.md
zivkan: I don't see a `--verbosity` option.   NuGet.Protocol's messages about GET, CACHED, OK, NOTFOUND, etc are spammy and not useful except when trying to investigate errors. Similar for when using a credential provider, its verbose output is only desirable when investigating auth failures.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+By default only stable versions are selected.
+* `--framework <tfm>`: Specifies the target framework.
+* `--download-only`: Downloads the `.nupkg` file to the output directory without extracting it.
+No dependencies are downloaded either.
+* `--no-http-cache`: Prevents the usage of http cached packages.
+* `--direct-download`: download directly without populating any caches with metadata.
+* `--configfile <path>`: Used to specify a NuGet.Config file to use instead of the default configuration hierarchy.
+* `--dependency-version <policy>`: Controls how versions of dependencies are selected when a version range is specified.
+Values: Lowest(default), HighestPatch, HighestMinor, and Highest.
+* `--no-audit`: opts out of vulnerability auditing. Be default it is on.
```
zivkan: What does auditing mean in the context of this command?

If a package has a known vulnerability, will the command prevent downloading/extracting the package? If so, how do customers accept the risk and allow it anyway?

If the command will only warn (some output message, but process exit code == 0), how can customers treat it as an error instead?  nuget.exe install is primarily used in scripts, not human interactive use, so I think it needs to be something easily scriptable.

accepted/2025/dotnet-package-install.md
zivkan: @baronfel my memory is that System.CommandLine treats all arguments starting with an `@` as a "response file", no opt out is even possible.  So, if a customer creates a file packages.txt where they list one package per line, will `dotnet package install @packages.txt` automatically give them the equivilent of `dotnet package install pkg1 pkg2 pkg3`?

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+* `--prerelase`: Allows pre-release packages to be considered when resolving versions.
+By default only stable versions are selected.
+* `--framework <tfm>`: Specifies the target framework.
+* `--download-only`: Downloads the `.nupkg` file to the output directory without extracting it.
+No dependencies are downloaded either.
+* `--no-http-cache`: Prevents the usage of http cached packages.
+* `--direct-download`: download directly without populating any caches with metadata.
+* `--configfile <path>`: Used to specify a NuGet.Config file to use instead of the default configuration hierarchy.
+* `--dependency-version <policy>`: Controls how versions of dependencies are selected when a version range is specified.
+Values: Lowest(default), HighestPatch, HighestMinor, and Highest.
```
zivkan: Does nuget support configurable dependency version anywhere else?  Not in PackageReference.  Maybe in packages.config, but the .NET SDK only supports PackageReference, so it feels out of place to re-use the packages.config dependency resolver in the dotnet cli.  If the packages.config resolver doesn't have configurable version resolution, then this option will cause us to need to implement it in the PackageReference dependency resolver, or implement a new dependency resolver, both of which sound like a lot of work.

Also, I'm quite confident that `nuget.exe install` doesn't support downloading dependencies without a project file. Do we have customers asking for this?  It feels to me like something we can consider for a future improvement, but a first version can leave out.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+This document proposes adding a `dotnet package install` command to the .NET CLI.
+The command would bring parity with `nuget.exe install`, enabling package download and extraction scenarios without requiring a project file.
+
+## Motivation
+
+Developers often need to download NuGet packages outside of a project context:
+
+* Populating internal or offline feeds.
+* Security scanning of packages.
+* CI/CD workflows that consume `.nupkg` artifacts directly.
```
zivkan: A lot of customers are using nuget.exe install in CI scripts running on Linux and Mac, and therefore need to install mono just for it.  I think it's worthwhile calling out that this will allow customers to avoid needing mono once implemented.

accepted/2025/dotnet-package-install.md
nkolev92: I agree, the global tool that I have I called `download package download` because of this exact reason. 

.NET tools can be installed on the machine, packages can be installed in a project and the gesture is package add.

nkolev92: I agree. 

I'd say download the nupkg and potentially extract it. Don't lean to strongly about what the default should be.
It seems like an extracted version should be it, but don't know for sure.

accepted/2025/dotnet-package-install.md
nkolev92: I understand the motivation behind the whole, change the version, don't change the script thing, but I've seen the usage of this feature cause so many discrepancies and uncertainties, because packages change their structure from version to version.

I think we should not add this.

accepted/2025/dotnet-package-install.md
nkolev92: What other commands support --no-http-cache?

Is it all? Can we do without it for now?

accepted/2025/dotnet-package-install.md
nkolev92: nuget.exe instlal downloads the dependencies as well. 
tbh, I'm not sure this command should be downloading the dependencies at all.

That being said, the whole dependency version policy shouldn't apply IMO, at least not in V1. 

Let's make it minimal and then we can add functionality based on feedback and usage.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+* `--output-directory <path>`: Sets the directory where packages will be downloaded.
+The default is the global packages folder.
+* `--exclude-version`: Omits version number from the package folder name.
+* `--prerelase`: Allows pre-release packages to be considered when resolving versions.
+By default only stable versions are selected.
+* `--framework <tfm>`: Specifies the target framework.
+* `--download-only`: Downloads the `.nupkg` file to the output directory without extracting it.
+No dependencies are downloaded either.
+* `--no-http-cache`: Prevents the usage of http cached packages.
+* `--direct-download`: download directly without populating any caches with metadata.
```
nkolev92: Which caches are we talking about here? 

direct download was added as a means to bypass the gpf. 

IMO dotnet package download should always bypass the GPF, effectively treat the output folder as the only output location. 
http cache should be used as normal.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+* `--configfile <path>`: Used to specify a NuGet.Config file to use instead of the default configuration hierarchy.
+* `--dependency-version <policy>`: Controls how versions of dependencies are selected when a version range is specified.
+Values: Lowest(default), HighestPatch, HighestMinor, and Highest.
+* `--no-audit`: opts out of vulnerability auditing. Be default it is on.
+* `--allow-insecure-connections`: allows downloading from http sources(insecure).
```
nkolev92: My suggestion would be to avoid adding all of these options everywhere. 
There are ways to do it in the nuget.config and we'd respect that.

accepted/2025/dotnet-package-install.md
nkolev92: Low technical complexity + no changes to existing files or anything like that. 
For example, if this was a project feature, and it was low technical complexity, but still assets file changes, I'd still want to see that in the spec. 

I agree in this case, we don't need this, imo it's just busy work for the author.

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+
+A `dotnet package install` improves developer productivity.
+
+## Explanation
+
+### Functional explanation
+
+The new command lives under `dotnet package`.
+
+`dotnet package` will then be the natural home for **search -> install -> add/remove** since it unifies all package-related workflows under one surface.
```
nkolev92: What are the exit codes?

0 - success
1 - fail.

What about when the package is already on disk? IMO just 0 is enough for now,, but it's worth explicitly clarifying since it's part of the UX.

aortiz-msft:
I thought dotnet was moving from nouns to verb. IOW, from dotnet package install to dotnet install <package>. How does that fit with this command? @baronfel, is that your understanding as well?

baronfel:
No, the other way around. We prefer nouns in order to group actions by the kind of thing they act upon. In this case 'dotnet package install' is aligned with the overall guidance (though as others note in the comments already we may discuss the use of install compared to other terms).

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+
+Developers often need to download NuGet packages outside of a project context:
+
+* Populating internal or offline feeds.
+* Security scanning of packages.
+* CI/CD workflows that consume `.nupkg` artifacts directly.
+
+Currently, the only way is either using `nuget.exe` or creating dummy projects with `dotnet restore`.
+
+A `dotnet package install` improves developer productivity.
```
aortiz-msft: Do we have links to GitHub issues or FeedbackTickets that we can add here?

accepted/2025/dotnet-package-install.md
```diff
@@ -0,0 +1,175 @@
...
+   * Otherwise, default to the global packages folder.
+
+1. **Configure folder layout**
+
+   * If `--exclude-version` is set, package folders are named `<id>` only.
+   * Otherwise, use `<id>.<version>` layout.
+
+1. **Handle target framework**
+
+   * If `--framework` is provided, constrain resolution to that TFM.
```
kartheekp-ms: When a package has different dependencies for different TFMs, does specifying `--framework` mean only the dependencies for that specific TFM are downloaded?

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: Discussed offline. Now that we have simplified the command to download just the package a user specified, this option has been removed from the design

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: Discussed offline.
The command will now follow a similar extraction pattern to global packages folder and this option will not be supported
The command Stores each package in its **own folder named after the package ID**, with subfolders for each version.

Example:

```ps1
<output-directory>/
  Contoso/
    13.0.3/
    12.0.3/
```

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: removed

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: added

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: removed. Will be added in the future if needed/requested

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: removed. The command will download to CWD by default unless specified by users to download to a different location. The packages won't be cached in GPF too.

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: Since this command will be used primarily in CI scenarios, I think having this option is critical. In cases where a dynamic source is required, relying on a `nuget.config` file can be cumbersome. For example, users encountered similar challenges with the `push` command, as noted in https://github.com/NuGet/Home/issues/14047.

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: This option has now been removed as we agree to make the command simple and only download the package a user specified without its dependencies

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: Updated based on offline discussion
By default, this command runs in **audit mode set to `warn`**.

* If a downloaded package is identified as vulnerable, a vulnerability message is written to **stderr**, but the exit code remains **0** (as long as the download itself was successful).
* `--audit off`: Disables vulnerability auditing.
* `--audit error`: Writes vulnerability messages to **stderr** and sets the exit code to **1** if any vulnerabilities are found.

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: 0 - success
1- fail as you mentioned for now. Added in the doc

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: For now, just -s ource and -v erbosity

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: yes, the original issue is linked at the top of the page: https://github.com/NuGet/Home/issues/12513. Do we also have other GitHub issues or Feedback Tickets that we should add here?

accepted/2025/dotnet-package-install.md
Nigusu-Allehu: Discussed offline
Summary
* The command will be a simple command that downloads a package to CWD unless specified otherwise
* Multiple versions will be handled similar to how multiple versions of the same package are stored in the GPF. ParentDirectory/PackageID/PackageVersion path will be used.
* For now, we won't support a specific manifest file but as @zivkan mentioned response file is also an option

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,98 @@
...
+
+This layout allows multiple versions of the same package to coexist and makes it easy for scripts or tools to locate the correct version.
+
+#### Audit
+
+By default, this command runs in **audit mode set to `warn`**.
+
+* If a downloaded package is identified as vulnerable, a vulnerability message is written to **stderr**, but the exit code remains **0** (as long as the download itself was successful).
+* `--audit off`: Disables vulnerability auditing.
+* `--audit error`: Writes vulnerability messages to **stderr** and sets the exit code to **1** if any vulnerabilities are found.
```
nkolev92: nit: idk if we need this yet. 
My concern is primarily around the fact that it introduces new syntax for audit that doesn't exist in other command, but I'm not 100% confident I see the value yet.

IMO, audit should start as inform only (so no elevation to error).

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,98 @@
...
+The command follows standard exit code conventions:
+
+* **0** – The package was successfully downloaded, or the requested version was already present.
+* **1** – The download failed, or audit mode was set to `error` and vulnerabilities were detected.
+
+#### **Deferred (Not in MVP)**
+
+* Download a package with its dependencies too
+* `--framework <tfm>` : target framework handling
+* `--no-http-cache` : ignore cached packages
```
nkolev92: I think this is a never tbh.

nkolev92 (commented):
LGTM, my consideration is audit. 

If people feel really strongly about it, I'm ok with it,

Nigusu-Allehu:
Hey @seaniyer , any feedback on this PR?

seaniyer:
> Hey @seaniyer , any feedback on this PR?

Thanks for sharing @Nigusu-Allehu. Looks good to me! I am wondering if we have any user input that MVP scope with no dependency downloads is enough for real-world use.

baronfel (dismissed):

accepted/2025/dotnet-package-download.md
nkolev92: I know we mentioned it during a sync meeting, but I'm not really sure what the resolution here.

Nigusu-Allehu: The plan is to keep this as it is. However, the implementation won't be part of the MVP implementation. I will add this as a follow up and that will also give room for more discussions on the feature.

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,98 @@
...
+```ps1
+Description:
+  Downloads a NuGet package to a local folder without requiring a project file.
+
+Usage:
+  dotnet package download <PackageId> [options]
+
+Arguments:
+  PackageId    Package in the form of a package identifier (e.g. 'Newtonsoft.Json') 
+               or identifier and version separated by '@' (e.g. 'Newtonsoft.Json@13.0.3').
```
kartheekp-ms, line 34: NuGet.exe install doesn't have `packageID@version` format but instead has `-Version` option. [NuGet CLI install command | Microsoft Learn](https://learn.microsoft.com/nuget/reference/cli-reference/cli-ref-install#usage). Should we have an explicit `--version` option to this command.

zivkan: relevant comment on an earlier iteration of this PR: https://github.com/NuGet/Home/pull/14495#discussion_r2289186574

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,98 @@
...
+
+Options:
+  -?, -h, --help                Show command line help.
+  --allow-insecure-connections  Allows downloading from HTTP sources.
+  --configfile <path>           Path to a NuGet.config to use.
+  --download-only               Download only the .nupkg file without extraction.
+  --interactive                 Enables interactive authentication if required.
+  --audit <mode>                Set log level for vulnerability report. Mode values: `off`, `warn`, `error`
+  --output-directory <path>     Directory where the package will be placed. Defaults to the current working directory.
```
zivkan: `dotnet pack` and `dotnet publish` both have `--output` with an `-o` alias. I think we should use that, to be consistent. The contexts for these commands are close enough that I don't think it will be confusing.

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,98 @@
...
+* If a downloaded package is identified as vulnerable, a vulnerability message is written to **stderr**, but the exit code remains **0** (as long as the download itself was successful).
+* `--audit off`: Disables vulnerability auditing.
+* `--audit error`: Writes vulnerability messages to **stderr** and sets the exit code to **1** if any vulnerabilities are found.
+
+#### Exit Codes
+
+The command follows standard exit code conventions:
+
+* **0** – The package was successfully downloaded, or the requested version was already present.
+* **1** – The download failed, or audit mode was set to `error` and vulnerabilities were detected.
```
zivkan: System.CommandLine also uses exit code 1 when CLI arguments fail parsing.

Some other obvious errors are things like network errors, or if the package source can't find the package or the requested version. Rather than trying to enumerate every possible error, you could just say "All other errors"

accepted/2025/dotnet-package-download.md
nkolev92: fwiw, we did talk about adding that conversation into the doc itself.

accepted/2025/dotnet-package-download.md
Nigusu-Allehu: Updated as per offline discussion.

Nigusu-Allehu:
> > Hey @seaniyer , any feedback on this PR?
> 
> Thanks for sharing @Nigusu-Allehu. Looks good to me! I am wondering if we have any user input that MVP scope with no dependency downloads is enough for real-world use.

The decision was based on the user discussion from https://github.com/NuGet/Home/issues/12513 and in hopes of keeping the first iteration simple and reliable. However, if we get signal from users on scenarios where the current implementation comes short we will update the design as per.

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,114 @@
...
+| **Dependency resolution**            | Downloads **only the explicitly requested package** (no transitive dependencies).                 | Downloads the requested package **and all its dependencies** by default.                    |Current discussions on the issue https://github.com/NuGet/Home/issues/12513 don't indicate a need for downloading transitive dependencies. |
+| **Default output location**          | Downloads to the **current working directory** unless `--output` is specified.                    | Also downloads to the **current working directory** unless `-OutputDirectory` is specified. | N/A |
+| **Folder layout**                    | Creates a folder per **package ID**, with **version subfolders** inside (e.g. `Contoso/13.0.3/`). | Creates a folder per **package-version pair** (e.g. `Contoso.13.0.3/`).                     | The new layout allows multiple versions of the same package to coexist while remaining easy to locate. |
+| **Default Version**                  | The latest version is selected if no version is specified                                         | The latest version is selected if no version is specified| N/A |
```
nkolev92: There's no change here, so this can be removed I think.

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,114 @@
...
+
+* **0** – The package was successfully downloaded, or the requested version was already present.
+* **1** – The download failed, network errors, requested package version not found, parsing of the command arguments failed, all other errors during the command operation.
+
+#### **Deferred (Not in MVP)**
+
+* Download a package with its dependencies too
+* `--framework <tfm>` : target framework handling
+* `--dependency-version <policy>` : control dependency resolution strategy
+* **Manifest file support** : install multiple packages from a single manifest
```
nkolev92: I'm not sure what manifest files are referring to.

If someone wants to specify the files,t hey provide them as args to the command, or they can use PackageDownload which behaves pretty much equivalently to this command: https://learn.microsoft.com/en-us/nuget/consume-packages/packagedownload-functionality

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,114 @@
...
+
+Providing `dotnet package download` improves developer productivity, simplifies pipelines, and ensures a cross-platform experience within the .NET CLI.
+
+## Explanation
+
+### Command Overview
+
+```ps1
+Description:
+  Downloads a NuGet package to a local folder without requiring a project file.
```
nkolev92, line 27: Is `dotnet package download packageA@1.0.0 and packageB@1.0.0`  support?

accepted/2025/dotnet-package-download.md
Nigusu-Allehu: The command design currently supports multiple packages as an argument. A manifest file for listing packages would basically be doing the same thing. So I have removed it.

accepted/2025/dotnet-package-download.md
Nigusu-Allehu: Yes, specified now

nkolev92 (dismissed):

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,121 @@
...
+
+* Populating internal or offline feeds.
+* CI/CD workflows that consume `.nupkg` artifacts directly.
+* **Cross-platform builds** where many customers today rely on `nuget.exe install` in CI scripts running on Linux or macOS, forcing them to install Mono just to get this functionality.
+
+Providing `dotnet package download` improves developer productivity, simplifies pipelines, and ensures a cross-platform experience within the .NET CLI.
+
+## Explanation
+
+### Command Overview
```
nkolev92, line 23: Did we ever discuss how PackageSourceMapping integrates with `dotnet package download`?

I think `dotnet tool install` is supposed to support it https://github.com/dotnet/sdk/issues/41871 and it might've regressed. 

Think we need to support source mapping in package download too.

Nigusu-Allehu: Great catch! We did not discuss it. But I also think it is something we should support

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,121 @@
...
+This document proposes adding a `dotnet package download` command to the .NET CLI.
+The command would enable package download to a local folder scenarios without requiring a project file.
+
+## Motivation
+
+Developers often need to download NuGet packages outside of a project context:
+
+* Populating internal or offline feeds.
+* CI/CD workflows that consume `.nupkg` artifacts directly.
+* **Cross-platform builds** where many customers today rely on `nuget.exe install` in CI scripts running on Linux or macOS, forcing them to install Mono just to get this functionality.
```
donnie-msft, line 17: `nuget install` is a bit sloppy with how it handles package source mapping.
My testing gave me
> Package 'Newtonsoft.json 13.0.4' is not found in the following primary source(s): 'https://api.nuget.org/v3/index.json,C:\Program Files (x86)\Microsoft SDKs\NuGetPackages\'. Please verify all your online package sources are available (OR) package id, version are specified correctly.

At best, the error message isn't giving me a clear understanding of the problem (likely the logic finding the unmapped package is happening later than other commands like Restore). At worst, I wonder if install is trying to retrieve the package from the source first, then failing due to PSM being enabled.

accepted/2025/dotnet-package-download.md
```diff
@@ -0,0 +1,121 @@
...
+  dotnet package download <PackageId> [options]
+
+Arguments:
+  PackageId    Package in the form of a package identifier (e.g. 'Newtonsoft.Json') 
+               or identifier and version separated by '@' (e.g. 'Newtonsoft.Json@13.0.3').
+
+Options:
+  -?, -h, --help                Show command line help.
+  --allow-insecure-connections  Allows downloading from HTTP sources.
+  --configfile <path>           Path to a NuGet.config to use.
```
donnie-msft, line 39: I find it weird that some commands support both config files _and_ override flags like `--source`, etc. 
It makes it hard to decide here if we should also have some `--ignore-package-source-mappings` or similar - but my opinion is we should either use a config or use the CLI options. Mixing the concepts is likely confusing and error-prone.

Perhaps we can error if `--configFile` is specified in addition to any other configuration related option?

baronfel: That's not a bad idea at all, but please compare and contrast this proposed behavior against all of the other dotnet CLI commands that interact with the NuGet config file and that have explicit source management flags. 

We need to bias very hard in the direction of consistency with the rest of the CLI. Anytime there's a command that behaves inconsistently with the rest of the CLI surface area we get a long tail of issues and confusion from end users.

Nigusu-Allehu: *My understanding is that `--source` option and `nuget.config` are designed to complement each other in most of our commands.*
Users can define their package sources through configuration files, but in dynamic environments: such as CI scenarios they may prefer to specify sources directly via the `--source` argument in scripts.

In other cases, when users already have multiple sources configured in their `nuget.config` but want to target one specifically, they can run `--source <sourceName>` to select it from their existing configuration.

In terms of precedence:

* The `--source` option always takes priority over sources(replaces them basically) defined in `nuget.config`.
* When `--source <packageSource>` is used, the CLI first checks whether the given value corresponds to a named source in the configuration file.

  * If it does, that configured source is used.
  * Otherwise, it’s treated as a direct source URI.
* If `--source` is not specified, the CLI falls back to the sources defined in the user’s configuration files.

This pattern is consistent across commands like `dotnet nuget push`, `dotnet package search`, and others.

When package source mapping is involved:

* If `--source <packageSource>` is specified, the CLI should follow the same behavior as above resolving the named source if present or treating it as a URI. In this case, **package source mapping is not applied.**
* If `--source` is not specified, the CLI reads the configuration file and uses package source mapping to determine the appropriate source.

For example, `dotnet tool install` does support package source mapping. However, in my testing, I observed a bug where the `--source` option is ignored when a nearby `nuget.config` file is present, and I’ve reported this issue for further investigation https://github.com/dotnet/sdk/issues/51412.

When it comes to the package download, command I think we should follow the same pattern too.
* If --source <packageSource> is specified, the CLI should follow the same behavior as above: resolving the named source if present or treating it as a URI. In this case, package source mapping should not be applied.

* If --source is not specified, the CLI should read the configuration file and use package source mapping to determine the appropriate source.

What do you think @nkolev92

accepted/2025/dotnet-package-download.md
Nigusu-Allehu: I have added PSM to the doc https://github.com/NuGet/Home/blob/dev-nyenework-install/accepted/2025/dotnet-package-download.md#source-selection-and-package-source-mapping.

Nigusu-Allehu:
FYI, the new thing here is Package source mapping : https://github.com/NuGet/Home/blob/dev-nyenework-install/accepted/2025/dotnet-package-download.md#source-selection-and-package-source-mapping

accepted/2025/dotnet-package-download.md
nkolev92: works for me.

nkolev92 (dismissed):

aortiz-msft (changes requested):
Do we want to deliver the functionality under the "Deferred (not in MVP)" section at some point in the future? If not, we should remove items from that section or the section itself altogether. (Deferred implied intent to deliver later.)

nkolev92 (approved):

aortiz-msft (approved):

Nigusu-Allehu:
> Do we want to deliver the functionality under the "Deferred (not in MVP)" section at some point in the future? If not, we should remove items from that section or the section itself altogether. (Deferred implied intent to deliver later.)

I have removed the section.

aortiz-msft (changes requested):
Please address comment on deferred improvements.

aortiz-msft (approved):
:shipit:

Issue #14583, opened 2025-10-06
Spec: Support for add/update/delete auditsources in Visual Studio options UI


donnie-msft:
Spec for https://github.com/NuGet/Home/issues/13955

PR #14589, opened 2025-10-10
Spec: NuGet Audit Sources in Visual Studio Options

donnie-msft:
Part of: https://github.com/NuGet/Home/issues/14583

Visual Studio currently supports NuGet Audit to help developers identify vulnerable packages. As adoption grows, customers need a convenient way to manage **Audit Sources** directly from the Visual Studio Options UI, rather than editing `nuget.config` manually. This proposal introduces first-class support for adding, removing, and updating Audit Sources in Visual Studio.

A dropdown in Visual Studio, **"Choose how NuGet Audit retrieves vulnerability data"**, will only show the Audit Sources table in the options page when relevant. That is, when there are already `<auditSources>`, or when the customer wants to create their first audit source. The goal is to emphasize that the act of creating an audit source means package sources are no longer used for vulnerability data by NuGet Audit.

[Rendered spec](https://github.com/NuGet/Home/blob/dev-donnie-msft-auditSourceVSOptions/accepted/2025/audit-sources-visual-studio-options.md)


nkolev92 (dismissed):

accepted/2025/audit-sources-visual-studio-options.md
```diff
@@ -0,0 +1,99 @@
...
+  - Audit Sources (shown when explicitly configured),
+  - Machine-wide Package Sources (shown when explicitly configured)
+
+#### Describe Package versus Audit source
+
+Before each table, introduce descriptive text to inform customers of how Package Sources and Audit Sources work together.
+
+- **Package sources**:
+
+  > Configure the sources NuGet will use to for displaying and downloading packages. NuGet Audit will also reference vulnerability data from sources that support it. Alternatively, dedicated Audit Source(s) can be configured below.
```
zivkan, line 61: I think the message about package sources being used for audit only when no audit sources are configured can be more clear.

```suggestion
  > Configure the sources NuGet will use to for displaying and downloading packages. When there are no Audit Source(s) configured (see below), then package sources will also reference vulnerability data.
```

accepted/2025/audit-sources-visual-studio-options.md
```diff
@@ -0,0 +1,99 @@
...
+
+#### Telemetry
+
+- Track management actions (add/remove/enable) of audit sources to better understand customer needs.
+
+## **Drawbacks**
+
+- Potential confusion for package sources that act as audit sources implicitly by having a vulnerability resource.
+
+## **Rationale and Alternatives**
```
zivkan, line 81: Another alternative could be to have a single sources list, and have a new column to allow customers to choose whether it's an package source, audit source, or both.  This might make it more important to clearly explain that no audit sources means that audit falls back to package sources.  But the advantage is that it becomes clearer to customers that when you disable a source with a specific name (or URL? I can't remember) it disables both audit and package sources with that name (or url).  We don't (yet?) have a GUI to edit credentials, but if we did, they similar apply to both package and audit sources.

accepted/2025/audit-sources-visual-studio-options.md
```diff
@@ -0,0 +1,99 @@
...
+## **Unresolved Questions**
+
+None at this time
+
+## **Future Possibilities**
+
+1. Consider UI for Package Sources that are implicitly audit sources
+    1. For implicit audit sources, modify the Package Sources table to show a checkbox or similar indicator that the package source provides vulnerability data.
+    1. Validate audit source URLs by checking for `VulnerabilityInfo` resource in the service index.
+1. Should there be a button to add `nuget.org` as an audit source?
```
zivkan, line 137: Related, but tangential, is that it would be great to detect customers still using nuget.org's legacy v2 andpoint, and ask them to migrate to the v3 endpoint.

zivkan (dismissed):

accepted/2025/audit-sources-visual-studio-options.md
donnie-msft: At first, I didn't think this was a possibility, but thinking through your comment, I believe we could implement it that way.
For `Both`, we would have to add the source to `<packageSources>` and to `<auditSources>`, effectively removing the concept of an implicit audit source. Is that your line of thinking as well?

jeffkl (dismissed):

accepted/2025/audit-sources-visual-studio-options.md
donnie-msft: Added as Alternative 3.
We also discussed the main proposal offline, so my understanding is you agree this is an alternative and not the primary proposal.

accepted/2025/audit-sources-visual-studio-options.md
donnie-msft: Totally - these are sample strings, not final strings. I believe we can discuss and have UX/PM review the strings when the spec implementation begins.

accepted/2025/audit-sources-visual-studio-options.md
donnie-msft: I believe this should be a separately filed issue. Would you like me to file that, or would you like to since you have the context?

donnie-msft:
Updated the proposal with a new dropdown: "Choose how NuGet Audit retrieves vulnerability data"

nkolev92 (approved):

donnie-msft:
Met with UX team and they will be providing updated Strings for the descriptions, and for what will become a Checkbox instead of the current Combobox. Will update once those resources are ready.
Thanks!

accepted/2025/audit-sources-visual-studio-options.md
donnie-msft: https://github.com/NuGet/Home/issues/14614

accepted/2025/audit-sources-visual-studio-options.md
```diff
@@ -0,0 +1,140 @@
...
+- Audit Sources table will be **shown**.
+- **Pre-selected**  when one or more audit sources are already configured.
+  - The dropdown will be disabled as well since the presence of audit sources takes away the behavior of reading vulnerability data from package sources.
+- **User-selectable** only when no audit sources exist, enabling  a customer to explicitly configure their first `<auditSource>` using the `Add` button.
+
+![New dropdown expanded with both options shown"](../../meta/resources/AuditSources-VS/choose-nuget-audit-dropdown.png)
+
+![Table of Audit Sources shown below Package Sources because Combobox option "Configure sources to read vulnerabilities" is selected](../../meta/resources/AuditSources-VS/audit-sources-visible-table.png)
+
+- Switching the dropdown back to "**Read vulnerabilities from my package sources**" would not be supported in this iteration.
```
aortiz-msft, line 52: Why not? Seems straightforward to implement.

donnie-msft: NuGet Restore will use audit sources if they exist. 
What isn't straightforward is making the uncheck automatically remove all of those audit sources. 
I don't expect more than an average of 1 audit source, maybe 2-3 audit sources in the more extreme cases. It's also probably less common to go from audit sources back to package source vulnerability resources.

Therefore, customers can press Remove on all their audit sources and end up back in the original state. Additionally, the Unified Settings team is exploring building in a Remove All button on array settings.

aortiz-msft: In that case, could we simplify the UX by always showing a list of Audit Sources (even if empty)? Having a dropdown that only works for opting in but not for opting out seems a bit broken.

donnie-msft: We discussed this option first, and we believe it's confusing to have an empty table taking up space that isn't used by Audit in most cases. We agreed on hiding it entirely behind a user's choice to explicitly configure Audit. This makes it a tangible act that you're trying to manage your audit sources yourself.

It'll be a Checkbox instead of a Dropdown, as mentioned - but it doesn't opt you in to anything. It just enables you to choose audit sources. We do not have any setting today that opts you in to using audit sources; we use them if they are present.

jebriede:
> Met with UX team and they will be providing updated Strings for the descriptions, and for what will become a Checkbox instead of the current Combobox. Will update once those resources are ready. Thanks!

That's great. I was going to suggest a checkbox in lieu of the combo box since there are essentially two states. I look forward to seeing what the UX Team comes up with! :-)

accepted/2025/audit-sources-visual-studio-options.md
jebriede: I like the part of the spec on the next line that covers how to revert back to using only package sources as the audit sources:
> If customers want the ability to switch back from their audit sources to only package sources for Vulnerability data, a future iteration could support this and automatically clear `<auditSources>` after showing a warning messagebox that can be cancelled.

I think this would be a great addition for the first iteration. Allowing users to clear their audit sources directly in the UI creates a complete experience. Is there a way to programmatically iterate over the list of audit sources and remove each one from the nuget.config file(s) when the user confirms toggles the checkbox and confirms in the dialog?

accepted/2025/audit-sources-visual-studio-options.md
```diff
@@ -0,0 +1,140 @@
...
+Rather than share the page with Package Sources, create a new section, **Audit Sources**
+
+- Advantages:
+
+  - Less clutter on the Package Sources page.
+
+- Disadvantages:
+
+  - Separate UI for Package & Audit sources is potentially confusing, or at least more cumbersome for customers, as they are conceptually tightly related.
+  - More duplication of code to setup a separate page which is nearly identical to the Package Sources page.
```
jebriede, line 100: Agreed. Since package sources and audit sources are related in that package sources can act as the audit sources when audit sources are not provided, having them configurable in the same dialog window would present a better experience I think.

jebriede (commented):
This is a great write-up that thoroughly explains the way audit sources will be added to the options page. I look forward to seeing the wording and behavior of the checkbox for enabling and disabling the audit sources UX is designed. I really like the idea of the user being able to clear the audit sources if present, and use package sources only, from within the options page UI.

jeffkl (approved):

donnie-msft:
@aortiz-msft @jebriede we discussed offline. Please either resolve comments or let me know if you have further ideas. I'll update screenshots/strings once I hear back from UX. Alternatively, I can merge this as a conceptual spec and the implementation specific details can be implemented based on the UX team feedback. Let me know your thoughts!

accepted/2025/audit-sources-visual-studio-options.md
jebriede: We talked offline and I understand why we want to consider other options for clearing all the audit sources later. My input should be considered resolved and not block this thread but I'm going to rely on the comment author to resolve the comment when ready.

jebriede:
> @aortiz-msft @jebriede we discussed offline. Please either resolve comments or let me know if you have further ideas. I'll update screenshots/strings once I hear back from UX. Alternatively, I can merge this as a conceptual spec and the implementation specific details can be implemented based on the UX team feedback. Let me know your thoughts!

Here's are some thoughts to merge the conceptual spec before we hear back from the UX team on the specific wording and behavior of the checkbox:
- Update the spec text to replace all mentions of a combobox with a checkbox.
- Add to "unresolved questions" that the specific text and behavior of the checkbox is to be determined. Something like this:
  - What will the text of the checkbox be?
  - How will the checkbox behave? Will it be checked to show the audit sources or unchecked?

And then we can merge this iteration of the spec and modify it once we get more details on the checkbox. What are your thoughts?

jebriede (approved):
Great spec! Approving with suggestions to update the mentions of the combobox to checkbox for this iteration and add unresolved questions for the text and behavior of the checkbox.

donnie-msft:
> Great spec! Approving with suggestions to update the mentions of the combobox to checkbox for this iteration and add unresolved questions for the text and behavior of the checkbox.

Glad you agree with the plan! As soon as I hear back from UX, I'll let folks know. If it's not soon then I'll go ahead and update this spec and merge it.

PR #14720, opened 2026-01-15
proposal for new dns hostname for audit

zivkan:
Design spec for:
* https://github.com/NuGet/NuGetGallery/issues/10656

[Read the rendered spec](https://github.com/NuGet/Home/blob/users/zivkan/new-audit-hostname/accepted/2026/audit-only-nuget.org-host.md)

### DIFF
--- a/accepted/2026/audit-only-nuget.org-host.md
+++ b/accepted/2026/audit-only-nuget.org-host.md
@@ -1,0 +1,110 @@
+# ***Untitled***
+
+- Author: [zivkan](https://github.com/zivkan/)
+- GitHub Issue: [NuGetGallery#10656](https://github.com/NuGet/NuGetGallery/issues/10656)
+
+## Summary
+
+Create a new URL that can be used with NuGet.Config file `<auditSources>`, in networks where www.nuget.org and api.nuget.org are blocked, where this new host contains only the vulnerable package information, nothing else.
+
+## Motivation
+
+Package substitution, or dependency confusion, is a security risk where a package manager is configured to download packages from multiple sources, and a package that is intended to be downloaded from a company's private source is instead downloaded from another source.
+If the package's payload is malicious, it is a vector for remote code execution (RCE) attacks.
+Package Source Mapping is one way to mitigate the risk, and another is to block api.nuget.org and www.nuget.org, so that only company controled package sources may be used.
+Teams using the block approach either need to mirror packages they need from nuget.org onto their private feed, or use a NuGet server that can automate it for them, while still providing the protection they desire.
+
+NuGet also added a feature, NuGet Audit, which will generate warnings during restore, when packages with known vulnerabilities are used.
+To make onboarding onto the feature easy, a `VulnerabilityInfo` resource was added to the V3 service index, so that anyone using https://api.nuget.org/v3/index.json as a package source (which NuGet uses by default when no config file says otherwise) can use NuGet Audit automatically.
+
+The combination of the two mean that companies that block api.nuget.org to mitigate risk of package substituion can no longer use NuGet Audit, unless the NuGet server they use for their private feed also supports it.
+
+## Explanation
+
+### Functional explanation
+
+If audit.nuget.org is chosan as the new hostname, then companies that block nuget.org for package substition reasons, could put audit.nuget.org on their DNS/network allow-list.
+Then, teams could configure their solution NuGet.Config files as follows:
+
+```xml
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="contoso" value="https://pkgs.dev.azure.com/contoso/public/_packaging/my-project-feed/nuget/v3/index.json" />
+  </packageSources>
+  <auditSources>
+    <clear />
+    <add key="nuget.org" value="https://audit.nuget.org/v3/index.json" />
+  </auditSources>
+</configuration>
+```
+
+By using a single package source, they delegate the risk of package substituion to the server configration of the package source they've configured
+
+### Technical explanation
+
+The service index on this new hostname should contain only the the `VulnerabilityInfo` resource, and nothing else.
+It's absolutely critical that this host name cannot serve arbitrary binaries, otherwise the host name will not be trusted for the same reason that api.nuget.org gets blocked.
+
+The way that NuGet's server protocol is designed, the service index file contains absolute URLs to individual resources.
+NuGet Audit uses the VulnerabilityInfo resource:
+
+```json
+{
+  "@id": "https://api.nuget.org/v3/vulnerabilities/index.json",
+  "@type": "VulnerabilityInfo/6.7.0",
+  "comment": "The endpoint for discovering information about vulnerabilities of packages in this package source."
+},
+```
+
+Similarly, the vulnerability index page contains links to vulnerability pages:
+
+```json
+{
+    "@name": "base",
+    "@id": "https://api.nuget.org/v3-vulnerabilities/2026.01.15.05.28.15/vulnerability.base.json",
+    "@updated": "2026-01-15T05:28:15.6714127Z",
+    "comment": "The base data for vulnerability update periodically"
+  },
+```
+
+Both of these files need the URLs listed to be changed to the new hostname.
+
+## Drawbacks
+
+This only helps in situations where DNS level blocks are used.
+If the new host name is hosted on the same CDN that api.nuget.org is hosted on, then they're likely to resolve to the same IP address, and therefore an IP level block will block both.
+
+However, I think that IP level blocks are less likely than DNS blocks for the following resions:
+
+- nuget.org does not publish what IP addresses the service is available on ([currently?](https://github.com/NuGet/NuGetGallery/issues/8883))
+- CDNs will resolve the hostname to different IP addresses in different geographic regions
+- CDNs might change the IP address within the same geographic region, either intentionally, or a developer's machine might take a different network path to the CDN.
+- CDNs might use the same IP address to host multiple websites, so blocking api.nuget.org's website can block other websites that the company wants to use
+
+## Rationale and alternatives
+
+NuGet servers can implement the VulnerabilityInfo resource.
+However, at best the NuGet team can contribute to implementations of open source NuGet servers.
+Otherwise we can't do anything more than ask closed source NuGet servers to implement it.
+
+The NuGet client team had a customer report a problem where their 3rd party NuGet server added the VulnerabilityInfo resource, by pointing it to api.nuget.org.
+This means that NuGet and Visual Studio were having network problems because their package source was directing NuGet to access a URL that was being blocked at their firewall.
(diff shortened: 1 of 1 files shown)


accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,110 @@
...
+Rust's `cargo-audit`, and Python's `pypi` are similar.
+They download packages from one server, and download an audit database from a different server (both appear to use github.com as their CDN).
+
+NPM does not appear to have similar functionality.
+Information I've been able to find is that NPM uses a security endpoint on the configured package registries, so the same hostname is used for downloading packages and auditing.
+However, NPM sends the package graph to the server, rather than downloading an audit database to check locally, so there are different tradeoffs to consider.
+
+## Unresolved Questions
+
+Should the https://api.nuget.org/v3/index.json service index use the new hostname for the VulnerabilityInfo resource endpoint?
```
donnie-msft: I was going to ask you this question, but you already changed my mind with your explanation. Let's say api.nuget.org is blocked, then what good is it that the index.json lists audit.nuget.org? It seems to me that the index should stick to the same endpoint, and then what this spec is offering is a 2nd endpoint which essentially accesses the same data, but from a different endpoint to support blocking api.nuget.org.

accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Both of these files need the URLs listed to be changed to the new hostname.
+
+## Drawbacks
+
+This only helps in situations where DNS level blocks are used.
+If the new host name is hosted on the same CDN that api.nuget.org is hosted on, then they're likely to resolve to the same IP address, and therefore an IP level block will block both.
+
+However, I think that IP level blocks are less likely than DNS blocks for the following resions:
```
jeffkl:
```suggestion
However, I think that IP level blocks are less likely than DNS blocks for the following reasons:
```

accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,110 @@
...
+    "@updated": "2026-01-15T05:28:15.6714127Z",
+    "comment": "The base data for vulnerability update periodically"
+  },
+```
+
+Both of these files need the URLs listed to be changed to the new hostname.
+
+## Drawbacks
+
+This only helps in situations where DNS level blocks are used.
```
donnie-msft: Should we blogpost or otherwise inform customers that DNS level blocks are a good strategy, and specify when to do so? I'd guess you are right, and am just wondering if we should post official guidance around that.

accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,110 @@
...
+  </auditSources>
+</configuration>
+```
+
+By using a single package source, they delegate the risk of package substituion to the server configration of the package source they've configured
+
+### Technical explanation
+
+The service index on this new hostname should contain only the the `VulnerabilityInfo` resource, and nothing else.
+It's absolutely critical that this host name cannot serve arbitrary binaries, otherwise the host name will not be trusted for the same reason that api.nuget.org gets blocked.
```
jeffkl: I wonder if there's any way to enforce this?

jeffkl (dismissed):

accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,110 @@
...
+Information I've been able to find is that NPM uses a security endpoint on the configured package registries, so the same hostname is used for downloading packages and auditing.
+However, NPM sends the package graph to the server, rather than downloading an audit database to check locally, so there are different tradeoffs to consider.
+
+## Unresolved Questions
+
+Should the https://api.nuget.org/v3/index.json service index use the new hostname for the VulnerabilityInfo resource endpoint?
+That's similar for how search works, search is on a different DNS name than api.nuget.org.
+However, the risk is that companies that deny by default and then put api.nuget.org and the search endpoint on an allow list will start having errors if the new audit hostname is being used before the company DNS allow list is updated.
+
+## Future Possibilities
```
donnie-msft, line 113: In VS Options, I could introduce a "Default" button that adds audit.nuget.org, or it could offer both api.nuget.org and audit.nuget.org as options with a link to docs talking about the scenarios/trade-offs. Thoughts?

donnie-msft (dismissed):
I like it! Left a few ideas in comments. 
Some spelling mistakes that I'll leave to Copilot to address...

donnie-msft (dismissed):

nkolev92 (dismissed):

accepted/2026/audit-only-nuget.org-host.md
zivkan: I see that as an implementation detail of the deployment process. Either, the server team has a pre-deployment check that blocks deployment if the precondition is not met. Or, if the service index is in source control, then a test that runs on the pull request CI pipeline to validate.

Enforcement doesn't make sense from a client point of view. Partly because it's perfectly valid to use a package source as an audit source.  Well, unless we add something new into the client. For example, if you add `auditOnly="true"` on your audit source, then client could throw if the service index had anything other than the vulnerability resource. But trying to put myself into the position of working in security, I'm going to block the dns name if I don't trust the host in order to avoid developers accidentally violating company policy, so client tooling validation doesn't seem like it really helps anything.

accepted/2026/audit-only-nuget.org-host.md
zivkan: I don't understand your comment, as it's currently written. If api.nuget.org is blocked, it doesn't matter what URL its VulnerabilityInfo resource uses, because within my company, I can't use api.nuget.org's service index.

But if a company allows api.nuget.org, I figure they'll also allow audit.nuget.org.

The nuget.org server team could possibly reduce deployments if api.nuget.org's VulnerabilityInfo resource points to audit.nuget.org because they'd need to deploy the vulnerability database to one place only.  If api.nuget.org's service index keeps pointing to audit.nuget.org, they'd either need to deploy the vulnerability database to two locations, or have complex CDN configuration to get the vulnerability pages from the same location, while having different content for the service index and vulnerability index files.

In any case, it's really a server team implementation detail and doesn't change the functional aspect of the spec in any way.

accepted/2026/audit-only-nuget.org-host.md
zivkan: It's a very interesting idea, but I think it's orthogonal to this spec's feature. So, for this pull request I'm going to say it's out of scope. But I think it's definitely a feature that we should consider.

accepted/2026/audit-only-nuget.org-host.md
donnie-msft: > Should the https://api.nuget.org/v3/index.json service index use the new hostname for the VulnerabilityInfo resource endpoint?

Maybe you can clarify what this unresolved question means? As I mentioned, I thought it was to support blocking api.nuget.org but allow the 2nd endpoint for audit.nuget.org.
I understand it's just a question, but if it's going to be included in the spec, I think it should be clear why we're putting it there and what the trade-off is if we don't resolve the question.

accepted/2026/audit-only-nuget.org-host.md
donnie-msft: You followed-up offline and mentioned there were some secure supply chain docs already in this area. What do you think about including that link in the spec for reference?

accepted/2026/audit-only-nuget.org-host.md
joelverhagen: I think it would make a lot of sense to point api.nuget.org/v3/index.json to audit.nuget.org for `VulnerabilityInfo `. Less stuff for us to maintain. Great idea!

accepted/2026/audit-only-nuget.org-host.md
```diff
@@ -0,0 +1,111 @@
...
+If the new hostname is hosted on the same CDN that api.nuget.org is hosted on, then they're likely to resolve to the same IP address, and therefore an IP-level block will block both.
+
+However, I think that IP-level blocks are less likely than DNS-level blocks for the following reasons:
+
+- nuget.org does not publish what IP addresses the service is available on ([currently?](https://github.com/NuGet/NuGetGallery/issues/8883))
+- CDNs will resolve the hostname to different IP addresses in different geographic regions
+- CDNs might change the IP address within the same geographic region, either intentionally, or a developer's machine might take a different network path to the CDN.
+- CDNs might use the same IP address to host multiple websites, so blocking api.nuget.org's website can block other websites that the company wants to use
+
+## Rationale and alternatives
```
joelverhagen, line 86: Could you expand on why this concern is not deferred to the 3P package source that is being used by the customer? It is supported to Azure Artifacts in some capacity (I saw it show up on public Artifact feeds mid October of 2025). We spoke about it on the call but this context might be useful for the community or for history.

joelverhagen (dismissed):

accepted/2026/audit-only-nuget.org-host.md
zivkan: I'm updating the spec locally. Hopefully I'll push a commit soon. But can you elaborate on what Azure Artifacts functionality you see?

When nuget.org is configured as an upstream source, they get registration pages (package metadata resource) from nuget.org and they pass through any vulnerability data that way. So, it shows up in VS's PM UI. But restore (NuGet Audit) doesn't use registration.  To the best of my knowledge, there isn't any service index VulnerabilityInfo resource support available at this time.

joelverhagen: https://github.com/joelverhagen/data-nuget-service-index/commit/c623ae50aa970a7decd8719218fd62662bf59542
On October 14, this resource appeared on a public dnceng feed. I think it was turned off later. I am assuming this is either because of an Artifacts rollout or someone turning on a feature toggle.
I am imagining other customers may have it on too.

zivkan: Interesting that it was available briefly on this Microsoft run public feed. However, the feature is not available to non-Microsoft Azure Artifacts customers. Searching the Azure DevOps blog and public docs, I can't find any public information, so I think it hasn't been publicly announced. Therefore, I think it would be a violation of my employment contract NDA to say anything more 😁

joelverhagen: Sounds good.

accepted/2026/audit-only-nuget.org-host.md
zivkan: > Maybe you can clarify what this unresolved question means?

Honestly, I realize now that it was mostly a way for me to share an implementation idea with the server team. I don't think it belongs in the feature spec, so I've removed it.

joelverhagen (approved):

nkolev92 (approved):

Issue #8601, opened 2019-09-17
Revisit deterministic pack


nkolev92:
The pack implementation for deterministic was reverted in 5.3 due to https://github.com/NuGet/Home/issues/8599. 

This issue tracks re-enabling that feature. 

Important: 

The implementation of this feature should account for https://github.com/NuGet/Home/issues/7001. 
Consider the #8603 scenario.

nkolev92:
Important scenario in https://github.com/NuGet/Home/issues/8603, hit because of https://github.com/NuGet/Home/issues/7001. 

tl;dr; 1980-1-1 might not be the best standard date because of timezones.

poke:
I personally think that having a _real_ date is somewhat useful. Since NuGet probably won't be able to figure out a deterministic _real_ date for a package, I think having a way to externally set the date would be useful.

@tmat said in the dotnet issue the following:
> Making the timestamp settable just shifts the problem elsewhere but it does not address the problem.

But I actually think that's exactly a valid way to solve this. NuGet does not have this information, nor has MSBuild. But there are systems that _do_ have this kind of information: version control systems.

When generating a deterministic build and pack output, I would expect that to always happen based on a specific version in the VCS. So when packing deterministically, we could pass the commit date from the source revision the build is based on. That way, there would be a fixed and deterministic date (1) that makes sense (because it's real), (2) and that's even able to identify the source a pack is based on.

I could see that actually figuring out that date is not a responsibility for NuGet (since it would have to interact with the VCS then), so the default could be a different behavior. But then I would really like there to be some way to externally set the date during pack, so that e.g. a build process can pass the revision date to the pack step.

olivier-spinelli:
Yes, having a real date is somewhat useful, and yes the commit date is the best candidate (when a VCS is used), otherwise the date can be derived from... whatever you can imagine.

Honnestly a simple **PackageDate[Time]Utc** or a more general **BuildDate[Time]Utc** property that could be initialized with UtcNow seems perfect to me... 

...or did I miss something huge?

tmat:
I don't see why is there a need for timestamps on the items zipped in the NuGet package. Why wouldn't the version stored in these files be sufficient/better.
NuGet packages already have commit SHA metadata from VCS, so it is easy to find out what source snapshot were they built from.

nkolev92:
@olivier-spinelli Using UTC now would break determinism. 

[...]

PR #14785, opened 2026-03-03
Add a design doc for deterministic pack

omajid:


omajid:
cc @nkolev92 @zivkan

accepted/2026/deterministic-pack.md
```diff
@@ -0,0 +1,244 @@
...
+Bit-by-bit reproducible nuget packages are more important than they might
+appear at first because the hash of a nuget package is added to the deps.json
+of a .NET application that depends on the nuget package. In other words, one
+package not being reproducible can impact things that build on it.
+
+This is also one of the pieces needed to make the .NET SDK build itself
+deterministic and reproducible. For more details, see
+https://github.com/dotnet/source-build/issues/4963
+
+Note: this proposal uses "deterministic" and "reproducible" and interchangeably
```
Frulfump: Duplicate and
```suggestion
Note: this proposal uses "deterministic" and "reproducible" interchangeably
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,244 @@
...
+
+This is also one of the pieces needed to make the .NET SDK build itself
+deterministic and reproducible. For more details, see
+https://github.com/dotnet/source-build/issues/4963
+
+Note: this proposal uses "deterministic" and "reproducible" and interchangeably
+as synonyms. .NET uses the term deterministic. The wider ecosystem uses
+reproducible too.
+
+For more information on deterministic and reproducible builds, see:
```
Frulfump, line 48: There's also the dotnet reproducible build repo. https://github.com/dotnet/reproducible-builds

accepted/2026/deterministic-pack.md
```diff
@@ -0,0 +1,244 @@
...
+Abstractly, being deterministic or reproducible simply means the same inputs
+produces the same outputs.
+
+This proposal is for moving packages built by NuGet.Client from being
+non-deterministic to being deterministic in package contents, and optionally to
+being bit-by-bit deterministic.
+
+### Functional explanation
+
+<!-- Explain the proposal as if it were already implemented and you're teaching it to another person. -->
```
Frulfump: Not sure if these comments should be left in

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,244 @@
...
+- We enabled this in the past with fixed 1980-based timestamps. Customers
+  reported deployments failing. Their tools used the time to determine whether
+  a file was newer, which returned bad results with fixed timestamps. For more
+  details see [#3388](https://github.com/dotnet/core/issues/3388). To mitigate
+  this, this proposal uses real wall-clock-based-timestamps by default, and
+  allows developers to override it.
+
+- This feature makes the code more complex.
+
+- The implicit use of `SOURCE_DATE_EPOCH` as an environment variable can lead
```
baronfel, line 201: If this enviroment variable is read during evaluation or during the build execution, it will be present in the 'environment variables' section of the binlog. This isn't immediately 'in your face' to end users, but it is observable and verifiable.

accepted/2026/deterministic-pack.md
```diff
@@ -0,0 +1,244 @@
...
+   - For msbuild project files:
+
+     Use the property `Deterministic`. For example:
+
+     `<Deterministic>true</Deterministic>`
+
+     This is already the default for recent versions of .NET, at least as far
+     as .NET Core 3.0.
+
+   - For `nuget`:
```
zivkan: We usually refer to it as NuGet.exe. I understand that on Linux & Mac a script might be named `nuget` to launch NuGet.exe with mono, but most, if not all, of our docs on learn.microsoft.com/nuget refer to it as NuGet.exe or nuget.exe.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,244 @@
...
+  sources, reducing risks of backdoors and other vulnerabilities.
+
+- Auditing and Compliance: It becomes easier to verify that the sources and
+  binaries match for compliance reasons.
+
+Users are already trying to make packages deterministic on their own; providing
+a first class feature will make it easier for them and everyone else to adapt
+this too.
+
+Bit-by-bit reproducible nuget packages are more important than they might
```
zivkan, line 35: I'm not sure where the best place in the doc is to put it, but I think it would be worthwhile adding a short section somewhere, perhaps technical explanation or drawbacks, saying that package signing breaks the possibility of bit by bit reproducable, due to embedded timestamps. However, NuGet has the concept of "content hash" that is not affected by package signatures, and should still be able to compare two builds of a nupkg. A package's content hash (although probably base64 encoded, rather than showing the hex encoding) can be viewed with `dotnet nuget verify`, from .NET 10.0.100

zivkan (dismissed):

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks!

accepted/2026/deterministic-pack.md
omajid: Thanks!

accepted/2026/deterministic-pack.md
omajid: Thanks. Removed.

accepted/2026/deterministic-pack.md
omajid: Thanks! I renamed the references to this tool as `NuGet.exe`, and left the CLI commands as `nuget` following what I see in https://learn.microsoft.com/en-us/nuget/reference/cli-reference/cli-ref-pack

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks!

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks.

omajid:
I added a note about the `PackageBuilder` API change too.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,263 @@
...
+     <PropertyGroup>
+       <DeterministicTimestamp>$(DATE_TIME)</DeterministicTimestamp>
+     </PropertyGroup>
+     ```
+
+     If `DeterministicTimestamp` is not set, but `SOURCE_DATE_EPOCH` is set
+     (eg, from environment variable), then `DeterministicTimestamp` is set to
+     the value of `SOURCE_DATE_EPOCH`.
+
+   - For `NuGet.exe:
```
akoeplinger:
```suggestion
   - For `NuGet.exe`:
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,263 @@
...
+
+- This feature makes the code more complex.
+
+- The implicit use of `SOURCE_DATE_EPOCH` as an environment variable can lead
+  to action-at-a-distance issues. To mitigate this, we will reading this
+  variable using msbuild which should make the variable's usage and value
+  available through the binlog.
+
+- Package signing breaks the possibility of bit-by-bit reproducibility, due to
+  embedding a timestamp. Nuget has the concept of a contnet hash, that can
```
akoeplinger:
```suggestion
  embedding a timestamp. NuGet has the concept of a content hash, that can
```

akoeplinger: you could still strip the signature before comparing the packages right? if yes, that doesn't sound like a big deal to me

omajid: Good point. I am worried about the sha hashes in the .deps.json files, though.

akoeplinger: .deps.json shouldn't be in the nuget packages though? or do you mean when you build an app with the sdk?

akoeplinger: ah they could be for tools packages. I wonder how that works today?

omajid: They also end up in the .NET SDK itself:

```
$ find /usr/lib64/dotnet -iname '*.deps.json'
/usr/lib64/dotnet/packs/Microsoft.AspNetCore.App.Runtime.fedora.43-x64/10.0.3/runtimes/fedora.43-x64/lib/net10.0/Microsoft.AspNetCore.App.deps.json
/usr/lib64/dotnet/packs/Microsoft.NETCore.App.Runtime.fedora.43-x64/10.0.3/runtimes/fedora.43-x64/lib/net10.0/Microsoft.NETCore.App.deps.json
/usr/lib64/dotnet/shared/Microsoft.NETCore.App/10.0.3/Microsoft.NETCore.App.deps.json
/usr/lib64/dotnet/shared/Microsoft.AspNetCore.App/10.0.3/Microsoft.AspNetCore.App.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Containers/tasks/net10.0/Microsoft.NET.Build.Containers.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-dev-certs/10.0.3-servicing.26075.103/tools/net10.0/any/dotnet-dev-certs.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-format/BuildHost-netcore/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-format/dotnet-format.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-user-jwts/10.0.3-servicing.26075.103/tools/net10.0/any/dotnet-user-jwts.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-user-secrets/10.0.3-servicing.26075.103/tools/net10.0/any/dotnet-user-secrets.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-watch/10.0.103/tools/net10.0/any/BuildHost-netcore/Microsoft.CodeAnalysis.Workspaces.MSBuild.BuildHost.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-watch/10.0.103/tools/net10.0/any/dotnet-watch.deps.json
/usr/lib64/dotnet/sdk/10.0.103/DotnetTools/dotnet-watch/10.0.103/tools/net10.0/any/dotnet.deps.json
/usr/lib64/dotnet/sdk/10.0.103/FSharp/fsc.deps.json
/usr/lib64/dotnet/sdk/10.0.103/FSharp/fsi.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Roslyn/bincore/VBCSCompiler.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Roslyn/bincore/csc.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Roslyn/bincore/vbc.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Roslyn/Microsoft.Build.Tasks.CodeAnalysis.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.Build.Tasks.Git/tools/net/Microsoft.Build.Tasks.Git.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.NET.Sdk.BlazorWebAssembly/tools/net10.0/Microsoft.NET.Sdk.BlazorWebAssembly.Tool.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.NET.Sdk.Razor/source-generators/Microsoft.CodeAnalysis.Razor.Compiler.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.NET.Sdk.Razor/tools/rzc.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.NET.Sdk.StaticWebAssets/tools/net10.0/Microsoft.NET.Sdk.StaticWebAssets.Tool.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.SourceLink.AzureRepos.Git/tools/net/Microsoft.SourceLink.AzureRepos.Git.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.SourceLink.Bitbucket.Git/tools/net/Microsoft.SourceLink.Bitbucket.Git.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.SourceLink.Common/tools/net/Microsoft.SourceLink.Common.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.SourceLink.GitHub/tools/net/Microsoft.SourceLink.GitHub.deps.json
/usr/lib64/dotnet/sdk/10.0.103/Sdks/Microsoft.SourceLink.GitLab/tools/net/Microsoft.SourceLink.GitLab.deps.json
/usr/lib64/dotnet/sdk/10.0.103/MSBuild.deps.json
/usr/lib64/dotnet/sdk/10.0.103/NuGet.CommandLine.XPlat.deps.json
/usr/lib64/dotnet/sdk/10.0.103/datacollector.deps.json
/usr/lib64/dotnet/sdk/10.0.103/dotnet.deps.json
/usr/lib64/dotnet/sdk/10.0.103/redist.deps.json
/usr/lib64/dotnet/sdk/10.0.103/testhost.deps.json
/usr/lib64/dotnet/sdk/10.0.103/vstest.console.deps.json
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,263 @@
...
+  packages for reproduciblity and share the live status:
+  https://reproducible-builds.org/citests/
+
+- Some Linux distributions like Fedora expect all distro-packages, including
+  .NET, to be deterministic. This proposal will support that.
+  - https://fedoraproject.org/wiki/Changes/Package_builds_are_expected_to_be_reproducible
+
+## Unresolved Questions
+
+- Is `DeterministicTimestamp` the best name?
```
zivkan: I know other designs in the repo have left unresolved questions before getting merged, but my thought is what happens when someone in 5 years from now looks at the design doc and still sees the unresolved question.  So, I think unresolved questions are great to have discussions in the PR, but I like to see them resolved and removed so the unresolved section is empty before merging the PR.

Otherwise this spec looks good to me.

Another idea for the name is `PackTimestamp`, but since the timestamp is tied to the `Deterministic` setting, I think `DeterministicTimestamp` is better, even though it's longer and much harder to type.

omajid: > I like to see them resolved and removed so the unresolved section is empty before merging the PR.

:100:

omajid: Another proposal was `DeterminisiticPackTimestamp` but that is even longer.

akoeplinger: I like `DeterministicTimestamp`, it's also better if we later on need to use this in scenarios outside of nuget pack

accepted/2026/deterministic-pack-revisited.md
akoeplinger: I was thinking of the hash in the .deps.json, we sign the .dlls in the Microsoft build which presumably changes the hash which was recorded in .deps.json at build time?

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks. I have removed this unanswered question, and the entire section too.

omajid:
Please feel free to merge if this looks okay now.

accepted/2026/deterministic-pack-revisited.md
omajid: I don't know the answer. This will need more investigation.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,264 @@
...
+     This will accept a string-ified version of `{DATE_TIME}`.
+
+   `DeterministicTimestamp` must be either a full date/time string specified in
+   the RFC3339 format, or a single number indicating the number of seconds
+   since the unix epoch (`Jan 1 1970, 00:00:00 UTC`). The behaviour is
+   unspecified if the value can not be parsed.
+
+### Technical explanation
+
+- `Deterministic` is a boolean. When `Deterministic` is set to true, a single
```
nkolev92: fwiw, the technical implementation is probably too much in this case, and not something we'd normally need unless there were some special things to consider.  I'm sorry you had to go through writing this part.
I think in this case, all of the explanation functional.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,264 @@
...
+Abstractly, being deterministic or reproducible simply means the same inputs
+produces the same outputs.
+
+This proposal is for moving packages built by NuGet.Client from being
+non-deterministic to being deterministic in package contents, and optionally to
+being bit-by-bit deterministic.
+
+### Functional explanation
+
+From an implementation point of view there are 3 levels to deterministic-ness
```
nkolev92, line 67: So the summary is that we're proposing respecting the same deterministic property that the build supports, thus making deterministic enabled for all new packages generated with the new tooling correct?

I'd have included that in the summary. 

The by default part matters, since we need to communicate that to everyone.

nkolev92 (commented):
One clarifying question about the default. 

Worth consider whether we enable this by default in 10.0.300/10.0.400 or 11.0.100.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,264 @@
...
+         init { ... }
+     }
+     ```
+
+     This will accept a string-ified version of `{DATE_TIME}`.
+
+   `DeterministicTimestamp` must be either a full date/time string specified in
+   the RFC3339 format, or a single number indicating the number of seconds
+   since the unix epoch (`Jan 1 1970, 00:00:00 UTC`). The behaviour is
+   unspecified if the value can not be parsed.
```
akoeplinger: > The behaviour is unspecified if the value can not be parsed.

This is meant to say that the value will be ignored right? not in the C++ undefined behavior sense where it's up to the implementation what happens

accepted/2026/deterministic-pack-revisited.md
omajid: I see. Is there some user-facing document that I need to update with the possible values of `Deterministic` and `DeterministicTimestamp`?

accepted/2026/deterministic-pack-revisited.md
omajid: I did mean [unspecified behaviour](https://eel.is/c++draft/defns.unspecified) in the C++ way, giving us wiggle room to either throw an exception or ignore the input. I didn't mean the [undefined behaviour](https://eel.is/c++draft/defns.undefined) like C++, which allows anything to happen, even run`rm -rf`.

I can switch this to a specified implementation. We can throw exceptions (easier/no-op to implement, good feedback to user) or the ignore the value (which is perhaps more flexible for complex projects). Which would you prefer?

akoeplinger: I'd prefer to be explicit rather than unspecified, but I don't have a strong preference on exception vs. ignore so I'll let the nuget team decide :)

omajid:
> Worth consider whether we enable this by default in 10.0.300/10.0.400 or 11.0.100.

My primary focus is on making the .NET SDK itself reproducible. To do that, I will need to consume this in arcade. I don't think arcade can depend on 10.0.3xx or 10.0.4xx. So I am okay with targeting this to 11.0.100.

accepted/2026/deterministic-pack-revisited.md
omajid: Done.

nkolev92:
The question that I have is whether it'll be enabled by default. 

Does everyone that's installing the .NET 11 SDK now get the new deterministic behavior by default given that it is set it the projects. 

That's the part that needs documented/announced and probably the thing that would not meet the 10.0.* bar.

accepted/2026/deterministic-pack-revisited.md
nkolev92: Yes, 

https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/msbuild-targets.md and https://github.com/dotnet/docs/blob/main/docs/core/tools/dotnet-pack.md
https://github.com/NuGet/docs.microsoft.com-nuget/blob/main/docs/reference/cli-reference/cli-ref-pack.md

I might be missing something, but I *think* that's all.

nkolev92 (dismissed):
LGTM. 

To me 11.0.100 as default make sense, and maybe something we can include in the spec, but not too worried as long as the documentation makes sense. SDKAnalysisLevel is what we can use in this case in the pack targets.

It's also worth considering how we're changing the legacy pack default now that it's auto imported in VS too.  This is on the client team though.

akoeplinger:
> The question that I have is whether it'll be enabled by default.

That's basically explained in the "Functional explanation" section. With Deterministic=true (default in most projects) some previously random GUID-based file names like `<foo.nupkg>/package/services/metadata/core-properties/<GUID>.psmdcp` will now be deterministic. This is very low risk.

The slightly more risky change of having deterministic timestamps always requires opting in via `DeterministicTimestamp`/`SOURCE_DATE_EPOCH` even for 11.0+

nkolev92:
Thanks for the clarification. 

Honestly, package/services/metadata/core-properties is probably good enough to do even when deterministic is not requested, since we took https://github.com/NuGet/NuGet.Client/pull/6963. 

Either way, LGTM

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,243 @@
...
+   Some things that help make nuget packages more deterministic are already
+   enabled and the default in already-released versions of NuGet.Client. Some
+   can't be turned off. For example, the order of files within the nuget
+   package archive is already fully deterministic
+   ([NuGet.Client#6963](https://github.com/NuGet/NuGet.Client/pull/6963)).
+
+1. New: Enable more determinism by default
+
+   This includes using a single timestamp for all files added to the archive.
+   Currently, this also includes the making the names of psmdcp files
```
Frulfump: Extra "the"
```suggestion
   Currently, this also includes making the names of psmdcp files
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,243 @@
...
+
+     Use the `-Deterministic` argument. For example:
+
+     `nuget pack packageA.nuspec -Deterministic`
+
+     The `-Deterministic` flag is **not** the default.
+
+   - For `PackageBuilder` API:
+
+     There is a **behavioural change** in the `NuGet.Packaging.PackageBuilder`
```
Frulfump: Not sure but I think the repo uses American English


```suggestion
     There is a **behavioral change** in the `NuGet.Packaging.PackageBuilder`
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -77,9 +77,12 @@ in NuGet:
 
 1. New: Enable more determinism by default
 
-   This includes the names of psmdcp files, which are otherwise random and
-   based on a GUID. The names will now be based on a hash that's going to be
-   deterministic.
+   This includes using a single timestamp for all files added to the archive.
```
nkolev92: This is now suggesting we're making unified timestamp by default, but I don't tihnk that's the case.

omajid: This is the plan, indeed: https://github.com/NuGet/Home/issues/8601#issuecomment-3874876706

I forgot to spell this out in the doc earlier.

omajid: In other words, everything will have the current wall-clock-time as the default timestamp.

With `DeterministicTimestamp` also set, that will be used as the timestamp instead of the current-wall-clock-time.

nkolev92: So this isn't correct then: 

https://github.com/NuGet/Home/pull/14785#issuecomment-4055173024

nkolev92: I think Andy's comment is clearer. 

IMO this is now confusing since it's not necessarily deterministic if it chooses current time. 

We're not aligned in the exact language, but aligned in the UX.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,243 @@
...
+- Author Name: [omajid](https://github.com/omajid)
+- GitHub Issue: [#8601](https://github.com/NuGet/Home/issues/8601)
+
+## Summary
+
+Make NuGet packaging more deterministic by default. NuGet will start respecting
+the existing `Deterministic` property, which already defaults to `true`.
+
+Optionally, support making NuGet packages bit-by-bit deterministic and
+reproducible through the new `DeterministicTimestamp` property.
```
zivkan, line 12: I was thinking about this, `DeterministicTimestamp` might not make sense unless you already know about pack, and are thinking about it. Consider someone new to .NET and sees `DeterministicTimestamp` in a Directory.Build.props or a build script.

So, maybe `PackTimestamp` makes more sense.

zivkan (dismissed):

accepted/2026/deterministic-pack-revisited.md
akoeplinger: if you're not familiar with it you'd need to google both terms anyway. I like DeterministicTimestamp better since it's less tied to nuget and could be used in other parts of the stack too if necessary (e.g. timestamps in msbuild ZipDirectory task)

baronfel: I agree about the more general naming - it's very likely that future determinism work will need the ability to set timestamps in generated files, and having one unified property (and maybe task-specific sub-properties _if necessary_) makes that easier.

accepted/2026/deterministic-pack-revisited.md
omajid: Sorry, I am not sure what you would like to see changed.

> IMO this is now confusing since it's not necessarily deterministic if it chooses current time.

It can't be fully deterministic without setting the time, which we don't want to do by default. But this still *increases* determinism. And like Andy says, if you can find what time was used in the previous pack operation, you can set the same value and hopefully get a fully matching build again.

akoeplinger: I guess one question is whether we want a way to revert back to the previous behavior of using File.LastWriteTime in case we discover any unforeseen issue with using DateTime.UtcNow (e.g. by setting  DeterministicTimestamp=false)

nkolev92: > Sorry, I am not sure what you would like to see changed.

No worries. 

What I'd add in this sentence is what you said here in the comments, that the unifying timestamp is current wall-clock time. 

>    This includes using a single timestamp for all files added to the archive.

omajid: > whether we want a way to revert back to the previous behavior of using File.LastWriteTime in case we discover any unforeseen issue with using DateTime.UtcNow (e.g. by setting DeterministicTimestamp=false)

Are you suspecting that `DateTime.UtcNow` is not available/reliable, or that there is a risk of using the utc time itself?

Would you be okay with leaving this out until an implementation of this is used throughout the .NET SDK? And then based on our findings we make a decision? Or would you prefer to be cautious in advance and implement it right now?

omajid: > What I'd add in this sentence is what you said here in the comments, that the unifying timestamp is current wall-clock time.

Thanks, done.

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks!

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,246 @@
...
+   package archive is already fully deterministic
+   ([NuGet.Client#6963](https://github.com/NuGet/NuGet.Client/pull/6963)).
+
+1. New: Enable more determinism by default
+
+   This includes using a single timestamp, based on the current wall clock
+   time, for all files added to the archive. See the `DeterministicTimestamp`
+   property below on how to override the timestamp.
+
+   Currently, this also includes the making the names of psmdcp files
```
Frulfump: Extra "the"
```suggestion
   Currently, this also includes making the names of psmdcp files
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,246 @@
...
+
+   This is tied to the existing `Deterministic` property. Use it like this:
+
+   - For `dotnet pack`:
+
+     Use the `/p:Deterministic=true` argument. For example:
+
+     `dotnet pack /p:Deterministic=true`
+
+     This property is already the set to `true` in recent versions of .NET, at
```
Frulfump: Extra "the"
```suggestion
     This property is already set to `true` in recent versions of .NET, at
```

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks!

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks!

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,246 @@
...
+  It becomes easier to verify items like SBOMs.
+
+- Increasing user trust: Users can trust the binaries they have matches the
+  sources, reducing risks of backdoors and other vulnerabilities.
+
+- Auditing and Compliance: It becomes easier to verify that the sources and
+  binaries match for compliance reasons.
+
+Users are already trying to make packages deterministic on their own; providing
+a first class feature will make it easier for them and everyone else to adapt
```
MattKotsenas: nit: typo

```suggestion
a first-class feature will make it easier for them and everyone else to adopt
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,246 @@
...
+- Some Linux distributions like Fedora expect all distro-packages, including
+  .NET, to be deterministic. This proposal will support that.
+  - https://fedoraproject.org/wiki/Changes/Package_builds_are_expected_to_be_reproducible
+
+## Future Possibilities
+
+- With `Deterministic=true` by default, support for `Deterministic=false` could
+  be fully dropped, and the code paths simplified.
+
+- Should turning off `Deterministic=true` produce warnings or errors?
```
MattKotsenas, line 254: Feels like a natural place to start with a BuildCheck and then escalate over time?

nkolev92 (dismissed):

accepted/2026/deterministic-pack-revisited.md
Frulfump: Interesting idea, I do wonder if we could get numbers from telemetry how often Build Checks are ran, I feel like that possibility(option) isn't advertised like at all. So what I'm getting at is that it might be used too rarely to be a good guard (should probably still include it but just wanted to highlight the "fact")

accepted/2026/deterministic-pack-revisited.md
akoeplinger: > Are you suspecting that DateTime.UtcNow is not available/reliable, or that there is a risk of using the utc time itself?

The latter. Basically what if you had a file that was last modified 5 years ago which was preserved before but now when it gets packed it has UtcNow timestamp instead.

> Would you be okay with leaving this out until an implementation of this is used throughout the .NET SDK? And then based on our findings we make a decision? Or would you prefer to be cautious in advance and implement it right now?

My gut feeling says it's better to have some escape hatch to get back to the old behavior, otherwise we'd risk needing to revert the whole thing if we find something unexpected. But it's up to the nuget team to decide.

accepted/2026/deterministic-pack-revisited.md
omajid: @nkolev92 @zivkan Any thoughts?

accepted/2026/deterministic-pack-revisited.md
omajid: @nkolev92 @zivkan what do you think?

accepted/2026/deterministic-pack-revisited.md
omajid: Sounds like we have a rough consensus on `DeterministicTimestamp`, then?

accepted/2026/deterministic-pack-revisited.md
omajid: Thanks. I have added this to my todo list.

accepted/2026/deterministic-pack-revisited.md
nkolev92: I prefer to fail the operation. 
Thinking you've configured something, but it not really doing what you thought it'd do can be pretty infuriating to users.

accepted/2026/deterministic-pack-revisited.md
nkolev92: I thinking having a revert makes sense. 

If I'm reading it correctly, isn't the only new thing that `Deterministic` does is unify the timestamps?

Isn't that the switch to revert things?

>  Basically what if you had a file that was last modified 5 years ago which was preserved before but now when it gets packed it has UtcNow timestamp instead.

Yeah, this was the primary thing. I think @zivkan had the conversations and did some analysis here.

akoeplinger: > If I'm reading it correctly, isn't the only new thing that Deterministic does is unify the timestamps?

yes but given `Deterministic=true` is the default since .net core 3.0 and influences other places in the SDK/compiler I'm not sure turning that one off is a good escape hatch.

What if we had:

- `DeterministicTimestamp=true`: default fallback value in the net11 sdk, uses DateTime.UtcNow
- `DeterministicTimestamp=<explicit value>` or `SOURCE_DATE_EPOCH` set in the environment: optin to the fixed timestamp
- `DeterministicTimestamp=false`: escape hatch to turn the feature off

omajid: Added to spec.

nkolev92: I don't mind that. 

I was gonna suggest a DeterministicPack option potentially too in case someone wants to decouple it from the general determinism, but I think the timestamp approach is plenty powerful.

nkolev92 (approved):

omajid:
Sounds like this is good to merge now?

akoeplinger (approved):

accepted/2026/deterministic-pack-revisited.md
omajid: I slept on this, but now I am wondering if it might be better to have a few literals other than `true` and `false`:

- `DeterministicTimestamp=wall-clock-time`  replaces `true`. It's more explicit about what it's doing
- `DeterministicTimestamp=file-time` to replace `false`. It's more explicit about the behaviour.
- `DeterministicTimestamp=<explicit value>` or `SOURCE_DATE_EPOCH` handled as before

Using some fixed literals makes it easier to understand and also allows us to add more knobs if we have to down the line. The values above are just suggestions, I am happy to use another convention/style to make it more consistent with other things.

Is there precedence for this in nuget or .NET ?

omajid:
What are the next steps here? Can this be merged? Or are there any other thoughts/feedback?

nkolev92:
Sorry we've been busy with some other things. 
We're just waiting on some folks to be back and we'll get back to you.

omajid:
Okay, thanks for the update!

accepted/2026/deterministic-pack-revisited.md
nkolev92: I don't know of too many precedents like that, which is why I had mentioned hinging it off of DeterministicPack potentially, so having 2 different options. 

DeterministicPack = true is DeterministicTimestamp=wall-clock
DeterministicPack = false is DeterministicTimestamp=file-time
DeterministicTimestamp can only take an explicit value and it requires DeterministicPack to be true.

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+  sources, reducing risks of backdoors and other vulnerabilities.
+
+- Auditing and Compliance: It becomes easier to verify that the sources and
+  binaries match for compliance reasons.
+
+Users are already trying to make packages deterministic on their own; providing
+a first-class feature will make it easier for them and everyone else to adopt
+this too.
+
+Bit-by-bit reproducible nuget packages are more important than they might
```
Frulfump, line 35:
```suggestion
Bit-by-bit reproducible NuGet packages are more important than they might
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+optionally to being bit-by-bit deterministic.
+
+### Functional explanation
+
+From an implementation point of view there are 3 levels to deterministic-ness
+in NuGet:
+
+0. Always enabled and already the default.
+
+   Some things that help make nuget packages more deterministic are already
```
Frulfump, line 72:
```suggestion
   Some things that help make NuGet packages more deterministic are already
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+     This property is already set to `true` in recent versions of .NET, at
+     least as far back as .NET Core 3.0.
+
+   - For msbuild project files:
+
+     Use the property `Deterministic`. For example:
+
+     `<Deterministic>true</Deterministic>`
+
+     This is property is already set to `true` in recent versions of .NET, at
```
Frulfump, line 108:
```suggestion
     This property is already set to `true` in recent versions of .NET, at
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+     `Deterministic` is explicitly set to `false`, this will not be used.
+
+   `DeterministicTimestamp` must be one of:
+
+     - The literal string `true` to indicate using `DateTime.UtcNow` as the
+       timestamp for all files added to the archive.
+     - The literal string `false` to use the original file modification times.
+     - A full date/time string specified in the [RFC
+       3339](https://www.rfc-editor.org/rfc/rfc3339) format
+     - A a single number indicating the number of seconds since the unix epoch
```
Frulfump, line 185:
```suggestion
     - A single number indicating the number of seconds since the Unix epoch
```

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+
+- This feature makes the code more complex.
+
+- The implicit use of `SOURCE_DATE_EPOCH` as an environment variable can lead
+  to action-at-a-distance issues. To mitigate this, we will read this variable
+  using msbuild which should make the variable's usage and value available
+  through the binlog.
+
+- Package signing breaks the possibility of bit-by-bit reproducibility, due to
+  embedding a timestamp. Nuget has the concept of a content hash, which can
```
Frulfump, line 207:
```suggestion
  embedding a timestamp. NuGet has the concept of a content hash, which can
```

Frulfump (commented):
Just some typo nits

accepted/2026/deterministic-pack-revisited.md
```diff
@@ -0,0 +1,254 @@
...
+- We considered making what is now called `DeterministicTimestamp`
+  automatically infer the timestamp of the last commit in the source repository
+  through a change in sourcelink. This was deemed as too much code to maintain
+  for a small benefit: https://github.com/dotnet/sourcelink/pull/1552.
+
+- As an alternative, we can simply not implement this. This would make .NET's
+  security story and positioning weaker than many other programming language
+  stacks.
+
+## Prior Art
```
kartheekp-ms, line 227: Thanks for the spec. I am curious how deterministic pack compares to [source link](https://learn.microsoft.com/dotnet/standard/library-guidance/sourcelink) technology. I think it might be   helpful if we explain the similarities and differences between these 2 options in the prior art.

omajid: source-link is about associating binaries to sources. It involves adding enough metadata to the built binaries to allow tools to figure out what sources were used to build this binary. There's no guarantee/evidence that this information is correct.

deterministic builds (of which deterministic pack is one part) are about producing the same binaries when the same sources are rebuilt. There's no additional metadata in the build itself for this; it's up to the build runner (eg, CI, or official build machines) to record and make that data available separately.

I believe these are orthogonal. source-link makes debugging easier. deterministic builds provide security (including some defence against a subset of supply chain attacks).

They appear related because they both talk about matching sources and binaries. Unfortunately, source-link only claims that a certain set of source code was used to build something, while deterministic builds are about verifying that claim. On the other hand, deterministic builds don't directly help debugging.

zivkan (approved):
Sorry about the very long delays. I was working on something high priority that took much longer than expected, and I was on vacation for a few weeks as well. I'm now catching up on other things.

omajid:
Thank you for working through this with me and accepting this change!

Issue #14762, opened 2026-02-17
List package should use TargetFramework property as a key it its output instead of the effective TargetFramework; Introduce a V2 version of the list package output


nkolev92:
The TargetFramework property has been an alias since .NET 5. 
That means you could author projects with a target framework value like apple.

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>apple</TargetFrameworks>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <PropertyGroup>
    <TargetFrameworkIdentifier Condition=" '$(TargetFramework)' == 'apple' ">.NETCoreApp</TargetFrameworkIdentifier>
    <TargetFrameworkVersion Condition=" '$(TargetFramework)' == 'apple' ">v10.0</TargetFrameworkVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="NuGet.Configuration" Version="6.4.20" Condition=" '$(TargetFramework)' == 'apple' "/>
  </ItemGroup>

</Project>
```

With #5154, this is becoming even more important and pretty much necessary. 

This will be a change in the table output where we'll go from say the following for a maui project to: 

```console
Project 'maui' has the following package references
   [net9.0-android35.0]:
   Top-level Package                            Requested    Resolved
   > Microsoft.Extensions.Logging.Debug         10.0.0       10.0.0
   > Microsoft.Maui.Controls                    9.0.120      9.0.120
   > Microsoft.NET.ILLink.Tasks           (A)   [9.0.12, )   9.0.12

   [net9.0-ios26.0]:
   Top-level Package                            Requested    Resolved
   > Microsoft.Extensions.Logging.Debug         10.0.0       10.0.0
   > Microsoft.Maui.Controls                    9.0.120      9.0.120
   > Microsoft.NET.ILLink.Tasks           (A)   [9.0.12, )   9.0.12

   [net9.0-maccatalyst26.0]:
   Top-level Package                            Requested    Resolved
   > Microsoft.Extensions.Logging.Debug         10.0.0       10.0.0
   > Microsoft.Maui.Controls                    9.0.120      9.0.120
   > Microsoft.NET.ILLink.Tasks           (A)   [9.0.12, )   9.0.12

   [net9.0-windows10.0.19041]:
   Top-level Package                         Requested   Resolved
[...]

PR #14786, opened 2026-03-03
Add proposal for the list package output in aliasing scenarios

nkolev92:
Design for https://github.com/NuGet/Home/issues/14762. 

Update to the general spec. 

This is pre-req to the implementation of https://github.com/NuGet/NuGet.Client/pull/7146.

@aortiz-msft 

In here we talk about the breaking change considerations to the different outputs.

### DIFF
--- a/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
+++ b/accepted/2025/Multiple-Equivalent-Framework-Support-TFM-As-Aliases.md
@@ -457,12 +457,12 @@ The actual name is under consideration in all these options.
 
 | Option | Changes | Details | Considerations |
 | ------ | ------- | ------- | -------------- |
-| 1 | Add `targetFramework` property. Change the schema version to 2. When needed in .NET 10, default in .NET 11. | The new property allows for disambiguation in aliasing scenarios. | Should the name be something else? Does the schema version even matter? |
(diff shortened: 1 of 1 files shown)


Nirmal4G:
Don't you have `targetAlias` in some other lock file? Why not keep the `framework` key same on both V1 and V2 and add `targetAlias` as the new key/property on V2?

As a dev/contributor, you'd instantly recognize and as a user, it'd be a less cognitive load when coming from older versions and the former benefit applies if you maintain build pipelines.

nkolev92:
> Don't you have targetAlias in some other lock file?

`targetAlias` is part of the assets file, and it's not something meant for the end user, it's a restore intermediary, compared to dotnet list package, which has a json output explicitly for that.

> Why not keep the framework key same on both V1 and V2 and add targetAlias as the new key/property on V2?

I thought the name targetAlias is something that we use internally, but not something a machine readable user of the dotnet list package command should know about.

Nirmal4G:
> it's not something meant for the end user

May be but whatever you guys design on a need to know basis always ends up on people needing to know in order to make their build pipelines work properly on edge cases.

How do you think I came to know about `targetAlias`? And I'm not even a full time .NET dev, you know!?

> I thought the name `targetAlias` is something that we use internally

True but what do you call the custom build target name?

1. `{Target}Alias` – It already exists internally
2. `{Target}Profile` – I don't think we're ready to use that. Too many legacy references.
3. `{Target}Build` – Too generic yet it does match the intent but you can't use it without `Target`.

I would go with 1 since I suspect most of them who have been tinkering with build stuff knows this.

martinrrm (approved):

PR #14793, opened 2026-03-07
Refactor GitOps policies into separate files

donnie-msft:
### Summary

Split the monolithic `resourceManagement.yml` policy file into 7 focused files, organized by concern with a unique ID given to each item across all items under the `policies/` directory.

- Created 7 files, each handling one logical concern
- To make for easier identification/communication, with room to insert new files:
  - Numbered the files with gaps of 10 for easy insertion in the future.
  - Re-numbered the rules with the file using **`[file#-item#]` format** 
- Removed some redundant rules and some contradictory filters (see notes in the table below).

### New file structure in _.github/policies/_
- 10-issue-breaking-change.yml # Breaking change label instructions
- 20-issue-label-reminder.yml # Type label enforcement
- 30-issue-stale-p3-cleanup.yml # Icebox/P3 stale issue cleanup
- 40-issue-transferred-triage.yml # Auto-triage transferred issues
- 50-issue-waiting-for-customer.yml # WaitingForCustomer lifecycle
- 60-pr-stale-community.yml # Community PR labeling + staleness
- 70-pr-stale-general.yml # Team PR staleness

### Rule mapping: old → new

Below is a table summarizing all the old rules and the new IDs.
For the few rules I deleted or change, it's described in the Notes.

#### Scheduled searches

| Old ID | Old description | New ID | Notes |
|---|---|-----|---|
| `[stale PR] [5-1]` | Search for PRs with no activity over 30 days and warn (except draft PRs and Community PRs) | `[70-10]` | Fixed: removed contradictory `hasLabel: Community` filter |
| `[stale PR] [5-5]` | Close PRs with no activity over 15 days after warn (except Community PRs) | `[70-20]` | |
| `[Manage stale WaitingForCustomer issues]` | Search for WaitingForCustomer issues with no activity over 14 days and warn | `[50-10]` | |
| `[Close stale WaitingForCustomer issues]` | Close stale WaitingForCustomer issues with no activity over 14 days | `[50-20]` | |
| `[Type Label Reminder][4-1]` | Remind HotSeat to Add Type Label | `[20-10]` | |
| `[Close stale icebox][3-1]` | Mark inactive issues as icebox cleanup candidates and warn | `[30-10]` | |
| `[Close stale icebox][3-2]` | Close "Icebox cleanup candidate" if no activity within 14 days | `[30-20]` | |
| `[stale community PR] [5-1]` | Search for community PRs with no activity over 30 days and warn | `[60-10]` | |
[...]

PR #14795, opened 2026-03-09
Remove Icebox cleanup candidate label on activity

donnie-msft:
Depends on https://github.com/NuGet/Home/pull/14793


krishnaprasadkartha3-droid (commented):
U

PR #14798, opened 2026-03-10
Proposal: Add --allow-untrusted-signing flag to nuget sign commands

elantiguamsft:
## Summary

Add a new `--allow-untrusted-root` flag to `nuget sign` and `dotnet nuget sign` that allows signing NuGet packages with certificates whose root CA is not installed in a trusted root certificate store. When this flag is set, the `UntrustedRoot` X509 chain status is treated as a warning instead of a fatal error. The certificate chain is still fully built and validated for structural correctness.

## Motivation

In enterprise and CI/CD environments, signing certificates are often provisioned from Azure Key Vault where the root CA is not (and should not be) installed into the machine's trusted root certificate store. Installing root CAs requires admin elevation, which fails in non-elevated CI containers and triggers UAC prompts on developer machines.

Today, `nuget sign` performs internal chain validation that treats `UntrustedRoot` as a fatal error with no workaround. Other signing tools (signtool.exe, VsixSigntool) already handle this scenario.

## Implementation

A reference implementation is available in [NuGet/NuGet.Client#7201](https://github.com/NuGet/NuGet.Client/pull/7201). The approach:

- Adds `AllowUntrustedRoot` boolean to `SignArgs`, `CertificateSourceOptions`, and `SignPackageRequest`
- New `GetCertificateChain` overload with `allowUntrustedRoot` parameter
- Moves `UntrustedRoot` from error flags to warning flags in `GetChainStatusFlags()`
- Works on **net472** (nuget.exe) and net8.0+ (dotnet nuget sign)
- CLI: `--AllowUntrustedSigning` (nuget.exe) / `--allow-untrusted-signing` (dotnet CLI)

See the full proposal in `accepted/2026/allow-untrusted-signing.md`.

### DIFF
--- a/accepted/2026/allow-untrusted-root.md
+++ b/accepted/2026/allow-untrusted-root.md
@@ -1,11 +1,11 @@
-# Add --allow-untrusted-signing flag to NuGet sign commands
+# Add --allow-untrusted-root flag to NuGet sign commands
 
 - [@elantiguamsft](https://github.com/elantiguamsft)
 - [NuGet/NuGet.Client#7201](https://github.com/NuGet/NuGet.Client/pull/7201)
 
 ## Summary
 
-Add a new `--allow-untrusted-signing` flag to `nuget sign` and `dotnet nuget sign` that allows signing NuGet packages with certificates whose root CA is not installed in a trusted root certificate store. When this flag is set, the `UntrustedRoot` X509 chain status is treated as a warning instead of a fatal error. The certificate chain is still fully built and validated for structural correctness.
+Add a new `--allow-untrusted-root` flag to `nuget sign` and `dotnet nuget sign` that allows signing NuGet packages with certificates whose root CA is not installed in a trusted root certificate store. When this flag is set, the `UntrustedRoot` X509 chain status is treated as a warning instead of a fatal error. The certificate chain is still fully built and validated for structural correctness.
 
 ## Motivation
 
@@ -28,7 +28,7 @@ NuGet's `sign` command is the only commonly used signing tool that cannot sign w
 
 ### Functional explanation
 
-When you need to sign a NuGet package with a certificate whose root CA is not installed in your machine's trusted root certificate store, you can now use the `--allow-untrusted-signing` flag:
+When you need to sign a NuGet package with a certificate whose root CA is not installed in your machine's trusted root certificate store, you can now use the `--allow-untrusted-root` flag:
 
 **dotnet CLI:**
 ```bash
@@ -36,7 +36,7 @@ dotnet nuget sign MyPackage.nupkg \
   --certificate-fingerprint <SHA256-fingerprint> \
   --certificate-store-name My \
   --timestamper http://timestamp.example.com \
-  --allow-untrusted-signing
+  --allow-untrusted-root
 ```
 
 **nuget.exe (classic):**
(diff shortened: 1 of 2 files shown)


dtivel (dismissed):
Thanks for writing this, @elantiguamsft.  I agree with this proposal.

accepted/2026/allow-untrusted-signing.md
```diff
@@ -0,0 +1,181 @@
+# Add --allow-untrusted-signing flag to NuGet sign commands
+
+- [@elantiguamsft](https://github.com/elantiguamsft)
+- [NuGet/NuGet.Client#7201](https://github.com/NuGet/NuGet.Client/pull/7201)
+
+## Summary
+
+Add a new `--allow-untrusted-signing` flag to `nuget sign` and `dotnet nuget sign` that allows signing NuGet packages with certificates whose root CA is not installed in a trusted root certificate store. When this flag is set, the `UntrustedRoot` X509 chain status is treated as a warning instead of a fatal error. The certificate chain is still fully built and validated for structural correctness.
```
kartheekp-ms: I suggest using `--allow-untrusted-root` option for dotnet nuget sign command and `allowUntrustedRoot` option for NuGet.exe sign command. Adding signing word to the option is redundant IMO because from the command it is clear that user is trying to sign a package and would like the operation to succeed even though the certificate chains to untrusted root.

[dotnet nuget trust](https://learn.microsoft.com/dotnet/core/tools/dotnet-nuget-trust) command has `--allow-untrusted-root` to allow untrusted root during package signature verification.

dtivel (approved):

accepted/2026/allow-untrusted-root.md
```diff
@@ -0,0 +1,181 @@
...
+- **NuGet's existing self-signed certificate handling**: `CertificateChainUtility.GetChainStatusFlags` already treats `UntrustedRoot` as a warning for self-issued certificates. This proposal extends that same mechanism.
+- **NuGet `allowUntrustedRoot` in `nuget.config`**: NuGet's [`trustedSigners` configuration](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#trustedsigners-section) already supports an `allowUntrustedRoot` attribute on `<certificate>` elements. When set to `true`, it allows a trusted signer's certificate to chain to an untrusted root during signature verification. This proposal brings the same concept to the `sign` side as a CLI flag.
+- **CoseSignTool `X509ChainTrustValidator`**: Microsoft's [CoseSignTool](https://github.com/microsoft/CoseSignTool) uses `CustomRootTrust` with `CustomTrustStore` for similar scenarios, but requires .NET 5+.
+
+## Unresolved Questions
+
+- **Warning message verbosity**: Should the warning logged when `UntrustedRoot` is encountered include remediation guidance (e.g., "consider installing the root CA certificate to suppress this warning")?
```
kartheekp-ms, line 174: I think we can leave the log message as is and update [NU3018](https://learn.microsoft.com/nuget/reference/errors-and-warnings/nu3018) documentation with your suggestion, i.e. to install the root CA certificate to suppress the error/warning.

accepted/2026/allow-untrusted-root.md
```diff
@@ -0,0 +1,181 @@
...
+- **Interaction with `nuget verify`**: The existing `allowUntrustedRoot` attribute on `<certificate>` elements in the [`trustedSigners` section of `nuget.config`](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#trustedsigners-section) serves a similar purpose for the verification side. Should the new `--allow-untrusted-root` flag and the existing config-level `allowUntrustedRoot` be documented together as complementary features?
```
kartheekp-ms, line 175: Given that the link you mentioned points to the `nuget.config` reference section, and the `allow-untrusted-root` option is only related to `nuget sign` and `dotnet nuget sign`, we should be fine as long as we update the documentation for the signing commands.

kartheekp-ms (approved):

PR #14872, opened 2026-04-26
Proposal: refresh HTTP cache on miss for exact-version restore (#3116)

Saibamen:
## Summary

Design proposal addressing [#3116](https://github.com/NuGet/Home/issues/3116), the long-standing publish-then-consume failure where NuGet's cached versions list causes spurious `NU1102` errors for up to 30 minutes after a package is published.

The proposal introduces a one-time, opt-out, **refresh-on-miss** heuristic in the V3 HTTP source stack: when a restore asks for an exact version that the cached versions list does not contain, NuGet refreshes that one cached document from origin before failing.

- Common case (cache up-to-date): zero extra HTTP traffic.
- Failure case: exactly one extra HTTP round-trip per `(package id, source)` per restore, on a path that was about to fail anyway.
- Default-on, opt-out via `nuget.config`, environment variable, or MSBuild property.

## Test plan

This is a design proposal, not code, so the "test plan" is review:

- [ ] Functional and technical sections are sufficient for an implementation review.
- [ ] Drawbacks / alternatives section captures the trade-offs raised on #3116 (especially [@nkolev92's perf analysis](https://github.com/NuGet/Home/issues/3116#issuecomment-540879988) and [@binki's offline-cache concern](https://github.com/NuGet/Home/issues/3116#issuecomment-2855244557)).
- [ ] Prior art comparison to Cargo / npm / pip is accurate.
- [ ] Naming of the opt-out knobs is acceptable (open question called out in "Unresolved Questions").

## Companion implementation

A first implementation sketch (without the `nuget.config` / MSBuild knobs, env-var only) is open at [NuGet/NuGet.Client#7321](https://github.com/NuGet/NuGet.Client/pull/7321) so reviewers can see what the proposal looks like in code.

### DIFF
--- a/accepted/2026/refresh-http-cache-on-miss.md
+++ b/accepted/2026/refresh-http-cache-on-miss.md
@@ -1,0 +1,187 @@
+# Refresh the HTTP cache when a requested package version is missing
+
+- Author: [@AdamStachowicz](https://github.com/AdamStachowicz)
+- GitHub Issue: [NuGet/Home#3116](https://github.com/NuGet/Home/issues/3116)
+
+## Summary
+
+When NuGet restore is asked to install a specific package version and the cached `index.json` (registration / versions list) for that package on a given V3 source does not contain the requested version, NuGet should treat the cache entry as potentially stale and transparently refresh it from the origin once before failing the restore. Today the cached document is reused for its full TTL (30 minutes by default), which causes spurious "package not found" failures for users who push and immediately consume packages from private feeds. This proposal adds a single, bounded, opportunistic refresh on a "version miss" so that the common case stays fast while the failing case becomes self-healing without users having to discover `--no-cache`, `RestoreNoCache`, or manual cache clearing.
+
+## Motivation
+
+Issue [#3116](https://github.com/NuGet/Home/issues/3116) has been open since 2016 with 87+ 👍 reactions and frequent comments from teams whose CI pipelines and developer inner loops break in the same way:
+
+1. A package version `X 1.0.2` is published to a private feed (Azure Artifacts, GitHub Packages, ProGet, MyGet, internal NuGet.Server, etc.).
+2. A consuming repository updates its `<PackageReference>` to `1.0.2` and runs `dotnet restore` / `nuget restore` / restore-on-build.
+3. Some machines fail with `NU1102` ("Unable to find package X with version (>= 1.0.2)") even though the package is available on the feed, because the machine has a cached versions list from before publication and the cache TTL has not expired.
+
+The HTTP cache for the V3 protocol stores the `index.json` (registration index) and the versions list per `(package id, source)`. The default TTL is 30 minutes (`HttpSourceCacheContext.DefaultMaxAge`) and the cache is keyed by URI, so until the TTL expires every restore on that machine sees the same stale list. The cache was designed to amortize the cost of repeated lookups, not to be authoritative about whether a specific version exists.
+
+The only existing workarounds are all unsatisfactory:
+
+- `restore --no-cache` / `RestoreNoCache=true` — works, but every consumer has to know to set it, it skips the HTTP cache for *every* request (not just the missing one), and it has to be applied per build / per project / per machine. It also is not honoured by the MSBuild SDK resolver, so it does not help SDK references ([#7777](https://github.com/NuGet/Home/issues/7777)).
+- `dotnet nuget locals http-cache --clear` — needs to be re-run on every affected machine and build agent, often by an administrator; users in [#3116](https://github.com/NuGet/Home/issues/3116) report doing this multiple times per week across fleets of build agents.
+- Waiting 30 minutes for the cache to expire — wastes developer and pipeline time and is not viable on shared CI agents that keep producing failures during that window.
+
+The expected outcome of this proposal: when a user asks for an exact version that the cached document says does not exist, NuGet does what every other cache does — it consults the origin once before declaring the item missing.
+
+## Explanation
+
+### Functional explanation
+
+#### Before
+
+```
+> dotnet restore
+error NU1102: Unable to find package Contoso.Foo with version (>= 1.0.2)
+  - Found 5 version(s) in contoso [ Nearest version: 1.0.1 ]
+```
+
+The user pushed `Contoso.Foo 1.0.2` to `contoso` two minutes ago, the feed already serves it, but the local HTTP cache from the previous restore still lists only versions up to `1.0.1`. The restore fails. The user has to know to run `dotnet nuget locals http-cache --clear` or rerun with `--no-cache`.
+
+#### After
+
+```
+> dotnet restore
+  Restored C:\src\app\app.csproj (in 4.2 sec).
+```
+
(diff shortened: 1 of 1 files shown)


accepted/2026/refresh-http-cache-on-miss.md
```diff
@@ -0,0 +1,187 @@
+# Refresh the HTTP cache when a requested package version is missing
+
+- Author: [@AdamStachowicz](https://github.com/AdamStachowicz)
+- GitHub Issue: [NuGet/Home#3116](https://github.com/NuGet/Home/issues/3116)
+
+## Summary
+
+When NuGet restore is asked to install a specific package version and the cached `index.json` (registration / versions list) for that package on a given V3 source does not contain the requested version, NuGet should treat the cache entry as potentially stale and transparently refresh it from the origin once before failing the restore. Today the cached document is reused for its full TTL (30 minutes by default), which causes spurious "package not found" failures for users who push and immediately consume packages from private feeds. This proposal adds a single, bounded, opportunistic refresh on a "version miss" so that the common case stays fast while the failing case becomes self-healing without users having to discover `--no-cache`, `RestoreNoCache`, or manual cache clearing.
```
nkolev92, line 8: Thanks for this work @Saibamen. 

A lot of the notes here are not really represented in the PR, so not really sure what to make of it.

I took a quick look and less than 0.3% of failed are restores are due to NU1102, while 5% of all projects have a calculated NU1603 (basically one where a restore actually happens and not relying on cached scenarios) which basically means the refreshing is more likely to affect many projects that may not need it.

There's definitely gotchas in those numbers, like where the data is collected, the assumption that all NU1102 errors would even be resolved by an extra refresh, but one thing is for sure that the impact of the extra work will be felt in many projects, even if they don't need it.

accepted/2026/refresh-http-cache-on-miss.md
```diff
@@ -0,0 +1,187 @@
...
+- The Global Packages Folder (`%userprofile%/.nuget/packages`) is unaffected — refresh-on-miss only operates against the HTTP cache layer (`%localappdata%/NuGet/v3-cache`).
+- If the package exists in the GPF, NuGet still resolves it from there as today; the HTTP cache is only consulted when the GPF does not satisfy the request.
+
+#### Performance
+
+The added cost in the failure-prone case is exactly one extra HTTP round-trip for the affected `(id, source)`. In the common case (cache is up to date) there is no extra HTTP traffic at all. The refresh is bounded:
+
+- At most one refresh per `(id, source)` per restore.
+- Only when the cached document was actually used (no double-fetch when the document was already fetched fresh).
+- Only when the requested version is not present — i.e. the restore was *about* to fail anyway.
```
donnie-msft, line 133: Is the restore necessarily about to fail here when multiple HTTP sources are configured?  

`ResolverUtility` queries the source providers concurrently, and the implementation refreshes after an individual source misses. For example, if package X exists on nuget.org but not on a configured private feed, could the private feed refresh even though another source successfully resolves X and the overall restore succeeds? 

If so, this may add traffic on successful restores for packages absent from one source. Should refresh happen only after the aggregate source lookup fails, or should this behavior be explicitly accepted and covered by a multiple-source test?

donnie-msft (commented):
I'm driving the PR https://github.com/NuGet/NuGet.Client/pull/7321 and had this design question.

accepted/2026/refresh-http-cache-on-miss.md
```diff
@@ -0,0 +1,187 @@
...
+- **npm** issues a metadata request to the registry on each install by default and uses HTTP cache validators (`If-None-Match` / 304); a stale cached document does not prevent finding a newly published version.
+- **pip** with `--index-url` revalidates the simple index on demand; the package finder does not treat a cached page as authoritative for "version does not exist".
+- Within NuGet itself, the **PackageManagement UI** in Visual Studio already performs a refresh on the version list when the user explicitly browses for a version, which is why the bug is far less visible in the UI than at the command line — exactly what users keep pointing out on the issue thread.
+
+## Unresolved Questions
+
+1. Should the refresh also fire when the cached document indicates the package id is entirely absent (i.e. the registration index 404'd and that 404 was cached), or should that case be handled by a follow-up proposal? The current proposal only addresses the "version not in list" case, which is the one reported in #3116.
```
nkolev92: > Should the refresh also fire when the cached document indicates the package id is entirely absent

There's no cached document for 404s. 
There's internal resource caching that a package was not available, but that's only per restore session and is always preceeded by an http call.

accepted/2026/refresh-http-cache-on-miss.md
Saibamen: Fair on both points; the doc had drifted ahead of the implementation and has been rewritten in https://github.com/NuGet/Home/pull/14872/commits/40ec3626de7a10264d89870c5f62e48e897dc069 to describe only what ships in.

Removed: `nuget.config` and MSBuild knobs, restore telemetry, and floating-range refresh. What remains is a single env-var opt-out (`NUGET_HTTP_CACHE_REFRESH_ON_MISS=false`/`0`), exact non-floating ranges only, and the `OpenedFromDisk` vs `OpenedFromNetwork` distinction the implementation now actually makes.

On your numbers, the `NU1603` population is the important one, and it does not reach this path. The refresh only runs when a source returns no match at all for the range. A request for `[1.0.2, )` that resolves to 1.0.3 - the `NU1603` shape - returns a match and never triggers a refresh. On top of that the refresh requires the versions list to have come from the on-disk HTTP cache, requires every HTTP source to have missed, and runs at most once per `(id, source)` per restore. A steady-state successful restore issues zero extra origin GETs; the extra request only happens on a restore that was already heading for `NU1102`.

accepted/2026/refresh-http-cache-on-miss.md
Saibamen: I have corrected the doc. The `cached document indicates the id is entirely absent` question was based on a wrong assumption, so it is no longer an open question. The Unresolved Questions section now states that 404s are not stored in the on-disk HTTP cache and that the per-session "not available" memory is always preceded by an HTTP call, so this change adds no second origin GET there.

accepted/2026/refresh-http-cache-on-miss.md
Saibamen: Fixed.

When two or more HTTP sources are configured, the first pass through `ResolverUtility.FindLibraryByVersionAsync` runs with `SuppressHttpCacheRefreshOnMiss` set. If any source (HTTP or local) satisfies the version, restore continues and nothing is refreshed. Only when every HTTP source has missed does a second pass run without suppression, allowing the refresh. With a single HTTP source there is no clone and no second pass, so nothing is paid for the common case.

Multi-source tests:

- `FindLibraryByVersionAsync_WhenOneHttpSourceHits_DoesNotQueryMissSourceAgain`
- `FindLibraryByVersionAsync_WhenAllHttpSourcesMiss_QueriesEachSourceAgainWithoutSuppress`
- `..._WhenSingleHttpSourceMisses_DoesNotSuppressAndQueriesOnce`
- `..._WhenOneHttpAndOneLocalMiss_QueriesHttpOnceWithoutSuppress` for the single-feed path.

The proposal's algorithm and drawbacks sections were updated to match.

PR #14875, opened 2026-04-28
Add design for packed project references

Saibamen:
## Links
- Tracking issue: NuGet/Home#3891
- Implementation PR: NuGet/NuGet.Client#7325

## Summary
- Adds a NuGet/Home design proposal for explicitly packing opted-in `ProjectReference` outputs.
- Documents scenarios, anti-scenarios, implementation shape, and unresolved questions for NuGet/Home#3891.
- Aligns the proposal with the current NuGet.Client implementation branch so maintainers can review the UX before the code.

## Test plan
- Not run; documentation-only proposal.
- Verified markdown diagnostics with Cursor lint check.

### DIFF
--- a/accepted/2026/pack-project-reference-outputs.md
+++ b/accepted/2026/pack-project-reference-outputs.md
@@ -2,6 +2,7 @@
 
 - Author Name: [Saibamen](https://github.com/Saibamen)
 - GitHub Issue: [#3891](https://github.com/NuGet/Home/issues/3891)
+- Implementation PR: [NuGet.Client#7325](https://github.com/NuGet/NuGet.Client/pull/7325)
 
 ## Summary
 
--- a/accepted/2026/pack-project-reference-outputs.md
+++ b/accepted/2026/pack-project-reference-outputs.md
@@ -1,0 +1,128 @@
+# Pack opted-in project reference outputs
+
+- Author Name: [Saibamen](https://github.com/Saibamen)
+- GitHub Issue: [#3891](https://github.com/NuGet/Home/issues/3891)
+
+## Summary
+
+Add an explicit opt-in for SDK-style pack to include selected `ProjectReference` build outputs in the parent `.nupkg`, instead of always representing project references as package dependencies. The feature is intended for projects that deliberately ship several implementation assemblies as one NuGet package, while preserving the existing default behavior for all project references.
+
+## Motivation
+
+`dotnet pack` and `msbuild /t:pack` currently treat project references as package dependencies. This is the right default for projects that are independently packaged and versioned, but it is painful for packages that intentionally expose one package while their implementation is split across multiple projects.
+
+Common scenarios from [NuGet/Home#3891](https://github.com/NuGet/Home/issues/3891) include:
+
+- A public client package that has one or more private implementation projects. The implementation assemblies are required at runtime, but the project owners do not want to publish, version, and support each implementation project as an independent package.
+- A package that contains a tightly versioned set of assemblies, such as a facade/client assembly plus contracts or shared implementation that must always ship in the same package version.
+- Migration from `nuget.exe pack -IncludeReferencedProjects`, nuspec-based packaging, or custom `TargetsForTfmSpecificBuildOutput` workarounds to SDK-style pack.
+- Packages that need referenced project outputs in non-default package folders, such as analyzer or build-task packages, where the package author must choose the final package layout explicitly.
+
+There are also important anti-scenarios. This feature should not make it easy to accidentally redistribute arbitrary assemblies, hide dependencies that should be modeled as packages, or bundle the same implementation assembly into several packages that can be installed together. The default behavior therefore remains unchanged, and package authors must opt in per reference.
+
+## Explanation
+
+### Functional explanation
+
+By default, a `ProjectReference` continues to be represented as a package dependency when packing an SDK-style project:
+
+```xml
+<ProjectReference Include="..\Implementation\Implementation.csproj" />
+```
+
+To bundle the referenced project's output assemblies into the parent package, the package author sets `Pack="true"` on the project reference:
+
+```xml
+<ProjectReference Include="..\Implementation\Implementation.csproj" Pack="true" />
+```
+
+For this opted-in reference:
+
+- The referenced project's build outputs are added to the parent package.
+- The referenced project is not emitted as a nuspec dependency.
+- Package dependencies needed by the bundled project are still emitted as package dependencies of the parent package, so consumers can restore the external packages needed at runtime.
+- Symbols are included consistently with existing `IncludeSymbols` behavior.
+- The default package location follows the existing pack convention for build output, for example `lib/<tfm>/Implementation.dll`.
+
+The package author may override the destination with `PackagePath`:
+
+```xml
+<ProjectReference
+  Include="..\Analyzer\Analyzer.csproj"
+  Pack="true"
+  PackagePath="analyzers/dotnet/cs" />
+```
+
+When `PackagePath` is specified, NuGet places the referenced project's outputs under that path. The package author is responsible for choosing a path that is meaningful for the package type. For normal library assemblies, omitting `PackagePath` is preferred so outputs remain under the target framework's `lib` folder.
+
+For multi-targeting projects, pack applies the behavior per target framework. A project reference that only applies to a subset of target frameworks is bundled only for those frameworks.
+
+### Technical explanation
+
+The implementation should keep restore as the source of truth for the project graph. Pack should not rediscover an independent project graph or infer project reference metadata from only the entry project.
+
+At restore time, NuGet should preserve pack-specific project reference metadata in the restore graph and `project.assets.json`:
+
+- `Pack`
+- `PackagePath`
+
+This metadata must flow through the existing restore entry points:
+
+- MSBuild restore
+- Static graph restore
+- Visual Studio nomination
+- `project.assets.json` read/write round-tripping
+
+At pack time, NuGet reads the assets file to identify the project references that opted in to bundling. It then collects build outputs through the existing pack output group targets, using the target framework from the restore graph and avoiding a rebuild of project references during output collection.
+
+Dependency generation changes only for opted-in project references. The bundled project itself is suppressed as a nuspec dependency. Package dependencies reachable through the bundled project are promoted into the parent package dependency group so package consumers still restore required external packages.
+
+The initial implementation should be conservative about the packaging surface:
+
+- Include build outputs that existing pack output groups already understand.
+- Do not include arbitrary content, source, native assets, or runtime-specific assets from referenced projects unless a follow-up design explicitly defines those behaviors.
+- Do not change nuspec pack.
+- Do not change `nuget.exe pack -IncludeReferencedProjects`.
+- Do not change the default handling of project references.
+
+## Drawbacks
+
+Bundling referenced assemblies can create packages that are harder for consumers to reason about. Assemblies inside a package are not truly private in .NET's normal load context; consumers can still reference them, observe type identities, or encounter binding/version conflicts.
+
+The feature can encourage package authors to avoid creating proper packages for independently useful components. If two packages bundle different versions of the same implementation assembly, a consuming application can see duplicate assembly conflicts that NuGet cannot resolve as package version conflicts.
+
+The feature also makes pack more complex because restore metadata, assets-file serialization, dependency generation, and output collection all need to agree on the same project-reference metadata.
+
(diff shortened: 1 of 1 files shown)


accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+- GitHub Issue: [#3891](https://github.com/NuGet/Home/issues/3891)
+- Implementation PR: [NuGet.Client#7325](https://github.com/NuGet/NuGet.Client/pull/7325)
+
+## Summary
+
+Add an explicit opt-in for SDK-style pack to include selected `ProjectReference` build outputs in the parent `.nupkg`, instead of always representing project references as package dependencies. The feature is intended for projects that deliberately ship several implementation assemblies as one NuGet package, while preserving the existing default behavior for all project references.
+
+## Motivation
+
+`dotnet pack` and `msbuild /t:pack` currently treat project references as package dependencies. This is the right default for projects that are independently packaged and versioned, but it is painful for packages that intentionally expose one package while their implementation is split across multiple projects.
```
zivkan, line 13: I think that describing it as painful is unhelpful opinionated commentary.

In the list of common scenarios below, the first two are personal preferences. So, I don't believe it's fair to say it's painful. Both scenarios would work just fine from a technical perspective with multiple nupkgs.

The third talks about needing a migration, but doesn't explain why it's difficult or painful. If the projects are simple classlibs, the only migration cost of `nuget.exe pack -IncludeReferencedProjects` to `msbuild -t:pack` is migrating from packages.config to PackageReference. But if the developer is willing to accept their single package becoming multiple packages, there is no additional migration cost.  On the other hand, if they want to onboard onto the feature this spec is proposing, then there is additional migration cost.

The example in the list of common scenarios is a good scenario, and if we can make this easier for package authors, it will be a great win. I don't have experience creating packages with roslyn analyzers, so I don't know how painful it is, but if it's achievable with a `<None Include="../../My.Analyzer/bin/$(Configuration)/netstandard2.0/My.Analyzer.dll" Pack="true" PackagePath="analyzers/dotnet/cs" />`, then that doesn't feel very painful to me.

From a package consumer point of view, there's pretty much no difference between a package that has multiple assemblies, or a package with one assembly and one or more dependencies. NuGet will download all the packages and hook them into the build either way.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+
+For this opted-in reference:
+
+- The referenced project's build outputs are added to the parent package.
+- The referenced project is not emitted as a nuspec dependency.
+- Package dependencies needed by the bundled project are still emitted as package dependencies of the parent package, so consumers can restore the external packages needed at runtime.
+- Symbols are included consistently with existing `IncludeSymbols` behavior.
+- The default package location follows the existing pack convention for build output, for example `lib/<tfm>/Implementation.dll`.
+
+If a bundled project references another project, the default behavior should be to include the project-reference closure under the opted-in project. This is required for runtime correctness in common "private implementation assembly" scenarios. The package author opts in at the first redistributed edge; NuGet then follows the restored project graph from that edge so the package contains the implementation assemblies needed by that bundled project.
```
zivkan, line 48: Edge case: the child project has a ProjectReference to a project that the parent project also has a ProjectReference to, but the parent doesn't set `Pack="true"` on the transitive project.

I guess possible options are:
* Add the transitive assembly to the package, and suppress it as a package dependency
* Do not include the transitive assembly to the package, only add its package as a dependency
* Both add the transitive assembly to the package, and also add it as a package dependency. Consider this an authoring error that NuGet will not try to prevent/solve.

I don't know which is best.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+```xml
+<ProjectReference
+  Include="..\Analyzer\Analyzer.csproj"
+  Pack="true"
+  PackagePath="analyzers/dotnet/cs" />
+```
+
+When `PackagePath` is specified, NuGet places the referenced project's outputs under that path instead of under `BuildOutputTargetFolder` and the target framework folder. The package author is responsible for choosing a path that is meaningful for the package type. For normal library assemblies, omitting `PackagePath` is preferred so outputs remain under the target framework's `lib` folder.
+
+For multi-targeting projects, pack applies the behavior per target framework. A project reference that only applies to a subset of target frameworks is bundled only for those frameworks. The feature is also controlled by existing pack switches: if `IncludeBuildOutput=false`, bundled project build outputs are not added to the package.
```
zivkan, line 61: I think this should be rephrased, and possibly an example given for clarity. I think I know what you're trying to say, but I don't know if I'm just guessing based on my own knowledge.

For example, something like:

> NuGet will use normal restore and build asset selection to choose child project assemblies to copy into the package.  For example, ProjectA targeting net10.0 and net481 with a `ProjectReference` to ProjectB, which targets net8.0 and net472 will have ProjectA's net10.0 and ProjectB's net8.0 dlls in `lib/net10.0/`, and ProjectA's net481 and ProjectB's net472 dlls in `lib/net481/`. If ProjectB targets only netstandard2.0, then that assembly will be duplicated in the package's `lib/net10.0/` and `lib/net481/` directories.

zivkan, line 61: Should there be a warning when `IncludeBuildOutput` is `false`, and one or more `ProjectReference` has `Pack="true"`?

On one hand, I don't like it when mutually exclusive features are configured at the same time, and there's no feedback about this probable configuration mistake. If I'm new to the technology, I don't yet know enough to debug why it's silently failing.

On the other hand, the scenario should be simple enough to debug, since the project's own assembly won't be making it into the package. Also trying to consider every permutation of incompatible features and emit warnings could be noisy, slow down the product from an explosion of validation checks, make docs more confusing with an overabundance of warning messages.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+The package author may override the destination with `PackagePath`:
+
+```xml
+<ProjectReference
+  Include="..\Analyzer\Analyzer.csproj"
+  Pack="true"
+  PackagePath="analyzers/dotnet/cs" />
+```
+
+When `PackagePath` is specified, NuGet places the referenced project's outputs under that path instead of under `BuildOutputTargetFolder` and the target framework folder. The package author is responsible for choosing a path that is meaningful for the package type. For normal library assemblies, omitting `PackagePath` is preferred so outputs remain under the target framework's `lib` folder.
```
zivkan, line 59: I'm wondering if this could be problematic for multi-targeting projects.

Say I have project Contoso with a ProjectReference to Contoso.Analyzers, which I want to put in `analyzers/dotnet/cs`. 

If both Contoso and Contoso.Analyzers multi-target, we can't have two files with the same name with different contents. If NuGet's instructed to put both copies in the same location, it needs to fail with a duplicate file error. I did a search, but wasn't able to quickly find what the NU error code is for this.

Only if Contoso.Analyzers is single targeting (probably netstandard2.0), then it can make sense to allow it. It becomes an implementation detail to de-duplicate, as long as the source path is the same for duplicate target paths. I don't believe NuGet does this at the moment.

edit: this appears to be called out already in unresolved questions.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+
+- `nuget.exe pack -IncludeReferencedProjects` supports a broad version of this behavior for legacy pack workflows.
+- SDK-style pack already supports adding additional files through `TargetsForTfmSpecificBuildOutput` and `TfmSpecificPackageFile`, but users must hand-author the graph and dependencies.
+- NuGetizer has prior art for distinguishing project references that should become dependencies from projects whose outputs should be bundled.
+- The original MSBuild pack design states that project-to-project references are treated as NuGet package references by default.
+
+## Unresolved Questions
+
+- Should NuGet warn when an opted-in project reference points to a project that is itself packable or has a `PackageId`, since that may indicate the project should remain a package dependency?
+- Should transitive bundled project outputs inherit the root opted-in reference's `PackagePath`, or should only the directly opted-in project's outputs use that override? Inheriting the path is simple, but may be wrong for mixed asset types.
```
zivkan, line 128: There's also the question of diamond dependencies.

If I'm packing ProjectA, which have references to ProjectB and ProjectC, both of which depend on ProjectD, if I use a different PackagePath for ProjectB and ProjectC, where does ProjectD's output go?

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+The initial implementation should be conservative about the packaging surface:
+
+- Include build outputs that existing pack output groups already understand.
+- Include debug symbols through the existing debug symbol output groups when symbols are requested.
+- Do not include arbitrary content, source, native assets, or runtime-specific assets from referenced projects unless a follow-up design explicitly defines those behaviors.
+- Do not change nuspec pack.
+- Do not change `nuget.exe pack -IncludeReferencedProjects`.
+- Do not change the default handling of project references.
+
+The current implementation shape uses two assets-file reads during pack: one to collect all project references for the existing package-version lookup, and one to collect the subset of project references whose outputs should be packed. The packed project-reference items carry `TargetFramework`, `BuildProjectReferences=false`, and, when specified, the requested package path.
```
zivkan, line 92: are you referring to the pull request you already opened, or are you referring to how pack is already implemented in the .NET SDK that ships today?

In either case, it sounds like an implementation detail not really relevant for a feature spec.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+For multi-targeting projects, pack applies the behavior per target framework. A project reference that only applies to a subset of target frameworks is bundled only for those frameworks. The feature is also controlled by existing pack switches: if `IncludeBuildOutput=false`, bundled project build outputs are not added to the package.
+
+### Technical explanation
+
+The implementation should keep restore as the source of truth for the project graph. Pack should not rediscover an independent project graph or infer project reference metadata from only the entry project.
+
+At restore time, NuGet should preserve pack-specific project reference metadata in the restore graph and `project.assets.json`:
+
+- `Pack`
+- `PackagePath`
```
zivkan, line 70: Please provide a json snippet of the part of the assets file that will change, and what it looks like.  I highly recommend using a diff code block, so github can highlight the changes, making it easier for people reading the spec.

accepted/2026/pack-project-reference-outputs.md
```diff
@@ -0,0 +1,137 @@
...
+
+When `PackagePath` is specified, NuGet places the referenced project's outputs under that path instead of under `BuildOutputTargetFolder` and the target framework folder. The package author is responsible for choosing a path that is meaningful for the package type. For normal library assemblies, omitting `PackagePath` is preferred so outputs remain under the target framework's `lib` folder.
+
+For multi-targeting projects, pack applies the behavior per target framework. A project reference that only applies to a subset of target frameworks is bundled only for those frameworks. The feature is also controlled by existing pack switches: if `IncludeBuildOutput=false`, bundled project build outputs are not added to the package.
+
+### Technical explanation
+
+The implementation should keep restore as the source of truth for the project graph. Pack should not rediscover an independent project graph or infer project reference metadata from only the entry project.
+
+At restore time, NuGet should preserve pack-specific project reference metadata in the restore graph and `project.assets.json`:
```
zivkan, line 67: Thinking out loud:  restore is NuGet's primary use case. Far more people restore packages and never pack, than people that pack. Let alone people who will want to pack and use this spec's feature.

Adding more content to the assets file means that reading the file has more work to do, and therefore will be slower, even in scenarios (restore) where the additional metadata isn't used.

As long as the assets file is effectively unmodified in projects that don't use this spec's feature, then it should be fine, no measurable perf degradation. But within Microsoft we have some repos using 1000+ projects per restore, so even a few additional memory allocations can add up to additional garbage collections and slower restores.

If we can't avoid perf regressions for restore by adding this feature, while using the assets file, then I think it'll be better to make pack slower by needing to run additional msbuild targets, or whatever is needed, to collect the additional project information it needs.

zivkan (commented):
I've been busier than usual the last few weeks, so I haven't made time to read and think about this much until now.  Some of it is still churning in my head, so there are parts I don't have an opinion on yet.

I'm generally supportive of trying to aim small, get the simple scenarios working, and later we'll see if we need to extend it. I just hope that it's not too limited on first release and people get frustrated that it's not usable. But for this feature, I think that being able to pack multiple simple class libraries will be a good first start.

Issue #14949, opened 2026-06-16
Pack Warning when Package ID doesn't meet the standards


nkolev92:
See: https://github.com/NuGet/Announcements/issues/75

nkolev92:
The pack warning will be introduced in `dotnet pack` and .NET 11.

nkolev92:
Warning adding for .NET 11, tbd on MSBuild. I'll open a new issue for that.

PR #14950, opened 2026-06-16
Warn when creating packages with non-ASCII characters in the package ID

nkolev92:
Design for https://github.com/NuGet/Home/issues/14949

Related: https://github.com/NuGet/Announcements/issues/75

### DIFF
--- a/accepted/2026/warning-non-ascii-characters.md
+++ b/accepted/2026/warning-non-ascii-characters.md
@@ -1,13 +1,13 @@
-# ***Warn when creating packages with that do not adhere to the restricted set of characters in the package ID***
+# ***Warn when creating packages with package IDs that do not adhere to the allowed set of characters***
 
 - Nikolche Kolev <https://github.com/nkolev92>
 - [Pack Warning when Package ID doesn't meet the standards](https://github.com/NuGet/Home/issues/14949)
 
 ## Summary
 
-Raise a NuGet pack warning when the package ID being authored does not adhere to the new restricted set of allowed characters.
+Raise a NuGet pack warning when the package ID being authored does not adhere to the new allowed set of characters.
 The warning is opt-out via `NoWarn`, framework agnostic, and applies to both the MSBuild `Pack` target (`dotnet pack` / `msbuild /t:Pack`) and `nuget.exe pack`.
-This is the client-side companion to the nuget.org change that blocks new pushes with package IDs that do not adhere to the restricted set of characters.
+This is the client-side companion to the nuget.org change that blocks new pushes with package IDs that do not adhere to the allowed set of characters.
 See [Strengthening NuGet Security with Package ID Standards](https://github.com/NuGet/Announcements/issues/75).
 
 ## Motivation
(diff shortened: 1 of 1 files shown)


accepted/2026/warning-non-ascii-characters.md
```diff
@@ -0,0 +1,90 @@
+# ***Warn when creating packages with non-ASCII characters in the package ID***
+
+- Nikolche Kolev <https://github.com/nkolev92>
+- [Pack Warning when Package ID doesn't meet the standards](https://github.com/NuGet/Home/issues/14949)
+
+## Summary
+
+Raise a NuGet pack warning when the package ID being authored contains characters outside of the ASCII range.
```
jeffkl:
```suggestion
Raise a NuGet pack warning when the package ID being authored does not adhere to the new restricted set of allowed characters.
```

jeffkl (dismissed):
Probably also worth mentioning that other feed implementations will probably enforce the new naming conventions, but we aren't 100%.

Also make sure to call out that if/when its ever changed to an error, we'll always need an escape hatch for Pack so users can't create packages that use invalid characters just in case.

accepted/2026/warning-non-ascii-characters.md
nkolev92: Applied throughout.

accepted/2026/warning-non-ascii-characters.md
```diff
@@ -0,0 +1,90 @@
...
+See [Strengthening NuGet Security with Package ID Standards](https://github.com/NuGet/Announcements/issues/75) for the broader rollout plan and timelines.
+
+## Explanation
+
+### Functional explanation
+
+When a user runs `dotnet pack` (or any other entry point that invokes the NuGet `Pack` task) on a project whose effective package ID contains characters outside the restricted set, NuGet emits a warning:
+
+```text
+warning NU5052: The package ID 'Contöso.Utilities' does not adhere to the restricted set of characters allowed in package IDs. Rename the package to use only characters from the restricted set.
```
Nigusu-Allehu: This warning does make it clear that the customer is using a package id that could constitute characters that are outside of the allowed set. But it does not make it clear what the allowed set is. Should we consider adding some information about the allowed set or an aka.ms link that leads to more information about the allowed list.

accepted/2026/warning-non-ascii-characters.md
```diff
@@ -0,0 +1,90 @@
...
+Warning at pack time gives authors the earliest possible signal, before any artifact is uploaded, and avoids late-stage release surprises.
+
+The longer-term plan is to promote this warning to an error in a later .NET SDK release (targeted at .NET 12).
+See [Strengthening NuGet Security with Package ID Standards](https://github.com/NuGet/Announcements/issues/75) for the broader rollout plan and timelines.
+
+## Explanation
+
+### Functional explanation
+
+When a user runs `dotnet pack` (or any other entry point that invokes the NuGet `Pack` task) on a project whose effective package ID contains characters outside the restricted set, NuGet emits a warning:
```
Nigusu-Allehu: nit/question: Should we use the phrase "allowed set" instead of "restricted set"?

Nigusu-Allehu (dismissed):
LGTM bar a few nits and suggestions I had

accepted/2026/warning-non-ascii-characters.md
nkolev92: The message has diverted from the client implementation. 

I'll bring them up to date. 

The variation in the PR is: 

> The package ID '{0}' is invalid. Package IDs must start with a letter, digit, or underscore, and contain only ASCII letters, digits, dots (.), dashes (-), and underscores (_), with no consecutive dots or dashes

accepted/2026/warning-non-ascii-characters.md
nkolev92: Yep, can be tweaked, that improves it.

Nigusu-Allehu (approved):

PR #12874, opened 2023-09-08
Proposal for package grouping to enable release staging and release deprecation

mkArtakMSFT:
This proposal introduces a concept for logical grouping of packages published to NuGet. This will allow one to easily identify all the packages released as part of a single product version release as well as request bulk-processing for a set of packages which belong to the same release. For example, all packages published as part of .NET 8 Preview 5 release will be staged together and then also can be deprecated together by a single call to NuGet APIs.

[Rendered Proposal](https://github.com/NuGet/Home/blob/mkArtakMSFT/releaseGrouping/accepted/2024/release-staging-and-deprecation.md)

PR #14978, opened 2026-07-07
Add refreshed nuget staging proposal

japarson:
Design spec for package staging on nuget.org. Builds on the [2024 Release Staging and Deprecation proposal](https://github.com/NuGet/Home/pull/12874) with a narrower scope and a security-first approach.

Related: https://github.com/NuGet/NuGetGallery/issues/3931
Rendered document: https://github.com/NuGet/Home/blob/japarson/staging/accepted/2026/package-staging.md


accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+- Max 1000 packages per group
+- Max 5000 ungrouped staged packages per owner
+
+#### CLI commands
+
+All commands discover the staging endpoint via the `PackageStaging/1.0.0` resource type in the service index. Sources that do not advertise this resource return an error indicating staging is not supported.
+
+All commands support `--format json` for machine-readable output.
+
+**`dotnet nuget stage push`**
```
Nigusu-Allehu: Question: I see we are adding a new command. What happens to the exisitng `dotnet nuget push` command?

I would assume we won't be deprecating it as it would still be used by other package sources.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+
+Package staging lets authors push packages to nuget.org ahead of time, run them through full validation (malware scanning, signing), and publish them later with a single action through the Gallery UI. Staged packages are invisible to consumers until published. Publishing requires logging into nuget.org, which enforces two-factor authentication as a proof of presence check.
+
+This builds on the [2024 Release Staging and Deprecation proposal](../2024/release-staging-and-deprecation.md) with a narrower scope focused on staging and publishing, informed by direct feedback from the .NET release team and security requirements around supply chain protection.
+
+## Motivation
+
+### Supply chain protection
+
+If a CI/CD pipeline is compromised or credentials leak, an attacker can push malicious packages to nuget.org today. Staging introduces a separation between pushing and publishing. CI/CD can only stage packages. A human must log into nuget.org (which requires two-factor authentication) to publish them. Even if an attacker compromises a build pipeline, they can only stage invisible packages that never reach consumers.
```
Nigusu-Allehu: what happens if a customer runs `dotnet nuget push`(How does the server react to these requess)? They could be on an old dotnet sdk.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
+# Package Staging
+
+- Author: James Parsons ([@japarson](https://github.com/japarson))
+- GitHub Issue: [NuGet/NuGetGallery#3931](https://github.com/NuGet/NuGetGallery/issues/3931)
+
+## Summary
+
+Package staging lets authors push packages to nuget.org ahead of time, run them through full validation (malware scanning, signing), and publish them later with a single action through the Gallery UI. Staged packages are invisible to consumers until published. Publishing requires logging into nuget.org, which enforces two-factor authentication as a proof of presence check.
```
Nigusu-Allehu: nit: it would be nice if we could have a sentence per line. Makes commenting during reviews much easier.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Pre-validation
+
+The .NET team pushes 2000+ packages to nuget.org every Patch Tuesday. Validation (malware scanning, certificate checks, repository co-signing) takes 40 to 60 minutes for that volume. If a package fails validation, the team discovers it on release day and has to rebuild, re-sign, and re-push during a tight embargo window.
+
+Staging lets authors push packages days or weeks early. Validation runs during that window. Failures are caught and fixed before release day, not during it.
+
+### Publish latency
+
+Today, the time from push to all packages being restorable includes validation plus the V3 pipeline (catalog, flat container, registration, search). For 2000 packages, total time was about 2 hours during the March 2026 Patch Tuesday. By removing validation from the critical path, the only remaining cost is the V3 pipeline, which should bring the total time from publish to restorable down significantly.
```
chabiss: We should call out that we made work to bring the V3 ingestion down from 1 hour to around 5 minutes.  So I overall, we will be able to manage to bring down the published time from 2 hours to only 5 min, assuming package are properly staged and validated prior to release

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+    Available --> V3["V3 Pipeline"]
+```
+
+Authors push packages to nuget.org using a new staging endpoint. The CLI command looks like:
+
+```
+dotnet nuget stage push MyPackage.1.0.0.nupkg
+```
+
+The package goes through the same validation pipeline as a normal push: malware scanning, author signature checks, certificate revocation checks, and repository co-signing. The difference is what happens after validation succeeds. Instead of becoming publicly available, the package enters a "staged" state. It is not restorable, not searchable, and not visible on nuget.org to anyone except the package owner.
```
chabiss: I don't think the package are visible at all even for the package owners.  Package Owner would only be able to see the staged package from the UI,   not through search. Let's clarify that.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+If validation fails, the author is notified the same way as a normal push failure. They can fix the issue and re-push. Since this happens days or weeks before release, there is no time pressure.
+
+Re-pushing the same content is a no-op. Pushing different content for the same ID and version replaces the staged package and re-runs validation.
+
+#### Staging groups
+
+Packages can optionally be organized into staging groups. A group is a named collection of staged packages owned by a user or organization.
+
+```
+dotnet nuget stage group create my-release --name "My Release"
```
chabiss, line 71: I don't think we need to have a create command.  Let's simplify this!

If you do 'dotnet nuget stage --group "BCL Release 11.0" , the push will automatically stage the package under that group.  If you do not supply the --group,  the package will be stage under no group

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+
+[Trusted Publishing](../2024/trusted-publishers-oidc-for-nuget-push.md) (OIDC) is the recommended authentication method for automated package publishing on nuget.org. API keys are strongly discouraged and should be replaced with Trusted Publishing.
+
+Trusted Publishing configurations will support a new "stage only" permission scope. When configured with this scope, a GitHub Actions workflow (or other supported CI/CD provider) can stage packages but cannot publish them. This aligns with the security model: CI/CD stages, humans publish.
+
+#### Gallery UI
+
+The nuget.org Manage Packages page adds two new sections:
+
+![Manage Packages - Staged Packages and Staging Groups](images/staging-manage-packages.png)
```
chabiss: I think we should only have one category: "Staged Packages"  When expanding Stage Packages, staged packages would be grouped by "Groups", one of the group being "none".  It's important for the groups to be expanded  in order to see the packages visible under that group.  For instance, if a package fails to validate, the ability to re-kick a validation needs to be available, even for packages under a group

accepted/2026/package-staging.md
chabiss: we do not change the behavior for 'dotnet nuget push'

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+```
+dotnet nuget stage push <PACKAGE_PATH> [options]
+
+Arguments:
+  PACKAGE_PATH    Path to the .nupkg file to stage.
+
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  -k, --api-key <KEY>       The API key for the source. Not recommended; use Trusted Publishing instead.
+  --group <GROUP_ID>        Add the package to a staging group. The group must already exist.
```
chabiss: I think the requirements of creating the group first is an overhead.  if you pass --group "Group_ID", if the group doesn't already exist, it will create it

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+dotnet nuget stage push MyPackage.1.0.0.nupkg
+dotnet nuget stage push MyPackage.1.0.0.nupkg --group my-release
+```
+
+**`dotnet nuget stage list`**
+
+List staged packages for the authenticated user.
+
+```
+dotnet nuget stage list [options]
```
chabiss: to be clear, 'dotnet nuget stage list' will list all stage package by group

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  -k, --api-key <KEY>       The API key for the source.
+  --no-confirm              Skip the confirmation prompt.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+**`dotnet nuget stage group create`**
+
+Create a new staging group.
```
chabiss: I don't think we need that command.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+Microsoft.Extensions.DependencyInjection  10.0.0-preview.4     my-release       Ready      Jul 10, 2026
+Contoso.Utilities                         3.1.0                (ungrouped)      Validating Jul 8, 2026
+```
+
+**`dotnet nuget stage delete`**
+
+Delete a staged package. Fails if the package is currently validating.
+
+```
+dotnet nuget stage delete <PACKAGE_ID> <VERSION> [options]
```
chabiss: I which context this command would be needed?

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  -k, --api-key <KEY>       The API key for the source.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+Example output:
+```
+Staging Groups (2 groups)
+
+Group ID           Name                Packages   Status      Expires
```
chabiss: (Ungroup) should an implied group for ungrouped packages

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+my-release         My Release          2 ready    Ready       Jul 10, 2026
+hotfix-june        June Hotfix         0          Empty       Jul 1, 2026
+```
+
+**`dotnet nuget stage group delete`**
+
+Delete a staging group and all its staged packages. Fails if any package in the group is currently validating.
```
chabiss: Maybe a --force command to force deletion of a group no matter what?

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,281 @@
...
+  -k, --api-key <KEY>       The API key for the source.
+  --no-confirm              Skip the confirmation prompt.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+## Drawbacks
+
+- **Publish is not automatable.** Requiring Gallery UI login means publishing cannot be scripted or run from CI/CD. This is intentional for security but adds a manual step to release workflows. Someone has to log in and click "Publish" on release day.
+
+- **No atomic publish.** We are not guaranteeing that all packages in a group become restorable at the exact same instant. The V3 pipeline processes packages in batches and there will be a small window (minutes) where some packages are available and others are not. Package authors who need strict ordering should manage dependency sequencing on their side.
```
chabiss: By dependencies sequencing, this means proper grouping.   For instance, BCLs groups should be release first, then SDK Libraries and then Tools

chabiss (dismissed):

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+The Upload page also supports staging via a Publishing Mode option:
+
+![Upload Package - Publishing Mode](images/upload-package.png)
+*Figure 1. Upload Package page with Publishing Mode selector.*
+
+#### Revalidation
+
+Validation results go stale over time. Malware signature databases update daily and certificate revocation status can change. A package that passed validation weeks ago may no longer be clean.
+
+After a configurable staleness window (default: a few days), the package transitions to a "Stale" state in the UI. Stale packages cannot be promoted until the owner triggers revalidation. Only external-state checks rerun: certificate revocation and malware scanning. Content-only checks (signature structure, package format) are skipped since the bytes have not changed.
```
nkolev92: Have you considered revalidating automatically? 
At least like once or twice. 
This would allow a Tuesday release to be able to be prepped on like Thursday/Friday.

Maybe like revalidate if that package had been pushed within a week?

I get that it creates a bit of a potential for race conditions, but just an idea.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+  PACKAGE_ID      The package ID to delete.
+  VERSION         The package version to delete.
+
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  --no-confirm              Skip the confirmation prompt.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+**`dotnet nuget stage group create`**
```
nkolev92: This is a fun long name :D 

So long it basically has me thinking of alternatives, that are may less wordy. 

I'd honestly consider whether we even need the group creating in the CLI mvp if group promotion requires 2FA. 


Beyond that, I think having a higher level example of what `dotnet nuget stage group` has as sub commands would make it to parse from the spec.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+
+Create a new staging group.
+
+```
+dotnet nuget stage group create <GROUP_ID> [options]
```
nkolev92: I wonder if it's the right time to start "rebranding" the `nuget  push` commands as `package push`. 

cc @baronfel I'd love to get your thoughts on this.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+```
+dotnet nuget stage list [options]
+
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  --group <GROUP_ID>        Filter to packages in a specific group.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+**`dotnet nuget stage status`**
```
nkolev92: You can consider merging the list and status into 1 command since they basically do the same thing, one for packge, the other for group. 

I think using familiar dotnet CLI verbs would be preferred as well, unless absolutely necessary.

nkolev92 (commented):
Great write-up @japarson!

Most of my questions are really about CLI command naming. 

Some of them are pretty long, so I wonder if we have more concise approaches. 

Beyond that, I wonder if `dotnet nuget` is something we still need to keep using, or whether this can be another one of those commands that we need to move towards package. I have add the .NET PM @baronfel for feedback here.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  --name <DISPLAY_NAME>     Optional display name for the group. Defaults to the group ID.
+  --organization <ORG>      Assign ownership to an organization instead of the calling user.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+**`dotnet nuget stage group list`**
+
+List staging groups for the authenticated user.
```
leecow: Is there a group detail option? This would enable automation to track the status of a given stage group.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,393 @@
...
+  VERSION         The package version.
+
+Options:
+  -s, --source <SOURCE>     The package source to use. Defaults to nuget.org.
+  --format <FORMAT>         Output format: text (default) or json.
+```
+
+## Drawbacks
+
+- **Promotion is not automatable.** Requiring Gallery UI login with step-up 2FA means promotion cannot be scripted or run from CI/CD. This is intentional for security but adds a manual step to release workflows. Someone has to log in and complete a 2FA challenge on release day.
```
leecow: I understand the concerns and need for the 2FA challenge. However, since we already have an authenticated human kicking off the release pipeline on the .NET side, this is an added stoppage. It would be great to see if there's a way to achieve the security goal without the additional human touch.

japarson: It might be better to view the additional "stoppage" as an additional layer of security. Keep in mind this feature is being developed to help raise the security floor of the entire .NET ecosystem. I sympathize with the fact that the .NET team already has good security practices in place, but many developers and organization do not and I don't think we should pick and choose who has this requirement.

accepted/2026/package-staging.md
japarson: Updated the spec:
`dotnet nuget stage group list [<GROUP_ID>] [options]`

accepted/2026/package-staging.md
japarson: Updated.

accepted/2026/package-staging.md
japarson: @Nigusu-Allehu

accepted/2026/package-staging.md
japarson: @Nigusu-Allehu

accepted/2026/package-staging.md
japarson: I updated the spec to clarify the staleness window is 7 days which should avoid weekend scenarios. Maybe the feature can be added to v2 if requested enough.

japarson:
> Great write-up @japarson!
> 
> Most of my questions are really about CLI command naming.
> 
> Some of them are pretty long, so I wonder if we have more concise approaches.
> 
> Beyond that, I wonder if `dotnet nuget` is something we still need to keep using, or whether this can be another one of those commands that we need to move towards package. I have add the .NET PM @baronfel for feedback here.

I had a chat with @Nigusu-Allehu and we decided to remove the CLI design from the spec so that Nigusus can focus on writing his own spec for that next sprint.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,201 @@
...
+```
+dotnet nuget stage group create my-release --name "My Release"
+dotnet nuget stage push MyPackage.1.0.0.nupkg --group my-release
+```
+
+##### Managing groups
+
+Groups are useful for coordinating releases across many packages. A team can stage hundreds of packages into a single group over several days, verify all of them pass validation, and then promote the entire group at once.
+
+Groups are ephemeral. They are deleted after a successful promotion and expire after a configurable TTL. Each release gets its own group.
```
aortiz-msft, line 87: There should be a maximum allowable TTL. Any thoughts on what the value should be?

japarson: The current design is 30 days

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
+# Package Staging
+
+- Author: James Parsons ([@japarson](https://github.com/japarson))
+- GitHub Issue: [NuGet/NuGetGallery#3931](https://github.com/NuGet/NuGetGallery/issues/3931)
+
+## Summary
+
+Package staging lets authors push packages to nuget.org ahead of time, run them through full validation (malware scanning, signing), and promote them later with a single action through the Gallery UI. Staged packages are invisible to consumers until promoted. Promotion requires that the user is logged into nuget.org with two-factor authentication enabled on their account.
```
richlander:
```suggestion
Package staging lets authors push packages to nuget.org ahead of time, run them through full validation (malware scanning, signing), and promote them later with a single action through the Gallery UI. Staged packages are invisible to consumers until promoted. Promotion is only available for users with two-factor authentication enabled.
```

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+
+This builds on the [2024 Release Staging and Deprecation proposal](../2024/release-staging-and-deprecation.md) with a narrower scope focused on staging and promotion, informed by direct feedback from the .NET release team and security requirements around supply chain protection.
+
+## Motivation
+
+### Supply chain protection
+
+If a CI/CD pipeline is compromised or credentials leak, an attacker can push malicious packages to nuget.org today. Staging introduces a separation between pushing and promoting. CI/CD can only stage packages. A human must log into nuget.org with 2FA enabled to promote them. Even if an attacker compromises a build pipeline, they can only stage invisible packages that never reach consumers.
```
richlander: This claim is unclear/unsubstantiated. Is the suggestion that you can switch your account to staging only? The text doesn't say that.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+
+Staging is entirely opt-in. The existing `dotnet nuget push` command and push API continue to work exactly as they do today, including for older SDK versions. Staging uses a separate endpoint that newer SDKs may expose through commands such as `dotnet nuget stage push`. The exact command names and syntax shown in this proposal are illustrative and not final.
```
richlander:
```suggestion
Staging is opt-in. The existing `dotnet nuget push` command and push API continue to work exactly as they do today, including for older SDK versions. Staging uses a separate endpoint that newer SDKs may expose through commands such as `dotnet nuget stage push`. The exact command names and syntax shown in this proposal are illustrative and not final.
```

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Security embargo timing
+
+.NET security releases are coordinated with public disclosure of vulnerabilities. The team cannot push packages before 10 AM Pacific because early availability would give attackers time to analyze changes and derive exploits before users can patch. But validation takes 40 to 60 minutes after push, so the packages are not available until well after disclosure. This creates a window where the vulnerability is publicly known but the fix is not yet restorable.
+
+Staging eliminates this gap. Packages are pre-validated and sitting in a staged state. At the coordinated disclosure time, promotion makes them available without waiting for validation.
```
richlander:
```suggestion
Staging eliminates this gap. Packages are pre-validated, in a staged state. At the coordinated disclosure time, promotion makes them available without waiting for validation.
```

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+
+### Pre-validation
+
+The .NET team pushes 2000+ packages to nuget.org every Patch Tuesday. Validation (malware scanning, certificate checks, repository co-signing) takes 40 to 60 minutes for that volume. If a package fails validation, the team discovers it on release day and has to rebuild, re-sign, and re-push during a tight embargo window.
```
richlander:
```suggestion
The .NET team pushes 2000+ packages to nuget.org every Patch Tuesday. Validation (malware scanning, certificate checks, repository co-signing) takes 40 to 60 minutes for that volume. If a package fails validation, the team discovers it on release day and has to rebuild, re-sign, and re-push, delaying the release of the patches but not the disclosure. Delay can create a 0-day opportunity.
```

richlander: The main point is that embargo is blown. MSRC may publish disclosure before we do.

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+After promotion, the packages enter the normal V3 pipeline and become restorable through the standard process. The public "published" timestamp is set at promotion time, not at stage time.
+
+The package detail page shows staged packages to the owner with validation status, promote, download (for inspection), replace, and delete actions.
+
+![Package Display - Staged Package](images/display-package.png)
+*Figure 4. Package detail page for a staged package showing promotion and staging group controls.*
+
+#### Trusted Publishing integration
+
+[Trusted Publishing](../2024/trusted-publishers-oidc-for-nuget-push.md) (OIDC) is the recommended authentication method for staging, but API keys are also supported. A new "Stage packages" API key scope is added alongside Push, Push only new versions, and Unlist.
```
richlander, line 126: Is there a strong reason to support API keys if we're trying to get rid of them? Or is it more work to dissallow them?

accepted/2026/package-staging.md
```diff
@@ -0,0 +1,206 @@
...
+
+![API Key Scopes - Stage packages](images/api-key-scopes.png)
+
+*Figure 5. API key management showing the new "Stage packages" scope.*
+
+Trusted Publishing configurations support a "Restrict to staging only" scope. When configured with this scope, a GitHub Actions workflow (or other supported CI/CD provider) can stage packages but cannot push directly to the public feed. This is enforced in both directions: the staging endpoints accept staging-only credentials, and the normal push endpoint rejects them with 403.
```
richlander: > Trusted Publishing configurations support a "Restrict to staging only" scope. 

Nice!

agr: How is it different from selecting only "Stage packages" option in API key scope settings? Why do we need another option here?

agr:
I don't see details about how multi-ownership is handled: does any package owner can promote staged packages or only the one that had it staged? What about potential ownership discrepancies between packages in the group (e.g. Package1 owned by A and B and Package2 owned by B and C)? Who can add packages to a group?

accepted/2026/package-staging.md
japarson: It's mentioned later that there will be a new staging-only API scope, but I'll add it here as well.

accepted/2026/package-staging.md
japarson: We're still in the beginning phase of phasing out api keys and there are still many users not onboarded to trusted publishing who can benefit from staging. For example, Trusted Publishing support for GitLab pipelines was only just recently (?) released.

accepted/2026/package-staging.md
japarson: I edited the doc to more clearly delineate that api keys will have staging scope and trusted publishing will have a staging only "constraint".

japarson:
> I don't see details about how multi-ownership is handled: does any package owner can promote staged packages or only the one that had it staged? What about potential ownership discrepancies between packages in the group (e.g. Package1 owned by A and B and Package2 owned by B and C)? Who can add packages to a group?

I added some details in the spec.

JonDouglas:
Few comments

1. I think step up 2FA needs to happen in this definition of done, not a future item.
2. I think npm makes staged content immutable, which makes most sense to me. Otherwise a compromised staging key can swap content between ready -> promote.
3. I worry about say a 30 day TTL situation. Day 0 it is scanned and cleeared, Day 29/30 when it is about to go live may be stale given how fast vulnerabilities travel nowadays. May need to validate on promotion too.
4. Approving with 2FA now but later time still keeps HITL, a human shouldn't have to be manually clicking the promote button.

Lots of prior art in App Store Connect / etc that I've seen some solid patterns in.

Issue #14738, opened 2026-02-04
Query dependencies for sponsorship opportunities


pranathibora14:
### NuGet Product(s) Involved

Other/NA

### The Elevator Pitch

Users want to be able see which of their project dependencies are in need of sponsorship using tools like Copilot, MCP Server, and CLI.

Additionally, improving AI-powered dependency sponsorship discovery can be a strong focus.

Currently, running a prompt like `Show me a list of all the nuget package dependencies in this solution that seek sponsorship` results in the agent (example: Opus 4.5) primarily looking through funding.xml in GitHub repos for the packages. When NuGet MCP is engaged it's slightly better as it looks at readme grounding for every package.

PR #14992, opened 2026-07-13
Proposal to Surface Sponsorship Information in the CLI

### LINKED ISSUE #14999: Design Spec for Sponsorships in CLI
Functional + Technical Specs for Sponsorships in CLI

kalebfik:
[Rendered Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorCliSpec/accepted/2026/sponsorship-cli.md)

### DIFF
--- a/accepted/2026/sponsorship-cli.md
+++ b/accepted/2026/sponsorship-cli.md
@@ -58,7 +58,7 @@ When a user runs `dotnet package list --sponsor`, output is grouped by project a
 
 ```text
 // sample response
-Project 'Contso.App' has the following sponsorable packages
+Project 'Contoso.App' has the following sponsorable packages
 Top-level Package        Sponsor
 > Contoso.Tools          Source: https://api.example.org/v3/index.json
                            https://github.com/sponsors/contoso
@@ -70,7 +70,7 @@ Transitive Package
 ```
 
 Sponsorship information is applied to the package ID rather than the version of a package.
-Therefore, sponsorship information remains the same for a package regardless of what version package is published, as long as those links have not been changed.
+Therefore, sponsorship information remains the same for a package regardless of which package version is published, as long as those links have not changed.
 When a package has multiple sponsorship links, the CLI will display the links in the order returned by nuget.org.
 
 Both top-level and transitive packages will be included by default when using `--sponsor`.
@@ -148,6 +148,7 @@ When a source is specified using `--source <SOURCE>`, only that source is querie
 A successful Registration response with a missing or empty `sponsorshipUrls` field is treated as a successful empty result.
 
 Console output includes only sources that return one or more sponsorship URLs.
+`https://api.nuget.org/v3/index.json` will be recommended when it is not part of the user's configured sources.
 If none of the selected sources return sponsorship details for a project, the CLI displays:
 
 ```text
@@ -199,7 +200,7 @@ The exact Registration version will be finalized with the server implementation.
 ```json
 {
   "@id": "https://api.nuget.org/v3/registration5-semver1/",
-  "@type": "RegistrationsBaseUrl/<NEW_VERSION>",
+  "@type": "RegistrationsBaseUrl/{NEW_VERSION}",
   "comment": "Base URL of Azure storage where NuGet package registration info is stored. This base URL does not include SemVer 2.0.0 packages."
 }
 ```
@@ -239,7 +240,7 @@ This includes sources that do not propagate sponsorship metadata from an upstrea
 
 ### Technical Explanation
 
-The CLI will retrieve sponsorship information through the Registration API, similar to how other package metadata is passed to the CLI ([package ID-level metadata in the Registration API](https://github.com/NuGet/Home/issues/15038)).
+The CLI will retrieve sponsorship information through the Registration API, similar to how other package metadata is passed to the CLI.
 The difference is that sponsorship information will be scoped by package ID only, whereas other metadata is scoped by package ID and version.
 
 A successful nuget.org response with one or more URLs will appear in the report, while packages with no sponsorship URLs will be absent.
@@ -268,7 +269,7 @@ A package source will indicate sponsorship support through a `metadata.sponsorsh
       "https://github.com/sponsors/contoso"
     ]
   }
-  // ** End of proposal **//
+  // ** End of proposal ** //
 }
 ```
 
@@ -277,8 +278,6 @@ A package source will indicate sponsorship support through a `metadata.sponsorsh
 - **Multiple sponsorship URLs:** The CLI displays sponsorship URLs in the order received from the package source.
 - **nuget.org URL limit:** nuget.org currently enforces a maximum of 10 URLs per package.
 - **Empty state:** A missing, null, or empty `sponsorshipUrls` value is treated as a successful empty result.
-
-A detailed implementation proposal can be found [here](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/sponsorship-CLI-Spec.md).
 
 ## Drawbacks
 
@@ -306,7 +305,6 @@ If sponsorship grows into a bigger feature, a dedicated command could be revisit
 - [**Companion server spec**](https://devdiv.visualstudio.com/DevDiv/_git/NuGet.Services/pullrequest/763096?_a=files&iteration=2&base=1): Proposes implementation for supporting package ID-level metadata in the Registration API (internal link).
 - **[`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)**: `npm fund` provides precedent for this pattern. The `funding` field lives in package metadata, and npm's guidance suggests keeping funding links at the package or author level.
   npm notes that funding information can be noisy in the CLI and that stale information could be problematic.
-- **[Gallery UI Sponsorship Implementation](https://github.com/NuGet/Engineering/blob/prabora-needs-sponsorship-feature/Server.Specs/2025/NeedsSponsorship.md)**: companion server-side proposal for surfacing sponsorship needs directly in the nuget.org Gallery UI; relevant precedent/parallel effort for surfacing the same underlying sponsorship data.
 - **`--deprecated`/`--vulnerable`**: precedent for opt-in report-style information that consumers already use and understand, informing this proposal's output/UX conventions.
 - **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
 - **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: Open issue for a PM UI sponsorship button.
--- a/accepted/2026/sponsorship-cli.md
+++ b/accepted/2026/sponsorship-cli.md
@@ -24,17 +24,20 @@ Providing sponsor links from nuget.org to the client is the key motivation for t
 ### Why not restore?
 
 Restore runs on every build; adding informational (non-actionable) content there, even opt-in, risks alarm fatigue and encourages developers to disable the feature entirely rather than engage with it. 
-`dotnet list package --sponsor` is opt-in per invocation, run only when a developer actually wants this information, keeping sponsorship messaging out of the default build/restore loop while still making the data fully discoverable on demand.
+Restore is also performance-sensitive: checking sponsorship requires reaching out over the network for every package, and doing so as part of restore would slow down builds and add repeated network calls to the frequent background restores that happen automatically while editing code (e.g., to keep IntelliSense/autocomplete up to date), not just when a developer explicitly builds.
+
+Combining sponsorship into an existing report flag would blur `--sponsor`'s informational nature with report flags that call out actionable problems (deprecations, vulnerabilities), for limited benefit over a dedicated flag.
 
 ## Who this affects
 
 - **Package consumers** get a way to discover, from the command line, which of their dependencies are seeking sponsorship — without having to visit nuget.org for every package individually.
 - **Package authors** who've already added sponsorship links to their packages get more visibility for those links, since consumers can now see them as part of a workflow they already use (`dotnet list package`), not just on the package's web page.
-- **Repository/org admins** get transparency into which of their organization's dependencies have funding needs, which can support internal conversations about sponsoring critical open-source dependencies.
+- **Repository/org admins** get transparency into which of their organization's dependencies have funding needs, which can support internal conversations about sponsoring critical open-source dependencies. 
 
 ## Explanation
 
 This proposal adds `--sponsor` to `dotnet list package`, following the same opt-in report pattern as `--deprecated`, `--outdated`, and `--vulnerable`.
+Adding a new `dotnet sponsor` verb would cost more than introducing a new flag, and it would be inconsistent with how consumers already discover `--deprecated`, `--outdated`, and `--vulnerable`.
 
 When you run `dotnet list package`, it restores the project and lists package dependencies:
 
@@ -55,48 +58,106 @@ Project 'Contoso.App' has the following sponsorship information:
       https://opencollective.com/contoso-utilities
 ```
 
-When there are no packages looking for sponsors, there will be a message: 
+In the case that a source either supports sponsorships but has no packages looking for them or a source does not support sponsorships, a respective message will be shown: 
 
 ```text
+// No packages looking for sponsorship
 No packages in project 'Contoso.Tools' are looking for sponsors.
+```
+
+```text
+// Source does not provide sponsorship information
+Source 'X' does not provide sponsorship information
 ```
 
 For private/internal feeds, users can explicitly query nuget.org sponsorship data:
 
 ```text
-dotnet list package --sponsor --source nuget.org
+dotnet list package --sponsor --source https://api.nuget.org/v3/index.json
+```
+
+`--sponsor` also supports JSON output via `--format json`, following the same base envelope as other report types (`version`, `parameters`, `sources`, `problems`). Since sponsorship is a per-package-ID concept independent of target framework, sponsored packages are listed per-project under `sponsoredPackages`, rather than nested under `frameworks` like `--deprecated`/`--outdated`/`--vulnerable`:
+
+```json
+{
+  "version": 1,
+  "parameters": "--sponsor",
+  "sources": [
+    "https://api.nuget.org/v3/index.json"
+  ],
+  "projects": [
+    {
+      "path": "/path/to/Contoso.App.csproj",
+      "sponsoredPackages": [
+        {
+          "id": "Contoso.Forms",
+          "urls": [
+            "https://github.com/sponsors/contoso",
+            "https://patreon.com/sponsor/contoso-forms",
+            "https://opencollective.com/contoso-forms"
+          ]
+        },
+        {
+          "id": "Contoso.Utilities",
+          "urls": [
+            "https://opencollective.com/contoso-utilities"
+          ]
+        }
+      ]
+    }
+  ]
+}
+```
+
+### Sponsorship document format
+
+| Name | Type | Required | Notes |
+| --- | --- | --- | --- |
+| packageId | string | yes | The package ID this document describes |
+| sponsorshipUrls | array of objects | yes | The package's current sponsorship links |
+
+Each `sponsorshipUrls` element:
+
+| Name | Type | Required | Notes |
+| --- | --- | --- | --- |
+| url | string | yes | The sponsorship URL |
+| timestamp | string | yes | When this URL was added, in ISO 8601 |
+
+Sample response:
+
+```json
+{
+  "packageId": "Contoso.Forms",
+  "sponsorshipUrls": [
+    {
+      "url": "https://github.com/sponsors/contoso",
+      "timestamp": "2025-01-15T00:00:00Z"
+    }
+  ]
+}
 ```
 
 - **Scope:** solution-wide across all projects.
 - **Transitive packages:** respects `--include-transitive`; no sponsor-specific behavior is required.
-- **Multiple sponsorship URLs:** up to 10 URLs are stored per package.
+- **Multiple sponsorship URLs:** The CLI will display the same sponsorship URLs that are received from NuGet.org
 - **Empty state:** if a project has no sponsored packages, show a per-project message indicating none were found.
 
 ## Drawbacks
 
-- **No link priority/reorder support** With append/remove semantics, default ordering is oldest-added first. Not a v1 design need per this proposal's scope.
-- **CLI-only scope for v1.** PM UI (Phase 2), popup link-out (Phase 3), Copilot/IDE, and Azure Search-backed surfaces are deferred.
-- **Freshness is cache-policy dependent.** Origin updates can be immediate, but user visibility depends on cache behavior.
+- **No link priority/reorder support** The CLI will display sponsorship information in the same order as the queried package source, not performing any sorting of its own 
+- **CLI-only scope.** PM UI (Phase 2), popup link-out (Phase 3), Copilot/IDE, and NuGet.org/PM UI Search surfaces are deferred.
+- **Freshness is cache-policy dependent.** Sponsorship data is fetched through the local HTTP cache, which currently has a fixed 30 minute TTL (time to live) 
+
 
 ## Rationale and alternatives
 
-Proposed solution: a `--sponsor` flag on `dotnet list package`, so developers can discover which dependencies are seeking sponsors.
-
-Alternatives considered:
-
-1. **Discovery alternatives**
-- Local cache or assets-file discovery: stale/non-authoritative for the current dependency graph — could show consumers outdated sponsorship information.
-- Separate tool outside `dotnet`/`nuget.exe`: inconsistent with existing CLI patterns consumers already know.
-
-2. **Command alternatives**
+1. **Command alternatives**
 - New `dotnet sponsor` verb: higher cost than adding a report flag, and inconsistent with how consumers already discover `--deprecated`/`--outdated`/`--vulnerable` information.
-- Combining sponsorship into an existing report flag: would blur `--sponsor`'s informational nature with report flags that call out actionable problems (deprecations, vulnerabilities), for limited benefit over a dedicated flag.
 
-3. **Display alternatives**
-- JSON-only output: inconsistent with existing report UX that consumers already expect from `dotnet list package`.
-- Link-out popup flow: depends on separate nuget.org web work, deferred to Phase 3.
+2. **Display alternatives**
+- a default JSON-only output: inconsistent with existing report UX that consumers already expect from `dotnet list package`. **`--format json`** will be supported.
 
-4. **Restore-time messaging**
+3. **Restore-time messaging**
 - Restore is a noise-sensitive surface run on every build; An explicit, per-invocation `--sponsor` flag keeps this information fully opt-in without polluting the default build/restore loop.
 
 **Impact of not doing this:** sponsorship links remain visible only on the nuget.org website; there is no way for `dotnet list package`, PM UI, or other client tooling to programmatically surface which dependencies are seeking sponsorship, limiting discoverability for package authors relying on this signal for funding.
@@ -104,17 +165,17 @@ Alternatives considered:
 ## Prior Art
 
 - **[`sponsor-link.md` (2023 community proposal)](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)**: closest precedent for the overall idea — introduced the `--sponsor` flag concept, sponsor metadata authors could set, a Visual Studio UI link, and an opt-in restore-time summary message.
-- **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: open customer ask for a Package Manager UI sponsorship button, confirming real demand for Phase 2.
-- **`--deprecated`/`--vulnerable`**: precedent for opt-in report flags that consumers already use and understand.
+- **`--deprecated`/`--vulnerable`**: Direct precedent for opt-in report flags that consumers already use and understand. `--sponsor` will follow the same invocation pattern.
 - **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
+- **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: Open issue for PM UI sponsorship button
 
 ## Unresolved Questions
 
-
+- Sponsorship lookups use NuGet's local disk cache (30-minute default TTL) with no way to force a refresh today — should a `--no-cache`-equivalent be wired into `--sponsor` for v1, or is that deferred?
 
 ## Future Possibilities
 
 - **Phase 2: VS Package Manager UI hyperlinks** for sponsorship links, directly addressing [nuget/home#14739](https://github.com/nuget/home/issues/14739).
 - **Phase 3: Compact link-out CLI display** using template-driven website links (`nuget.org/.../{id}/sponsor`) once sponsor popup URLs are independently addressable.
-- **Azure Search-backed visibility** for PM UI Browse and nuget.org search experiences, helping consumers discover sponsorship needs while browsing for packages.
-- **Copilot/IDE surfaces**, per original customer interest cited in Motivation.
+- **Search and Browse Visibility** for PM UI Browse and nuget.org search experiences
+- **Copilot/IDE surfaces**: See [Issue 14738](https://github.com/NuGet/Home/issues/14738)
--- a/accepted/2026/sponsorship-cli.md
+++ b/accepted/2026/sponsorship-cli.md
@@ -160,7 +160,7 @@ A source that does not support sponsorships will have a separate message:
 
 ```text
 These sources do not provide sponsorship support:
-  https://packages.contoso.com/v3/index.json
+  C:/Program Files (x86)/Microsoft SDKs/NuGetPackages/
 ```
 
 The JSON output includes every selected source in the top-level `sources` list, which includes supported and unsupported sources, along with local sources.
--- a/accepted/2026/sponsorship-cli.md
+++ b/accepted/2026/sponsorship-cli.md
@@ -6,7 +6,7 @@
 ## Summary
 
 Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org. Developers only see this today if they visit the package's nuget.org page. 
-This proposal brings that existing data into developer tooling — starting with a new `--sponsor` flag on `dotnet list package` — so developers can discover which dependencies are seeking sponsorship without leaving their workflow.
+This proposal brings that existing data into developer tooling — starting with a new `dotnet package sponsor` command — so developers can discover which dependencies are seeking sponsorship without leaving their workflow.
 
 ## Motivation
 
@@ -15,7 +15,7 @@ The core problem is getting existing nuget.org sponsorship data into developer t
 
 This proposal adds a non-disruptive, opt-in way to surface sponsorship information, and is scoped as the first of three phases:
 
-1. **Phase 1 (this proposal): CLI** — `dotnet list package --sponsor`.
+1. **Phase 1 (this proposal): CLI** — `dotnet package sponsor`.
 2. **Phase 2: Visual Studio Package Manager UI** — surfacing sponsorship links using the same hyperlink pattern PM UI already uses for project/license/report-abuse URLs, directly addressing the demand in [nuget/home#14739](https://github.com/nuget/home/issues/14739).
 3. **Phase 3: CLI popup link-out** — opening a nuget.org-hosted sponsorship page (`nuget.org/.../{id}/sponsor`) directly from the CLI, so developers get a richer, always-current destination without the CLI needing to hardcode or format every payment provider's link itself.
 
@@ -26,27 +26,17 @@ Providing sponsor links from nuget.org to the client is the key motivation for t
 Restore runs on every build; adding informational (non-actionable) content there, even opt-in, risks alarm fatigue and encourages developers to disable the feature entirely rather than engage with it. 
 Restore is also performance-sensitive: checking sponsorship requires reaching out over the network for every package, and doing so as part of restore would slow down builds and add repeated network calls to the frequent background restores that happen automatically while editing code (e.g., to keep IntelliSense/autocomplete up to date), not just when a developer explicitly builds.
 
-Combining sponsorship into an existing report flag would blur `--sponsor`'s informational nature with report flags that call out actionable problems (deprecations, vulnerabilities), for limited benefit over a dedicated flag.
-
 ## Who this affects
 
 - **Package consumers** get a way to discover, from the command line, which of their dependencies are seeking sponsorship — without having to visit nuget.org for every package individually.
-- **Package authors** who've already added sponsorship links to their packages get more visibility for those links, since consumers can now see them as part of a workflow they already use (`dotnet list package`), not just on the package's web page.
+- **Package authors** who've already added sponsorship links to their packages get more visibility for those links, since consumers can now see them as part of a workflow (`dotnet package sponsor`), not just on the package's web page.
 - **Repository/org admins** have an easier way to see sponsorship status alongside the other package information they already manage day-to-day (versions, licenses, vulnerabilities).
 
 ## Explanation
 
-This proposal adds `--sponsor` to `dotnet list package`, following the same opt-in report pattern as `--deprecated`, `--outdated`, and `--vulnerable`.
-Adding a new `dotnet sponsor` verb would cost more than introducing a new flag, and it would be inconsistent with how consumers already discover `--deprecated`, `--outdated`, and `--vulnerable`.
+This proposal adds a new `dotnet package sponsor` command, giving sponsorship discovery a dedicated and purpose-built surface. 
 
-When you run `dotnet list package`, it restores the project and lists package dependencies:
-
-```text
-Top-level Package       Requested           Resolved
-<PACKAGE_NAME>          <REQUESTED_VERSION> <RESOLVED_VERSION>
-```
-
-With `--sponsor`, output is grouped by project and includes sponsorship links for each matching package:
+When you run `dotnet package sponsor`, output is grouped by project and includes sponsorship links for each matching package:
 
 ```text
 Project 'Contoso.App' has the following sponsorship information:
@@ -73,22 +63,23 @@ Source 'X' does not provide sponsorship information
 For private/internal feeds, users can explicitly query nuget.org sponsorship data:
 
 ```text
-dotnet list package --sponsor --source https://api.nuget.org/v3/index.json
+dotnet package sponsor --source https://api.nuget.org/v3/index.json
 ```
 
-`--sponsor` also supports JSON output via `--format json`, following the same base envelope as other report types (`version`, `parameters`, `sources`, `problems`). Since sponsorship is a per-package-ID concept independent of target framework, sponsored packages are listed per-project under `sponsoredPackages`, rather than nested under `frameworks` like `--deprecated`/`--outdated`/`--vulnerable`:
+`dotnet package sponsor` also supports JSON output via `--format json`. 
+Since sponsorship is a per-package-ID concept independent of target framework, sponsorsable packages are listed per-project under `sponsorablePackages`
 
 ```json
 {
   "version": 1,
-  "parameters": "--sponsor",
+  "parameters": "dotnet package sponsor",
   "sources": [
     "https://api.nuget.org/v3/index.json"
   ],
   "projects": [
     {
       "path": "/path/to/Contoso.App.csproj",
-      "sponsoredPackages": [
+      "sponsorablePackages": [
         {
           "id": "Contoso.Forms",
           "urls": [
@@ -140,7 +131,7 @@ Sample response:
 - **Scope:** solution-wide across all projects.
 - **Transitive packages:** respects `--include-transitive`; no sponsor-specific behavior is required.
 - **Multiple sponsorship URLs:** The CLI will display the same sponsorship URLs that are received from NuGet.org
-- **Empty state:** if a project has no sponsored packages, show a per-project message indicating none were found.
+- **Empty state:** if a project has no sponsorable packages, show a per-project message indicating none were found.
 
 ## Drawbacks
 
@@ -152,26 +143,39 @@ Sample response:
 ## Rationale and alternatives
 
 1. **Command alternatives**
-- New `dotnet sponsor` verb: higher cost than adding a report flag, and inconsistent with how consumers already discover `--deprecated`/`--outdated`/`--vulnerable` information.
+- New `dotnet package fund` verb: mirrors `npm fund` naming, kept as an alternative naming option to `dotnet package sponsor`
+
+- `dotnet list package --sponsor`: Reuses existing report-flag pattern as `--vulnerabilities` or `--outdated`.
+
 
 2. **Display alternatives**
 - a default JSON-only output: inconsistent with existing report UX that consumers already expect from `dotnet list package`. **`--format json`** will be supported.
 
 3. **Restore-time messaging**
-- Restore is a noise-sensitive surface run on every build; An explicit, per-invocation `--sponsor` flag keeps this information fully opt-in without polluting the default build/restore loop.
+- Restore is a noise-sensitive surface run on every build; an explicit, per-invocation command keeps this information fully opt-in without polluting the default build/restore loop.
 
-**Impact of not doing this:** sponsorship links remain visible only on the nuget.org website; there is no way for `dotnet list package`, PM UI, or other client tooling to programmatically surface which dependencies are seeking sponsorship, limiting discoverability for package authors relying on this signal for funding.
+4. **Bake sponsorship into package metadata**: `npm fund` has precedent for this.
+- The `funding` field lives directly in package metadata.
+- npm's own guidance suggests keeping funding links at the package or author level.
+- Doing the same for NuGet could be noisy for the CLI, and stale sponsorship information could be an issue.
+
+**Impact of not doing this:** sponsorship links remain visible only on the nuget.org website; there is no way for the CLI, PM UI, or other client tooling to programmatically surface which dependencies are seeking sponsorship, limiting discoverability for package authors relying on this signal for funding.
 
 ## Prior Art
 
-- **[`sponsor-link.md` (2023 community proposal)](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)**: closest precedent for the overall idea — introduced the `--sponsor` flag concept, sponsor metadata authors could set, a Visual Studio UI link, and an opt-in restore-time summary message.
-- **`--deprecated`/`--vulnerable`**: Direct precedent for opt-in report flags that consumers already use and understand. `--sponsor` will follow the same invocation pattern.
+- **[2023 community proposal](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)**: Initial community proposal for the overall idea — introduced the `--sponsor` flag concept, sponsor metadata authors could set, a Visual Studio UI link, and an opt-in restore-time summary message.
+- **[`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)**:  `npm fund` has precedent for this.
+`funding` field lives in metadata. 
+NPM suggests to have funding links live on package or author level.
+They also state that it is noisy to the CLI, and stale funding information could be an issue
+- **[Gallery UI Sponsorship Implementation](https://github.com/NuGet/Engineering/blob/prabora-needs-sponsorship-feature/Server.Specs/2025/NeedsSponsorship.md)**: companion server-side proposal for surfacing sponsorship needs directly in the nuget.org Gallery UI; relevant precedent/parallel effort for surfacing the same underlying sponsorship data.
+- **`--deprecated`/`--vulnerable`**: precedent for opt-in report-style information that consumers already use and understand, informing this proposal's output/UX conventions even though it uses a dedicated verb rather than a flag.
 - **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
 - **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: Open issue for PM UI sponsorship button
 
 ## Unresolved Questions
 
-- Sponsorship lookups use NuGet's local disk cache (30-minute default TTL) with no way to force a refresh today — should a `--no-cache`-equivalent be wired into `--sponsor` for v1, or is that deferred?
+None at this time
 
 ## Future Possibilities
 
--- a/accepted/2026/sponsorship-cli.md
+++ b/accepted/2026/sponsorship-cli.md
@@ -2,7 +2,7 @@
 
 - Author: [@kalebfik](https://github.com/kalebfik)
 - GitHub Issue: [10703](https://github.com/NuGet/NuGetGallery/issues/10703)
-- [Accompanying Technical Implentation Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/implementing-sponsorship-in-cli.md)
+- [Accompanying Technical Implementation Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/implementing-sponsorship-in-cli.md)
 
 ## Summary
 
@@ -57,11 +57,11 @@ When a user runs `dotnet package list --sponsor`, output is grouped by project a
 `dotnet package list --sponsor` output:
 
 ```text
-//sample response
+// sample response
 Top-level Package        Sponsor
 > Contoso.Tools          Source: https://api.example.org/v3/index.json
                            https://github.com/sponsors/contoso
-
+Transitive Package
 > Contoso.Utility        Source: https://api.example.org/v3/index.json
                            https://github.com/sponsors/contoso
                          Source: https://www.myget.org/F/contoso/api/v3/index.json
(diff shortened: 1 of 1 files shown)


accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+      https://patreon.com/sponsor/contoso-forms
+      https://opencollective.com/contoso-forms
+   > Contoso.Utilities
+      https://opencollective.com/contoso-utilities
+```
+
+When there are no packages looking for sponsors, there will be a message: 
+
+```text
+No packages in project 'Contoso.Tools' are looking for sponsors.
```
donnie-msft: You talked about a message for "your sources don't provide any sponsorship information". That should be called out here as well.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+For private/internal feeds, users can explicitly query nuget.org sponsorship data:
+
+```text
+dotnet list package --sponsor --source nuget.org
```
donnie-msft: The sample should be the URI for the source. If the source were already configured to use a name, then users wouldn't need to specify the `--source` in the first place.

```suggestion
dotnet list package --sponsor --source https://api.nuget.org/v3/index.json
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+- **Scope:** solution-wide across all projects.
+- **Transitive packages:** respects `--include-transitive`; no sponsor-specific behavior is required.
+- **Multiple sponsorship URLs:** up to 10 URLs are stored per package.
```
donnie-msft: Storage is happening on the server already. I'm not clear what this means for client. Do we limit to 10 or we display whatever comes back from the server, etc?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **Empty state:** if a project has no sponsored packages, show a per-project message indicating none were found.
+
+## Drawbacks
+
+- **No link priority/reorder support** With append/remove semantics, default ordering is oldest-added first. Not a v1 design need per this proposal's scope.
```
donnie-msft: Order should probably just match what the server provides? nuget.org could change ordering or allow publishers to change the order in the future, right?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **CLI-only scope for v1.** PM UI (Phase 2), popup link-out (Phase 3), Copilot/IDE, and Azure Search-backed surfaces are deferred.
```
donnie-msft: Client focused reviewers may not be familiar with "Azure Search" terminology. I'd rephrase this in Client terms.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **Freshness is cache-policy dependent.** Origin updates can be immediate, but user visibility depends on cache behavior.
```
donnie-msft: Be specific about what cache this is referring to and the implications of that.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Rationale and alternatives
+
+Proposed solution: a `--sponsor` flag on `dotnet list package`, so developers can discover which dependencies are seeking sponsors.
+
+Alternatives considered:
+
+1. **Discovery alternatives**
+- Local cache or assets-file discovery: stale/non-authoritative for the current dependency graph — could show consumers outdated sponsorship information.
```
donnie-msft: I don't really know what this is saying. I'd rewrite or remove it.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- Separate tool outside `dotnet`/`nuget.exe`: inconsistent with existing CLI patterns consumers already know.
```
donnie-msft: We discussed a potential new dotnet command, which you called out below. I'd just remove this. Not sure we even need to say "nuget.exe is out of scope". Others can chime in on that.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+2. **Command alternatives**
+- New `dotnet sponsor` verb: higher cost than adding a report flag, and inconsistent with how consumers already discover `--deprecated`/`--outdated`/`--vulnerable` information.
```
donnie-msft: This is great rationale for your design choice. I'd move this up under your "proposed solution" as the rationale for it.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- Combining sponsorship into an existing report flag: would blur `--sponsor`'s informational nature with report flags that call out actionable problems (deprecations, vulnerabilities), for limited benefit over a dedicated flag.
+
+3. **Display alternatives**
+- JSON-only output: inconsistent with existing report UX that consumers already expect from `dotnet list package`.
+- Link-out popup flow: depends on separate nuget.org web work, deferred to Phase 3.
```
donnie-msft: I know what this means but it's probably too vague for others to know what the popup is about. I also would say that's a future possibility with Phase 3, not necessarily an alternative proposal.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+1. **Discovery alternatives**
+- Local cache or assets-file discovery: stale/non-authoritative for the current dependency graph — could show consumers outdated sponsorship information.
+- Separate tool outside `dotnet`/`nuget.exe`: inconsistent with existing CLI patterns consumers already know.
+
+2. **Command alternatives**
+- New `dotnet sponsor` verb: higher cost than adding a report flag, and inconsistent with how consumers already discover `--deprecated`/`--outdated`/`--vulnerable` information.
+- Combining sponsorship into an existing report flag: would blur `--sponsor`'s informational nature with report flags that call out actionable problems (deprecations, vulnerabilities), for limited benefit over a dedicated flag.
+
+3. **Display alternatives**
+- JSON-only output: inconsistent with existing report UX that consumers already expect from `dotnet list package`.
```
donnie-msft: Does this mean you're not supporting something like `--sponsor --format json`?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+4. **Restore-time messaging**
+- Restore is a noise-sensitive surface run on every build; An explicit, per-invocation `--sponsor` flag keeps this information fully opt-in without polluting the default build/restore loop.
+
+**Impact of not doing this:** sponsorship links remain visible only on the nuget.org website; there is no way for `dotnet list package`, PM UI, or other client tooling to programmatically surface which dependencies are seeking sponsorship, limiting discoverability for package authors relying on this signal for funding.
+
+## Prior Art
+
+- **[`sponsor-link.md` (2023 community proposal)](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)**: closest precedent for the overall idea — introduced the `--sponsor` flag concept, sponsor metadata authors could set, a Visual Studio UI link, and an opt-in restore-time summary message.
+- **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: open customer ask for a Package Manager UI sponsorship button, confirming real demand for Phase 2.
```
donnie-msft: I would focus on the Developer Community link https://developercommunity.visualstudio.com/t/Add-sponsorlink-visibility-in-the-NuGet/10456467 and just make tracking issue 14739 a sidenote here.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **`--deprecated`/`--vulnerable`**: precedent for opt-in report flags that consumers already use and understand.
```
donnie-msft: highlight this more, it's a great point!

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
+
+## Unresolved Questions
+
+
```
donnie-msft: nit/tip: I usually put something like "None at this time" when blank.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Future Possibilities
+
+- **Phase 2: VS Package Manager UI hyperlinks** for sponsorship links, directly addressing [nuget/home#14739](https://github.com/nuget/home/issues/14739).
+- **Phase 3: Compact link-out CLI display** using template-driven website links (`nuget.org/.../{id}/sponsor`) once sponsor popup URLs are independently addressable.
```
donnie-msft: I'd file a nuget/home issue for this. We can set it to P3 for now.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+- **Azure Search-backed visibility** for PM UI Browse and nuget.org search experiences, helping consumers discover sponsorship needs while browsing for packages.
+- **Copilot/IDE surfaces**, per original customer interest cited in Motivation.
```
donnie-msft: This is the MCP server, right? We have this issue about it https://github.com/nuget/home/issues/14738

accepted/2026/sponsorship-cli.md
kalebfik: Will add!

accepted/2026/sponsorship-cli.md
kalebfik: Will update!

accepted/2026/sponsorship-cli.md
kalebfik: We would be displaying whatever comes back from the server. The reason it is 10 in the spec is because that is what is enforced on the server-side. I will update the spec to reflect this.

accepted/2026/sponsorship-cli.md
kalebfik: Great point! I didn't consider that a future implementation could change this, I will update accordingly

accepted/2026/sponsorship-cli.md
kalebfik: Will update accordingly

accepted/2026/sponsorship-cli.md
kalebfik: Will update!

accepted/2026/sponsorship-cli.md
kalebfik: Good call, will update accordingly

accepted/2026/sponsorship-cli.md
kalebfik: This was intended as the default output rather than the proposed solution, I will update the spec to reflect that

accepted/2026/sponsorship-cli.md
kalebfik: Yes! I will update the spec to include the issue.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+1. **Phase 1 (this proposal): CLI** — `dotnet list package --sponsor`.
+2. **Phase 2: Visual Studio Package Manager UI** — surfacing sponsorship links using the same hyperlink pattern PM UI already uses for project/license/report-abuse URLs, directly addressing the demand in [nuget/home#14739](https://github.com/nuget/home/issues/14739).
+3. **Phase 3: CLI popup link-out** — opening a nuget.org-hosted sponsorship page (`nuget.org/.../{id}/sponsor`) directly from the CLI, so developers get a richer, always-current destination without the CLI needing to hardcode or format every payment provider's link itself.
+
+Providing sponsor links from nuget.org to the client is the key motivation for this proposal. 
+
+### Why not restore?
+
+Restore runs on every build; adding informational (non-actionable) content there, even opt-in, risks alarm fatigue and encourages developers to disable the feature entirely rather than engage with it.
```
aortiz-msft, line 19: Restore performance is the main consideration and should be explicitly called out.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+When you run `dotnet list package`, it restores the project and lists package dependencies:
+
+```text
+Top-level Package       Requested           Resolved
+<PACKAGE_NAME>          <REQUESTED_VERSION> <RESOLVED_VERSION>
+```
+
+With `--sponsor`, output is grouped by project and includes sponsorship links for each matching package:
+
+```text
```
aortiz-msft, line 36: Dotnet list package supports json output and we are missing a json sample in the spec.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,120 @@
...
+
+## Unresolved Questions
+
+
+
+## Future Possibilities
+
+- **Phase 2: VS Package Manager UI hyperlinks** for sponsorship links, directly addressing [nuget/home#14739](https://github.com/nuget/home/issues/14739).
+- **Phase 3: Compact link-out CLI display** using template-driven website links (`nuget.org/.../{id}/sponsor`) once sponsor popup URLs are independently addressable.
+- **Azure Search-backed visibility** for PM UI Browse and nuget.org search experiences, helping consumers discover sponsorship needs while browsing for packages.
```
aortiz-msft: The NuGet client does not directly talk directly to any Azure Services. Not sure about this comment.

accepted/2026/sponsorship-cli.md
kalebfik: Will update spec to highlight this!

accepted/2026/sponsorship-cli.md
kalebfik: Will add tables and samples to visualize details

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,181 @@
...
+{
+  "version": 1,
+  "parameters": "--sponsor",
+  "sources": [
+    "https://api.nuget.org/v3/index.json"
+  ],
+  "projects": [
+    {
+      "path": "/path/to/Contoso.App.csproj",
+      "sponsoredPackages": [
```
nkolev92: sponsoredPackages has a different connotation than what we have here. 

I think this is really sponsorablePackages or packages with sponsors.

kalebfik:
@dotnet-policy-service agree company="Microsoft"

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Who this affects
+
+- **Package consumers** get a way to discover, from the command line, which of their dependencies are seeking sponsorship — without having to visit nuget.org for every package individually.
+- **Package authors** who've already added sponsorship links to their packages get more visibility for those links, since consumers can now see them as part of a workflow (`dotnet package sponsor`), not just on the package's web page.
+- **Repository/org admins** have an easier way to see sponsorship status alongside the other package information they already manage day-to-day (versions, licenses, vulnerabilities).
+
+## Explanation
+
+This proposal adds a new `dotnet package sponsor` command, giving sponsorship discovery a dedicated and purpose-built surface.
```
nkolev92: Consider providing a sample output of all sub commands.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,185 @@
...
+Project 'Contoso.App' has the following sponsorship information:
+   > Contoso.Forms
+      https://github.com/sponsors/contoso
+      https://patreon.com/sponsor/contoso-forms
+      https://opencollective.com/contoso-forms
+   > Contoso.Utilities
+      https://opencollective.com/contoso-utilities
+```
+
+In the case that a source either supports sponsorships but has no packages looking for them or a source does not support sponsorships, a respective message will be shown:
```
nkolev92: Sounds like we're leaning the direction of using the configured sources right?

If that's the case, to me the different form dotnet list package is much much smaller, so I'd be ok with `dotnet list package --sponsor` too.

More so than what we had discussed before.

nkolev92 (commented):
Looks pretty good to me. 

Just final details to iron out during the meeting.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+        }
+      ]
+    }
+  ]
+}
+```
+
+### Distinguishing "No Sponsors" from "Source does not Provide Sponsorship Information" 
+
+For package sources implementing this proposal, the sponsorshipUrls field is expected to be present in registration documents.
```
zivkan: As per [the docs on the Package Metadata (registration) resource, there can be pages when there are many versions of a package](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#registration-pages-and-leaves).  For example, NuGet.Versioning: https://api.nuget.org/v3/registration5-gz-semver2/nuget.versioning/index.json

Will `sponsorshipUrls` be on just the root json, just the pages, or both?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
+# **Surface Package Sponsorship Links to Developers**
+
+- Author: [@kalebfik](https://github.com/kalebfik)
+- GitHub Issue: [10703](https://github.com/NuGet/NuGetGallery/issues/10703)
+
+## Summary
+
+Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org. Developers only see this today if they visit the package's nuget.org page.
```
zivkan: nitpick: This line has 2 sentences

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+## Motivation
+
+Sponsorship links already exist and are maintained by package owners on nuget.org.
+The data is real and current, it's just invisible outside the website. 
+
+This proposal adds a non-disruptive, opt-in way to surface sponsorship information, and is scoped as the first of three phases:
+
+1. **Phase 1 (this proposal): CLI** — `dotnet package list --sponsor`.
+2. **Phase 2: Visual Studio Package Manager UI** — surfacing sponsorship links using the same hyperlink pattern PM UI already uses for project/license/report-abuse URLs, directly addressing the demand in [nuget/home#14739](https://github.com/nuget/home/issues/14739).
+3. **Phase 3: CLI popup link-out** — opening a nuget.org-hosted sponsorship page (`nuget.org/.../{id}/sponsor`) directly from the CLI, so developers get a richer, always-current destination without the CLI needing to hardcode or format every payment provider's link itself.
```
zivkan: This line needs clarification.

1. I'm not sure if "CLI popup link-out" is supposed to be sufficiently clear just from the name, or if it's the proposed name for the feature. In case the name is intended to be self-describing, I can't guess what it means.
2. "opening a nuget.org-hosted .. page .. directly from the CLI". To me this sounds like it's proposing a text-based browser to get the nuget.org html and then render it directly in the CLI.
3. "so develpers get ...." to me, this sounds like a non-technical sales pitch which doesn't tell me anything about the features or implementation.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+
+Providing sponsor links from nuget.org to the client is the key motivation for this proposal. 
+
+### Why not restore?
```
zivkan, line 17: pedantic: this seems like something that should be under rationale and alternaties, not motivation.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+
+For package sources implementing this proposal, the sponsorshipUrls field is expected to be present in registration documents. 
+
+If sponsorshipUrls is present and contains values, sponsorship information is displayed. 
+
+If sponsorshipUrls is present but empty, the package is considered to have no sponsorship links.
+
+If sponsorshipUrls is absent, the client treats sponsorship information as unavailable from that source. 
+
+**Sample registration index response using NewtonSoft.Json:**
```
zivkan: I don't think there's any need to use a real package as an example. You can obfuscate it using something like Contoso.WebApi.Client

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+          }
+        }
+      ]
+    }
+  ]
+}
+```
+
+- **Scope:** solution-wide across all projects.
+- **Transitive packages:** Included by default; users can use `--exclude-transitive` to exclude them.
```
zivkan: Firstly, I think it's very reasonable that when asking what packages are looking for sponsors, to automatically include transitive packages by default. But thinking about it from the point of view of someone looking at `dotnet package list --help`, there's going to be both `--include-transitive` and `--exclude-transitive`. How will customers know when to use which? Which of the "modes" (outdated, vulnerable, sponsor, deprecated, or default list) are include by default and which are exclude by default?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+- **Multiple sponsorship URLs:** The CLI displays sponsorship URLs in the order received from the package source.
+- **Empty state:** When a project has no sponsorable packages, display a per-project message indicating none were found.
+
+## Drawbacks
+
+- **No link priority/reorder support:** The CLI will display sponsorship information in the same order as the queried package source, not performing any sorting of its own.
```
zivkan, line 177: It's common for nuget.config files to list multiple package sources. Ideally just two, nuget.org and a company private feed, but it's also common for companies to have one feed per group. On one hand sponsorship is probably going to be nuget.org only, but if you're querying all package sources, and more than one of them return sponsorship info for a package (so, they're not using package source mapping to limit packages to a single source each), what order will they be displayed if the sponsor list is different?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+
+```bash
+dotnet package list --sponsor
+```
+
+Report-style experience aligned with existing commands.
+
+## Why not mirror npm exactly?
+
+1. Sponsorship information is treated as package metadata that's exposed by a package source instead of stored directly.
```
zivkan:
```suggestion
1. Sponsorship information needs to be queried from the package source, rather than embedded in the package.
```

I think this is what this line means. I find the original wording confusing. Sponsorship links is metadata, it's not the package contents, so I feel like saying "is treated as package metadata" is a truism that adds words without clarifying. I also find "exposed by a package source instread of stored directly" to not be very clear, but that's definitely personal preference.

donnie-msft: 👍

accepted/2026/sponsorship-cli.md
donnie-msft: One thought is does this option allow `--include-transitive false` ? If so, the default can be true as it is today and then for your command, if customers want only top-level they can provide false?

accepted/2026/sponsorship-cli.md
donnie-msft: For v1, should we hard-code a check for nuget.org, then in a future client, implement a merged metadata approach where other feeds could provide their own sponsorUrls?
It'd be ideal for v1 to merge, but at the same time, we have no evidence customers will implement sponsorship URLs on their own feeds.

accepted/2026/sponsorship-cli.md
kalebfik: `sponsorshipUrls` would just be on the root json

accepted/2026/sponsorship-cli.md
kalebfik: I can see where it is unclear. 
The pop-up link is referring to the server-side implementation of this feature [(blog post for the feature)](https://devblogs.microsoft.com/dotnet/announcing-sponsorship-on-nugetdotorg-for-maintainer-appreciation/). 
I will update the spec to provide more clarity.

accepted/2026/sponsorship-cli.md
kalebfik: This is a good question. I am actively exploring how a separate source (in this case, Azure Artifacts) works with upstreaming registration information. There could be cases where that information is never reflected in other sources.

accepted/2026/sponsorship-cli.md
kalebfik: Looking at the different modes, none of them have transitive packages showing by default. In this case, sponsorships would be the first (at least to my findings). One solution to users knowing that it includes transitive packages by default could be adding to --help signaling that they are on by default. Sponsorships would stand out, but I also think having to change the other flag descriptions to indicate whether they include/exclude transitive packages would be a bit much. 

I am open to other suggestions/solutions! This is just a floating idea specific to your question.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,345 @@
...
+This was chosen because sponsorship information is currently treated as package metadata rather than a standalone workflow.
+
+Rationale: 
+- Aligns with existing package reporting patterns
+- Sponsorhsip information is retrieved alongside other package metadata.
+- Allows reuse of existing package discovery and reporting infrastructure
+
+The current proposal focuses on displaying sponsorship information rather than providing sponsorship-specific actions. 
+
+1. **Command alternatives**
```
aortiz-msft: I think this entire list (lines 219 to 231) can be removed. 

1. The fund naming can be mentioned in 191. Example:  Why not `dotnet package sponsor` or `dotnet package fund` ?
2. JSON-only output was never really an option, so let's remove it.
3. Restore-time messaging was already mentioned at the top of the document. @zivkan has a comment saying that it could be moved to this section. Either approach sounds good as long as it's not repeated in the doc.
4. We will not revisit the sponsorship information being baked into the package metadata as part of this document. That was covered as part of the initial specification and out of scope for this design, so we should remove this.
 
Impact of not doing this section can also be merged with Motivation at the top of this document.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,337 @@
...
+npm notes that funding information can be noisy in the CLI and stale information could be problematic.
+- **[Gallery UI Sponsorship Implementation](https://github.com/NuGet/Engineering/blob/prabora-needs-sponsorship-feature/Server.Specs/2025/NeedsSponsorship.md)**: companion server-side proposal for surfacing sponsorship needs directly in the nuget.org Gallery UI; relevant precedent/parallel effort for surfacing the same underlying sponsorship data.
+- **`--deprecated`/`--vulnerable`**: precedent for opt-in report-style information that consumers already use and understand, informing this proposal's output/UX conventions even though it uses a dedicated verb rather than a flag.
+- **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
+- **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: Open issue for PM UI sponsorship button
```
aortiz-msft, line 247: Could you also please add a link to the NuGet.org specification? (Maybe change the title to Prior Art / Related?)

kalebfik: Will do! That said, it does live in the Server repo which is private so those who don't have permissions won't be able to access it

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
+# **Surface Package Sponsorship Links to Developers**
+
+- Author: [@kalebfik](https://github.com/kalebfik)
+- GitHub Issue: [10703](https://github.com/NuGet/NuGetGallery/issues/10703)
+
+## Summary
+
+This spec proposes adding a `--sponsor` flag to the exisitng `dotnet package list` command to expose sponsorship links in the CLI using the Registration API.
```
donnie-msft, line 8: My understanding is this is a design spec, and the https://github.com/NuGet/Home/pull/14997 spec is a technical spec.
With that mindset, I'd expect the summary here to be more about the _why_ and less about the _how_.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+
+## Motivation
+
+Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org.
```
donnie-msft, line 13: Link to the spec or blogpost integrating sponsors on nuget.org?
Example, https://learn.microsoft.com/nuget/nuget-org/package-sponsorship-on-nuget-org

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+## Summary
+
+This spec proposes adding a `--sponsor` flag to the exisitng `dotnet package list` command to expose sponsorship links in the CLI using the Registration API.
+
+
+## Motivation
+
+Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org. 
+Developers only see this today if they visit the package's nuget.org page. 
+This feature will assist in increasing discoverability for packages that need or are seeking funding.
```
donnie-msft, line 15: nit: wording?

```suggestion
Surfacing sponsorship data in the Client will assist in increasing discoverability for packages that need or are seeking funding. 
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Motivation
+
+Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org. 
+Developers only see this today if they visit the package's nuget.org page. 
+This feature will assist in increasing discoverability for packages that need or are seeking funding. 
+
+### Why not restore?
+
+Restore runs on every build; adding informational (non-actionable) content there, even opt-in, risks alarm fatigue and encourages developers to disable the feature entirely rather than engage with it.
```
donnie-msft, line 19: Is the feature capable of being disabled?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+## Motivation
+
+Package maintainers can already add sponsorship links (GitHub Sponsors, Patreon, Open Collective, etc.) to their packages on nuget.org. 
+Developers only see this today if they visit the package's nuget.org page. 
+This feature will assist in increasing discoverability for packages that need or are seeking funding. 
+
+### Why not restore?
+
+Restore runs on every build; adding informational (non-actionable) content there, even opt-in, risks alarm fatigue and encourages developers to disable the feature entirely rather than engage with it. 
+Restore is also performance-sensitive: checking sponsorship requires reaching out over the network for every package, and doing so as part of restore would slow down builds and add repeated network calls to the frequent background restores that happen automatically while editing code (e.g., to keep IntelliSense/autocomplete up to date), not just when a developer explicitly builds.
```
donnie-msft, line 20: nit: sentence is a bit lengthy. Condense?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+                         https://domain3.com/sponsors/user
+                         https://domain3.com/sponsors/user2
+                         https://domain4.com/123
+                         https://domain4.com/456
+                         https://domain4.com/789
+```
+
+Two distinct empty states are possible, and they are given distinct per-package markers so a user scanning only the table can distinguish "this package has no sponsors" with "this source can't tell us either way":
+
+- `(none)` — the source supports sponsorship data and confirms this package has no active sponsorship links.
```
donnie-msft, line 58: I believe you've talked about only showing packages that have sponsors in the output.
Do we still need the `(none)` status in that case?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+- `(unsupported)` — the source does not provide sponsorship data at all, so no package from that source can be evaluated; this is a source-level limitation, not information about the package itself.
```
donnie-msft, line 59: When would a package be in this scenario?
My thought is it would have to be part of a configuration where at least 1 package source supports sponsorship and 1 package source which does not. 
Then, this package was found on package source which does not support sponsorship.

I'm wondering if the idea of "just show packages that we found sponsors for" and "output a message indicating no package sources provide sponsorship data" is enough to cover the concern?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+**Zero Sponsorship Links** (source supports sponsorship data; package has none)
+
+```text
+Top-level Package        Sponsor
+> Contoso.Tools          (none)
+```
```
donnie-msft, line 66: Would we output this or would we just show a message? This implies that we are listing 100% of all packages and putting `(none)` in every Sponsor column.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+**Source doesn't provide sponsorship information** 
+
+```text
+Top-level Package        Sponsor
+> Contoso.Tools          (unsupported)
+
```
donnie-msft, line 73: See other feedback. Table/JSON output may not be needed for this scenario.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+Source 'X' does not provide sponsorship data
+```
+
+Users may explicity specify a package source using `--source`.
```
donnie-msft, line 77: We should mention that if Package Source Mapping is enabled, then this will fail unless the provided source is mapped (But in that case, there was no need to specify `--source` in the first place)

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+Users may explicity specify a package source using `--source`. 
+
+```text
+dotnet package list --sponsor --source https://api.nuget.org/v3/index.json
+```
+
+`dotnet package list --sponsor` also supports JSON output via `--format json`. 
+Since sponsorship is a per-package-ID concept independent of target framework, sponsorable packages are listed per-project under `sponsorablePackages`
+
+```json
```
donnie-msft, line 86: Let's show the exact command prior to each output.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+```bash
+dotnet package sponsor
+```
+
+This approach treats sponsorships similarly to npm's `npm fund` command. 
+
+Benefits considered: 
+- Clear intent and discoverability
+- Provides room for future sponsorship-focused experiences. 
+- Naturally supports interactive workflows if introduced in future implementations.
```
donnie-msft, line 194: I know you had considered the differences between `dotnet package list` and `dotnet package sponsor` and decided to focus on the filtering mechanism we have already in `dotnet package list`. This section captures that but the extra details here feel a bit subjective/distracting. I'd make it a little more focused.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+```bash
+dotnet package list --sponsor 
+```
+
+This was chosen because sponsorship information is currently treated as package metadata rather than a standalone workflow.
+
+Rationale: 
+- Aligns with existing package reporting patterns
+- Sponsorhsip information is retrieved alongside other package metadata.
+- Allows reuse of existing package discovery and reporting infrastructure
```
donnie-msft, line 212: Someone repeats the above. I'd consolidate this entire section.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+While using the Search API would be a easier technical implementation, installed-package workflows retrieve metadata through the Registration API. 
+Having sponsorship information published through the Registration API would allow that data available without requiring additonal endpoint lookups.
+
+Additionally, the command `dotnet package list` looks at packages that are installed and relies on Registration metadata. 
+Using the Search API would require a second retrieval path, while publishing sponsorships through Registration allows the command to retrieve sponsorship data alongside existing package metadata using the same data flow. 
+
+
+
+
+
```
donnie-msft, line 236: nit: blank lines

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+This was chosen because sponsorship information is currently treated as package metadata rather than a standalone workflow.
+
+Rationale: 
+- Aligns with existing package reporting patterns
+- Sponsorhsip information is retrieved alongside other package metadata.
+- Allows reuse of existing package discovery and reporting infrastructure
+
+The current proposal focuses on displaying sponsorship information rather than providing sponsorship-specific actions. 
+
+### PMUI Integration
```
donnie-msft, line 216: We could also have a separate spec for PM UI. The title of this spec is Sponsorship in CLI so it's probably not worth including PM UI details here.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+Endpoint Registartion:
```
donnie-msft, line 218:
```suggestion
Endpoint Registration: 
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+| Browse/Search | Search |
+| Install/Update/Details | Registration |
+
+**Justification for Registration API** 
+
+While using the Search API would be a easier technical implementation, installed-package workflows retrieve metadata through the Registration API. 
+Having sponsorship information published through the Registration API would allow that data available without requiring additonal endpoint lookups.
+
+Additionally, the command `dotnet package list` looks at packages that are installed and relies on Registration metadata. 
+Using the Search API would require a second retrieval path, while publishing sponsorships through Registration allows the command to retrieve sponsorship data alongside existing package metadata using the same data flow.
```
donnie-msft, line 231: This seems like technical detail? Also, I believe you're creating a dedicated spec for Registration API right?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+Having sponsorship information published through the Registration API would allow that data available without requiring additonal endpoint lookups.
+
+Additionally, the command `dotnet package list` looks at packages that are installed and relies on Registration metadata. 
+Using the Search API would require a second retrieval path, while publishing sponsorships through Registration allows the command to retrieve sponsorship data alongside existing package metadata using the same data flow. 
+
+
+
+
+
+**Impact of not doing this:** Sponsorship links remain visible only on the nuget.org website. There would be no way for the CLI, PM UI, or other client tooling to programmatically surface which dependencies are seeking sponsorship, limiting discoverability for package authors relying on this signal for funding.
```
donnie-msft, line 237: Maybe move this to Motivation or another primary section?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+- **[`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)**: `npm fund` provides precedent for this pattern. The `funding` field lives in package metadata, and npm's guidance suggests keeping funding links at the package or author level. 
+npm notes that funding information can be noisy in the CLI and stale information could be problematic.
+- **[Gallery UI Sponsorship Implementation](https://github.com/NuGet/Engineering/blob/prabora-needs-sponsorship-feature/Server.Specs/2025/NeedsSponsorship.md)**: companion server-side proposal for surfacing sponsorship needs directly in the nuget.org Gallery UI; relevant precedent/parallel effort for surfacing the same underlying sponsorship data.
+- **`--deprecated`/`--vulnerable`**: precedent for opt-in report-style information that consumers already use and understand, informing this proposal's output/UX conventions even though it uses a dedicated verb rather than a flag.
+- **PM UI's existing project/license/report-abuse links**: precedent for Phase 2 — PM UI already shows package-supplied links to consumers using an established, low-risk pattern.
```
donnie-msft, line 246: Good point here, but there's not much PM UI detail in the spec. If you decide not to separate the PM UI spec, I'd put more of this detail into the PM UI section above.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+- **[`nuget/home#14739`](https://github.com/nuget/home/issues/14739)**: Open issue for PM UI sponsorship button
+
+## Unresolved Questions
+
+None at this time
```
donnie-msft, line 251: 👀

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+## Future Possibilities
+
+While `dotnet package list --sponsor` is the proposed experience, future iterations could expand functionality:
+
+**Enhanced CLI Experiences:**
+- Interactive sponsor selection experiences.
```
donnie-msft, line 258: I'd be specific. This just means navigating the list of URLs and the ability to "Open in Browser", right?

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+- Filtering, grouping, or prioritizing sponsorship opportunities.
```
donnie-msft, line 259:
```suggestion
- Filtering, grouping, or prioritizing sponsorship ..(providers?)....
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+- Additional ecosystem engagement scenarios inspired by commands like `npm fund`.
+
+- **Phase 2: VS Package Manager UI hyperlinks** for sponsorship links, directly addressing [nuget/home#14739](https://github.com/nuget/home/issues/14739).
+- **Phase 3: Link to Package Specific Sponsor Page** using template-driven website link (`nuget.org/.../{id}/sponsor`) to reach the sponorship page specific to the package on nuget.org.
+- **Copilot/IDE surfaces**: See [Issue 14738](https://github.com/NuGet/Home/issues/14738)
```
donnie-msft, line 264: I think #14738 is about AI Agents, not specifically Copilot and not related to IDEs.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+
+- Funding information is stored within package metadata.
+- Funding information is distributed through the package ecosystem.
+- Funding discovery has dedicated command surface (`npm fund`).
+
+## Comparison with Proposed Approach
+
+### Metadata Source
+
+#### NPM
```
donnie-msft, line 317: There may be too many section headings overall, but especially here. I'd only use a new heading level if there's going to be 2 distinct sections at that level.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+Report-style experience aligned with existing commands.
+
+## Why not mirror npm exactly?
+
+1. Sponsorship information needs to be queried from the package source, rather than embedded within the package. 
+2. Sponsorship discovery follows existing NuGet resolution behavior 
+3. Restore output remains unchanged to avoid additional noise.
+Users explicitly opt into sponsorship discovery instead of receiving sponsorship messaging during restore operations.
```
donnie-msft, line 348: Should this be the 4th bullet?

donnie-msft (changes requested):
Great progress! I know this doc has gotten edited with many iterations of feedback. I think it needs to be consolidated a bit and structured so that it highlights your key design ideas.

accepted/2026/sponsorship-cli.md
kalebfik: I intended to frame it as an opt-in feature, is disabling opt-in features standard? If so, then yes

accepted/2026/sponsorship-cli.md
kalebfik: That could be a solution (for sources that don't support sponsorships), at least in this first iteration. I wanted to find a way to visually differentiate the two so that users don't automatically assume the package isn't seeking funding. 

Specifically, for `(none)`, I realize it could pollute the CLI more than I intended, especially for people who have many packages that don't have sponsorship links. It was there because of the proposed `(unsupported)` message, and it is difficult to imagine a scenario where we show that sponsorships aren't supported and not show that there are no sponsorship links in a specific package.

accepted/2026/sponsorship-cli.md
kalebfik: I think this is the most straightforward approach to acknowledging us not having sponsorship information for a specific package or package source. 

Similar to your example, I was thinking of it from the perspective that nuget.org could support sponsorship information, but someone getting their packages from Azure Artifacts or a third party that doesn't support sponsorship information might get the same package without that information.

accepted/2026/sponsorship-cli.md
kalebfik: See my comment [above](https://github.com/NuGet/Home/pull/14992#discussion_r3751739087)

accepted/2026/sponsorship-cli.md
kalebfik: That is true, PSM would take out the need to specify a source, and I am not opposed to that.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+| ❌ | ✅ | ✅ | Not provided | No sponsorship lookup occurs because no packages are mapped to a supported sponsorship source. |
+| - | - | - | NuGet.org | All package IDs are queried against the specified source. |
+| - | ✅ | - | Non-NuGet.org source | The CLI reports that the source does not support sponsorship reporting in V1. |
+
+
+
+### Technical Explanation
+
+The CLI will retrieve sponsorship information through the Registration API, similar to how other package metadata is passed to the CLI. 
+The difference is that sponsorship information will be scoped by PackageID-only, whereas other metadata is scoped by its PackageID + version.
```
donnie-msft: Need a reference somewhere to the spec for that Registration API change. 
Could use the issue for now? https://github.com/NuGet/Home/issues/15038

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+| - | ✅ | - | Non-NuGet.org source | The CLI reports that the source does not support sponsorship reporting in V1. |
+
+
+
+### Technical Explanation
+
+The CLI will retrieve sponsorship information through the Registration API, similar to how other package metadata is passed to the CLI. 
+The difference is that sponsorship information will be scoped by PackageID-only, whereas other metadata is scoped by its PackageID + version. 
+
+A successful Nuget.org response with one or more URLs (maximum 10 URLs per server enforcement) will appear in the report, while packages with no sponsorship URLs will be absent.
```
donnie-msft:
```suggestion
A successful nuget.org response with one or more URLs (maximum 10 URLs per server enforcement) will appear in the report, while packages with no sponsorship URLs will be absent. 
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+  // ** End of proposal **//
+}
+```
+
+- **Scope:** solution-wide across all projects.
+- **Transitive packages:** Included by default; When using `dotnet package list -help`, the description will include that transitive packages are included by default.
+- **Multiple sponsorship URLs:** The CLI displays sponsorship URLs in the order received from the package source.
+- **Empty state:** When a project has no sponsorable packages, display a per-project message indicating none were found.
+
+A detailed implementation proposal can be found [here](https://github.com/NuGet/Home/blob/2974a40664a05cedda0f33ab6fb9ca40692153f0/accepted/2026/sponsorship-CLI-Spec.md).
```
donnie-msft: If you use the branch name instead of the commit, it'll stay current with new commits.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Drawbacks
+
+- **Nuget.org-only support**
```
donnie-msft: Why is this a drawback? eg,

```suggestion
- **Currently only implemented by NuGet.org, meaning customers using other, non-NuGet.org package sources will not see sponsors, even for packages which have links configured on NuGet.org**
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+
+NuGet: Sponsorship information is retrieved from the selected package source and surfaced through client tooling.
+
+```bash
+dotnet package list --sponsor
+```
+
+Report-style experience aligned with existing commands.
+
+## Why not mirror npm exactly?
```
donnie-msft: I'd look over the heading levels in this document. For example, this sounds like a sub-section but it's the same level `##` as the Appendix that it's supposed to be in.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,322 @@
...
+## Unresolved Questions
+
+- If future implementation extends to other sources, how should packages with no sponsorship information and sources that don't provide sponsorships be differentiated? 
+
+## Future Possibilities
+
+- **Multiple package sources:** Allow other feeds to provide sponsorship information 
+- **Dedicated sponsorship command:** Introduce a `dotnet package sponsor` or `dotnet package fund` command if sponsorships develop into a larger workflow
+- **Interactive CLI experience:** Allow users to navigate a package's sponsorship links using their arrow keys and open one in their native browser
+- **Filter top-level and transitive packages**
```
donnie-msft: These sound like sub-bullets of the dedicated command possibility.

accepted/2026/sponsorship-cli.md
donnie-msft: Unresolving as somewhere along the way this feedback has been lost.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,367 @@
...
+When PSM is disabled, every package ID is queried against every selected source.
+For privacy reasons, PSM is honored for sponsorship reporting.
+When PSM is enabled, each package ID is queried only against sources mapped to that package.
+
+Specifying a source with `--source <SOURCE>` replaces the configured source selection but does not bypass PSM.
+If a package is not mapped to a selected source, the package is not queried, and the console displays:
+
+```text
+Package `Contoso.Private` was not queried for sponsorship details because none of the selected package sources match its Package Source Mapping.
+```
```
donnie-msft: Left feedback around this on the technical spec, which should be changed here as well.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,367 @@
...
+
+Currently, nuget.org is the only source that exposes sponsorship information through its Registration response.
+Other sources will be queried but will produce empty results until they implement the same contract.
```
donnie-msft: We have been discussing an improvement on this idea, so reminder to change this when that's clarified.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,367 @@
...
+Sponsorship details from an upstream source are not shown unless that source exposes the sponsorship details through its own Registration response.
```
donnie-msft:
```suggestion
Sponsorship details from an Azure Artifacts upstream source are not shown unless that source exposes the sponsorship details through its own Registration response.
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,367 @@
...
+
+For the initial implementation of this feature, nuget.org will be the only source used to retrieve sponsorship information.
+While the server design is intended to be extensible to other feeds, we don't know when or if other feeds will implement sponsorship. To avoid complicating the client logic, results from multiple feeds won't be merged.
```
donnie-msft: We no longer need to merge, as we are just listing per-source.
Additionally, we may have a way to let sources declare if they support id-level metadata.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,386 @@
...
+No sponsorship details were returned using the following package sources:
+  https://api.example.org/v3/index.json
+
+Consider specifying another package source with `--source <SOURCE>`.
+```
+
+A source that does not support sponsorships will have a separate message:
+
+```text
+These sources do not provide sponsorship support:
```
donnie-msft, line 165: I think this is fine. We could only show this for `--verbosity detailed` or something.
No strong opinion from me.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,386 @@
...
+```
+
+The JSON output includes every selected source for each project.
+For sources that return no sponsorship details, the `sponsorablePackages` field remains empty, as follows:
+
+```json
+{
+  "version": 1,
+  "parameters": "--sponsor --format json",
+  "problems": [
```
donnie-msft, line 181: For `--vulnerable`, we output a list of sources here. Perhaps we should do the same and then in the "projects" below we could only list sources which have sponsors? This would reduce noise in the output while still providing the list of sources we checked against.

(note, in this case the local package feed could be listed as an enabled source, but we wouldn't query it since it's local and therefore not able to provide Registration)

```
"sources": [
    "https://data.nuget.org/v3/index.json",
     "C:/Program Files (x86)/Microsoft SDKs/NuGetPackages/"
  ],
````

File: accepted/2026/sponsorship-cli.md [hunk 70]
```diff
@@ -0,0 +1,386 @@
...
+    {
+      "project": "/path/to/Contoso.App.csproj",
+      "level": "warning",
+      "text": "This source does not provide sponsorship support: https://packages.contoso.com/v3/index.json"
+    }
+  ],
+  "projects": [
+    {
+      "path": "/path/to/Contoso.App.csproj",
+      "sources": [
```
donnie-msft: FYI, Mentioned above that we may want to only list sources here that support Sponsorship metadata in Registration.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,386 @@
...
+> Contoso.Utility        Source: https://api.example.org/v3/index.json
+                           https://github.com/sponsors/contoso
+                         Source: https://www.myget.org/F/contoso/api/v3/index.json
+                           https://buymeacoffee.com/contoso
+```
+
+Sponsorship information is applied to the package ID rather than the version of a package, which is why packages are listed once per package ID.
+When a package has multiple sponsorship links, the CLI will display the links in the order returned by nuget.org.
+
+Both top-level and transitive packages will be included by default when using `--sponsor`.
```
donnie-msft, line 76: We group by top level & transitive for `--vulnerable`. 

I propose that we remove "sources" and replace it with `topLevelPackages` & `transitivePackages`.
This proposed grouping is now project->top/transitive -> source providing 1 or more sponsor -> packageId which has 1 or more sponsor.

```json
{
  "projects": [
    {
      "path": "/path/to/Contoso.App.csproj",
      "topLevelPackages": [
        {
          "source": "https://api.example.org/v3/index.json",
          "sponsorablePackages": [
            {
              "id": "Contoso.Tools",
              "urls": [
                "https://github.com/sponsors/contoso"
              ]
            },
            {
              "id": "Contoso.Utility",
              "urls": [
                "https://github.com/sponsors/contoso"
              ]
            }
          ]
        },
        {
          "source": "https://www.myget.org/F/contoso/api/v3/index.json",
          "sponsorablePackages": [
            {
              "id": "Contoso.Utility",
              "urls": [
                "https://buymeacoffee.com/contoso"
              ]
            }
          ]
        }
      ],
      "transitivePackages": [
        {
          "source": "https://api.example.org/v3/index.json",
          "sponsorablePackages": [
            {
              "id": "Contoso.Tools.OpenApi",
              "urls": [
                "https://github.com/sponsors/contoso"
              ]
            }
          ]
        }
      ]
    }
  ]
}
```

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,386 @@
...
+
+Console output includes only sources that return one or more sponsorship URLs.
+If none of the selected sources return sponsorship details for a project, the CLI displays:
+
+```text
+//sample response
+No sponsorship details were returned using the following package sources:
+  https://api.example.org/v3/index.json
+
+Consider specifying another package source with `--source <SOURCE>`.
```
donnie-msft: Sounds fine to me IFF package source mapping is disabled.

Should we go ahead and use nuget.org in the example?

```suggestion
Consider specifying an additional package source that provides sponsorship metadata, for example: `--source https://api.nuget.org/v3/index.json`.
```

accepted/2026/sponsorship-cli.md
zivkan: If we assume most sources will accurately mirror the source of truth, then we could de-duplicate lists and I think that'll make the average size go down.  Something like:

```json
{
  "packages": [
    {
      "package": "Contoso.Utility",
      "sponsorUrls": [
        {
          "sources": [ "feed1", "feed2" ],
          "urls": [ "https://whatever", "https://something" ]
       }
    }
   ]
}
```

This way the only time that `sponsorUrls` is an array with `length != 1` is when a source disagrees about the complete list. But it avoids duplicating the package name, and the URL list

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,386 @@
...
+
+The command also supports `dotnet package list --sponsor --format json` and will produce output such as:
+
+```json
+{
+  "version": 1,
+  "parameters": "--sponsor --format json",
+  "projects": [
+    {
+      "path": "/path/to/Contoso.App.csproj",
```
zivkan: The text output didn't propose grouping by project, and I don't think the json should either. This is a format specific for sponsor, not a shared format.  If we think it's useful for customers to know which projects a package is used by, when getting sponsorship info, I think adding a `projects` array to a package object makes more sense.

martinrrm (dismissed):
LGTM

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,398 @@
+# **Proposal to Surface Sponsorship Information in the CLI**
+
+- Author: [@kalebfik](https://github.com/kalebfik)
+- GitHub Issue: [10703](https://github.com/NuGet/NuGetGallery/issues/10703)
+- [Accompanying Technical Implementation Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/implementing-sponsorship-in-cli.md)
```
donnie-msft, line 5: I'd just remove this reference. Once merged, this branch will be deleted and the link will become invalid.

accepted/2026/sponsorship-cli.md
```diff
@@ -0,0 +1,398 @@
...
+}
+```
+
+- **Scope:** solution-wide across all projects.
+- **Transitive packages:** Included by default; when using `dotnet package list --help`, the description will state that transitive packages are included by default.
+- **Multiple sponsorship URLs:** The CLI displays sponsorship URLs in the order received from the package source.
+- **nuget.org URL limit:** nuget.org currently enforces a maximum of 10 URLs per package.
+- **Empty state:** A missing, null, or empty `sponsorshipUrls` value is treated as a successful empty result.
+
+A detailed implementation proposal can be found [here](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/sponsorship-CLI-Spec.md).
```
donnie-msft: Same as above, I'd remove this.

donnie-msft (dismissed):

donnie-msft (approved):

Nigusu-Allehu (approved):

PR #14997, opened 2026-07-16
Technical implementation proposal for Sponsorship in CLI

### LINKED ISSUE #14999: Design Spec for Sponsorships in CLI

kalebfik:
Technical Design Spec for adding sponsorship urls into the CLI

[Rendered Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorTechSpec/accepted/2026/implementing-sponsorship-in-cli.md)


accepted/2026/sponsorship-Technical-Spec.md
```diff
@@ -0,0 +1,327 @@
...
+#### Proposed `dotnet package sponsor` design
+
+`dotnet package sponsor` fetches sponsorship data via `PackageSearchResource` against the feed's `SearchQueryService`.
+
+**Current flow:**
+
+```mermaid
+flowchart LR
+    A["<b>Gallery Database</b><br/>Stores package state, including sponsorship URLs"]
+    A --> B["<b>Catalog Writer Job</b><br/>Builds append-only catalog entries"]
```
pranathibora14: I believe mutable properties like owners skip the catalog. I think it's read straight from the DB and compared with prior snapshot.

accepted/2026/sponsorship-Technical-Spec.md
```diff
@@ -0,0 +1,327 @@
...
+    E --> F["<b>NuGet Client</b><br/>PackageSearchResource / SearchQueryService"]
+
+    A --> G["<b>Sponsorship URL Service</b><br/>Direct DB writes for website use"]
+    G --> H["<b>Gallery Website</b><br/>Displays sponsorship links"]
+```
+
+**Proposed server flow:**
+
+- Add a **sub-command**, e.g. `UpdateSponsorshipCommand`, to the existing `Auxiliary2AzureSearchCommand` array.
+- Following the `UpdateOwnersCommand` pattern, read `PackageRegistration.SponsorshipUrls` per package ID from the Gallery database, diff against the last-indexed snapshot, and push partial-document updates to Azure Search for changed IDs only, setting a new `SponsorshipUrls` field on `SearchDocument`.
```
pranathibora14: I think you should also mention how the first snapshot is created before this job takes over with comparisons with prior snapshots. I believe it's a separate job that does it.

accepted/2026/sponsorship-Technical-Spec.md
```diff
@@ -0,0 +1,327 @@
...
+    A["<b>Client</b>"]
+    A --> B["GET {service-index}/index.json"]
+    B --> C["GET {search-query-service}?q={packageId}"]
+    C --> D["Search Result incl. sponsorshipUrls"]
+    D --> E["Attach sponsorship metadata"]
+    E --> F["Render 'dotnet package sponsor' output"]
+```
+
+### Proposed alternative design:
+ A dedicated `SponsorshipUriTemplate` V3 resource with per-package-ID fetch, consumed by a new `--sponsor` flag within `dotnet list package`
```
pranathibora14: Our aim with alternative approach is still `dotnet package sponsor` right?

accepted/2026/sponsorship-Technical-Spec.md
```diff
@@ -0,0 +1,327 @@
...
+- **Response:** the `packageId`/`sponsorshipUrls[]` document already defined below in "Sponsorship document format."
+- **Missing ID / no sponsorship data:** returns `404`, matching the existing behavior of `registration5-semver1` for unknown/unlisted package IDs — the client treats `404` as "no sponsorship data," not an error.
+- **Hosting:** served as a static blob from the same Azure Storage account/CDN tier used for registration and flat-container blobs today, not from the Gallery web app — keeps it cacheable and off the SQL-backed request path.
+
+#### Current flow
+
+```mermaid
+flowchart LR
+    A["<b>Gallery Database</b><br/>Stores package state, including sponsorship URLs"]
+    A --> B["<b>Catalog Writer Job</b><br/>Builds append-only catalog entries"]
```
pranathibora14: Can you confirm if catalog writer job will get involved here?

accepted/2026/sponsorship-Technical-Spec.md
```diff
@@ -0,0 +1,327 @@
...
+    D --> E["<b>Registration API</b><br/>Static files served to clients"]
+    E --> F["<b>NuGet Client</b><br/>Reads package metadata"]
+
+    A --> G["<b>Sponsorship URL Service</b><br/>Direct DB writes for website use"]
+    G --> H["<b>Gallery Website</b><br/>Displays sponsorship links"]
+```
+
+#### Proposed server flow: DB → API data flow
+
+- A new, dedicated **Sponsorship Snapshot Job**:
```
pranathibora14: It might be good to mentioned that you want to consider reusing the `Auxiliary2AzureSearch` job here to get opinions on that and see if it's feasible.

accepted/2026/sponsorship-Technical-Spec.md
kalebfik: Yes! After further conversations, `dotnet package sponsor` is going to be the command used. The updated spec will reflect thsi

.github/workflows/icebox-watcher.yml
```diff
@@ -1,31 +0,0 @@
-name: IceBox Upvote Watcher
```
Nigusu-Allehu, line 1: assuming these removes are accidental and can be brought back

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+- `ListPackageConsoleRenderer.cs`
+  - Render `Sponsor`, `(none)`, and `(error)`, along with aligned continuation lines for multiple URLs.
+- `ListPackageJsonRenderer.cs`
+  - Emit ordered `sponsorshipUrls` and include non-fatal fetch issues in `problems` while preserving package entries.
+- `RegistrationIndex.cs` and `PackageMetadataResourceV3.cs`
+  - Propagate `sponsorshipUrls` from registration index payloads to metadata objects consumed by the list pipeline.
+
+#### Data assumptions
+
+- The client expects optional `sponsorshipUrls` as `IReadOnlyList<string>`.
```
Nigusu-Allehu: We could be more specific about what the new json in the registration resource would look like after including this new `sponsorshipUrls` property

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+```json
+{
+  "id": "Contoso.Tools",
+  "sponsorshipUrls": [
+    "https://github.com/sponsors/username"
+  ]
+}
+```
+
+**Ten sponsors** (matching NuGet.org's current policy cap)
```
nkolev92: Call out we're not validating anything client side.,

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+#### Data assumptions
+
+- The client expects optional `sponsorshipUrls` as `IReadOnlyList<string>`.
+- Missing, `null`, or empty values all mean no sponsorship data.
+- Console output uses `(none)`; JSON output uses `[]`.
+- Multiple URLs preserve server order and show as extra rows directly under the previous sponsorship link. 
+
+#### Source transport dependency and client abstraction
+
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
```
nkolev92, line 100: How do we detect that registration supports sponsorship?

I'm sure you have a plan, but just calling it out here would nbe good.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+
+- [2023 community sponsor-link proposal](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)
+- [`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)
+- Existing report-style CLI surfaces: `--deprecated`, `--vulnerable`, `--outdated`
+- PM UI's existing package link surfaces (project/license/report abuse)
+- [nuget/home#14739](https://github.com/nuget/home/issues/14739) (PM UI sponsorship button request)
+
+## Unresolved Questions
+
+1. Which command surface ships: `dotnet package list --sponsor` or `dotnet package sponsor`?
```
nkolev92: If we're not making dotnet package sponsor default to nuget.org only, then I don't mind `dotnet package list --sponsor`. Seems simnpler.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+2. Should `--include-transitive` be supported?
```
nkolev92: IMO, yes. 

Real question si what's the default.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+Packages with no sponsorship data display `(none)` in console output and `[]` in JSON.
+
+### Technical explanation
+
+Both command candidates share one sponsor-report model and one metadata pipeline.
+
+- Candidate A extends `ListPackageCommand` with a new report flag.
+- Candidate B registers `dotnet package sponsor` but still routes into the same metadata-fetch, model, and renderer internals.
+
+Planned code updates:
```
Nigusu-Allehu: nit: This is a very thorough write-up, and I appreciate the level of detail in the planned code updates. For a design spec, this may be more detail than necessary. Since implementation details can change, including too much code-level detail can make the spec stale over time.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+- Existing report-style CLI surfaces: `--deprecated`, `--vulnerable`, `--outdated`
+- PM UI's existing package link surfaces (project/license/report abuse)
+- [nuget/home#14739](https://github.com/nuget/home/issues/14739) (PM UI sponsorship button request)
+
+## Unresolved Questions
+
+1. Which command surface ships: `dotnet package list --sponsor` or `dotnet package sponsor`?
+2. Should `--include-transitive` be supported?
+For example, if package A depends on package B, should sponsorship data for package B also be shown?
+3. Should the command special-case NuGet.org by default?
```
nkolev92: If we default to nuget.org, imo we should choose `dotnet package sponsor`. 
If we don't, then I think both are an option.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+
+**Zero sponsors**
+
+```text
+Top-level Package        Sponsor
+> Contoso.Tools          (none)
+```
+
+```json
+{
```
Nigusu-Allehu: Should we have a version for this json

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+### Functional explanation
+
+Both command candidates produce the same sponsorship data:
+
+```bash
+# Candidate A
+dotnet package list --sponsor
+
+# Candidate B
+dotnet package sponsor
```
Nigusu-Allehu: If we go with this one let's make sure to add an allow insecure connections option like dotnet package list.

nkolev92: I'd rather have someone ask for this.

The expectation is that this would only be used against nuget.org in practice right?
that should always be https only.

accepted/2026/sponsorship-CLI-Spec.md
donnie-msft: If we support `--include-transitive`, that implies the default is top-level only.

accepted/2026/sponsorship-CLI-Spec.md
donnie-msft: The issue with defaults is what's the privacy risk of exposing package IDs to nuget.org unintentionally?

Do we simply respect package source mapping, so if it's enabled, we only check sponsorship for packages that were mapped to "https://api.nuget.org/v3/index.json"?
All other packages IDs mapped to other sources, we say "N/A" or something for the sponsorship URL output?

accepted/2026/sponsorship-CLI-Spec.md
donnie-msft: Yes, I'd rephrase to say Client is showing whatever the server provides. And as of the time of this spec, that's max 10 URLs.

accepted/2026/sponsorship-CLI-Spec.md
nkolev92: > If we support --include-transitive, that implies the default is top-level only.

Correct. 

Maybe we keep it simple and we just show them all? 

I honestly don't know if people would sponsor transitive packages and how they'd feel about that.

accepted/2026/sponsorship-CLI-Spec.md
nkolev92: I think if we add it to `dotnet package list --sponsor` then we should implement PSM. 
Or at least assume we will implement it at some point. 

> All other packages IDs mapped to other sources, we say "N/A" or something for the sponsorship URL output?

Assuming we can detect the difference between no sponsorship details and no sponsorship period, then we can say N/A or whatever. 
We can also just not show it at all. IMO, I'd just keep it simple and not show it at all.

accepted/2026/sponsorship-CLI-Spec.md
kalebfik: updating spec to reflect this!

accepted/2026/sponsorship-CLI-Spec.md
donnie-msft: Your proposal sounds more like, show all by default, then have `--exclude-transitive` to only output the top-level?

nkolev92: no, more like it lists all by default enumerated and that's it.

I'm probably overthinking it  though :)

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+    "https://github.com/sponsors/user",
+    "https://github.com/sponsors/user2",
+    "https://ko-fi.com/123",
+    "https://ko-fi.com/456",
+    "https://ko-fi.com/789"
+  ]
+}
+```
+
+Packages with no sponsorship data display `(none)` in console output and `[]` in JSON.
```
zivkan: Using [OrchardCore](https://github.com/OrchardCMS/OrchardCore) as an example, and using copilot to help me de-duplicate the `dotnet package list --include-transitive` package list, there are 355 packages used, and between 1/3 to 1/2 are Microsoft owned. There are also some AWS packages, and some Docker packages, and I don't think that these corporate-backed projects are going to ask for sponsorship.

I think we should reconsider outputting anything when a package doesn't have sponsorship information, given the large number of packages in the solution and the high percentage that are unlikely to request sponsorship.

It may be worthwhile looking for other open source .NET projects on github and doing a similar analysis to see if OrchardCore is an outlier.

accepted/2026/sponsorship-CLI-Spec.md
zivkan:
```suggestion
- The client reads optional `sponsorshipUrls` as `IReadOnlyList<string>`.
```

I think "expects" and "optional" are contradictory 😆

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+- Sources that do not emit sponsorship data cannot display sponsor URLs.
+- Scope is CLI-only in v1; PM UI and IDE surfaces are deferred.
+
+## Rationale and alternatives
+
+| Dimension | `dotnet package list --sponsor` | `dotnet package sponsor` |
+|---|---|---|
+| Pros | Aligns with existing list report flags (`--deprecated`, `--outdated`, `--vulnerable`) and established user workflow; no new command registration. | Improves discoverability; gives sponsorship a dedicated, tab-completable verb. |
+| Cons | Adds another mutually exclusive flag to an already busy command surface; not visible unless users already check `dotnet package list -h`. | Adds a new command entry point even though execution reuses list-package internals; creates two paths to identical data. |
+| Implementation cost | Lowest: one new flag, one new `ReportType` value, and an extended mutex-count check in `ListPackageCommand.GetReportType`. | Medium: new command registration in `Program.cs`, a routing shim into `ListPackageCommandRunner`, and duplicate console/JSON coverage. |
```
zivkan: nitpick: I don't think the 3 things listed for `dotnet package sponsor` make it medium implementation. I'd be shocked if copilot can't get it right on a single prompt. The new command registration in Program.cs should be just a single line, it can't be more trivial than that. There will be more command line parser scaffolding, but AI will trivialize it. I don't know what details you're thinking about for "duplicate console/JSON coverage", but the console and JSON output for `dotnet package list --sponsor` will be different to all the other commands, so there's going to be a bunch of console and JSON formatting and test code that needs to be written anyway.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+                         https://github.com/sponsors/user2
+                         https://ko-fi.com/123
+                         https://ko-fi.com/456
+                         https://ko-fi.com/789
+```
+
+```json
+{
+  "id": "Contoso.Utility",
+  "sponsorshipUrls": [
```
zivkan, line 90: nitpick: Some of these URLs go to real accounts, I don't think there's anything special about this spec that needs to go to real domain names (patreon, opencollective, github, ko-fi), so you could use fake domains in the examples to avoid accidentally linking to real accounts.

accepted/2026/sponsorship-CLI-Spec.md
nkolev92: I agree. We should treat it like a filter, the way `dotnet list package --vulnerable` and the likes work.

aortiz-msft (commented):
In this document, I'm hoping to learn which class(es) will need to be changed to hold the sponsorship information, and which classes will need to be modified for console and json output. By enumerating the changes needed, we can help you make sure no relevant places in the code are missed before the implementation starts.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+
+The command surface will be presented as `dotnet package list --sponsor`
+
+It will reuse the same metadata-fetch and rendering pipeline, with:
+
+- A `Sponsor` column in console output
+- A `sponsorshipUrls` array in JSON output
+
+## Explanation
+
```
donnie-msft, line 18: Missing `### Functional explanation`. It'd be a good place to walk through the overall functionality you're proposing. eg, what happens if there are multiple sources with sponsorship data? No sponsorship data? etc

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+### Technical explanation
+
+`dotnet package list --sponsor` extends the exsisting list-package report pipeline with a new report type.
```
donnie-msft, line 21: Let's mention the modes (CLI and json) and show sample output of each report.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+## Explanation
+
+### Technical explanation
+
+`dotnet package list --sponsor` extends the exsisting list-package report pipeline with a new report type. 
+
+Planned updates:
+
+- Add `--sponsor` handling in `ListPackageCommand`, `ListPackageArgs`, and `ReportType`, executed through `ListPackageCommandRunner`
+- Flow `sponsorshipUrls` through: `IPackageSearchMetadata`, `PackageSearchMetadata`, `LocalPackageSearchMetadata`, `PackageSearchMetadataV2Feed`, `PackageSearchMetadataBuilder`, `RegistrationIndex`, and `PackageMetadataResourceV3`
```
donnie-msft, line 26: Can you clarify why `LocalPackageSearchMetadata` will need to have sponsorship data?

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+- Preserve sponsorship data through `InstalledPackageReference` and `ListReportPackage`.
+- Render console sponsorship output through `ListPackageConsoleRenderer` and `ProjectPackagesPrintUtility`.
+- Render JSON sponsorship output through `ListPackageJsonRenderer`.
+
+#### Data assumptions
+
+- The client reads optional `sponsorshipUrls` as `IReadOnlyList<string>`.
+- A status message will be shown in the json reflecting whether sponsorship information is `available`, `none`, or `unsupported`.
```
donnie-msft, line 34: Do we need to list every package just to say sponsorship is "not supported"?
Is this about merging data from multi-source configurations? I think if that's being implemented, more detail is needed on how that merging will work. An example scenario with 1 feed that supports sponsorship and 1 feed that doesn't may help.

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+- Console output uses `(none)`; JSON output uses `[]`.
+- Multiple URLs preserve server order and show as extra rows directly under the previous sponsorship link.
```
donnie-msft, line 36: I know what this means, but reading it is a bit confusing. Maybe reword?

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+- Render JSON sponsorship output through `ListPackageJsonRenderer`.
+
+#### Data assumptions
+
+- The client reads optional `sponsorshipUrls` as `IReadOnlyList<string>`.
+- A status message will be shown in the json reflecting whether sponsorship information is `available`, `none`, or `unsupported`.
+- Console output uses `(none)`; JSON output uses `[]`.
+- Multiple URLs preserve server order and show as extra rows directly under the previous sponsorship link. 
+- At this time, there is a maximum of 10 URLs shown. 
+This is due to nuget.orgs current policy cap; client will not be validating urls, it will only show what the server provides.
```
donnie-msft, line 38: Can these sentences be combined/condensed for easier readability?

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+  ]
+  // ** End of proposal ** //
+  // ... additional information
+}  
+```
+
+#### Source transport dependency and client abstraction
+
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
+This document defines client behavior only when that data is available.
```
donnie-msft, line 101: Doesn't it define behavior when data is missing/unsupported?

accepted/2026/sponsorship-CLI-Spec.md
```diff
@@ -0,0 +1,146 @@
...
+
+#### Testing strategy
+
+- Validate command output for packages with no sponsorship information, a single sponsorship URL, and multiple sponsorship URLs.
+- Verify sponsorship rendering behavior in console output.
+- Verify JSON output for empty, single, and multiple sponsorship URL scenarios while preserving URL order.
+- Validate sponsorship metadata is correctly applied during package metadata enrichment.
+- Verify partial-result scenarios where sponsorship metadata retrieval succeeds for some package sources and fails for others.
+
+#### Extending to NuGet.exe
```
donnie-msft, line 111: This seems like too much detail to say we're not supporting nuget.exe

donnie-msft (changes requested):
Great progress! Let's make sure everything you've talked about gets reflected here.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+## Explanation
+
+### Functional Explanation
+
+When a user runs `dotnet package list --sponsor`, the CLI will examine the projects/solutions packages and retrieve sponsorship links from nuget.org. 
+The results are then grouped by project and displayed in the console, and optionally in JSON format. 
+Packages that have multiple links are returned in the order recieved by nuget.org
+
+For the initial implementation, nuget.org will be the only source that provides sponsorship information.
+If there are other package sources configured, the CLI will not query or merge sponsorship data from those sources.
```
donnie-msft: Let's investigate Azure Artifacts feeds at least to understand how upstream sources will work. Many customers won't have nuget.org configured as a direct package source, but will have it as an upstream source.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+    "https://github.com/sponsors/contoso"
+    ]
+  // ** End of proposal **//
+}
+```
+
+**Extending to NuGet.exe**
+
+`NuGet.exe` is outside the scope for this proposal. 
+The list and search commands are using the Search API, rather than the Registration API pipeline and supporting sponsorships would require a seperate design and implementation.
```
donnie-msft: I thought list used registration? I'd just remove this anyway as I'm not sure we need to explain nuget.exe in this spec.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
+Missing or empty `sponsorshipUrls` use the empty-result behavior.
+
+#### Testing strategy
+
+- Validate command output for packages with no sponsorship information, a single sponsorship URL, and multiple sponsorship URLs.
+- Verify top-level and transitive packages are included by default.
+- Verify sponsorship rendering behavior in console output.
+- Verify JSON output for single and multiple sponsorship URL scenarios while preserving URL order.
+- Validate sponsorship metadata is correctly applied during package metadata enrichment.
```
donnie-msft: I'm not sure what metadata enrichment means.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
+Missing or empty `sponsorshipUrls` use the empty-result behavior.
+
+#### Testing strategy
+
+- Validate command output for packages with no sponsorship information, a single sponsorship URL, and multiple sponsorship URLs.
+- Verify top-level and transitive packages are included by default.
+- Verify sponsorship rendering behavior in console output.
+- Verify JSON output for single and multiple sponsorship URL scenarios while preserving URL order.
```
donnie-msft: nit: I understand the goal here to be repeating some steps for CLI output and for JSON output. There may be a way to write this list so it emphasizes that those 2 scenarios need to be validated for each test case.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+`dotnet package list --sponsor` keeps sponsorship as a read-only package metadata report alongside `--deprecated`, `--vulnerable`, and `--outdated`. 
+A dedicated `dotnet package sponsor` command was considered, but this experience reports information and does not perform a sponsorship action.
+A dedicated command can be reconsidered if interactive sponsorship workflows are added later.
+
+## Prior Art
+
+- [2023 community sponsor-link proposal](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)
+- [`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)
+- Existing report-style CLI surfaces: `--deprecated`, `--vulnerable`, `--outdated`
+
```
donnie-msft, line 279: nit: extra blank lines

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Unresolved Questions
+
+If future versions support multiple sponsorship sources, how should results be merged, and how should “no sponsorship URLs” be distinguished from “source does not support sponsorship”?
```
donnie-msft: Also, what do Azure Artifacts upstream sources do? If they work with the new registration data, then they should be merged as well.

donnie-msft (commented):
I think it's pretty close! Let's verify the Azure Artifacts upstream behavior and that will close some remaining gaps.

accepted/2026/implementing-sponsorship-in-cli.md
kalebfik: Will clarify!

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+- A `Sponsor` column in console output
+- A `sponsorablePackages` array in JSON output
+
+## Explanation
+
+### Functional Explanation
+
+When a user runs `dotnet package list --sponsor`, the CLI will examine the projects/solutions packages and retrieve sponsorship links from nuget.org. 
+The results are then grouped by project and displayed in the console, and optionally in JSON format. 
+Packages that have multiple links are returned in the order received by nuget.org
```
donnie-msft: nit
```suggestion
Packages that have multiple links are returned in the order received by nuget.org.
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+The current proposal will receive sponsorship information from nuget.org. 
+The new `metadata` and `sponsorshipUrls` property (shown in technical explanation) do not automatically propagate in other sources, resulting in those sources not receiving sponsorship information.   
+
+Sponsorship information is currently queried from nuget.org
```
donnie-msft: Sneaky but significant! Without this period, the spec reads "currently queried from nuget.org in debug builds only".

```suggestion
Sponsorship information is currently queried from nuget.org.
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+In debug builds only, the environment variable `NUGET_SPONSOR_ENDPOINT_OVERRIDE` can be configured to https://apidev.nugettest.org/v3/index.json (for DEV) and https://apiint.nugettest.org/v3/index.json (INT).
```
donnie-msft: I'd move this down to the end of the Functional explanation. Let's say something about this being a testing strategy for development of the feature. Maybe a new subsection like `#### Testing strategy` or something.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+The command resolves packages from NuGet configuration and --source. 
+Packages that do not contain sponsorship links or sources that don't support sponsorship data will be omitted from the report. 
+
+The console will display:
```
donnie-msft: Can the condition be written out here? 

eg,
```suggestion
When nuget.org is a configured package source or is specified with `--source`, but no installed packages provided sponsorship data, the console will display:
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+The results are then grouped by project and displayed in the console, and optionally in JSON format. 
+Packages that have multiple links are returned in the order received by nuget.org
+
+The current proposal will receive sponsorship information from nuget.org. 
+The new `metadata` and `sponsorshipUrls` property (shown in technical explanation) do not automatically propagate in other sources, resulting in those sources not receiving sponsorship information.   
+
+Sponsorship information is currently queried from nuget.org
+In debug builds only, the environment variable `NUGET_SPONSOR_ENDPOINT_OVERRIDE` can be configured to https://apidev.nugettest.org/v3/index.json (for DEV) and https://apiint.nugettest.org/v3/index.json (INT).
+The command resolves packages from NuGet configuration and --source. 
+Packages that do not contain sponsorship links or sources that don't support sponsorship data will be omitted from the report.
```
donnie-msft: Omitted is technically correct, but we want to show a message in CLI and `problems` in JSON output.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+```text
+There are no sponsorship details found at https://api.nuget.org/v3/index.json
+```
+
+When nuget.org is not configured as a source and the command is used, the console will display a message:
+
+```text
+For sponsorship details, configure nuget.org as a package source, or allow access to nuget.org by specifying `--source https://api.nuget.org/v3/index.json`
+```
+The JSON output will include an empty `sponsorablePackages` array for those projects.
```
donnie-msft: I think if nuget.org isn't configured, we should output an Error in the JSON.

I'd like to see CLI sample output and JSON sample output for all messages.
Something like:

```json
"problems": [
  {
    "level": "warning",
    "text": "There are no sponsorship details found at https://api.nuget.org/v3/index.json"
  }
]
```
```json
"problems": [
  {
    "level": "error",
    "text": "For sponsorship details, configure nuget.org as a package source, or allow access to nuget.org by specifying `--source https://api.nuget.org/v3/index.json`"
  }
]

File: accepted/2026/implementing-sponsorship-in-cli.md [hunk 39]
```diff
@@ -0,0 +1,173 @@
...
+
+The default console output will add a `Sponsor` column. 
+The first URL will appear besides the package ID and each additional URL appears below the previous one. 
+When using `--format json`, those same packages are returned in JSON format, identified by `sponsorablePackages` and with links stored in each package's `urls` array.
```
donnie-msft: Move this json commentary below the CLI output.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+      ]
+    }
+  ]
+}
+```
+
+**Package Source Mapping and using `--source`**
+
+The proposed implementation does not apply Package Source Mapping (PSM).
+When PSM is configured, the command will not make sponsorship metadata requests unless nuget.org is stated using `--source https://api.nuget.org/v3/index.json`.
```
donnie-msft: We talked about just erroring in this case. Please add CLI/JSON samples.

```suggestion
When PSM is configured, the command will error unless nuget.org is stated using `--source https://api.nuget.org/v3/index.json`.
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Technical explanation
+
+`dotnet package list --sponsor` extends the existing package-list report pipeline with a new report type. 
+
+Planned updates:
+
+- Add `--sponsor` handling in `ListPackageCommand`, `ListPackageArgs`, and `ReportType`, executed through `ListPackageCommandRunner`
+- Deserialize `sponsorshipUrls` through `RegistrationIndex`, and propagate them to `PackageSearchMetadataRegistration` in `PackageMetadataResourceV3`
```
donnie-msft: Before deserializing, will you be validating the configured sources for nuget.org?

We now also decided to support a test environment envvar, so mentioning use of `EnvironmentVariableWrapper.Instance` to override the hard-coded nuget.org would be good.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+  "@type": [
+    "catalog:CatalogRoot", 
+    "PackageRegistration", 
+    "catalog:Permalink"
+  ],
+  "commitId": "afa91af1-9505-41b8-ad75-eab8e613db14",
+  "commitTimeStamp": "2026-04-10T00:15:25.1492389+00:00",
+  "count": 2,
+  "metadata":
+    // ** Start of proposal ** //
```
donnie-msft: I know the server change isn't part of this spec, but I think we should highlight that "metadata" is a new node as well.

```suggestion
    // ** Start of proposal ** //
    "metadata":
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+      "https://github.com/sponsors/contoso"
+      ]
+    // ** End of proposal **//
+}
+```
+
+#### Source transport dependency and client abstraction
+
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
+Missing or empty `sponsorshipUrls` use the empty-result behavior.
```
donnie-msft: Above we say
>- Only packages with one or more sponsorship urls are included. 

Can this all be combined into a single note?

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+#### Testing strategy
```
donnie-msft: Let's note that the envvar can be configured to DEV or similar as well for aiding in testing on our own.
For automated testing, we should be able to mock registration and not use any live package source.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+- Validate command output for packages with no sponsorship information, a single sponsorship URL, and multiple sponsorship URLs.
+- Verify top-level and transitive packages are included by default.
+- Verify sponsorship rendering behavior in console output.
+- Verify JSON output for single and multiple sponsorship URL scenarios while preserving URL order.
+- Validate sponsorship metadata is propagated from the Registration index and consumed by the command runner.
+
+## Drawbacks
+
+
```
donnie-msft, line 267: Some ideas

1. nuget.org is hard-coded, so other sources are not supported today.
2. If using Package Source Mapping, the feature won't work
3. ...

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Rationale and alternatives
+
+`dotnet package list --sponsor` keeps sponsorship as a read-only package metadata report alongside `--deprecated`, `--vulnerable`, and `--outdated`.
```
donnie-msft: I think you're listing this as rationale. Maybe clarifying it like,


```suggestion
The rationale for our approach using `dotnet package list --sponsor` keeps sponsorship as a read-only package metadata report alongside `--deprecated`, `--vulnerable`, and `--outdated`. 
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Prior Art
+
+- [2023 community sponsor-link proposal](https://github.com/NuGet/Home/blob/sponsor-link/proposed/2023/sponsor-link.md)
+- [`npm fund`](https://docs.npmjs.com/cli/v10/commands/npm-fund/)
+- Existing report-style CLI surfaces: `--deprecated`, `--vulnerable`, `--outdated`
+
+## Unresolved Questions
+
+If future versions support multiple sponsorship sources, how should results be merged, and how should “no sponsorship URLs” be distinguished from “source does not support sponsorship”?
```
donnie-msft: > and how should “no sponsorship URLs” be distinguished from “source does not support sponsorship”?

We have listed 2 distinct error messages already in this spec.

donnie-msft (changes requested):

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,173 @@
...
+- Validate sponsorship metadata is propagated from the Registration index and consumed by the command runner.
+
+## Drawbacks
+
+
+
+## Rationale and alternatives
+
+`dotnet package list --sponsor` keeps sponsorship as a read-only package metadata report alongside `--deprecated`, `--vulnerable`, and `--outdated`. 
+A dedicated `dotnet package sponsor` command was considered, but this experience reports information and does not perform a sponsorship action.
```
donnie-msft: I think the reason here isn't accurate. We consider it a future possibility that can support more interactive UI (like selecting/launching a sponsor hyperlink from the CLI). The reason we chose `package list` was more around consistency with existing options deprecated/vulnerable.

```suggestion
A dedicated `dotnet package sponsor` command was considered, ....
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+- Cover missing or null `metadata` and `sponsorshipUrls` behavior, along with empty arrays or malformed JSON.
+- Verify enabled configured source selection including packages mapped to one source, multiple sources, and sources excluded by `--source`.
+- Validate sponsorship metadata is propagated from the Registration index and consumed by the command runner.
+
+## Drawbacks
+
+- At this time, nuget.org is the only source that supports sponsorship details in the Registration index root. 
+  - Other sources will produce empty results until they implement the same properties. 
+  - Sponsorship details from an upstream source are not shown unless that source exposes those details through its own Registration response. 
+- PSM and an explicit `--source` can intentionally exclude sources that may contain sponsorship details.
```
donnie-msft: We discussed instead just showing an error when using PSM + `--source`.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+      ]
+    }
+  ]
+}
+```
+
+**Package Source Mapping and using `--source`**
+
+Registration requests may disclose package IDs to package sources; therefore, sponsorship reporting will honor Package Source Mapping (PSM). 
+When PSM is disabled, each package is queried against each selected source.
```
donnie-msft:
```suggestion
When PSM is disabled, each package is queried against all configured package sources. 
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+    }
+  ]
+}
+```
+
+**Package Source Mapping and using `--source`**
+
+Registration requests may disclose package IDs to package sources; therefore, sponsorship reporting will honor Package Source Mapping (PSM). 
+When PSM is disabled, each package is queried against each selected source. 
+When PSM is enabled, the client queries only sources that are selected for the command and mapped to that package.
```
donnie-msft:
```suggestion
When PSM is enabled, the client queries only sources that are mapped to that package. 
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+  ]
+}
+```
+
+**Package Source Mapping and using `--source`**
+
+Registration requests may disclose package IDs to package sources; therefore, sponsorship reporting will honor Package Source Mapping (PSM). 
+When PSM is disabled, each package is queried against each selected source. 
+When PSM is enabled, the client queries only sources that are selected for the command and mapped to that package. 
+Sources that are specified with `--source` replace the configured source selection, but they will not bypass PSM.
```
donnie-msft: I recall there was disagreement, but PSM being enabled should just be an immediate error with `--source` in my mind.

It's a very ambiguous scenario to say "don't use my package sources but do use my source mappings".

However, I don't think we should _ignore_ the source mappings ever. This is why I think we should just "error" and say you've specified a source which isn't part of the configuration.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+
+**Package Source Mapping and using `--source`**
+
+Registration requests may disclose package IDs to package sources; therefore, sponsorship reporting will honor Package Source Mapping (PSM). 
+When PSM is disabled, each package is queried against each selected source. 
+When PSM is enabled, the client queries only sources that are selected for the command and mapped to that package. 
+Sources that are specified with `--source` replace the configured source selection, but they will not bypass PSM. 
+If there is a package not mapped to any selected source, the package is not queried and the console will output a warning:
```
donnie-msft: PSM by design requires mappings for all packages in order to restore.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+**Package Source Mapping and using `--source`**
+
+Registration requests may disclose package IDs to package sources; therefore, sponsorship reporting will honor Package Source Mapping (PSM). 
+When PSM is disabled, each package is queried against each selected source. 
+When PSM is enabled, the client queries only sources that are selected for the command and mapped to that package. 
+Sources that are specified with `--source` replace the configured source selection, but they will not bypass PSM. 
+If there is a package not mapped to any selected source, the package is not queried and the console will output a warning: 
+
+```text
+Package `Contoso.Private` was not queried for sponsorship details because none of the selected package sources match its Package Source Mapping.
```
donnie-msft: These seems a bit noisy. Maybe just 1 message saying that Package Source Mapping is applied at the start of the command?

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,267 @@
...
+- Add sponsorship-specific query path in `ListPackageCommandRunner`.
+  - Query each `(source, package id)` pair once and retain the source with the ordered sponsorship URLs. 
+- Extend `ListPackageProjectModel` to retain each source and its query results. 
+  - Extend `ListReportPackage` with grouped sponsorship details.
+- Render console sponsorship output through `ListPackageConsoleRenderer` and `ProjectPackagesPrintUtility`.
+- Render JSON sponsorship output through `ListPackageJsonRenderer`.
+
+#### Data assumptions
+
+- Proposed contract is an optional array of strings on the packages Registration index root exposed as an ordered `IReadOnlyList<string>`.
```
donnie-msft: nit

```suggestion
- Proposed contract is an optional array of strings on the package's Registration index root exposed as an ordered `IReadOnlyList<string>`.
```

martinrrm (dismissed):
LGTM

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,273 @@
+# Implementing NuGet.Client CLI Sponsorship Support
+
+- Author: [@kalebfik](https://github.com/kalebfik)
+- GitHub Issue: [NuGet/NuGetGallery#10703](https://github.com/NuGet/NuGetGallery/issues/10703)
+- [Accompanying Functional Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-sponsorCliSpec/accepted/2026/sponsorship-cli.md)
```
donnie-msft: nit: Your branch will get deleted post-merge

```suggestion
- [Functional Spec](https://github.com/NuGet/Home/blob/dev/accepted/2026/sponsorship-cli.md)
```

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,273 @@
...
+
+Console output includes only sources that return one or more sponsorship URLs.
+If none of the selected sources return sponsorship details for a project, the CLI displays:
+
+```text
+// sample response
+No sponsorship details were returned using the following package sources:
+  https://api.example.org/v3/index.json
+
+Consider specifying an additional package source that provides sponsorship metadata, for example: `--source https://api.nuget.org/v3/index.json`.
```
donnie-msft, line 46: Just to clarify this one, we should only include this line if `https://api.nuget.org/v3/index.json` is not already a configured source.

In other words, we don't want to say, no sponsorship details found for nuget.org, then immediately suggest including nuget.org.

donnie-msft (dismissed):

Nigusu-Allehu (dismissed):

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,274 @@
...
+When a source is specified using `--source <SOURCE>`, only that source is queried for sponsorship information.
+
+Console output includes only sources that return one or more sponsorship URLs.
+`https://api.nuget.org/v3/index.json` will be recommended when it is not part of the user's configured sources.
+If none of the selected sources return sponsorship details for a project, the CLI displays:
+
+```text
+// sample response
+No sponsorship details were returned using the following package sources:
+  https://api.source.org/v3/index.json
```
zivkan, line 44: Does this mean "none of the packages referenced are asking for sponsorship"?

The next few lines of the spec have a different message "these sources do not provider sponsorship support", so I assume this "no details were returned" message is for a different scenario, but it's not clear to me what that scenario is.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,274 @@
...
+The command surface will be presented as `dotnet package list --sponsor`.
+
+It will reuse the same metadata-fetch and rendering pipeline, with:
+
+- A `Sponsor` column in console output.
+- A `sponsorablePackages` array in JSON output.
+
+## Explanation
+
+### Functional Explanation
```
zivkan, line 20: I might have missed it, but I didn't see a sample output when a project doesn't have any package references. Especially if you keep the warning message to say "try --source https://api.nuget.org/v3/index.json", then it's an important distinction between zero packages and no sponsorship info for referenced packages.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,274 @@
...
+    "https://api.example.org/v3/index.json",
+    "https://www.myget.org/F/contoso/api/v3/index.json"
+  ],
+  "packages": [
+    {
+      "id": "Contoso.Tools",
+      "projects": [
+        {
+          "path": "/path/to/Contoso.App.csproj",
+          "relationship": "topLevel"
```
zivkan: I can't imagine there are relationships other than top level and transitive, so I think this could be changed to a bool type `isTransitive` property instead.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,274 @@
...
+This proposal assumes the CLI consumes sponsorship data from the Registration API.
+It also depends on each selected package source exposing sponsorship details through the `metadata.sponsorshipUrls` property at the root of its package Registration index.
+
+Root Registration metadata is exposed through `RegistrationResourceV3`, allowing future package-level properties added to the root `metadata` object to use that same abstraction.
+A source that does not expose Registration or has authentication, network, or protocol failures uses the same list-package failure behavior.
+
+#### Testing Strategy
+
+- Validate output for packages with no sponsorship information, a single sponsorship URL, and multiple sponsorship URLs in the CLI and JSON.
+- Verify top-level and transitive packages are included by default.
```
zivkan, line 257: None of the other `dotnet package list` options include transitive packages by default. While it probably makes sense for sponsorship info to include transitive by default, I thought the reason that `dotnet package list` was chosen over a new `dotnet package sponsor` command was familiarity and consistency. So, making this command behave different to the others surprises me a bit.

accepted/2026/implementing-sponsorship-in-cli.md
```diff
@@ -0,0 +1,274 @@
...
+- Cover missing or null `metadata` and `sponsorshipUrls` behavior, along with empty arrays or malformed JSON.
+- Verify enabled configured source selection including packages mapped to one source, multiple sources, and sources excluded by `--source`.
+- Validate sponsorship metadata is propagated from the Registration index and consumed by the command runner.
+
+## Drawbacks
+
+- At this time, nuget.org is the only source that supports sponsorship details in the Registration index root.
+  - Other sources will produce a message saying that source does not support sponsorship reporting.
```
zivkan: I didn't see in the spec how the command will detect whether a source supports sponsorship reporting.

kalebfik: This is mainly due to me not wanting to duplicate information across multiple specs (as best as I could). Ultimately the [RegistrationBaseUrl](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource#versioning) resource version will be updated with a new version that will have the metadata container above. Based on the version, client will know whether a source can or cannot support sponsorships.

accepted/2026/implementing-sponsorship-in-cli.md
kalebfik: I remember this being a topic of discussion when designing the feature, and the main reason why transitive packages are included by default is because of what `--sponsor` is reporting. The main case of someone using `--sponsor` is when they want to know which maintainers are supporting a package they use. 

In the case that a top-level package does not contain sponsorship links, they would receive a message to let them know there are no sponsorable packages for their project. If there are transitive packages that it depends on, a user may not know that those packages are looking for funding without knowing to use `--include-transitive`. 

I agree that it is odd that `--sponsor` differs from other flags, and I think that is exactly why it took a while to come to a decision about it.

accepted/2026/implementing-sponsorship-in-cli.md
kalebfik: Technically, it doesn't. The current design covers packages that do not have sponsorship links, but it doesn't explicitly state that a project does not have package references, rather it shows the same message as a package with no sponsorship links. 
It is a good idea to state the distinction though, and I will update the spec to reflect that.

accepted/2026/implementing-sponsorship-in-cli.md
kalebfik: Yup! A user receives this message when they don't have any packages looking for funding. The extra messaging about specifying a source only applies when api.nuget.org/v3/index.json is not used.

I've updated the spec to show a message for a project that does not reference any packages to try to clear it up a bit more though.

accepted/2026/implementing-sponsorship-in-cli.md
donnie-msft: I agree it makes sense here because we're trying to advertise a hint in some scenarios with `--source`.
However, I wanted to point out that, `dotnet package list --vulnerable  -v:D` does not tell me that I have no package references.

> The given project `ConsoleApp123` has no vulnerable packages given the current sources.

I'll discuss with Andy offline.

donnie-msft (approved):

PR #15002, opened 2026-07-16
PM UI VersionRange constraint updates

martinrrm:
rendered version: https://github.com/NuGet/Home/blob/d16985ab8369de7d3ac82c41843d6093d7097c60/accepted/2026/package-update-version-constraints.md
## Description

Adds a PM UI-only design for respecting authored `PackageReference` version constraints during updates.

The proposal defines how PM UI:

- keeps automatic updates within the authored range;
- preserves ranges for in-range selections;
- replaces the range with a minimum-only version for explicit out-of-range selections;
- applies the behavior per project in solution view and with Central Package Management.

### DIFF
--- a/accepted/2026/package-update-version-constraints.md
+++ b/accepted/2026/package-update-version-constraints.md
@@ -1,0 +1,233 @@
+# Respect package version constraints in Package Manager UI
+
+- [Martin Ruiz](https://github.com/martinrrm)
+- [Adam Stachowicz](https://github.com/Saibamen)
+- [NuGet/Home#6566](https://github.com/NuGet/Home/issues/6566)
+- [NuGet/NuGet.Client#7323](https://github.com/NuGet/NuGet.Client/pull/7323)
+
+## Summary
+
+Visual Studio Package Manager UI (PM UI) will respect upper-bounded and
+floating PackageReference version ranges when finding and applying updates.
+
+Automatic updates will stay inside the authored range. An in-range update will
+preserve the range, while setting its lower bound to the selected version when
+applicable.
+
+If the customer explicitly selects a concrete version outside the range, PM UI
+will replace the range with that minimum-only version. For example:
+
+```text
+[8.0.0, 9.0.0) + selected 9.1.0 -> 9.1.0
+```
+
+This explicit selection does not require an additional confirmation.
+
+This proposal applies only to PM UI. Package Manager Console,
+`dotnet package update`, and other update entry points are outside its scope.
+It is intentionally a partial solution to
+[NuGet/Home#6566](https://github.com/NuGet/Home/issues/6566).
+
+## Motivation
+
+Customers use upper-bounded ranges to stay on a supported package release
+train:
+
+```xml
+<PackageReference Include="Contoso.Logging"
+                  Version="[8.0.0, 9.0.0)" />
+```
+
+Today, PM UI can update the package to an allowed version but replace the range
+with a minimum-only version:
+
+```xml
+<PackageReference Include="Contoso.Logging" Version="8.0.4" />
+```
+
+This silently removes the `9.0.0` ceiling. Customers must then repair the
+project file or risk a later update moving to an unintended major version.
+
+PM UI should preserve the boundary for routine in-range updates, while still
+allowing a deliberate version selection outside the range.
+
+## Explanation
+
+### Functional explanation
+
+#### Scope
+
+This proposal covers:
+
+- project and solution Package Manager UI;
+- top-level PackageReference dependencies (transitive packages do not support `VersionRange`);
+- project-local `PackageReference` versions;
+- and centrally managed `PackageVersion` versions.
+
+#### In-range selection
+
+For a non-floating upper-bounded range, PM UI sets the lower bound to the
+selected version and preserves the original upper bound and inclusivity:
+
+| Before | Selected | After |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


accepted/2026/package-update-version-constraints.md
```diff
@@ -0,0 +1,233 @@
...
+
+- [Martin Ruiz](https://github.com/martinrrm)
+- [Adam Stachowicz](https://github.com/Saibamen)
+- [NuGet/Home#6566](https://github.com/NuGet/Home/issues/6566)
+- [NuGet/NuGet.Client#7323](https://github.com/NuGet/NuGet.Client/pull/7323)
+
+## Summary
+
+Visual Studio Package Manager UI (PM UI) will respect upper-bounded and
+floating PackageReference version ranges when finding and applying updates.
```
zivkan, line 11: we normally format markdown as one line per sentence, not manual line wrapping

accepted/2026/package-update-version-constraints.md
```diff
@@ -0,0 +1,233 @@
...
+```text
+[8.0.0, 9.0.0) + selected 9.1.0 -> 9.1.0
+```
+
+This explicit selection does not require an additional confirmation.
+
+This proposal applies only to PM UI. Package Manager Console,
+`dotnet package update`, and other update entry points are outside its scope.
+It is intentionally a partial solution to
+[NuGet/Home#6566](https://github.com/NuGet/Home/issues/6566).
```
zivkan, line 29: Are you suggesting that some time in the future when `dotnet package update` supports some kind of similar functionality, it's acceptable for it to do so in a way that behaves differently to PM UI?

Even if the initial implementation is only in PM UI and not `dotnet package update`, I would expect them both to have the same behavior and therefore this is a spec for all NuGet update scenarios, not just PM UI.

accepted/2026/package-update-version-constraints.md
```diff
@@ -0,0 +1,233 @@
...
+### Preserve ranges only for in-range updates
+
+Routine updates should not erase a customer's authored boundary. Setting the
+lower bound records the update while retaining the upper-bound policy.
+
+### Replace the range for an explicit out-of-range selection
+
+Selecting an out-of-range concrete version is an explicit request to leave the
+current constraint. Replacing it with the normal minimum-only PackageReference
+version is predictable and does not require PM UI to infer a new policy.
```
zivkan, line 190: Was this entire spec written by AI? I haven't yet started reading the other subheadings under "rationale and alternatives", but these two are repeating what the rest of the spec already explained. I suspect the AI saw that this is a rationale section and is now trying to justify it, but I feel like it's making it harder for me to read.

accepted/2026/package-update-version-constraints.md
```diff
@@ -0,0 +1,233 @@
...
+
+### Do not shift or widen the range automatically
+
+PM UI could infer `[9.1.0, 10.0.0)` or widen the old range, but either behavior
+guesses the customer's intended upper bound. The customer can author a new
+range separately when needed.
```
zivkan, line 196: this only makes sense if you remember that most of the previous examples used the example of updating to version 9.1.0 when the previous range was `[8.0.0, 9.0.0)`. Otherwise, this paragraph is quite confusing.

accepted/2026/package-update-version-constraints.md
```diff
@@ -0,0 +1,233 @@
...
+
+Selecting a concrete version in the package details pane is already an explicit
+update gesture. An additional confirmation would add friction to a deliberate
+selection.
+
+### Use the existing PackageReference range
+
+Using the existing `Version` or `PackageVersion` range avoids a new project
+model contract. A future update-only constraint could separate update policy
+from restore semantics, but that is outside this proposal.
```
zivkan, line 208: When a project is packed, the PackageReference version range becomes the package dependency's version range. Overloading the Version therefore prevents having separate controls.

For example, I don't believe that Microsoft.Extensions.DependencyInjection has ever had a breaking change. If I want to stay on the 10.* package versions until I retarget my project to net12..0, that doesn't mean my package is incompatible with 11.* packages. Only, I don't want NuGet to suggest those versions as an upgrade.

The prior art section already mentions packages.config's allowedVersions attribute, but this alternatives section says it doesn't want to introduce a "new project model contract", while failing to acknowledge that an upper range in the existing Version already has a semantic meaning.

Additionally, "A future update-only contraint could superate update policy from restore semantics" is basically saying we can introduce breaking changes in the future.  I think it's better to introduce something different now, so that we don't need breaking changes in the future.

Going back to customer intent, I think customers are trying to use `Version="[8.0.0, 9.0.0)"` not because they believe that version 9 and above are incompatible with their project, but as an attempt to get this "update range" behavior, before they find out that it doesn't actually work that way. So, I feel like this proposal is reinforcing customers believe of using the wrong tool for the job. I think we should implement a separate `allowedUpates="[8.0.0, 9.0.0)"` or `allowedUpdate="8.*"` instead, and leave the version to retain its existing compatible version semantics.

The behavior in this spec still makes sense when a package is incompatible with higher versions, so the developer wants to bump the version and keep the max compatible version.  But the motivation in this spec is not aligned with that scenario.

accepted/2026/package-update-version-constraints.md
martinrrm: I was trying to scope this spec only to PM UI, but yeah `dotnet package update` should behave the same, that is a bigger discussion that I was trying to avoid but it's needed.

Nigusu-Allehu:
Pr review meeting: @martinrrm  should we have a spec review discussion on this?

PR #15037, opened 2026-08-10
PM Spec for PackageID level Metadata in the Registration API 

### LINKED ISSUE #15038: Functional Design spec for adding PackageID level metadata into the Registration API

kalebfik:
[Rendered Spec](https://github.com/NuGet/Home/blob/dev-kalebfika-pkgIDInRegistrationSpec/accepted/2026/packageID-in-registration.md)

### DIFF
--- a/accepted/2026/packageID-in-registration.md
+++ b/accepted/2026/packageID-in-registration.md
@@ -1,119 +1,140 @@
-# **Adding PackageID Level Metadata into the Registration API**
+# **Adding Package ID-Level Metadata to the Registration API**
 
 - Author: [@kalebfik](https://github.com/kalebfik)
-- GitHub Issue <!-- GitHub Issue link -->
+- GitHub Issue: [NuGet/Home#15038](https://github.com/NuGet/Home/issues/15038)
 
 ## Summary
 
-This proposal adds ID-level metadata to the Registration API, published once at the root of the registration index rather than inside any version's metadata.
+This proposal adds package ID-level metadata to the Registration API.
+The metadata is published once in an optional `metadata` object at the root of the registration index.
 
 ## Motivation
 
-Currently, metadata in registration is scoped by PackageID and version. 
-For certain metadata, especially those that are not changed often, adding functionality for PackageID level metadata provides a way for that information to be expressed. 
-
-An example of this is sponsorship information. 
-Sponsorships are currently a PackageID scoped piece of metadata. 
-If we were to provide a way to show sponsorship information in the Registration API, it would require it to be package id and version scoped, and that would fan out to every version of the package, essentially requiring a new catalog leaf for every version of that package.
-
-Adding PackageID level metadata in the Registration API would allow there to be one place where that information is stored, with one one catalog event per change rather than multiple catalog leaf being written for each change.  
+The Registration API is currently scoped by version.
+Certain metadata, such as sponsorship information, applies to a package ID independently of any one package version.
+Adding package ID-level metadata through the existing version-scoped model would require any metadata changes to fan out to every version of a package, producing a catalog leaf and registration updates for each version.
+Publishing package ID-level metadata at the Registration index root provides one place where ID-specific information is stored, with one catalog event per change rather than multiple catalog leaves being written for each change.
 
 ## Goals
 
-- Enable package-ID data in the Registration API
-- Avoid embedding package-ID level data within existing metadata documents 
+- Support optional package ID-level metadata in the Registration index root.
+- Publish and persist package ID-level changes once per package ID rather than per version.
+- Keep package ID-level data separate from existing Search metadata.
+- Maintain compatibility with clients that do not recognize the optional package ID-level metadata.
+- Ensure package ID-level metadata is included in Registration reflow, backfill, replication, and failover processes.
 
 ## Non-Goals
 
-- Client-side consumption UX (changes to client side functionality)
-- Changes to authorization/validation of data
+- Client-side consumption UX (changes to client side functionality or user experience).
+- Changes to authorization or validation of data.
 
 ## Explanation
 
 ### Functional explanation
 
-This is going to be extending Registration API (specifically `RegistrationBaseUrl`) rather than adding anything new. 
-ID-level metadata is going to be published as an optional property on the root of the Registration Index. 
+This proposal extends the existing `RegistrationBaseUrl`.
+Package ID-level metadata will be published as an optional `metadata` object at the root of the Registration index.
 
-**Example addition of Sponsorships**
+**Example addition of metadata information**
 
 | Name | Type | Required | Notes |
 | --- | --- | --- | --- |
-| count | integer | yes | The number of registration pages in the index |
-| `sponsorshipUrls` | array of strings | no | *(new)* Funding URLs for the package ID |
-| items | array of objects | yes | The array of registration pages |
+| `count` | integer | yes | The number of registration pages in the index |
+| `metadata` | object | no | Package ID-level metadata |
+| `items` | array of objects | yes | The registration pages |
 
-
-**Example registration root entry with `sponsorshipUrls` added:**
+**Example `metadata` registration root entry with `sponsorshipUrls` added:**
 
 ```json
-{ 
+{
   "@id": "https://api.nuget.org/v3/registration5-gz-semver2/contoso.webapi.client/index.json",
   "@type": [
-    "catalog:CatalogRoot", 
-    "PackageRegistration", 
+    "catalog:CatalogRoot",
+    "PackageRegistration",
     "catalog:Permalink"
   ],
   "commitId": "afa91af1-9505-41b8-ad75-eab8e613db14",
   "commitTimeStamp": "2026-04-10T00:15:25.1492389+00:00",
   "count": 2,
   // ** Start of proposal ** //
-  "sponsorshipUrls": [
-    "count": 1, 
-    "items": [
-    "https://github.com/sponsors/contoso"
+  "metadata": {
+    "sponsorshipUrls": [
+      "https://github.com/sponsors/contoso"
     ]
-  ],
+  }
   // ** End of proposal **//
 }
 ```
-**Behavior** 
-- Property written once per Package ID.
-- Changes to data are reflected once, instead of per-version.
-- Additional cursors are created per data-type, to signal any changes to the database and are updated accordingly
-- Data is written to the exisitng blob storage at its root, and consumers fetching the registration index will see the property.
+
+**Behavior**
+
+- The `metadata` object is optional and written once per package ID on the Registration index root.
+- Changes update the index once rather than being reflected in every package version.
+- Rewrites must preserve package ID-level metadata when version-level metadata changes.
+- Metadata must participate in the same publication, replication, reflow, backfill, and recovery processes as version-level metadata.
+- Network, authentication, and protocol failures remain errors and are not represented as empty metadata.
 
 ### Technical explanation
 
-A detailed technical explanation can be found here: [Technical Implementation for adding PackageID level metadata into Registration](https://devdiv.visualstudio.com/DevDiv/_git/NuGet.Services?version=GBkalebfik-sponsorshipV3-spec&path=/docs/specs/2026/SponsorshipV3Spec.md&_a=preview) (internal)
+A detailed technical explanation can be found here: [Technical implementation for adding package ID-level metadata to Registration](https://devdiv.visualstudio.com/DevDiv/_git/NuGet.Services?version=GBkalebfik-sponsorshipV3-spec&path=/docs/specs/2026/SponsorshipV3Spec.md&_a=preview) (internal).
 
 ## Drawbacks
 
-Third party feeds may need to create their own implementation of PackageID level metadata in order to see these changes.
+- Third-party feeds need to implement the optional Registration root contract to publish their own package ID-level metadata.
 
 ## Rationale and alternatives
 
-### Publish through the Search endpoint
+The following client access patterns inform where package ID-level metadata should be published:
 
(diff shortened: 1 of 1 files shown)


accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+This proposal adds ID-level metadata to the Registration API, published once at the root of the registration index rather than inside any version's metadata.
+
+## Motivation
+
+Currently, metadata in registration is scoped by PackageID and version. 
+For certain metadata, especially those that are not changed often, adding functionality for PackageID level metadata provides a way for that information to be expressed. 
+
+An example of this is sponsorship information. 
+Sponsorships are currently a PackageID scoped piece of metadata. 
+If we were to provide a way to show sponsorship information in the Registration API, it would require it to be package id and version scoped, and that would fan out to every version of the package, essentially requiring a new catalog leaf for every version of that package.
```
donnie-msft: nit: consider rewording

```suggestion
The Registration API today is version-scoped, which fans out to each version, requiring a new catalog leaf for every version of that package.
```

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+
+Adding PackageID level metadata in the Registration API would allow there to be one place where that information is stored, with one one catalog event per change rather than multiple catalog leaf being written for each change.
```
donnie-msft: nit & also a typo fix

```suggestion
Adding PackageID level metadata in the Registration API would allow there to be one place where ID-specific information is stored, with one catalog event per change rather than multiple catalog leafs being written for each change.  
```

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+
+An example of this is sponsorship information. 
+Sponsorships are currently a PackageID scoped piece of metadata. 
+If we were to provide a way to show sponsorship information in the Registration API, it would require it to be package id and version scoped, and that would fan out to every version of the package, essentially requiring a new catalog leaf for every version of that package.
+
+Adding PackageID level metadata in the Registration API would allow there to be one place where that information is stored, with one one catalog event per change rather than multiple catalog leaf being written for each change.  
+
+## Goals
+
+- Enable package-ID data in the Registration API
```
donnie-msft:
```suggestion
- Support package ID-level metadata in the Registration API
```

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+- Avoid embedding package-ID level data within existing metadata documents
```
donnie-msft: nit & also, are there other documents besides Search?

```suggestion
- Avoid embedding package ID-level data within existing Search metadata
```

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+Currently, metadata in registration is scoped by PackageID and version. 
+For certain metadata, especially those that are not changed often, adding functionality for PackageID level metadata provides a way for that information to be expressed. 
+
+An example of this is sponsorship information. 
+Sponsorships are currently a PackageID scoped piece of metadata. 
+If we were to provide a way to show sponsorship information in the Registration API, it would require it to be package id and version scoped, and that would fan out to every version of the package, essentially requiring a new catalog leaf for every version of that package.
+
+Adding PackageID level metadata in the Registration API would allow there to be one place where that information is stored, with one one catalog event per change rather than multiple catalog leaf being written for each change.  
+
+## Goals
```
donnie-msft, line 18: You're also working on a data backfill plan and a failover plan for sponsors.
Might be worth mentioning a goal that ID-level metadata is part of the existing failover?

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+### Publish through the Search endpoint
+
+Add a new PackageID-level metadata field to the existing `SearchDocument`/Azure Search index schema, populated by a new sub-command in the existing `Auxiliary2AzureSearchCommand` job, modeled on `UpdateOwnersCommand.cs`.
+
+`SearchQueryService` fires on every Visual Studio Package Manager UI query.
+Adding this data to `SearchDocument` would increase the serialization cost of every one of those high-volume, uncached queries, even for the vast majority that never touch the data at all.
+
+### Publish all Package-ID level metadata (vulnerability-style)
+
+Add a new, standalone V3 service-index resource modeled on the publication mechanics of `BlobStorageVulnerabilityWriter`, but with an ID-only schema.
```
donnie-msft: I'd make it clear this is an Alternative that was considered.

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+| **Pros** | Naturally fits the V3 registration model clients already query per-ID; no new V3 resource type; publicly cacheable via existing CDN path | Reuses proven `UpdateOwnersCommand`-style DB-diff-vs-snapshot pattern — completely skips the catalog, so no new catalog `@type` needed; unrecognized field is safely ignored by existing clients | Simple cursor-driven full-republish recovery story; naturally ID-only |
+| **Cons/Trade-offs** | Catalog is per-version by design; needs new DB trigger since no existing signal fires on registration-only edits; every catalog consumer needs new parsing logic if a new `@type` is introduced | Adds cost to every query even when unused; still requires new seeding logic for `Db2AzureSearchCommand` full rebuilds | Disproportionate to a display-only, optional, low-frequency-lookup use case; forces every consumer into whole-dataset reconciliation for no safety-critical reason |
+
+
+## Unresolved Questions
+
+How should absense of data be treated? 
+When a source does not support the PackageID level metadata in Registration API, it would output the same result as if the package itself did not have the information.
```
donnie-msft: This answers the question above, right?

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+| **Cons/Trade-offs** | Catalog is per-version by design; needs new DB trigger since no existing signal fires on registration-only edits; every catalog consumer needs new parsing logic if a new `@type` is introduced | Adds cost to every query even when unused; still requires new seeding logic for `Db2AzureSearchCommand` full rebuilds | Disproportionate to a display-only, optional, low-frequency-lookup use case; forces every consumer into whole-dataset reconciliation for no safety-critical reason |
+
+
+## Unresolved Questions
+
+How should absense of data be treated? 
+When a source does not support the PackageID level metadata in Registration API, it would output the same result as if the package itself did not have the information. 
+How should the difference between the two be noted?
```
donnie-msft: I'd combine this all into 1 question. Do we need to note those differences?

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,119 @@
...
+## Unresolved Questions
+
+How should absense of data be treated? 
+When a source does not support the PackageID level metadata in Registration API, it would output the same result as if the package itself did not have the information. 
+How should the difference between the two be noted? 
+
+## Future Possibilities
+
+- Extension to Client tooling
+- Additions to PMUI
```
donnie-msft: Can combine these

japarson (dismissed):

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,140 @@
...
+
+**Example addition of metadata information**
+
+| Name | Type | Required | Notes |
+| --- | --- | --- | --- |
+| `count` | integer | yes | The number of registration pages in the index |
+| `metadata` | object | no | Package ID-level metadata |
+| `items` | array of objects | yes | The registration pages |
+
+**Example `metadata` registration root entry with `sponsorshipUrls` added:**
```
zivkan, line 63: This spec defines `metadata` as an object, but then doesn't define what the schema for `metadata` is. Is `sponsorshipUrls` the only property allowed? Are there others? Is it required or optional?

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,140 @@
...
+
+- Client-side consumption UX (changes to client side functionality or user experience).
+- Changes to authorization or validation of data.
+
+## Explanation
+
+### Functional explanation
+
+This proposal extends the existing `RegistrationBaseUrl`.
+Package ID-level metadata will be published as an optional `metadata` object at the root of the Registration index.
```
zivkan, line 53: There may be value in clients being able to detect when a server supports the `metadata` property, in particular the `sponsorshipUrls` sub-property. For example, to tell customers that certain client features are not available because the package source doesn't support it.

How can clients detect whether the server supports this or not?

donnie-msft: It's a great point! We may be able to update the Registration resource version in the index.json on the server. 
I think it's worth doing as it enables client to tell customers when features don't apply to their package source.
It also saves client from wasting effort checking properties on servers where they will never be available.

accepted/2026/packageID-in-registration.md
kalebfik: That's a good point; I left it a bit open due to us not knowing what metadata would actually be implemented here in the future. Sponsorships were used only because that is the first instance of package ID-level metadata being implemented, but if future ID-level metadata is added (owners for example), it can go within the same `metadata` container. The properties are designed to be optional, since this information is not required for someone to publish or manage their packages. 

I'll update the spec to reflect these details!

donnie-msft (dismissed):
Looks good to me! 🚀

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,163 @@
...
+## Motivation
+
+The Registration API is currently scoped by version.
+Certain metadata, such as sponsorship information, applies to a package ID independently of any one package version.
+Adding package ID-level metadata through the existing version-scoped model would require any metadata changes to fan out to every version of a package, producing a catalog leaf and registration updates for each version.
+Publishing package ID-level metadata at the Registration index root provides one place where ID-specific information is stored, with one catalog event per change rather than multiple catalog leaves being written for each change.
+
+## Goals
+
+- Support optional package ID-level metadata in the Registration index root.
```
agr: Just "registration index". Terminology should be consistent with [published documentation](https://learn.microsoft.com/en-us/nuget/api/registration-base-url-resource).

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,163 @@
...
+For `dotnet package list`, its existing metadata report modes already query Registration for installed package IDs, so sponsorship can extend that path rather than introduce a Search or full-dataset lookup.
+
+### Alternative considered: Publish through the Search endpoint
+
+One alternative is to add package ID-level metadata to `SearchDocument` and the Azure Search index schema.
+Initial Search index builds would populate the field, and later changes could be handled by a new subcommand modeled on the existing package-owner update path.
+This would reuse the existing Search publication pipeline, avoid a new package ID-level catalog leaf type, and make the metadata available to discovery experiences such as Package Manager UI Browse.
+
+This alternative was not selected mainly because it would add optional and infrequently consumed data to the high-frequency Search path.
+The field would be retrieved, serialized, and transferred with applicable Search results even when the client does not use it.
```
agr: High frequency compared to what? Package restore does not use search but uses registration index. And package restore is by far the most frequent operation, and it does not need sponsorship URLs for restore. With that reasoning, search should have been preferred.

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,163 @@
...
+Every participating client would download and maintain metadata for the full source even when it needs information for only a small set of packages.
+
+A separate full-source resource is justified for vulnerability data because that data is safety-critical and must be broadly available to restore and audit workflows.
+Sponsorship information is optional, non-blocking display metadata.
+
+### Comparing the approaches
+
+| | Registration | Search | Vulnerability-style dataset |
+| --- | --- | --- | --- |
+| **Implementation cost** | Add the package ID-level catalog lane, collector cursor, typed root model, and recovery support; no new service-index resource | Extend the Search schema, full-build seeding, and auxiliary update pipeline; can avoid a new catalog leaf type | Add a new V3 resource type, publication job, monitoring, and client resource implementation |
```
agr: What is "catalog lane"?

accepted/2026/packageID-in-registration.md
```diff
@@ -0,0 +1,163 @@
...
+| **Client fit** | Matches known and installed package workflows, including existing `dotnet package list` metadata reports | Matches discovery workflows such as Package Manager UI Browse | Matches clients that need to evaluate most or all packages from a source |
+| **Network shape** | Cached request per relevant package ID and source; can require many requests for large solutions | Adds the field to applicable Search results on a frequently used, non-disk-cached query path | One cacheable dataset, but clients download metadata for the entire source |
+| **Pros** | Fetches only requested IDs; reuses existing known-package client paths | Reuses the Search pipeline; can avoid a new catalog type | Explicit support signal; no per-ID disclosure; efficient repeated local lookups |
+| **Cons** | Requires a new ID-level Registration publication lane; unsupported and empty sources are indistinguishable | Repeated optional payload cost; installed clients need another lookup | Disproportionate whole-source transfer and new protocol surface for optional display metadata |
```
agr: What is "Registration publication lane"? The word "lane" appears only in this table in this document and doesn't seem to be defined.

accepted/2026/packageID-in-registration.md
agr: To clarify: I am not saying that search is actually had to be preferred, just that the reasoning sounds weird.

accepted/2026/packageID-in-registration.md
kalebfik: This was in relation to the prior implementation, where we were considering having a different process for registration level events, separate from version level events. We've since decided to use the same timestamp across both version level and registration level events.

donnie-msft (approved):

PR #15048, opened 2026-08-14
Add NuGet package staging CLI proposal

Nigusu-Allehu:
https://github.com/NuGet/Home/blob/dev-nyenework-stage-cli/accepted/2026/nuget-package-staging-cli.md

### DIFF
--- a/accepted/2026/nuget-package-staging-cli.md
+++ b/accepted/2026/nuget-package-staging-cli.md
@@ -14,13 +14,11 @@ Promotion is excluded from the initial CLI and remains a NuGet Gallery action.
 `dotnet nuget push` sends a package directly into the public publication pipeline.
 Publishers cannot use it to privately upload and validate packages before a release, discover validation failures early, or prepare a coordinated release group.
 
-Validation and signing may take significant time.
-Failures discovered only during publication delay package availability and downstream products.
-
 The staging server introduces a private package lifecycle, but users need an automation-friendly CLI to operate it.
 The expected outcome is that publishers can prepare and validate packages before release day, privately organize related packages, and retain an explicit human approval boundary before publication.
 
-This proposal builds on the accepted [`Release staging and deprecation`](../2024/release-staging-and-deprecation.md) design and narrows its scope to the client commands needed to operate private staging.
+This proposal focuses on the CLI parts of package staging.
+It builds on the earlier [`Release staging and deprecation`](https://github.com/NuGet/Home/pull/12874) design and the [`refreshed NuGet staging proposal`](https://github.com/NuGet/Home/pull/14978).
 
 ## Explanation
 
@@ -42,39 +40,56 @@ dotnet nuget stage
     `-- delete <GROUP_ID>
 ```
 
-#### Options supported by every command
+#### Common options
 
-```text
--s|--source <SOURCE>
--k|--api-key <API_KEY>
---configfile <FILE>
---interactive
--t|--timeout <SECONDS>
---allow-insecure-connections
---format <table|json>
-```
+**`-s|--source <SOURCE>`**
 
 The source may be a configured source name or a V3 service-index URL.
 If omitted, the CLI uses `DefaultPushSource`.
 The command fails before making a staging request when neither is available.
 
+**`-k|--api-key <API_KEY>`**
+
+The API key is used to authenticate with the staging resource.
+API-key resolution and precedence are defined in the technical explanation.
+
+**`--configfile <FILE>`**
+
 When `--configfile` is supplied, only that NuGet configuration file is used.
 Otherwise, the normal configuration hierarchy is used.
 
+**`--interactive`**
+
 `--interactive` allows NuGet credential providers to prompt while accessing the selected source or staging resource.
 
-`--timeout` sets the timeout for each HTTP request in seconds and defaults to 300 seconds.
+**`--allow-insecure-connections`**
 
 HTTP sources are rejected unless `--allow-insecure-connections` is supplied.
 The CLI warns when this option is used.
 
-`--format` matches `dotnet package search`: `table` is the default, and `json` selects the machine-readable CLI contract.
+**`--format <console|json>`**
+
+`console` is the default, and `json` selects the machine-readable CLI contract.
+
+#### Commands
+
+##### **`stage push`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage push <PACKAGE_PATH> [--group <GROUP_ID>] [--no-symbols]
+```
+
+**Options**
+
+- **`--group <GROUP_ID>`** uploads the package directly into the specified group.
+- **`--no-symbols`** prevents upload of a matching sibling `.snupkg`.
 
 `dotnet nuget stage push` uploads one `.nupkg` to private staging.
 It returns when the server accepts the upload and does not wait for validation to finish.
 If a matching sibling `.snupkg` exists, the CLI uploads it with the package unless `--no-symbols` is specified.
 Symbol discovery reuses the existing `dotnet nuget push` behavior.
-Initial-release automation polls `dotnet nuget stage view` until validation reaches a terminal state.
 
 ```console
 dotnet nuget stage push artifacts/Contoso.1.0.0.nupkg
@@ -86,6 +101,19 @@ A package can be uploaded directly into a group:
 dotnet nuget stage push artifacts/Contoso.1.0.0.nupkg \
   --group august-release
 ```
+
+##### **`stage list`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage list [--group <GROUP_ID> | --ungrouped]
+```
+
+**Options**
+
+- **`--group <GROUP_ID>`** limits the result to packages in the specified group.
+- **`--ungrouped`** limits the result to packages that are not in a group.
 
 `dotnet nuget stage list` returns the complete package inventory visible to the supplied API key.
 Each entry contains the package ID and version.
@@ -102,11 +130,31 @@ Group summaries are available separately through `dotnet nuget stage group list`
 Each group summary contains group metadata and a package count, but not package membership.
 Use `stage list --group <GROUP_ID>` to list a group's packages.
 
-`dotnet nuget stage view` displays the detailed server-provided state for one package using the normative CLI schema:
+##### **`stage view`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage view <PACKAGE_ID@VERSION>
+```
+
+`dotnet nuget stage view` displays the detailed server-provided state for one package:
 
 ```console
 dotnet nuget stage view Contoso@1.0.0
 ```
+
+##### **`stage delete`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage delete <PACKAGE_ID@VERSION> [--symbols-only]
+```
+
+**Options**
+
+- **`--symbols-only`** deletes only the symbols while leaving the parent package staged.
 
 Staged content can be deleted by identity.
 Symbols can be deleted independently while leaving the parent package staged:
@@ -119,13 +167,42 @@ dotnet nuget stage delete Contoso@1.0.0 --symbols-only
 Package, symbol, and group deletion do not prompt for confirmation.
 Authentication and authorization are enforced by the server.
 
+##### **`stage group create`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage group create <GROUP_ID> [--name <DISPLAY_NAME>]
+```
+
+**Options**
+
+- **`--name <DISPLAY_NAME>`** assigns a human-readable display name to the group.
+
 Groups use a user-selected, immutable ID for commands and API routes.
-They may also have a human-readable display name.
 The valid ID syntax, including characters, casing, length, normalization, uniqueness, and reserved values, must be agreed with the NuGet server team before implementation.
 
 ```console
 dotnet nuget stage group create august-release \
   --name "August Release"
+```
+
+##### **`stage group list`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage group list
+```
+
+`dotnet nuget stage group list` returns group summaries containing group metadata and a package count, but not package membership.
+
+##### **`stage group add`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage group add <GROUP_ID> <PACKAGE_ID@VERSION>
 ```
 
 An existing staged package can be added to a group:
@@ -134,10 +211,26 @@ An existing staged package can be added to a group:
 dotnet nuget stage group add august-release Contoso@1.0.0
 ```
 
+##### **`stage group remove`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage group remove <GROUP_ID> <PACKAGE_ID@VERSION>
+```
+
 `group remove` requests removal of a package's membership from a group:
 
 ```console
 dotnet nuget stage group remove august-release Contoso@1.0.0
+```
+
+##### **`stage group delete`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage group delete <GROUP_ID>
 ```
 
 `group delete` requests deletion of a group:
@@ -145,11 +238,6 @@ dotnet nuget stage group remove august-release Contoso@1.0.0
 ```console
 dotnet nuget stage group delete august-release
 ```
-
-The initial release does not expose group display-name updates, staged-content downloads, listed-intent changes, validation waiting, or promotion.
-
-The initial CLI does not perform promotion.
-After validation succeeds, an authorized owner reviews and promotes an individual package or group through the NuGet Gallery.
 
 ### Technical explanation
 
@@ -168,7 +256,6 @@ The first supported contract is:
 If the source does not advertise a supported staging resource, the command fails.
 It must never fall back to `PackagePublish/2.0.0` or invoke ordinary push, because doing so could publish content that the caller intended to keep private.
 
-Staging is implemented as a new protocol resource and command runner rather than a mode on `PackageUpdateResource` or `dotnet nuget push`.
 The implementation should introduce equivalents of:
 
 ```text
@@ -198,9 +285,8 @@ NuGet.config key for the default nuget.org Gallery URL
 ```
 
 The CLI treats the API key as opaque and provides no `--owner` or `--organization` option.
-The documented server contract defines ownership resolution, authorization, and the inventory visible to a key.
 
-Because staged content is private, list and view operations use the same authentication model as mutations.
+Because staged content is private, list and view operations require authentication and return only content the authenticated identity is authorized to access.
 
 #### API mapping
 
@@ -224,8 +310,6 @@ The package upload uses multipart form data.
 It sends the `.nupkg`, an optional matching `.snupkg`, and an optional `stagingGroup` field.
 
 Package identities supplied on the command line use `<PACKAGE_ID>@<VERSION>`.
-The CLI separates and URL-escapes the two route segments.
-Package IDs remain case-insensitive, and server-side NuGet version normalization is authoritative.
 
 The CLI does not define replacement behavior and must not issue a preflight read before upload.
 It sends the upload and reports the server result.
@@ -239,7 +323,7 @@ With no filter, it returns all staged packages visible to the supplied API key, 
 
 #### Output
 
-Table output is the default.
+Console output is the default.
 Structured output is selected with `--format json`.
 
 JSON output is a versioned CLI contract rather than a direct copy of the server response.
@@ -267,7 +351,7 @@ Successful commands and commands containing only `Warning` problems return exit 
 #### Errors and retries
 
 The implementation parses the staging server's error contract and translates problems into the CLI output model.
-Table output presents actionable messages.
+Console output presents actionable messages.
 
 The CLI reuses `HttpSource` for existing NuGet transport behavior.
 
--- a/accepted/2026/nuget-package-staging-cli.md
+++ b/accepted/2026/nuget-package-staging-cli.md
@@ -1,0 +1,309 @@
+# NuGet Package Staging CLI
+
+- [Nigusu](https://github.com/Nigusu-Allehu)
+- [GitHub Issue #15014](https://github.com/NuGet/Home/issues/15014)
+
+## Summary
+
+Add a `dotnet nuget stage` command family for preparing NuGet packages without publishing them immediately.
+The CLI allows users and CI systems to upload packages to private staging, list and inspect staged packages, organize packages into release groups, and delete staged content.
+Promotion is excluded from the initial CLI and remains a NuGet Gallery action.
+
+## Motivation
+
+`dotnet nuget push` sends a package directly into the public publication pipeline.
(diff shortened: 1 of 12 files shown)


accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+**Synopsis**
+
+```text
+dotnet nuget stage push <PACKAGE_PATH> [--group <GROUP_ID>] [--unlisted]
+```
+
+**Options**
+
+- **`--group <GROUP_ID>`** uploads the artifact directly into the specified package or symbol group based on the file extension.
+- **`--unlisted`** sets the package's listed intent to `false`.
```
zivkan, line 87: This feels weird to me. Unlisted exists because nuget.org doesn't want to break restores and therefore forbids deleting published packages. But if a package owner wants to stop their package being discovered for new installs, they can unlist it.  So, it's not clear to me why a package owner would want to publish a new version of a package, but have it unlisted immediately.

But I guess another point of view is that the protocol has the concept of unlisted, so by making commands support it, it enables scenarios that I haven't yet imagined a use-case for.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+`console` is the default, and `json` selects the machine-readable CLI contract.
+
+#### Commands
+
+##### **`stage push`**
+
+**Synopsis**
+
+```text
+dotnet nuget stage push <PACKAGE_PATH> [--group <GROUP_ID>] [--unlisted]
```
zivkan, line 81: It's common to want to push all packages in a directory, or in a repo. Does it support only a single package, or can it take a list?  For example `dotnet nuget stage push .\artifacts\nupkgs\*`?

Note that Windows shells don't expand the arguments, so on cmd.exe and PowerShell, it will look like a single argument of value `.\articts\nupkgs\*`, but POSIX shells expand file globs, so will appear as multiple arguments, each of which is a specific filename.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+```
+
+**Options**
+
+- **`--group <GROUP_ID>`** uploads the artifact directly into the specified package or symbol group based on the file extension.
+- **`--unlisted`** sets the package's listed intent to `false`.
+  This option is valid only when uploading a `.nupkg`.
+  Using it with a `.snupkg` fails before a staging request is sent.
```
zivkan, line 89: so it won't fail with the older .symbols.nupkg symbols package?

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+```text
+dotnet nuget stage list [--kind <package|symbols>]
+```
+
+**Options**
+
+- **`--kind <package|symbols>`** selects packages or symbols and defaults to `package`.
+
+`dotnet nuget stage list` returns the complete artifact inventory for the selected kind visible to the supplied API key.
```
zivkan, line 134: artifact inventory?  Did AI write this, or was this your choice of name?

Anyway, what's not clear is if this is intended to show every package the API key has access to (even if the package was published without staging), or if it should show only staged packages, even if they've been published, or only staged packages that have not yet been published.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+**Synopsis**
+
+```text
+dotnet nuget stage list [--kind <package|symbols>]
+```
+
+**Options**
+
+- **`--kind <package|symbols>`** selects packages or symbols and defaults to `package`.
```
zivkan, line 132: this feels weird to me.

The "normal" path is that symbols are pushed immediately after the package it's related to, so the two together effectively become a pair. So, in effect, a package either has symbols, or it doesn't. This means it's a boolean true/false flag on a package, and I don't think that a `--kind` option would be needed.

So, I guess the question comes down to edge cases:

* Is it a goal to allow staging symbols for a package that hasn't yet been staged? Or put another way, can we instead require the package to be staged before the symbols?
* Is it a goal to allow staging symbols for a package that is already published? (why would you stage it, rather than just pushing?)

Basically, is there a scenario where symbols will be staged without the package, and why should we allow that scenario?

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+```text
+dotnet nuget stage group create <GROUP_ID> [--name <DISPLAY_NAME>] [--kind <package|symbols>]
+```
+
+**Options**
+
+- **`--name <DISPLAY_NAME>`** assigns a human-readable display name to the group.
+  When omitted, the display name defaults to `<GROUP_ID>`.
+- **`--kind <package|symbols>`** selects the group kind and defaults to `package`.
```
zivkan, line 197: I don't think creating different groups for packages and symbols makes sense from a customer point of view. Under the normal workflow, they will both be staged and published together, not seperately.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+API-key resolution uses:
+
+```text
+-k|--api-key
+NUGET_STAGE_API_KEY
+NuGet.config key for the staging endpoint
+```
+
+`NUGET_API_KEY` is not used by staging commands.
```
zivkan, line 308: I think it should. Both `dotnet nuget push` and these staging commands are in the service of getting packages onto a nuget feed, either publishing immediately, or to staging so it can be published later. While I can't imagine why someone would want to push some package and stage others in a single workflow, the fact that NuGet will be sending HTTP requests to the same package source means it should be using the same credentials, and therefore use the same environment variable.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+PackageStagingResource
+PackageStagingResourceV3Provider
+StageRunner
+```
+
+The staging implementation reuses existing NuGet infrastructure for source resolution, V3 service-index access, API-key lookup, HTTPS enforcement, `HttpSource`, HTTP upload handling, timeout handling, logging, cancellation, and secret redaction.
+
+#### Authentication and ownership
+
+API-key resolution uses:
```
zivkan, line 300: I haven't memorized the order in which push uses. If it's the same, I think it should be called out, to show that's it's consistent. If it's different, why?

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+```text
+-k|--api-key
+NUGET_STAGE_API_KEY
+NuGet.config key for the staging endpoint
+```
+
+`NUGET_API_KEY` is not used by staging commands.
+The NuGet.config key is mapped to the `@id` of the discovered `PackageStaging` resource.
```
zivkan, line 309: Why not use the service index URL? Isn't that what push uses?

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+The CLI sends the resolved key in the `X-NuGet-ApiKey` header.
+Trusted Publishing must exchange its identity token for a short-lived staging API key before invoking the command.
+
+Because staged content is private, list and view operations require authentication and return only content the authenticated identity is authorized to access.
+
+#### API mapping
+
+The authoritative server routes, request fields, response schemas, and error behavior
+are defined in the
+[`PackageStaging/1.0.0` API contract](https://devdiv.visualstudio.com/DevDiv/_git/NuGet.Services?path=/docs/specs/2026/StagingAPIContracts.md&version=GBjaparson/staging-spec&_a=preview).
```
zivkan, line 323: one line per setence, please.

Also, this spec is on NuGet/Home and is therefore public, but the link you provided is private.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+| `stage view <id@version> --kind symbols` | `GET symbols/{id}/{version}` |
+| `stage delete <id@version> --kind package` | `DELETE package/{id}/{version}` |
+| `stage delete <id@version> --kind symbols` | `DELETE symbols/{id}/{version}` |
+| `stage group create <group> --kind <kind>` | `POST groups/{kind}` |
+| `stage group list --kind <kind>` | `GET groups/{kind}` |
+| `stage group add <group> <id@version> --kind <kind>` | `PUT groups/{kind}/{groupId}/items/{id}/{version}` |
+| `stage group remove <group> <id@version> --kind <kind>` | `DELETE groups/{kind}/{groupId}/items/{id}/{version}` |
+| `stage group delete <group> --kind <kind>` | `DELETE groups/{kind}/{groupId}` |
+
+Package and symbol uploads use separate multipart requests.
```
zivkan: Why use multipart form data if only one file is sent per http request? Seems to be unnecessary complexity and waste of CPU processing time, and makes file streaming harder.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+The package request sends one `.nupkg`.
+When `--unlisted` is supplied, the package request includes `listed: false`.
```
zivkan, line 344: includes how?  Is there a "metadata" part of the multipart http request?  Or is it an http header on the nupkg part?

accepted/2026/nuget-package-staging-cli.md
zivkan, line 342: Please clarify. HTTP multipart is specifically about sending multiple pieces of data with on request. If we're really using different requests, then using multipart is unnecessary complexity if we just send one part per request. So, I'm wondering if you're mixing up terminology and actually mean that the nupkg and snupkg are sent as different parts on the same request?

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+Otherwise, the CLI omits `listed` and the server default of `true` applies.
+The symbol request sends one `.snupkg`.
+
+Artifact identities supplied on the command line use `<PACKAGE_ID>@<VERSION>`.
+
+The CLI does not issue a preflight request or determine replacement behavior.
```
zivkan, line 350: The only context I'm aware of "preflight" in the HTTP space is CORS validation. But since staging packages is done on a command line, not javascript on another webpage, doing CORS checks doesn't make sense to me.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+Successful commands and commands containing only `Warning` problems return exit code `0`.
+
+#### Errors and retries
+
+Any `2xx` response is successful.
+Non-`2xx` responses use existing `HttpSource` failure and retry handling.
+The CLI does not require or parse a staging-specific error envelope.
+List, view, and group-list commands deserialize successful response bodies.
+
+## Rationale and alternatives
```
zivkan, line 400: I wonder if a standalone cli tool could be used while the feature is under development, and get early feedback on CLI arguments. After it gets some usage, then we can take the learnings into the dotnet cli.

accepted/2026/nuget-package-staging-cli.md
japarson: Currently for normal push an owner can push a package and unlist it while the package is being validated which allows the package to be published without ever being discoverable. We're just trying to support that behavior for staged promotion too.

In the gallery staging UI we will have a checkbox for each staged package that controls whether it will be listed/unlisted on promotion. The flag here will set the value for that checkbox on stage push.

We could remove the flag here and make it a UI control only, but if for some reason a person wants to promote a 1000 unlisted packages that would be a lot of clicks.

accepted/2026/nuget-package-staging-cli.md
japarson: Stage pushing a list of packages is making me reconsider exposing the `--listed` argument for normal packages. If you stage push `.\artifacts\nupkgs\*` which contains normal packages and symbol packages, would we throw an error because symbols can't be unlisted or just ignore that flag for symbols? The behavior isn't obvious to users either way.

accepted/2026/nuget-package-staging-cli.md
japarson: nuget.org will only support .snupkg packages for staging. Other staging implementations could choose to support the legacy format.

accepted/2026/nuget-package-staging-cli.md
japarson: It will not show packages that were published without staging, only currently staged packages that have not been promoted.

accepted/2026/nuget-package-staging-cli.md
japarson: Here are the rules we are enforcing on nuget.org in regard to package/symbol staging:
- The package must be staged before its symbols.
  - The client should wait for the stage push response and then send the symbol stage push as separate request.
- If a package and symbol are staged together, but the package is deleted, we keep the symbols staged in a "waiting for parent" state. If the user tries to replace the symbol package via stage push, we allow that.
- Symbol packages can be staged and promoted if the parent package is already publsihed.

accepted/2026/nuget-package-staging-cli.md
japarson: We reviewed this internally yesterday and decided to remove the separate groups. Packages and symbols can now be grouped together. There is a caveat however that the parent and symbol packages must be both ungrouped or exist in the same group.

accepted/2026/nuget-package-staging-cli.md
japarson: Normal packages and symbol packages are never sent as parts of the same request. Stage pushing both of them requires 2 separate requests.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+
+Staged content can be deleted by identity.
+Symbols can be deleted independently while leaving the parent package staged:
+
+```console
+dotnet nuget stage delete Contoso@1.0.0
+dotnet nuget stage delete Contoso@1.0.0 --kind symbols
+```
+
+Package, symbol, and group deletion do not prompt for confirmation.
```
japarson, line 182: Why should we not require confirmation? Can't we make interactive default and supply some non-interactive flag for people who don't want confirmation? For example, a mis-typed group ID could cause a lot of pain and could get caught by a confirmation.

accepted/2026/nuget-package-staging-cli.md
```diff
@@ -0,0 +1,414 @@
...
+Credentials and authorization headers must never appear in output, logs, or telemetry.
+
+Commands return exit code `1` for invalid usage or when `problems` contains an `Error`.
+Successful commands and commands containing only `Warning` problems return exit code `0`.
+
+#### Errors and retries
+
+Any `2xx` response is successful.
+Non-`2xx` responses use existing `HttpSource` failure and retry handling.
+The CLI does not require or parse a staging-specific error envelope.
```
japarson, line 397: The staging API provides a user facing error message, but this says the CLI will not use it. Why?

accepted/2026/nuget-package-staging-cli-json-output.md
```diff
@@ -0,0 +1,344 @@
...
+  "result": {
+    "id": "release",
+    "name": "release",
+    "kind": "symbols"
+  },
+  "problems": []
+}
+```
+
+## `stage group list --format json`
```
japarson, line 243: Group output should include the `blockers` array which is provided in the API. Otherwise `canPromote: false` doesn't provide any reason(s) why.

accepted/2026/nuget-package-staging-cli.md
zivkan: > unlist it while the package is being validated

So, customers were unlisting newly published packages, because no staging feature existed?  To me, that suggests that a formal staging feature has less reason to support unlisting on push.

This doesn't mean there isn't still another good reason to unlist on publish/push, but it's not obvious to me.

accepted/2026/nuget-package-staging-cli.md
zivkan: yes, and this is a client feature spec that needs to work with all servers. The spec currently only says snupkg, hence why I'm pointing out changes to the spec to make it not nuget.org specific 😄

accepted/2026/nuget-package-staging-cli.md
zivkan: Are multiple parts ever sent in any request? Say I'm pushing two independent packages?  Why is HTTP form multipart requests being used at all?

accepted/2026/nuget-package-staging-cli.md
japarson: I wasn't suggesting otherwise. I was just clarifying that nuget.org specifically only supports .snupkg. Sorry.

accepted/2026/nuget-package-staging-cli.md
japarson: Staging requests may contain the `listed` and `groupId` multi part form fields

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: Thanks both for digging into this. Andy, that's a fair point, if the main precedent for unlisting-on-push today is really just working around the lack of a proper staging feature, then baking `--unlisted` into `stage push` for v1 is worth reconsidering. And James, the bulk-promotion use case (avoiding hundreds/thousands of manual UI clicks) is a legitimate concern too 


Proposal: let's drop `--unlisted` from `stage push` for now. We'll move CLI support for this to Future Possibilities, and add it back if/when we see real customer demand for setting listed-intent from the CLI. unless you feel strongly otherwise, James?

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: Right now `stage push` stays single-artifact-per-call; multi-path/glob support is already tracked under Future Possibilities. In CI it would be a simple loop the users could wrap this command around with. If users ask for it we can make that abstraction too.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: Good catch. We'll support both `.snupkg` and the legacy `.symbols.nupkg` for `stage push`  the client shouldn't be nuget.org-specific here. Also, since `--unlisted` is being removed (moved to Future Possibilities per the other thread), the "fails with `.snupkg`" clause tied to it goes away too. I'll update the spec accordingly.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: definitely AI. As far as `stage list` just prints whatever the server's list API returns as a response; the CLI doesn't filter or assume promoted vs. not-promoted.  As James said nuget.org excludes promoted packages. And I would expect that to be the standard for other sources too, but from the clients point of view we are listing all result of the stage list api request

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: I will be updating this according to thew new api desings

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: I'll update the spec to reuse `NUGET_API_KEY`

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: It's the same. Will add a sentence mentioning its consistency.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: It uses both. I will update this to use both too

Nigusu-Allehu: https://github.com/NuGet/NuGet.Client/blob/5be395d325e05961189acb7dfdbb5173a1909773/src/NuGet.Core/NuGet.Commands/CommandRunners/PushRunner.cs#L124-L136

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: It is private but useful to this design spec and wanted to at least help people that have acces

accepted/2026/nuget-package-staging-cli.md
japarson: I can't speak exactly as to why people want to do this, but it is a scenario that people want supported: https://stackoverflow.com/questions/59946581/publish-unlisted-nuget-package

I would probably vote to keep it.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: it will be sent as its own multipart field alongside the package https://devdiv.visualstudio.com/DevDiv/_git/NuGet.Services?path=/docs/specs/2026/StagingAPIContracts.md&version=GBjaparson/staging-spec&line=81&lineEnd=84&lineStartColumn=1&lineEndColumn=62&lineStyle=plain&_a=contents

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: I didn't mean CORS validation (not sure what that is tbh). I just mean we won't send a separate existence check to the server before hand and we will leave that responsibility to the server in our operations. Example: If I am staging a package, we will blindly send it to the server and let the server handle the replacement or ignoring an existing package.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: To clarify : the CLI does surface the API's error message to the user; what it won't do is add case-by-case custom handling per error type. It just displays whatever failure message the server returns, rather than trying to interpret it into different CLI-specific behaviors. At least for now

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: Reading it though, that scenario is really about `push` + `delete` as a workaround for the absence of staging. rather than demand for `--unlisted` specifically within the new staging flow? Once staging exists, would we  expect that workaround to still be needed?

Nigusu-Allehu: For now I have moved this to future possibilities from the CLI side.

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: FYI, for rc2 we will strive to enable just `stage push` command.

accepted/2026/nuget-package-staging-cli-json-output.md
Nigusu-Allehu: Move json output to future possibility

accepted/2026/nuget-package-staging-cli.md
Nigusu-Allehu: Fair point. My thought process is if someone types a whole group id for deletion, the really want to delete it. However, if they have two similar group Ids, as you said, it is possible they delete the wrong group.

PR #15052, opened 2026-08-19
Add PowerShell script execution spec

zivkan:


accepted/2026/control-powershell-script-execution.md
```diff
@@ -0,0 +1,156 @@
...
+
+So when considering all these points together, the decision is to not add any settings to persist choices at this time.
+After the change ships, we can get customer feedback.
+
+## Prior Art
+
+Projects using `PackageReference` can avoid importing package MSBuild files by using `ExcludeAssets="build;buildTransitive"`.
+However, package MSBuild file import is per-project, unlike `init.ps1` in the Package Manager Console.
+
+npm has an `--ignore-scripts` option on the command line, or a `ignore-scripts=true` setting in the config file, to ignore post-install scripts in packages.
```
nkolev92, line 153: The default is now scripts are not run though correct?

accepted/2026/control-powershell-script-execution.md
```diff
@@ -0,0 +1,156 @@
...
+When installing multiple versions of a package, each of which have their own `tools/init.ps1`, NuGet has never attempted to "uninit" or undo any changes that `init.ps1` has made.
+`Import-NuGetPackagesInitScripts` will also treat this as an unsupported scenario and customers will need to restart Visual Studio to unload any changes that were made by previous invocations.
+
+## Drawbacks
+
+Packages that are using `install.ps1`, `uninstall.ps1` or `init.ps1` for legitimate purposes will be more difficult to use.
+Unfortunately, this is the nature of security hardening.
+
+## Rationale and alternatives
+
```
nkolev92, line 120: I am wondering if package installation for packages that have install.ps1 and uninstall.ps1 should fail first and then basically make the customer explicitly opt in or opt out. 

i wonder if the impact of just disabling the scripts may be too great for folks since they may miss this notice.

Or maybe at least show that scripts were not in a more meaningful way, since I think the message in the PMC or Package Manager output is really easy to miss.

accepted/2026/control-powershell-script-execution.md
```diff
@@ -0,0 +1,156 @@
...
+1. The package counts were non-zero, but they were so small that they round off to 0.0% when using a single decimal place.
+
+When filtering to packages that have at least one million downloads (across all versions), the percentages are similar.
+3.5% of the latest SemVer stable packages have install.ps1 and 0.3% contain init.ps1.
+
+Customer impact depends on popular package usage, not the percentage of packages that use this NuGet feature.
+But package contents and download counts are publicly available information.
+Also, if a large percentage of packages used this feature, then it would be a signal that it must naturally also have high usage from package consumption.
+Since few packages use PowerShell scripts, this suggests that disabling automatic execution will affect a relatively small portion of package consumption.
+
```
nkolev92, line 44: Do you know how many of the packages installed are using install/uninstall scripts? So rather than look at opportunity look at actual package installation as part of the totality of packages. 

We may need to add telemetry for this, but *I* think it's something that we may be able to figure out from existing tel for public packages.

nkolev92 (commented):
Great write-up. 

I am on board with the no persisted setting part, but I'm wondering if we need to make the notification more prominent.

accepted/2026/control-powershell-script-execution.md
zivkan: AI chatbots tell me that npm still runs package install scripts by default on clean machines. You can use the `.npmrc` config file to disable it by default, either at the user-level config, or repo-level. But it's something that someone has to set, on a clean machine in a clean directory, apparently it will still run package install scripts.

accepted/2026/control-powershell-script-execution.md
zivkan: For PMC, we could add `-IgnorePackageScripts`, and when 1. the project uses packages.config and 2. one or more package install.ps1 or uninsta..ps1 exist, then fail. This way PMC installs, removes and updates must either `-AllowPackageScripts` or `-IgnorePackageScripts`.  It's then a matter of how risk adverse we are about people having scripts, or packages having scripts which in turn run `Update-Package` or `Install-Package`, which will start failing.

For PM UI, we could pop up a dialog, similar to (or maybe reuse and extend) the preview install window, possibly allowing the user to choose package by package to allow or ignore the package scripts. However, as someone not very experienced with GUI development, I'm worried about the cost-benefit ratio. Maybe copilot can implement it easily, but we'll still need to accessibility for customers using screen readers and/or keyboard only. Also, our Apex tests.

Perhaps given the low usage (since this only affects projects using packages.config and a package happens to have install.ps1 or uninstall.ps1), maybe it's acceptable to just fail PM UI with a message telling them the only way to install/upgrade/remove the package is via PMC now.

So, there are options, but it's a question of time and effort vs cost and taking customer usage into account, as well has how much of a breaking change we're willing to make in these existing workflows.

accepted/2026/control-powershell-script-execution.md
zivkan: I didn't look at which packages are actually getting installed, but we emit telemetry when solutions are closed, and when VS is closed, on the number of PowerShell host executions happened via PMC and PMUI. The PMUI ones are obviously a package-sourced ps1 script, whereas PMC usage could be customers using PMC as a terminal. In both cases, the percentage of both scenarios are extremely small.  I'm sure it won't be of comfort to the customers who will be affected by whatever changes we implement, but I'm not sure our management chain will be supportive of us implementing polished features for such a small customer cohort.

accepted/2026/control-powershell-script-execution.md
```diff
@@ -0,0 +1,156 @@
...
+The [Microsoft.EntityFrameworkCore.Tools package](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Tools) has more downloads than the [EntityFramework (EF6) package](https://www.nuget.org/packages/EntityFramework), despite the EF package being available for 4-5 years longer.
+At the time this is being written, the "per day average" for EF Core's package is 20 times higher than EF6.
+This is relevant because EF Core also has a .NET (global) tool package, allowing `dotnet ef` commands from any terminal, not just PMC.
+Hopefully this means that migrating from the PMC tooling to the .NET tool will just be a minor inconvenience.
+While developers using EF6 won't have alternatives to running `Import-NuGetPackageInitScripts` every time they restart Visual Studio, the evidence above suggests it's a smaller percentage of customers that will be affected.
+
+Another angle to consider is that once a configuration value is set, it's rarely reconsidered.
+If the configuration is forgotten about, the value can persist for years.
+While temporary exemptions to security features are useful, a forgotten security setting isn't temporary and therefore could pose a risk.
```
AndriySvyryd, line 141: This can be strengthened by storing the version for which the exemption was created. So, after an upgrade the exemption would need to be explicitly upgraded as well.

zivkan: I was brainstorming with an AI. I think (agree?) that giving exemptions to specific versions only makes sense, never all versions of a package name. This raises design questions about whether this should apply to install.ps1 and uninstall.ps1, or only init.ps1.

If customers have to add an exemption to a new version of a package being installing/upgrading to it, then that seems like more work than just using `-AllowPackageScripts` on a PMC install/upgrade/remove. So, I think that exemptions that are limited to init.ps1 makes sense. Since init.ps1 is PMC only, then it also makes sense to manage the excemption list via PMC cmdlets, and not need a GUI to manage it.

We could persist the exemption list to the VS .suo file, which is ignored by source control normally. This means someone pushing a malicious commit shouldn't enable running potentially malicious ps1 scripts on anyone who clones/pulls the repo.

However, as I mentioned in the spec, in order to ship the change more quickly, let's keep the V1 scope small, and this persisted exemption can be a follow up. Maybe copilot will be able to implement it quickly, so maybe we can add this in the same month that the spec's changes are merged.

